PR #4, opened 2020-03-09
WIP: Use StatefulSet to run multiple runners

summerwind:


api/v1alpha1/runner_types.go
```diff
@@ -21,25 +21,52 @@ import (
...
+
 // RunnerSpec defines the desired state of Runner
 type RunnerSpec struct {
 	// +kubebuilder:validation:MinLength=3
 	// +kubebuilder:validation:Pattern=`^[^/]+/[^/]+$`
 	Repository string `json:"repository"`
 
 	// +optional
-	Image string `json:"image"`
+	Replicas *int32 `json:"replicas,omitempty"`
```
mumoshu: Could you use `RunnerSet` or something similar for managing a stateful sets instead of pods?
Doing so will allow me to deploy both pod-based and stateful-set based runners for comparison against real workloads.

// If that makes sense, I'll rename the current `RunnnerSet` that I've added in #1 to `RunnerReplicaSet` to not collide and make mine look like a less recommended way.

mumoshu: Just submitted #6 for the rename

summerwind: Thank you for your helpful comment and PR! Using the `RunnerSet` resource for StatefulSet based management is a great idea. I'm going to merge #6 after reviewing.

callum-tait-pbx:
@mumoshu PRs are included by the stale bot if any are they are going to hang around need one of the exempt labels.

toast-gear:
Let me close this due to inactivity and conflicts, but feel free to restart another pull request. Thanks anyway!

PR #6, opened 2020-03-10
Rename RunnerSet to RunnerReplicaSet

mumoshu:
To hand over the name `RunnerSet` to the new StatefulSet-based implementation of that being developed at #4


missedone:
could you have a minor release, i'd like to try out this change. 
(I just start my evaluation on github actions, this controller looks quite promising, given there's naming change in this PR, i'd prefer use the latest, rather than rename my manifest later)
thanks

missedone:
ok never mind, i see the latest tag in dockerhub: https://hub.docker.com/r/summerwind/actions-runner-controller/tags

Issue #5, opened 2020-03-09
[question]: Support of pod spec fields


alexandrst88:
Hi! Thanks, for this cool tool. I have few questions, do you have a plan to add pod annotations and servicaccount to runners?

```
spec:
  replicas: 3
  template:
    metadata:
      annotations:
        iam.amazonaws.com/role: role-arn```
```

I'd like to build docker images and push them to ecr or have ability to access k8s cluster api for example as use case.

If nope, I could work on this, thanks

mumoshu:
@alexandrst88 Hey! I believe this being a valid feature request. Probably adding `annotations` next to https://github.com/summerwind/actions-runner-controller/blob/0edf0d59f74bcbbc7c9bd899c6547c2a49a9b1b5/api/v1alpha1/runner_types.go#L34 would make sense.

cc/ @summerwind

summerwind:
Thank you for the feedback!
That's nice idea! I'm also considering a similar feature. In my use case, I'd like to set Resource Limits.

However, managing each field of a pod individually can be complicated. How about adding a podTemplate field to Runner? 

```
apiVersion: actions.summerwind.dev/v1alpha1
kind: Runner
metadata:
  name: example
spec:
  repository: summerwind/actions-runner-controller
  podTemplate:
    metadata:
      annotations:
        iam.amazonaws.com/role: role-arn
    spec:
      serviceAccountName: example
      ...
```

summerwind:
I'm currently working on StatefulSet based runner management in #4. After that, it might be a good time to implement `podTemplate`.

alexandrst88:
I might be wrong, but in runner controller,

alexandrst88:
I might me wrong, but in runner_controller.go

```	pod := corev1.Pod{
		ObjectMeta: metav1.ObjectMeta{
			Name:      runner.Name,
			Namespace: runner.Namespace,
		},
		Spec: corev1.PodSpec{
			RestartPolicy: "OnFailure",
			Containers: []corev1.Container{
				{
					Name:            containerName,
					Image:           runnerImage,
					ImagePullPolicy: "Always",
					Env:             env,
					VolumeMounts: []corev1.VolumeMount{
						{
							Name:      "docker",
							MountPath: "/var/run",
						},
					},
					SecurityContext: &corev1.SecurityContext{
						RunAsGroup: &group,
					},
				},
				{
					Name:  "docker",
					Image: r.DockerImage,
[...]

PR #7, opened 2020-03-13
Add variants of PodTemplate spec fields into the Runner spec

alexandrst88:
Addressing https://github.com/summerwind/actions-runner-controller/issues/5. 

But need you advise or decision on https://github.com/summerwind/actions-runner-controller/issues/5#issuecomment-598324941 :)


api/v1alpha1/runner_types.go
```diff
@@ -31,7 +31,7 @@ type RunnerSpec struct {
 	Image string `json:"image"`
 
 	// +optional
-	Env []corev1.EnvVar `json:"env,omitempty"`
+	PodTemplate corev1.PodTemplate `json:"podTemplate,omitempty"`
```
summerwind: Could you add a `PodTemplate` field instead of replacing `Env` to keep your custom resources backward compatible?

controllers/runner_controller.go
```diff
@@ -398,10 +397,16 @@ func (r *RunnerReconciler) newPod(runner v1alpha1.Runner) (corev1.Pod, error) {
 		},
 	}
 
+	if runner.Spec.PodTemplate.ObjectMeta.Size() > 0 {
+		runner.Spec.PodTemplate.ObjectMeta.DeepCopyInto(&pod.ObjectMeta)
+	}
+
+	if runner.Spec.PodTemplate.Template.Spec.Size() > 0 {
+		runner.Spec.PodTemplate.Template.Spec.DeepCopyInto(&pod.Spec)
+	}
```
summerwind: I think it's better to first set `PodTemplate.Spec` to `pod.Spec` and then override the specific settings as follows:

```
podSpec := Spec: runner.Spec.PodTemplate.Template.Spec.DeepCopy()
pod := corev1.Pod{
  Spec: *podSpec, 
}

pod.Spec.RestartPolicy = "OnFailure"
...
```

However, how to override the `pod.Spec.Containers` field is still open for consideration...

mumoshu: [As similar as K8s already do in strategic-merge patch, we can merge on `name ` for containers](https://github.com/kubernetes/client-go/blob/7ac12361944b3b98074594cf00977861b9b41e4b/pkg/api/v1/types.go#L2119).

// Too bad we can't reuse the strategic-merge itself in our case, as it seems like for generating patches, rather than merging objects :(

alexandrst88: @mumoshu, I missed the fact that @summerwind posted in the Issuehttps://github.com/summerwind/actions-runner-controller/issues/5#issuecomment-599968643. So seems like I will rewrite code to use specific fields of podspec

mumoshu: According to the discussion made at #5, can we just defer implementing merge on `containers`?

alexandrst88: yeah. I will just go through fields, and assign them if not empty to current pod creation or merge additional fields like env :)

alexandrst88:
check this https://github.com/elastic/cloud-on-k8s/issues/1822, I've wrote about this https://github.com/summerwind/actions-runner-controller/issues/5#issuecomment-598324941
https://github.com/summerwind/actions-runner-controller/issues/5#issuecomment-599538842

alexandrst88:
@summerwind , @mumoshu I did redesign for spec, if you want to completely remove containers that are creating in code, you can just specify runnerContainers, and go to own risk.

if not just use all the fields that I've specified.

controllers/runner_controller.go
```diff
@@ -398,6 +406,63 @@ func (r *RunnerReconciler) newPod(runner v1alpha1.Runner) (corev1.Pod, error) {
...
+		pod.Spec.Containers = runner.Spec.RunnerContainers
+	}
+
+	if len(runner.Spec.VolumeMounts) != 0 {
+		pod.Spec.Containers[0].VolumeMounts = append(pod.Spec.Containers[0].VolumeMounts, runner.Spec.VolumeMounts...)
+	}
+
+	if runner.Spec.RestartPolicy == "" {
+		pod.Spec.RestartPolicy = "OnFailure"
+	}
```
summerwind: Since the Controller is performing a restart upon completion of the runner's Pod, I don't think RestartPolicy should be overridable.

If RestartPolicy is set to "Always", the Controller will not be able to restart the Runner Pod properly.

https://github.com/summerwind/actions-runner-controller/blob/23538d43b310736ca5e7ebdfaeefa086ae7e7001/controllers/runner_controller.go#L183-L193

controllers/runner_controller.go
```diff
@@ -285,8 +290,8 @@ func (r *RunnerReconciler) unregisterRunner(ctx context.Context, repo, name stri
 	return true, nil
 }
 
-func (r *RunnerReconciler) listRunners(ctx context.Context, repo string) ([]GitHubRunner, error) {
-	runners := []GitHubRunner{}
+func (r *RunnerReconciler) listRunners(ctx context.Context, repo string) (GitHubRunnerList, error) {
+	runners := GitHubRunnerList{}
```
summerwind: Could you tell me why you changed to Struct?

alexandrst88:
```
{
  "total_count": 2,
  "runners": [
    {
      "id": 23,
      "name": "MBP",
      "os": "macos",
      "status": "online"
    },
    {
      "id": 24,
      "name": "iMac",
      "os": "macos",
      "status": "offline"
    }
  ]
} 
```
Github had changed API,  as result controller can't unregister the runner

https://github.com/summerwind/actions-runner-controller/issues/11, due to

```2020-03-18T18:34:44.737Z	ERROR	controller-runtime.controller	Reconciler error {"controller": "runner", "request": "actions-runner-system/my-repo-runner", "error": "json: cannot unmarshal object into Go value of type []controllers.GitHubRunner"}```

File: controllers/runner_controller.go (hunk 3 above)
Comment (alexandrst88) [reply controllers/runner_controller.go]: got it

**Comment by alexandrst88** [comment]
updated PR.

File: api/v1alpha1/runner_types.go [hunk 5]
```diff
@@ -28,10 +28,44 @@ type RunnerSpec struct {
 	Repository string `json:"repository"`
 
 	// +optional
-	Image string `json:"image"`
+	RunnerContainers []corev1.Container `json:"runnerContainers,omitempty"`
```
mumoshu: `RunnerContainers` sounds like it replaces runner-specific containers only. I prefer just `Containers` so that it becomes:

```
	}
```

api/v1alpha1/runner_types.go
```diff
@@ -28,10 +28,44 @@ type RunnerSpec struct {
...
 	// +optional
 	Env []corev1.EnvVar `json:"env,omitempty"`
+
+	// +optional
+	Volumes []corev1.Volume `json:"volumes,omitempty"`
+
+	// +optional
+	InitContainers []corev1.Container `json:"initContainers,omitempty"`
+	// +optional
+	Containers []corev1.Container `json:"containers,omitempty"`
```
mumoshu: I prefer `SidecarContainers` for this one so that it becomes:

```
	if len(runner.Spec.SidecarContainers) != 0 {
		pod.Spec.Containers = append(pod.Spec.Containers, runner.Spec.SidecarContainers...)
	}
```

We're appending this to `pod.Spec.Containers` rather than replacing that, which makes me think this is better named after sidecars.

alexandrst88: I like this idea, thank you @mumoshu

alexandrst88: done

api/v1alpha1/runner_types.go
alexandrst88: done

alexandrst88:
@mumoshu fixed your review changes requests 🙂

controllers/runner_controller.go
summerwind: That's good to know. Thank you!

PR #8, opened 2020-03-15
add role for runnerreplicasets to fix error:

missedone:
Failed to list *v1alpha1.RunnerReplicaSet: runnerreplicasets.actions.summerwind.dev is forbidden: User "system:serviceaccount:actions-runner-system:default" cannot list resource "runnerreplicasets" in API group "actions.summerwind.dev" at the cluster scope

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -27,6 +27,26 @@ rules:
   - patch
   - update
 - apiGroups:
+    - actions.summerwind.dev
+  resources:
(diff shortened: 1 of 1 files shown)


summerwind:
Thank you for your controbution!

The RBAC is generated from the controller code comments, so the code needs to be modified. 
I fixed it in a897eee .

Please let me know if you have any problems. 
Thanks!

missedone:
ah, good to know, thx

PR #3, opened 2020-03-05
Manual test feedbacks

mumoshu:
I'd like to add two fixes for the issues I've observed in manual testing. See each commit message for more details.

In nutshell, `RunnerDeployment` just works as explained in README after these fixes!

PR #9, opened 2020-03-17
Readme fix requirements

jorge07:


summerwind:
Thanks!
This should be fixed in #3

PR #14, opened 2020-03-20
merge runner spec required Env to custom Pod spec

### LINKED ISSUE #13: Runner failed to start with custom Pod spec
with version 0.4.0 (PR #7), Runner failed to start with custom Pod spec
```yaml
kind: RunnerDeployment
metadata:
  name: my-runner
  annotations:
    iam.amazonaws.com/role: my-role
spec:
  replicas: 1
  template:
    spec:
      repository: example/repo
      containers:
        - name: runner
          image: example/custom-action-runner:latest
          imagePullPolicy: Always
```

Error log
> kubectl logs muy-runner-8gcph-v7plj -f
RUNNER_NAME must be set

a fix PR will follow up soon.

missedone:
fixes #13

### DIFF
--- a/controllers/runner_controller.go
+++ b/controllers/runner_controller.go
@@ -409,6 +409,13 @@ func (r *RunnerReconciler) newPod(runner v1alpha1.Runner) (corev1.Pod, error) {
 
 	if len(runner.Spec.Containers) != 0 {
 		pod.Spec.Containers = runner.Spec.Containers
+		for i := 0; i < len(runner.Spec.Containers); i++ {
+			// merge runner spec required Env to custom Pod spec
+			if runner.Spec.Containers[i].Name == containerName {
+				runner.Spec.Containers[i].Env = append(runner.Spec.Containers[i].Env, env...)
+				break
+			}
+		}
 	}
 
 	if len(runner.Spec.VolumeMounts) != 0 {


controllers/runner_controller.go
```diff
@@ -409,6 +409,13 @@ func (r *RunnerReconciler) newPod(runner v1alpha1.Runner) (corev1.Pod, error) {
 
 	if len(runner.Spec.Containers) != 0 {
 		pod.Spec.Containers = runner.Spec.Containers
+		for i := 0; i < len(runner.Spec.Containers); i++ {
+			// merge runner spec required Env to custom Pod spec
+			if runner.Spec.Containers[i].Name == containerName {
+				runner.Spec.Containers[i].Env = append(runner.Spec.Containers[i].Env, env...)
+				break
+			}
+		}
```
mumoshu: Thanks a lot for the PR!

This is good functionality-wise, but for consistency, you'd better add the dynamic env to the first container:

```suggestion
		// merge runner spec required Env to custom Pod spec
		c := runner.Spec.Containers[0].DeepCopy()
		c.Env = append(c.Env, env...)

		pod.Spec.Containers = append(pod.Spec.Containers, *c)
		pod.Spec.Containers = append(pod.Spec.Containers, runner.Spec.Containers[1:]...)
```

It's okay as we assume that the first container must be that of the runner as seen in e.g. https://github.com/summerwind/actions-runner-controller/blob/d4c849ee091076d233e14217d09e5f014ba8efa8/controllers/runner_controller.go#L206-L208

mumoshu:
cc/ @summerwind We gonna fix it this way. Without this, I believe `RunnerSpec.Containers` is currently unusable because three's no way for the user to specify the required `env` externally.

controllers/runner_controller.go
missedone: it's is not OK to assume the first container is the runner!!
so that's why I have `if runner.Spec.Containers[i].Name == containerName {` to make sure the runner specific Env gets injected to the runner container.

missedone: this if statement is still not ideal, however, the code for now has hard dependency on container named `runner`, so i'd like to leave it as it is now.
but would like to refactor to make it the runner container name configurable via annotation.

mumoshu: So do you have strong opinion on we should treat either the first container or the container named `runner` as the runner container?

We already assume that it's the first container in https://github.com/summerwind/actions-runner-controller/blob/d4c849ee091076d233e14217d09e5f014ba8efa8/controllers/runner_controller.go#L206-L208. So if you're changing the assumption, you'd need to change that as well with explanation and benefits.

Mixing those two in the same code base seems worse to me.

mumoshu: > would like to refactor to make it the runner container name configurable via annotation.

What's the benefit of making the runner container name customizable?

mumoshu: Also, In Kubernetes we usually use the first container as the primary. Based on that assuming the first container as the runner doesn't involve any actual problem to me.

I.e. is there any real use-case that you want a runner pod to have the runner container as a sidecar?

mumoshu: Also - as we've discussed in #5, the point of supporting only a subset of the pod spec fields is to avoid this kind of confusion.

With the subset, you can just use `spec.image` and maybe `spec.imagePullPolicy` instead of trying hard with `spec.containers` for customizing the runner container.

missedone: > We already assume that it's the first container in

i would not make assumption like this. the manifest is designed for human, the application breaks if people has different assumption.

so as i said, i would have follow up PR to address this, if you still welcome...

mumoshu: I always welcome any contribution if it has any positive effect! That's why I'm asking why you prefer `spec.containers[]` for your specific use-case.

But the point here is that you're trying to do two things altogether. Can we split it into two?

First, the fix for #13 can be done without changing our assumption this way. I'd rather like to fix #13 with https://github.com/summerwind/actions-runner-controller/pull/14#discussion_r395940542.

Second, create a dedicated github issue to discuss about enhancing `spec.containers[]` and other things.

mumoshu: https://github.com/summerwind/actions-runner-controller/issues/15 is the issue.

summerwind:
The original code relied on the container name to use to restart the pod. However, I think it's a good idea to assume that the first item in the `containers` field is the 'Runner' container, since it's now possible to override the pod's containers field.

callum-tait-pbx:
@mumoshu PRs are included by the stale bot if any are they are going to hang around need one of the exempt labels.

toast-gear:
The project has changed substantially since this PR was raised, let me close this due to inactivity and conflicts, but feel free to restart another pull request. Thanks anyway!

PR #18, opened 2020-03-25
Clarify instructions for getting a token

acj:
Finding a token was confusing the first time -- is it a personal access token, or something else? But the process is straightforward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ NAME           READY   STATUS    RESTARTS   AGE
 example-runner 2/2     Running   0          1m
 ```
 
-The runner you created has been registerd to your repository.
+The runner you created has been registered to your repository.
 
 <img width="756" alt="Actions tab in your repository settings" src="https://user-images.githubusercontent.com/230145/73618667-8cbf9700-466c-11ea-80b6-c67e6d3f70e7.png">
 
--- a/README.md
+++ b/README.md
@@ -16,7 +16,9 @@ First, install *actions-runner-controller* with a manifest file. This will creat
 $ kubectl apply -f https://github.com/summerwind/actions-runner-controller/releases/latest/download/actions-runner-controller.yaml
 ```
 
-Set your access token of GitHub to the secret. `${GITHUB_TOKEN}` is the value you must replace with your access token. This token is used to register Self-hosted runner by *actions-runner-controller*.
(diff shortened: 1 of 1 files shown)


mumoshu:
@acj Hey! Thanks for this. But it should be a personal access token.

Please see https://github.com/summerwind/actions-runner-controller/blob/d4c849ee091076d233e14217d09e5f014ba8efa8/controllers/runner_controller.go#L251 - the whole purpose of this project is to automate creation of registration tokens in favor of you, and your personal access token, so that you don't have to.

I obtained my personal access token from https://github.com/settings/tokens, giving it `repo` and `workflow` scopes.

If that works for you as well, I think this part of the doc can be better updated to note that where to create the token and what scopes you should provide.

acj:
Ah, my mistake. Fixed.

I'm not having any luck with a PAT that has repo and workflow scopes either:

```
Events:
  Type     Reason                         Age                  From               Message
  ----     ------                         ----                 ----               -------
  Warning  FailedUpdateRegistrationToken  0s (x21 over 6m37s)  runner-controller  Updating registration token failed
```

I'll keep trying and will open a separate issue if I can't figure it out. Thanks for the fast reply!

summerwind:
@acj Thank you for the contribution!

To issue a Registration Token, the user who issued the personal access token must have **admin** permission for the repository. I think that only *repo* scope is required for personal access token.

acj:
Thanks, @summerwind. I updated the text to reflect that.

This can be done separately, but if the pods need to be privileged then we may want to note that in the readme as well:

>2020-03-25T12:04:14.250Z        ERROR   controller-runtime.controller   Reconciler error        {"controller": "runner", "request": "default/repo-runners-gdb69-6br77", "error": "Pod \"repo-runners-gdb69-6br77\" is invalid: spec.containers[1].securityContext.privileged: Forbidden: disallowed by cluster policy"}

mumoshu:
@acj Thanks for the update 👍 

> the pods need to be privileged then we may want to note that in the readme as well:

Sounds good. Would providing readers an example PodSecurityPolicy to allow priv pods would also help?

acj:
> Would providing readers an example PodSecurityPolicy to allow priv pods would also help?

@mumoshu Yes, I think that's a good idea.

Thanks, all, for the fast feedback and for building this controller :)

Issue #22, opened 2020-03-31
Docker caching


vitobotta:
Hi! I just installed this and am happy I can delete the dedicated VPS for my runner. I am adapting my workflows, but I am having problems with Docker builds, because it's not caching image layers. I exec'ed into the runner and saw that after each build `docker images` returns nothing. 

What can I do to prevent each build from starting from scratch?

Thanks!

vitobotta:
Also, it seems I cannot run commands with sudo?

mumoshu:
> Also, it seems I cannot run commands with sudo?

Probably we need to add the `runner` user to sudoers?

https://github.com/summerwind/actions-runner-controller/blob/b8b4ef4b60a45fe647a19a6f56130eb3449a9a99/runner/Dockerfile#L25

Would you mind giving it a shot? I'm more than happy yo review if you could submit a PR for that!

mumoshu:
> it's not caching image layers

This is by-design atm. The `--once` flag we're consistently passing in the runner entrypoint [here](https://github.com/summerwind/actions-runner-controller/blob/f2d3ca672f5a5a2c9deba852eab2286a5a3984e6/runner/entrypoint.sh#L22) instructs the runner to stop/restart from the scratch on each build.

Perhaps we can have an additional field in the Runner CRD to make it configurable.

Also - `--once` is preferable for reproducible builds. So an alternative way would be to deploy an in-cluster docker image registry for caching images, from where you can pull the image on each build.

Which way do you prefer?

vitobotta:
Hi, I can send a small PR for the user thing.

As for the caching, do you mean `--cache-from`?

mumoshu:
Yeah - More concretely, I believe adding your docker-build a `--cache-from` to a in-cluster docker registry, and adding `docker push` against the in-cluster registry in the very end of your build would help reducing the build time.

vitobotta:
OK, I will try installing Harbor in my cluster and use that. Thanks for the suggestion!

vitobotta:
I've been using Harbor as in cluster registry and --cache-from. It's not as fast as using the local cache due to the pull required, but it's a lot better than building from scratch. Thanks a lot again, this project saved me the time to do something similar :)

PR #23, opened 2020-03-31
Add runner user to sudoers and bump Docker version

vitobotta:
Hi, a small PR as discussed in #22. Thanks

### DIFF
--- a/runner/Dockerfile
+++ b/runner/Dockerfile
@@ -4,14 +4,16 @@ ARG RUNNER_VERSION
 ARG DOCKER_VERSION
 
 RUN apt update \
-  && apt install curl ca-certificates -y --no-install-recommends \
+  && apt install sudo curl ca-certificates -y --no-install-recommends \
   && curl -L -o docker.tgz https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_VERSION}.tgz \
   && tar zxvf docker.tgz \
   && install -o root -g root -m 755 docker/docker /usr/local/bin/docker \
   && rm -rf docker docker.tgz \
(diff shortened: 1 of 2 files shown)


mumoshu:
@vitobotta Thanks! I'll review this asap.

@summerwind Should we add a field to the Runner spec to make this configurable? WDYT?

vitobotta:
Ehm... I tested this image locally and I got the user in the sudoers group etc. But how can I use my image for the runner now, while this gets merged? I need to be root to investigate some issues, especially networking related. There seems to be a networking issue with Docker in Docker, and often my build fails in the runner because it cannot connect to some hosts to download packages etc. Any idea?

vitobotta:
I have found some references lke https://discuss.istio.io/t/connection-issues-with-docker-build-inside-a-kubernetes-pod/5761 where they say that using the host network would solve these networking issues. Is this possible?

vitobotta:
I solved the networking issue.... but I still don't know how to use my image for the runner :(

mumoshu:
Please specify the `image` field in your runner to the name of the image you've docker-pushed https://github.com/summerwind/actions-runner-controller/blob/master/api/v1alpha1/runner_types.go#L40

vitobotta:
@mumoshu   It worked, thannks!

summerwind:
Thanks for the PR! 

In the GitHub's runner environment allows us to use `sudo` with no password. So, this change looks good to me 👍 
See also: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#administrative-privileges-of-github-hosted-runners

In the future, it seems better to make the file system of the Runner container image the same as GitHub's runner environment.

PR #27, opened 2020-04-10
Bump golang to v1.14.2

chenrui333:


summerwind:
Thank you for the PR!
I think we need to stay on Go 1.13 to avoid known problems until kubernetes uses golang 1.14.
https://github.com/kubernetes/kubernetes/pull/88638

chenrui333:
@summerwind I dont think that is an issue, I actually already applied this change in my fork, https://github.com/chenrui333/actions-runner-controller/releases/tag/v0.4.5 and it works well for me.

Let me know if that makes more sense.

summerwind:
Thanks for testing with golang 1.14.

I'm concerned that the many Kubernetes-related packages that this controller depends on will work properly in go 1.14. Since many of the Kubernetes-related packages have not yet been migrated to golang 1.14, we'll keep golang 1.13 for a while.

When the Kubernetes-related packages are migrated to golang 1.14, this controller will be migrated as well.

I'm going to close this PR. Thanks!

PR #28, opened 2020-04-10
Update defaultRunnerImage to v2.168.0

chenrui333:
Good to have runner by default running the latest version.

### DIFF
--- a/main.go
+++ b/main.go
@@ -38,7 +38,7 @@ import (
 )
 
 const (
-	defaultRunnerImage = "summerwind/actions-runner:v2.165.2"
+	defaultRunnerImage = "summerwind/actions-runner:v2.168.0"
 	defaultDockerImage = "docker:19.03.6-dind"
 )
 


config/default/manager_auth_proxy_patch.yaml
```diff
@@ -23,3 +23,4 @@ spec:
         args:
         - "--metrics-addr=127.0.0.1:8080"
         - "--enable-leader-election"
+        - "--runner-image=summerwind/actions-runner:v2.168.0"
```
summerwind, line 26: If `defaultRunnerImage` is updated, you don't need to specify this option. Could you delete this line please?

chenrui333: Yeah, I was struggling a little bit with how to do the setting, maybe we should have it in the documentation somewhere.

With that being said, I think I can remove this line.

chenrui333:
@summerwind should be good to go 😄

config/default/manager_auth_proxy_patch.yaml
summerwind: I agree with you. I will add usage section to the document, thanks!

PR #36, opened 2020-04-16
Update runner to v2.169.1

summerwind:
This PR updates runner container image to v2.169.1. This also changes the tag of the container image that the controller executes so that it always uses the latest.


chenrui333:
Cool!

PR #38, opened 2020-04-17
Update runner version to v2.169.0

chenrui333:
relates to https://github.com/actions/runner/pull/435


summerwind:
Thank you for the PR!
I think this has been fixed in #36.

Issue #32, opened 2020-04-14
Issue with running `docker run` for integration testing


chenrui333:
I have run into some issue with running `docker run xx-image` after moving some of the workflow into self-hosted runner. I could not figure it out, but feel it might be helpful to post in here and discuss.

The flow looks like something below:
```
docker run -it --rm \
	-v $(PWD):/local \
	xx-image:latest \
	/local/e2e.sh
```

The error log that i keep getting:

```
docker: Error response from daemon: OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"/local/e2e.sh\": stat /local/e2e.sh: no such file or directory": unknown.
```

While it works before, I suspect it might be related to the `dind` docker image it used for the runner. Does the dind docker engine impact the volume mount?? 

Please shed some insight on fixing the issue and let me know if this makes sense. Thanks!!

mumoshu:
@chenrui333 Ah yes, the dind container and the runner container needs explicit volume & volume mounts configured for that to work.

Could you configure `spec.volumes` and `spec.volumeMounts` appropriately in your runner/runnerreplicaset/runnerdeployment spec?

If you could share me a few examples of paths returned by `$(PWD)`, I can guide you how you should configure volumes/volumeMounts.

FYI, the mapping between the runner spec and the runner pod volumes/mounts is implemented in https://github.com/summerwind/actions-runner-controller/blob/b8b4ef4b60a45fe647a19a6f56130eb3449a9a99/controllers/runner_controller.go#L416.

chenrui333:
Let me check on my environment really quick, thanks for clarifying the mystery.

chenrui333:
@mumoshu any advice or do you need more info from my side?

mumoshu:
Ah, that's too bad. We can't share `/` between the runner container and the find container cuz it breaks both sides. We need to change either the runner side or your workflow to use another directory.

I thought it was `/github/workspace` in the managed GitHub Actions runner.

Would you mind running `find -maxdepth 3 /github/workspace` in your workflow to figure out what's in your workflow env currently?

Depending on that, we can perhaps change the runner container to create necessary directories and set proper workdir so that you can just mount $(PWD) to docker containers.

chenrui333:
I dont think I have `/github/workflow` path there.

Here is the debug output:
```
Run find -maxdepth 3 /github/workspace
  find -maxdepth 3 /github/workspace
  shell: /bin/bash -e {0}
[...]

PR #40, opened 2020-04-21
Use systemd as entrypoint in runner container

summerwind:
This PR allows us to use of systemd-based runner container.

We've been running `dockerd` as a sidecar container as an initial implementation. However, it does create problems like #32 and constraints on the sidecar configuration of the Runner Pod. Also, the current self-hosted runner is implemented on the assumption that it will run on a virtual machine instance.

So, this PR allows us to use the systemd-based runner container image, which allows us to run Runner in an environment closer to a virtual machine instance. This image also run `dockerd` inside the runner container, which solves docker-related problems. The runner pod will also be consolidated into a single container, making it easier to customize the sidecar.

This approach may cause several problems, one of which is that systemd hides the runner's error so that the controller can't detect it. This problem may be solved in the future by continuously executing command inside the container using Liveness Probe.

The other issue is that the container cannot be stopped when the runner has finished running. This problem was solved by sending a signal to systemd to stop the container when it stopped successfully.

It may cause some problems, but overall it seems to be working properly in my environment.


summerwind:
@mumoshu Could you please take a look at this PR and give your feedback?

mumoshu:
@summerwind Thanks for your efforts! I believe this is a very interesting and practical approach to the problem.

What I'm not feeling good about though is that, the hard dependency to systemd. I know it may be a necessary evil but it seemed like an overly complex init system for our use-case.

Are there any simple init-like system, hopefully written in our used language(golang), that provides enough functionality for us?

I'd wish for one that can support the followings:

- Ability to stop the runner process when the dockerd stopped(so that an "incomplete" runner pod immediately stops without trying run any workflow/job run to fail in the middle of the process(=running docker client, which obviously fail due to the stopped dockerd
- Ability to not start the runner process(to let the whole pod recreated) when the dockerd failed to start
- Ability to propagate/coordinate signals like SIGTERM

controllers/runner_controller.go
```diff
@@ -171,15 +171,15 @@ func (r *RunnerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
-
-		if pod.Status.Phase == corev1.PodRunning {
+		if pod.Status.Phase == corev1.PodFailed {
 			for _, status := range pod.Status.ContainerStatuses {
 				if status.Name != containerName {
 					continue
 				}
 
-				if status.State.Terminated != nil && status.State.Terminated.ExitCode == 0 {
+				if status.State.Terminated != nil && status.State.Terminated.ExitCode == 130 {
```
mumoshu, line 180: Can we have a reference to any doc/sentence that helps making the sense of this number `130`?
I did a quick googling for `systemd exit status` and `systemd 130` but had no luck. So perhaps this is a different thing?

runner/Dockerfile
```diff
@@ -1,29 +1,40 @@
...
+
+RUN cp /tmp/files/entrypoint.sh /usr/local/bin/entrypoint \
+  && cp /tmp/files/actions-runner.sh /usr/local/bin/actions-runner \
+  && cp /tmp/files/actions-runner-halt.sh /usr/local/bin/actions-runner-halt \
+  && cp /tmp/files/actions-runner.service /etc/systemd/system/actions-runner.service \
+  && systemctl enable actions-runner.service \
+  && rm -rf /tmp/files /var/log/* /tmp/* /usr/share/doc* /usr/share/man/* /usr/share/local/*
 
-COPY entrypoint.sh /runner
+STOPSIGNAL 37
```
mumoshu, line 38: Aha! Neat - I've never realized that the existence of this operation in Dockerfile.

I'm still unsure how this works, though.

I thought that this instructs the container to send SIGALRM(37) to the init process (PID 1 = /sbin/init = systems) on `docker stop`, which might be called via a pod deletion.

But then, who responds to the signal? I "guess" systemd would respond to SIGTERM for stopping. Is it able to respond to 37 as well?

// I read that you've intended to send 37 from within the container itself via the runner unit on post stop using `actions-runner-halt.sh`. Impressive! But I'm still not sure who actually respond to 37 and how.

mumoshu:
> Are there any simple init-like system, hopefully written in our used language(golang), that provides enough functionality for us?

I had a quick googling on this and gave up :) I'm now convinced and feeling free to go ahead with systemd.

mumoshu:
@summerwind Do you think we can add a few `jounalctl -f` containers so that we can easily view logs of important processes via `kubectl logs`?

I mean I would like to have:

- `journalctl -f actions-runner` in a container called e.g. `runner`
- `journalctl -f dockerd` (Or was it `journalct -f docker-daemon`? Not sure) in e.g. `dockerd`

so that we can use `kubectl logs -f $RUNNER_POD -c runner` to view the runner log and `kubectl logs -f $RUNNER_POD -c dockerd` to view the dockerd log.

Thoughts?

controllers/runner_controller.go
summerwind: To be honest, I'm not sure about that either. systemd is always started as PID 1, so theres no way to know systemd's exit code without reading the code. Exit code `130` is based on the pod status after executing `kill 37 1` in the Pod.

runner/Dockerfile
summerwind: systemd is responsible for this signal in this case.

`STOPSIGNAL 37` indicates the signal to be sent when the container is stopped. The container runtime will use this signal to the container's main process (PID 1, systemd) when it stops the container.

The reason why we use `37 (SIGRTMIN+3)` is based on systemd's container interface document.
https://systemd.io/CONTAINER_INTERFACE/#advanced-integration

> If the container manager wants to cleanly shutdown the container, it might be a good idea to send SIGRTMIN+3 to its init process. systemd will then do a clean shutdown. Note however, that since only systemd understands SIGRTMIN+3 like this, this might confuse other init systems.

summerwind:
@mumoshu Thank you for your feedback!

> What I'm not feeling good about though is, the hard dependency to systemd. I know it may be a necessary evil but it seemed like an overly complex init system for our use-case.

This is totally agree. Putting runner and dockerd under systemd made it difficult to observe the state of the process from Kubernetes. This will create complexity on the controller side.

> Are there any simple init-like system, hopefully written in our used language(golang), that provides enough functionality for us?

Actually, I also looked into more lightweight init systems, such as supervisord or upstart. But the easiest way to get Docker to work was with Systemd. I know how you feel, but right now I think it's best to use systemd.

> Do you think we can add a few jounalctl -f containers so that we can easily view logs of important processes via kubectl logs?

That's a good idea! I also thought it would be nice to output journald's log as pod's log. However, if we add a sidecar, we lose the benefit of removing the sidecar from this change.

summerwind:
Rethinking based on the feedback, this change seemed like a not-so-good idea because it would make the internals of the pod more complex.

Would it be simpler to give up the sidecar simplicity this time and simply share the woking directory between `docker` sidecar and `runner` container?

summerwind:
I created PR #45 to simply solve the problem.

summerwind:
I'm going to close this. I'll revisit the sidecar integration at some time.
@mumoshu Thank you for reviewing and giving feedback!

PR #44, opened 2020-04-24
support for organization runners & labels

### LINKED ISSUE #26: Associate runners with multiple repos
I wonder if there is any way that we can associate action runners with multiple repos. 

(this would also allow to reduce the committed resource usage)

reiniertimmer:
Closes #26 

Add support for runners on the organization level, in addition to runners per repository.

In addition, this supports adding labels to runners so the organization runners can be used effectively. So you can select a specific set of runners by using `runs-on: my-label` to distinguish between the type of self-hosted runner that will be used

**Note:** this adds a _required_ organization field, in addition to the (now optional) repository field. While a relatively clean solution, this does break compatibility with previous versions. I'm up for suggestions on this part if this is not desired

I implemented API calls that are missing from `go-github` but I am thinking of adding the implementation there (if they're up for it) so we have a bit of a cleaner solution (though it would be mostly identical to this)


mumoshu:
@reiniertimmer Thanks! At glance this looks sooo nice.

I'll take a deeper look tonight.

Until then, would you mind sharing me how you tested or could test this?

Example:

- Create a personal access token of a github user who has access to the specific organization(?)
- Put `github.token` in a K8s secret
- Create a RunnerDeployment

Based on your info I'll test this myself, manually, so that we are sure that this doesn't break anything and only add benefits!

reiniertimmer:
Of course, manual testing is still underway here, though so far all looks to be fine.

How I tested it:
- make sure you have owner permissions to some organization
- create a personal access token in github with scope `admin:org`
- generate & install the updated manifests: `make install`
- run the controller locally: `GITHUB_TOKEN=<YOUR_TOKEN> make run`

Now you are ready to create organization runners using the `organization` field only, as well as repository runners (specify both `organization` and `repository` in the spec)

Note that the existing docker image does not support organizations yet, so I am using a custom image at the moment (for testing you could use this one, but I suggest you generate a new one using the updated `entrypoint.sh`):

```
spec:
  image: reiniert/actions-runner:v2.169.1
  organization: your-org-name
```

and 

```
spec:
  image: reiniert/actions-runner:v2.169.1
  organization: your-org-name
  repository: your-repo-name
```

runner/entrypoint.sh
```diff
@@ -5,8 +5,14 @@ if [ -z "${RUNNER_NAME}" ]; then
...
-if [ -z "${RUNNER_REPO}" ]; then
-  echo "RUNNER_REPO must be set" 1>&2
+if [ -n "${RUNNER_ORG}" -a -n "${RUNNER_REPO}" ]; then
+  ATTACH="${RUNNER_ORG}/${RUNNER_REPO}"
+elif [ -n "${RUNNER_ORG}" ]; then
+  ATTACH="${RUNNER_ORG}"
+elif [ -n "${RUNNER_REPO}" ]; then
+  ATTACH="${RUNNER_REPO}"
+else
+  echo "At least one of RUNNER_ORG or RUNNER_REPO must be set" 1>&2
```
reiniertimmer, line 15: This would allow to work both for old and new controllers (ie. those that do support organizations and those that don't) so you can keep using tag 'latest' as default in `main.go`

api/v1alpha1/runner_types.go
```diff
@@ -24,8 +24,15 @@ import (
 // RunnerSpec defines the desired state of Runner
 type RunnerSpec struct {
 	// +kubebuilder:validation:MinLength=3
-	// +kubebuilder:validation:Pattern=`^[^/]+/[^/]+$`
-	Repository string `json:"repository"`
+	// +kubebuilder:validation:Pattern=`^[^/]+$`
+	Organization string `json:"organization"`
+
+	// +optional
+	// +kubebuilder:validation:Pattern=`^[^/]*$`
```
summerwind, line 31: I think this would be a breaking change. The current user has set this field to a value such as "owner/repository". However, applying this change would make such a value invalid. Can you change it so that it remains compatible?

reiniertimmer: I totally agree. But what would be the best way then?

We could make both "optional", because you could either specify organization, repository or both. However, then there's no validation anymore on these fields - so you can make a runner without either of them and the resource will be applied (though the controller will fail).

Maybe the best option is then to allow the change, but let the controller set it to failed state if it's missing both the organization and the repository

Any thoughts?

reiniertimmer:
FYI: organization API has been merged into `go-github` as well https://github.com/google/go-github/pull/1506

So as soon as we can use that `github_beta.go` can be removed from this PR

api/v1alpha1/runner_types.go
summerwind: I agree with your thought.  I also think that if both `organization` and `repository` were not specified, the controller would need to set this resource to the failed state too.

There are four resource definitions that we should consider.

### Organization Runner Mode

With this manifest, the controller starts an organization runner.

```
spec:
  organization: summerwind
```

### Respository Runner Mode

With this manifest, the controller starts a repository runner.

```
spec:
  organization: summerwind
  repository: actions-runner-controller
```

### Compat Mode

With this manifest, the controller also starts a repository runner.

```
spec:
```

### Invalid resource

This is an invalid resource.

```
spec:
  organization: 
  repository: 
```

If such a resource is detected, I think it is better for the controller to set `ValidationError` to the runner state as follows. For a runner resource with a "ValidationError", the controller will not start the runner pod.

```
status:
  phase: ValidationError
  reason: "InvalidRepository"
  message: ""
```

I think that such a process should be validated using Validation Webhook, but the scope of the change would be larger, so it would be better to deal with it in a separate PR.

reiniertimmer: The webhook is an interesting idea, but probably a bit too much for now.

We could also consider phasing out the compat mode once you move a couple versions up (deprecate it from this version on) or when you move beyond v1alpha or something

I will log a meaningful error in the meantime. Also I will update the readme with the new preferred way of using the runner spec

summerwind: It's a good point to consider a phased out. The plan I think as follows:

We will release v0.5.0 that includes this changes. At the same time, we will mark the compat mode configuration as deprecated. Then remove the compat mode support in v0.6.0 completly.

reiniertimmer: I guess that's the cleanest solution.

In that case, I am thinking about renaming `organization` to `owner` in the RunnerSpec, because this mode has to be used for private repositories as well (and having `organization` suggests that the field is only usable for organizations - not for personal repositories)

reiniertimmer: However, the more I think about it, the more options there are.

1. allow `organization` for org, and keep `repository` as-is for repository mode (so either allow one or the other, but _not_ both). This is very explicit about the intent of the runner and backwards compatible (no need for a phase out) - but there will be a gotcha for users that supply both
2. introduce `owner` (instead of organization) and repository (both optional, but one should be supplied). Repository will then have a different meaning (and a different format) depending on if there is an owner or not. Basically, this is how it's implemented now, except for naming (organization implies that this is only for organizations, though it's really for both)

Any maybe even more..? 

Only issue is that all scenarios that I could think of have the same downside: spec validation cannot be done, because we have two fields that depend on each other (this would require a webhook to implement fully, but that's a different story)

@summerwind I'm leaning a bit towards option 1. now (the most explicit and most similar to current implementation), but I'd like you to have the final say in the matter. Or maybe there are even better options that I haven't though of?

summerwind: Thank you for your consideration!
I would choose option 1. This is the simplest choice right now.

I don't think we should introduce `owner`. I feel `owner` is an abstract and confusing word because users choose Organization-Level or Repository-Level, and I think `owner` is an expression for implementers.

reiniertimmer: Agreed :)

I'll implement option 1

github/github_beta.go
```diff
@@ -0,0 +1,95 @@
+package github
```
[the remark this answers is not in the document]
reiniertimmer: It's now merged in `go-github`, but since there's no release yet, I could update go.mod to use a tag instead of the last release if you prefer (so this file can be removed)

api/v1alpha1/runner_types.go
reiniertimmer: Ok, implementation is done, you can check if you like

reiniertimmer:
The pleasure was all mine! This is a very useful project for us so I am happy I could contribute

I ran into some small issues I was considering to fix as well, but I'll report them as separate issues then

chenrui333:
Looking forward to trying this. ❤️

PR #45, opened 2020-04-24
Share runner's working directory with docker sidecar

### LINKED ISSUE #32: Issue with running `docker run` for integration testing

summerwind:
This PR changes the runner container's working directory to be shared with the docker sidecar.
This will also fix #32.

### DIFF
--- a/controllers/runner_controller.go
+++ b/controllers/runner_controller.go
@@ -282,6 +282,10 @@ func (r *RunnerReconciler) newPod(runner v1alpha1.Runner) (corev1.Pod, error) {
 					EnvFrom:         runner.Spec.EnvFrom,
 					VolumeMounts: []corev1.VolumeMount{
 						{
+							Name:      "work",
+							MountPath: "/runner/_work",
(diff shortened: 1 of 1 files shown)


summerwind:
@mumoshu Thank you!

PR #47, opened 2020-04-29
Fix possible runner stuck in Terminating on various configuration errors

reiniertimmer:
When there is an invalid repository / organization in the pod spec a runner cannot be deleted, and will remain in Terminating state.

This is because the controller attempts (and fails) to remove the runner at a GitHub repository/organization that does not exist, and stops after this error. In this PR we check if the runner was ever registered before (if it has a token) and don't attempt to remove if the runner was never registered before.

Edit: clarifications :)

### DIFF
--- a/controllers/runner_controller.go
+++ b/controllers/runner_controller.go
@@ -88,16 +88,20 @@ func (r *RunnerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		}
 	} else {
 		finalizers, removed := removeFinalizer(runner.ObjectMeta.Finalizers)
-
+		
 		if removed {
-			ok, err := r.unregisterRunner(ctx, runner.Spec.Organization, runner.Spec.Repository, runner.Name)
-			if err != nil {
-				log.Error(err, "Failed to unregister runner")
-				return ctrl.Result{}, err
-			}
+			if len(runner.Status.Registration.Token) > 0 {
+				ok, err := r.unregisterRunner(ctx, runner.Spec.Organization, runner.Spec.Repository, runner.Name)
+				if err != nil {
+					log.Error(err, "Failed to unregister runner")
+					return ctrl.Result{}, err
+				}
 
-			if !ok {
-				log.V(1).Info("Runner no longer exists on GitHub")
+				if !ok {
+					log.V(1).Info("Runner no longer exists on GitHub")
+				}
+			} else {
+				log.V(1).Info("Runner was never registered on GitHub")
 			}
 
 			newRunner := runner.DeepCopy()
@@ -108,7 +112,7 @@ func (r *RunnerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 				return ctrl.Result{}, err
 			}
 
-			log.Info("Removed runner from GitHub", "repository", runner.Spec.Repository)
+			log.Info("Removed runner from GitHub", "repository", runner.Spec.Repository, "organization", runner.Spec.Organization)
 		}
 
 		return ctrl.Result{}, nil


controllers/runner_controller.go
```diff
@@ -88,16 +88,20 @@ func (r *RunnerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
-			ok, err := r.unregisterRunner(ctx, runner.Spec.Organization, runner.Spec.Repository, runner.Name)
-			if err != nil {
-				log.Error(err, "Failed to unregister runner")
-				return ctrl.Result{}, err
-			}
+			if len(runner.Status.Registration.Token) > 0 {
+				ok, err := r.unregisterRunner(ctx, runner.Spec.Organization, runner.Spec.Repository, runner.Name)
+				if err != nil {
+					log.Error(err, "Failed to unregister runner")
+					return ctrl.Result{}, err
```
reiniertimmer, line 97: Alternative (or additional) we could even continue on this error - just continue to remove the runner regardless of any failures.

mumoshu:
@reiniertimmer Thanks for the PR!

Still trying to understand this:

> This is because the controller attempts (and fails) to remove the runner at GitHub, and stops after this error.

How/Why does it stop, you think?

I thought we already handled that case gracefully with this guard:

https://github.com/summerwind/actions-runner-controller/blob/4c029f73ee366b2ac97373fe48387efae65bfcdc/controllers/runner_controller.go#L242-L244

I mean, unregisterRunner should just return `false, nil` when the list runners call is successful AND there was no runner registered yet.

mumoshu:
> When there is some error in the pod spec (like invalid repository URL) a runner cannot be deleted.

So maybe this observation was a bit incorrect /Perhaps your patch rather fix other things?

For example - I guess - this rather seem to prevent the runner resource/pod from hanging in Terminating state when the API token(not runner token) was invalid. In that case even the list runners API call should fail, and this patch will help skipping the API call. Am I correct?

reiniertimmer:
Hi @mumoshu 

Yes, that is true. The only problem I've been having is when the calls to list runners is _not_ successful (for example, when your token does not have access to the repo/org, or there's a typo in the url, etc). In that case, the controller reports an error and does not continue

https://github.com/summerwind/actions-runner-controller/blob/master/controllers/runner_controller.go#L93-L97

Maybe this is a bit of a hypothetical case, but it happened to me a lot already :)

reiniertimmer:
> For example - I guess - this rather seem to prevent the runner resource/pod from hanging in Terminating state when the API token(not runner token) was invalid. In that case even the list runners API call should fail, and this patch will help skipping the API call. Am I correct?

Ah, yes, sorry for the confusion. The _API_ token is unable to list runners (because either there's no access or the repo simply does not exist).

mumoshu:
Ahh, okay, thanks a lot for the clarification!

That does make sense.

So, indeed a list-runners call can persistently fail in various ways, on invalid token, repo, orgs, etc. As long as `len(runner.Status.Registration.Token) > 0` is true, the token, repo and org settings should be correct and the runner registration is more likely to work. With this patch, we try to unregister a runner when and only when it is likely to succeed eventually, so that it won't stuck in Terminating. Great!

controllers/runner_controller.go
mumoshu: Perhaps it can leave the github runner remain registered and never removed when there was a transient error, for example when GitHub API limit is met?

reiniertimmer:
Exactly!

Check if the _registration_ token exists beforehand is a best-effort way to prevent this.

It's maybe not the most subtle way to determine if remove will succeed (it may still fail, so the code could also check for various error codes from GitHub) but that would include a lot more code changes. And there's a lot of cases to be considered then

mumoshu:
@reiniertimmer Thanks! Yes, this should be much better than just leaving the stale runner resources when the cause of the error is obvious.

PR #50, opened 2020-04-30
Add validation webhooks

summerwind:
This PR adds and enables validation webhooks for custom resources. This webhook can be used to prevent the addition of resources that have invalid `organization` and `repository` fields.

To enable validation webhooks, cert-manager is required for webhook endpoint certificate management. Therefore, this PR also updates the installation instruction.


summerwind:
@mumoshu It would be great if you could review this when you have time. Thanks as always!

api/v1alpha1/runner_webhook.go
```diff
@@ -0,0 +1,84 @@
+/*
```
mumoshu, line 1: Is this file really supposed to be located in this `api/v1alpha1` package? I thought this better fit within somewhere else, e.g. `webhook` pkg.

summerwind: This is due to the behavior of `kubebuilder create webhook` command. I was a little uncomfortable with this at first, but I think it's acceptable because Webhook behavior can vary depending on the API version of the resource ;-)

mumoshu: That made total sense. Thanks for clarifying!

summerwind:
@mumoshu Thanks for the review!

dudicoco:
@summerwind what is the mutating webhook used for?

mumoshu:
@dudicoco I think it's no-op right now https://github.com/summerwind/actions-runner-controller/blob/acb1700b7ceb286e0095a9dd5aef428323460471/api/v1alpha1/runner_webhook.go#L43

dudicoco:
Thanks @mumoshu.

PR #52, opened 2020-05-01
Add dependencies from GitHub virtual environment to runner

### LINKED ISSUE #51: Add unzip/git to runner dependencies
I'm not sure if you're willing to consider this, but for some use cases it would be very helpful to have some additional dependencies installed in the runner. Although it is easy to get around by building your own runner image with the additional dependencies, maybe the following ones could have added value for the official runner image.

### unzip

Some tool installation packages use .zip instead of .tgz, which will fail on this runner

### git > 2.18

The official checkout action `actions/checkout` gives the following message when checking out a repository:

```
The repository will be downloaded using the GitHub REST API
To create a local Git repository instead, add Git 2.18 or higher to the PATH
```

If you want to work with a local git repo (analyze, commit, etc) then a higher git version is needed. At the moment, the base ubuntu:18.04 image contains git 2.17.1, so in my custom image I install this via ppa (I'm not sure if this is desirable or if there's a viable alternative)

Again - I'm not sure this should be in the official image - it's also very well possible to do without and use a custom runner image

reiniertimmer:
Closes #51 

This adds a subset of the packages from the official Ubuntu 18.04 virtual environment to the runner image:

- Basic CLI
- Git (2.26.2)
- build-essential
- Docker

for the complete list, see: https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1804-README.md

### DIFF
--- a/runner/Dockerfile
+++ b/runner/Dockerfile
@@ -3,9 +3,40 @@ FROM ubuntu:18.04
 ARG RUNNER_VERSION
 ARG DOCKER_VERSION
 
-RUN apt update \
-  && apt install sudo curl ca-certificates -y --no-install-recommends \
-  && curl -L -o docker.tgz https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_VERSION}.tgz \
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt update -y \
+  && apt install -y software-properties-common \
+  && add-apt-repository -y ppa:git-core/ppa \
+  && apt update -y \
+  && apt install -y --no-install-recommends \
+     build-essential \
+     curl \
+     ca-certificates \
+     dnsutils \
+     ftp \
+     git \
+     iproute2 \
+     iputils-ping \
+     jq \
+     libunwind8 \
+     locales \
+     netcat \
+     openssh-client \
+     parallel \
+     rsync \
+     shellcheck \
+     sudo \
+     telnet \
+     time \
+     tzdata \
+     unzip \
+     upx \
+     wget \
+     zip \
+     zstd \
+  && rm -rf /var/lib/apt/lists/*
+
+RUN curl -L -o docker.tgz https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_VERSION}.tgz \
   && tar zxvf docker.tgz \
   && install -o root -g root -m 755 docker/docker /usr/local/bin/docker \
   && rm -rf docker docker.tgz \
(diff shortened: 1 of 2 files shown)


runner/Dockerfile
```diff
@@ -3,9 +3,15 @@ FROM ubuntu:18.04
...
 
-RUN apt update \
-  && apt install sudo curl ca-certificates -y --no-install-recommends \
-  && curl -L -o docker.tgz https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_VERSION}.tgz \
+RUN apt update -y \
+  && apt install -y software-properties-common \
+  && add-apt-repository -y ppa:git-core/ppa \
+  && apt update -y \
+  && apt install sudo curl ca-certificates unzip git -y --no-install-recommends 
+  && apt clean \
```
summerwind, line 11: According to Docker's best practice, this is not necessary now ;)
> Official Debian and Ubuntu images automatically run apt-get clean, so explicit invocation is not required.


https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get

runner/Dockerfile
```diff
@@ -20,7 +26,9 @@ RUN mkdir -p /runner \
   && curl -L -o runner.tar.gz https://github.com/actions/runner/releases/download/v${RUNNER_VERSION}/actions-runner-linux-x64-${RUNNER_VERSION}.tar.gz \
   && tar xzf ./runner.tar.gz \
   && rm runner.tar.gz \
-  && ./bin/installdependencies.sh
+  && ./bin/installdependencies.sh \
+  && apt clean \
```
summerwind, line 30: ditto.

runner/Dockerfile
reiniertimmer: Thanks, I wasn't 100% sure about it. I did keep the `rm -rf /var/lib/apt/lists/*` though (saves about 30 Mb on the final image)

reiniertimmer:
@summerwind I installed all of the packages from the virtual environment document _except_ for the following:

* libcurl3 (this would uninstall the `curl` executable, as that uses libcurl4)
* libicu55 (could be installed from 16.04 repositories, but I'm not sure this is preferred)

I updated the documentation about using the official & custom docker images as well

runner/Dockerfile
summerwind: Yeah, I think `rm -rf /var/lib/apt/lists/*` is still needed today ;)

reiniertimmer:
Thanks for the feedback and for the merge. As always, I'm happy to contribute to this awesome project - we're already very happy users here!

PR #54, opened 2020-05-12
Update runner to v2.262.0

summerwind:
This PR updates runner's version to v2.262.0.
See: https://github.com/actions/runner/releases/tag/v2.262.0


summerwind:
It's a small change, so let me merge!

PR #57, opened 2020-05-26
feat: Repository-wide RunnerDeployment Autoscaling

mumoshu:
This adds `spec.maxReplicas`, `spec.minReplicas` and `status.desiredReplicas` to RunnerDeployment. If and only if `minReplicas` and `maxReplicas` are set to non-zero values, the controller computes and automatically sets desired `replicas` between the min and the max, depending on demand.

```yaml
kind: RunnerDeployment
metadata:
  name: summerwind-actions-runner-controller
spec:
  # replicas: 1
  # NEW: Set the minimum number of replicas
  minReplicas: 1
  # NEW: Set the maximum number of replicas
  maxReplicas: 3
  template:
    spec:
status:
  # NEW: Calculated depending on demand and propagated to RunnerReplicaSet.Spec.Replicas
  desiredReplicas: 2
```

The number of demanded runner replicas is computed by `queued workflow runs + in_progress workflow runs` for the repository.

We use GitHub Action's `List Repository Workflow Runs` API to list workflow runs whose statuses can be one of `queued`, `in_progress`, `completed`. We don't need to scale depending on number of `completed` runs so that's not considered in the calculation.

The support for organizational runners is a TODO, but shouldnt be that hard.

Ref https://github.com/summerwind/actions-runner-controller/issues/10

Notes:
- [x] Manual test pending. Early review comments are appreciated.
- [x] ~~TODO: I think we'd better have `scaleDownDelayAfterAdd` like e.g. `scale-down-delay-after-add` of [Cluster Autoscaler](https://github.com/uswitch/kubernetes-autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-the-parameters-to-ca) to avoid flapping(too frequent/fast addition/deletion loop)~~
  - `scaleDownDelayAfterSclaeUp` added
- [x] ~~TODO: We'd better to make [controller-runtime `SyncPeriod`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/manager?tab=doc#Options) configurable so that we can set it to something other than the default 10 hours. Without this, the controller can take up to 10 hours to start recalculating desired replicas, which makes this feature useless.~~
  - `--sync-period DURATION` added


summerwind:
Thank you for the PR! I will review this in a few days.

controllers/runnerdeployment_controller.go
```diff
@@ -94,7 +97,14 @@ func (r *RunnerDeploymentReconciler) Reconcile(req ctrl.Request) (ctrl.Result, e
...
 
-	desiredRS, err := r.newRunnerReplicaSet(rd)
+	desiredReplicas, err := r.determineDesiredReplicas(rd)
+	if err != nil {
+		log.Error(err, "Could not determine desired replicas")
+
+		return ctrl.Result{}, err
+	}
+
+	desiredRS, err := r.newRunnerReplicaSet(rd, *desiredReplicas)
```
mumoshu, line 107: Probably we'd better modify RunnerReplicaSet controller to prefer scaling down inactive runner first, so that scale down and happen more responsively.

Please also see how a similar project does it: https://github.com/philips-labs/terraform-aws-github-runner/blob/88bc451e5e36358666a220d9c737ad6639da03a1/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts#L79-L101

api/v1alpha1/runnerdeployment_types.go
```diff
@@ -22,14 +22,17 @@ import (
...
 
 	Template RunnerTemplate `json:"template"`
 }
 
 type RunnerDeploymentStatus struct {
-	AvailableReplicas int `json:"availableReplicas"`
-	ReadyReplicas     int `json:"readyReplicas"`
+	AvailableReplicas int  `json:"availableReplicas"`
+	ReadyReplicas     int  `json:"readyReplicas"`
+	DesiredReplicas   *int `json:"desiredReplicas"`
```
summerwind, line 35: Using the `scale` sub-resource here would be a more Kubernetes-like implementation. It looks like controller-runtime already supports the `scale` sub-resource. thoughts?

controllers/autoscaling_test.go
```diff
@@ -0,0 +1,136 @@
...
+		{
+			fixed: intPtr(3),
+			want:  3,
+		},
+	}
+
+	for i := range testcases {
+		tc := testcases[i]
+
+		log := klogr.New()
```
summerwind, line 84: It may be better to use the controller-runtime log package, as this will add a new dependency package.
https://godoc.org/sigs.k8s.io/controller-runtime/pkg/log.

controllers/autoscaling_test.go
```diff
@@ -0,0 +1,136 @@
...
+		r := &RunnerDeploymentReconciler{
+			Log: log,
+			GitHubClient: &github.Client{
+				ListRepositoryWorkflowRuns: func(ctx context.Context, owner, repo string, opts *github2.ListOptions) (runs *github2.WorkflowRuns, response *github2.Response, err error) {
```
summerwind, line 88: I'm sure what this code is doing. However, we use the fake server for tests that need to work with the GitHub API. Is it possible to modify this test to be dependent on the fake server?

api/v1alpha1/runnerdeployment_types.go
mumoshu: @summerwind Sounds good! But can we address that in another PR? AFAIK, a `scale` sub-resource doesn't make `DesiredReplicas` unnecessary, and the change would be additive.

mumoshu: BTW, do you have any opinion on whether we use `Replicas` or `DesiredReplicas` for this field?

Initially, I thought `Replicas` is better, but it turned out that we usually use status.Replicas for something other than what I mean here.

Here's the definition from apps/v1.DeploymentStatus:

```
// DeploymentStatus is the most recently observed status of the Deployment.
type DeploymentStatus struct {
	// The generation observed by the deployment controller.
	// +optional
	ObservedGeneration int64 `json:"observedGeneration,omitempty" protobuf:"varint,1,opt,name=observedGeneration"`

	// Total number of non-terminated pods targeted by this deployment (their labels match the selector).
	// +optional
	Replicas int32 `json:"replicas,omitempty" protobuf:"varint,2,opt,name=replicas"`

	// Total number of non-terminated pods targeted by this deployment that have the desired template spec.
	// +optional
	UpdatedReplicas int32 `json:"updatedReplicas,omitempty" protobuf:"varint,3,opt,name=updatedReplicas"`

	// Total number of ready pods targeted by this deployment.
	// +optional
	ReadyReplicas int32 `json:"readyReplicas,omitempty" protobuf:"varint,7,opt,name=readyReplicas"`

	// Total number of available pods (ready for at least minReadySeconds) targeted by this deployment.
	// +optional
	AvailableReplicas int32 `json:"availableReplicas,omitempty" protobuf:"varint,4,opt,name=availableReplicas"`

	// Total number of unavailable pods targeted by this deployment. This is the total number of
	// pods that are still required for the deployment to have 100% available capacity. They may
	// either be pods that are running but not yet available or pods that still have not been created.
	// +optional
	UnavailableReplicas int32 `json:"unavailableReplicas,omitempty" protobuf:"varint,5,opt,name=unavailableReplicas"`

	// Represents the latest available observations of a deployment's current state.
	// +patchMergeKey=type
	// +patchStrategy=merge
	Conditions []DeploymentCondition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,6,rep,name=conditions"`

	// Count of hash collisions for the Deployment. The Deployment controller uses this
	// field as a collision avoidance mechanism when it needs to create the name for the
	// newest ReplicaSet.
	// +optional
	CollisionCount *int32 `json:"collisionCount,omitempty" protobuf:"varint,8,opt,name=collisionCount"`
}
```

It does say that `status.replicas` is for `Total number of non-terminated pods targeted by this deployment (their labels match the selector).`, which implies to include outdated pods as well.

In contrast, RunnerDeploymnet's status.desiredReplicas is basically `total number of desired and non-terminated pods to be set for the primary runnerset`.

controllers/autoscaling_test.go
mumoshu: Migrated to the controller-runtime one in 7555b7e. Thanks!

mumoshu:
> Agreed. However, we need to take into account the rate limit of the API. I think that 10 minutes is a reasonable value.

Added `--sync-period DURATION` for that in 7adf772.

Should we change the default to e.g. 10 minutes, or not?

I think it's pretty safe to have a lower default because it should be rate-limited per app/token and doesn't affect other github apps. And I'd say it's better to have a lower default cuz it would help us reduce user support issues about scaling not happening due to high sync period. But I'm open to suggestions.

controllers/autoscaling_test.go
mumoshu: Great suggestion 👍

Addressed in 4ebc4e8. Please note that I've enhanced the fake server creation to customize the API response per test run. Hope you like it.

MPV:
Hi! 👋  
Thanks for your hard work on this. ❤️ 

I'm mainly interested in autoscaling for org-level runners.

Would something like this be needed/helpful for this? 👉    https://github.com/actions/runner/issues/454

api/v1alpha1/runnerdeployment_types.go
mumoshu: Thanks. Changed to `Replicas` in 0d45cef 😃

mumoshu:
> I completely agree with you on this. I think setting 10 minutes is a good idea for now.

Thanks for confirming! I've updated the default sync period to `10m`

mumoshu:
Finished updateing README about this: https://github.com/summerwind/actions-runner-controller/compare/master...7c522510c9b0ac3b321b1e991efd5c853e66d586#diff-04c6e90faac2675aa89e2176d2eec7d8R192-R229

mumoshu:
I've finished brief manual testing. At least it seems like working as intended.

@summerwind I believe this is good to go now. Can we merge this and start getting feedback from users?

mumoshu:
@summerwind Thanks for reviewing! I'm gonna squash everything into a single commit and merge this.

mumoshu:
@summerwind Ah, did you literally mean to use merge instead of squash/commit? If so, I'll redo it and ensure using merges in the future.

summerwind:
@mumoshu Oops, I meant to clean up the fixup commits and merge them. 
But it's fine as is. Thanks for developing a great feature!

Issue #10, opened 2020-03-18
Autoscaling support


iagomelanias:
It would be great if pods could be adjusted automatically based on the number of pending workflows. I'm not sure if that's possible using the the current version of the Github Actions API or we should use a metrics server and try to keep the number of idle runners below X, but there should be a way to implement this feature. 🙂

alexandrst88:
I think, It could be postpone, due to GitHub Actions will release Organizational runners officially.

iagomelanias:
Oh, didn't know about that! Let's wait then 🙂

mumoshu:
This might be possible today by poling the [list repository workflow runs API](https://developer.github.com/v3/actions/workflow_runs/#list-repository-workflow-runs) for runs w/ `status: queued` where `created_at` is before some threshold.

If anyone is willing to give it a shot, I'm more than happy to review and test!

As @alexandrst88 has noted, this is likely to change once the organization runners gets added though. Perhaps they'll add an another API to let us know which workflow runs is tied to which runner(repository-specific vs organizational vs etc)?

jay2610:
Any update on this since we got ORG runners added?

mumoshu:
Nothing on my end. Would you be willing to contribute it? The API to be used should now be https://developer.github.com/v3/actions/workflow-runs/#list-workflow-runs 😃

jay2610:
I would more than happy to contribute but I don't know Go 😞. I tried to explore API and going try run one more controller with runner-controller deployment in JavaScript or even bash maybe. 
This should be the logic from my initial search.
1. Set the env variable for minimum number of runners set on RunnerDeployment
2. Query the number of pending jobs every minute. 
3. Based on that scale up RunnerDeployment (that's what I use)
3. If there is no pending job and more runners are Ideal then minimum number, scale down RunnerDeployment to min number.

PR #66, opened 2020-07-03
feat: Organizational RunnerDeployment Autoscaling

mumoshu:
Enhances #57 to add support for organizational runners.

As GitHub Actions does not have an appropriate API for this, this is the spec you need:

```
kind: RunnerDeployment
metadata:
  name: myrunners
spec:
  minReplicas: 1
  maxReplicas: 3
  template:
    spec:
      organization: myorg
---
metadata:
  name: myrunners
spec:
  scaleTargetRef:
    name: myrunners
  minReplicas: 1
  maxReplicas: 2
  metrics:
  - type: TotalNumberOfQueuedAndInProgressWorkflowRuns
    repositoryNames:
    # Assumes that you have `github.com/myorg/myrepo1` repo
    - myrepo1
    - myrepo2

```

It works by collecting "in_progress" and "queued" workflow runs for the repositories `myrepo1` and `myrepo2` to autoscale the number of replicas, assuming you have this organizational runner deployment only for those two repositories.

For example, if `myrepo1` had 1 `in_progress` and 2 `queued` workflow runs, and `myrepo2` had 4 `in_progress` and 8 `queued` workflow runs at the time of running the reconcilation loop on the runner deployment, it will scale replicas to 1 + 2 + 4 + 8 = 15.

Perhaps we might be better add a kind of "ratio" setting so that you can configure the controller to create e.g. 2x runners than demanded. But that's another story.

Ref #10


avik-so:
When using `repositories` in the yaml file I get an error: `validating autoscaling metrics: spec.autoscaling.metrics[].repositoryNames is required and must have one more more entries for organizational runner deployment` (typo in the error message btw, "more" is repeated twice.

I had to change the yaml to `repositoryNames:` and then I got a new error:
`"GET https://api.github.com/repos/<org>/<repo>/actions/runs: 404 Not Found []"}

It's a private repository.

summerwind:
Thank you for the PR! I believe that this is a very useful feature for many users!

As an overall design, I'm not sure if we should add fields related to autoscaling metrics to `RunnerDeployment`. There is a concern that adding autoscaling capabilities to `RunnerDeployment` resource could be very complex.

I think it would be possible to separate the autoscaling functionality from `RunnerDeployment` by adding a custom resource with a name like `RunnerAutoscaler`, for example, `HorizontalPodAutoscaler`, to tie the metrics to the Runner.

avik-so:
As a user, I prefer as few YAML files as necessary :)
So hopefully the examples will make it easy to use without a lot of repetition or boilerplate.

mumoshu:
Easy doesn't mean simple. I mean, everyone would love separating CRDs for deployments and autoscaling 😄 

@summerwind So I'm willing to separate CRDs 👍 

Would you mind if I do that before adding this feature?

Also, I won't try to keep the backward-compatibility while doing that. It's only 1 minor release since the feature is added and therefore removing `RunnerDeployment.Spec.MinReplicas` and `MaxReplicas` won't result in too much confusion. Would you be ok with that direction?

summerwind:
@mumoshu Thanks for your comment! 
Yes. I think it's okay to make incompatible changes, such as removing the MinReplicas field at this time 😉

mumoshu:
@summerwind Done. Would you mind taking a look once again? :smi

avik-so:
> @summerwind Done. Would you mind taking a look once again?

Should I test it for my scenario, or is that not ready yet?

controllers/horizontalrunnerautoscaler_controller.go
```diff
@@ -0,0 +1,180 @@
...
+// HorizontalRunnerAutoscalerReconciler reconciles a HorizontalRunnerAutoscaler object
+type HorizontalRunnerAutoscalerReconciler struct {
+	client.Client
+	GitHubClient *github.Client
+	Log          logr.Logger
+	Recorder     record.EventRecorder
+	Scheme       *runtime.Scheme
+}
+
+// +kubebuilder:rbac:groups=actions.summerwind.dev,resources=runnerdeployments,verbs=get;list;watch;create;update;patch;delete
```
summerwind, line 47: I think we should remove `create` and `delete` from here 😉

controllers/horizontalrunnerautoscaler_controller.go
```diff
@@ -0,0 +1,180 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var scaleDownDelay time.Duration
+
+	if hra.Spec.ScaleDownDelaySecondsAfterScaleUp != nil {
+		scaleDownDelay = time.Duration(*hra.Spec.ScaleDownDelaySecondsAfterScaleUp) * time.Second
+	} else {
+		scaleDownDelay = 10 * time.Minute
```
summerwind, line 164: This value of `10` seems to be a constant.

README.md
```diff
@@ -201,11 +201,23 @@ kind: RunnerDeployment
...
-  minReplicas: 1
-  maxReplicas: 3
   template:
     spec:
       repository: summerwind/actions-runner-controller
+---
+apiVersion: actions.summerwind.dev/v1alpha1
+kind: HorizontalRunnerAutoscaler
+metadata:
+  name: summerwind-actions-runner-controller
```
summerwind, line 211: This name is confusing, so I think it would be better to use a name like `example-runner-deployment-autoscaler`

README.md
```diff
@@ -201,11 +201,23 @@ kind: RunnerDeployment
...
+spec:
+  scaleTargetRef:
+    name: summerwind-actions-runner-controller
```
summerwind, line 214: ditto. (eg: `example-runner-deployment`)

README.md
```diff
@@ -219,12 +231,24 @@ kind: RunnerDeployment
...
-  maxReplicas: 3
-  scaleDownDelaySecondsAfterScaleUp: 1m
   template:
     spec:
       repository: summerwind/actions-runner-controller
+---
+apiVersion: actions.summerwind.dev/v1alpha1
+kind: HorizontalRunnerAutoscaler
+metadata:
+  name: summerwind-actions-runner-controller
```
summerwind, line 241: ditto.

README.md
```diff
@@ -219,12 +231,24 @@ kind: RunnerDeployment
...
+spec:
+  scaleTargetRef:
+    name: summerwind-actions-runner-controller
```
summerwind, line 244: ditto.

mumoshu:
@summerwind Thanks for reviewing 😃 I've just finished updating code to address everything you've commented. Would you mind reviewing it once again?

PR #81, opened 2020-08-02
Add scaling-down scenario to integration test

mumoshu:
This enhances the integration test suite added in #76 to cover scaling down.
In the process, I've refactored verbose/fundamentally redundant code around customizing the fake GitHub API server, and enhanced it to allow mutation on the List Workflows API mock for testing.

I've manually verified that the integration test is passing on my machine.

### DIFF
--- a/controllers/integration_test.go
+++ b/controllers/integration_test.go
@@ -19,18 +19,32 @@ import (
 	actionsv1alpha1 "github.com/summerwind/actions-runner-controller/api/v1alpha1"
 )
 
+type testEnvironment struct {
+	Namespace *corev1.Namespace
+	Responses *fake.FixedResponses
+}
+
+var (
+	workflowRunsFor3Replicas = `{"total_count": 5, "workflow_runs":[{"status":"queued"}, {"status":"queued"}, {"status":"in_progress"}, {"status":"in_progress"}, {"status":"completed"}]}"`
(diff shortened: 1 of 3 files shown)


controllers/integration_test.go
```diff
@@ -249,6 +266,43 @@ var _ = Context("Inside of a new namespace", func() {
 					},
 					time.Second*5, time.Millisecond*500).Should(BeEquivalentTo(3))
 			}
+
+			// Scale-down to 1 replica
+			{
+				responses.ListRepositoryWorkflowRuns.Body = workflowRunsFor1Replicas
```
mumoshu, line 272: We alter the mock GitHub API response here, so that the next `force-update` on the HRA triggers an immediate resync on the resource, which result in catching change in the GH API and rescale RD accordingly.

controllers/integration_test.go
```diff
@@ -216,9 +233,9 @@ var _ = Context("Inside of a new namespace", func() {
 					},
 				}
 
-				err := k8sClient.Create(ctx, rs)
+				err := k8sClient.Create(ctx, hra)
 
-				Expect(err).NotTo(HaveOccurred(), "failed to create test RunnerDeployment resource")
```
mumoshu, line 221: This was a typo

PR #91, opened 2020-08-21
Add helm chart for controller and deployment

stackdumper:
❗️WIP❗️

This PR introduces two Helm charts:
- **actions-runner-controller**
- **actions-runner-deployment**

## actions-runner-controller
Installs actions-runner-controller manifest with no configuration options. Installing this chart is analogous to running:
```sh
 kubectl apply -f https://github.com/summerwind/actions-runner-controller/releases/latest/download/actions-runner-controller.yaml
```
For me, the need for a chart instead of `kubectl apply` arises from a need to manage Kubernetes using [Terraform](https://www.terraform.io/). Kubernetes plugin for Terraform doesn't support the installation of custom manifests from files. Introducing the Helm chart allows the installation of actions-runner-system via Terraform by using [Helm provider](https://registry.terraform.io/providers/hashicorp/helm/latest) and [helm_release](https://registry.terraform.io/providers/hashicorp/helm/latest/docs/resources/release). 

The snippet below provides an example use-case with Terraform.
```hcl
resource "kubernetes_namespace" "actions-runner-system" {
  metadata {
    name = "actions-runner-system"
  }
}


# secret for actions-runner
resource "kubernetes_secret" "controller-manager" {
  metadata {
    name      = "controller-manager"
    namespace = kubernetes_namespace.actions-runner-system.metadata[0].name
  }

  type = "generic"
  data = {
    "github_app_id"              = "..."
    "github_app_installation_id" = "..."
    "github_app_private_key"     = file("${path.module}/certs/actions-runner-private-key.pem")
  }
}

resource "helm_release" "actions-runner-controller" {
  name = "actions-runner-controller"
  namespace = kubernetes_namespace.actions-runner-system.metadata[0].name
  chart     = "https://github.com/stackdumper/actions-runner-controller/releases/download/v0.7.2/actions-runner-controller-0.1.0.tgz"
  version   = "0.1.0"
  depends_on = [kubernetes_secret.controller-manager]
}
```

An example use-case with Helm install:
```sh
helm install -n actions-runner-system actions-runner-controller https://github.com/stackdumper/actions-runner-controller/releases/download/v0.7.2/actions-runner-controller-0.1.0.tgz
```

#### Notes
1. This chart disables the automatic creation of the `actions-runner-system` namespace in favor of a manual namespace creation. This prevents an error if namespace already exists and allows for a more fine-grained installation. 
2. It would be good to enable custom configuration of actions-runner-controller. The simplest example that I can think of is a need to run actions-controller-manager only on a specific type of nodes by using `nodeSelector`. 

## actions-runner-deployment
Installs RunnerDeployment and, optionally, HorizontalRunnerAutoscaler. 
An example use-case with Terraform:

```hcl
// self-hosted github actions runner
resource "helm_release" "actions-runner" {
  name      = "actions-runner"
  namespace = kubernetes_namespace.actions-runner-system.id
  chart     = "https://github.com/stackdumper/actions-runner-controller/releases/download/v0.7.2/actions-runner-deployment-0.1.0.tgz"
  version   = "0.1.0"

  set {
    name  = "spec.template.spec.organization"
    value = "rotten-tech"
  }

  set {
    name  = "spec.replicas"
    value = 2
  }

  set {
    name  = "spec.template.spec.labels"
    value = "{staging.example.com,qa.example.com}"
  }

  depends_on = [helm_release.actions-runner-controller]
}
```

An example use-case with Helm install:

```sh
helm install -n actions-runner-system --set spec.template.spec.repository=rotten-tech/fs-api test https://github.com/stackdumper/actions-runner-controller/releases/download/v0.7.2/actions-runner-deployment-0.1.0.tgz
```

## GitHub Releases
1. During `make release`, the controller manifest (`actions-runner-controller.yaml`) is automatically copied into `charts/actions-runner-controller/templates`. 
2. I've extended `make release` to build an archive for each chart using `helm archive` and put it into `./release` dir to be included in the GitHub release files. An example release:
https://github.com/stackdumper/actions-runner-controller/releases/tag/v0.7.2

## To do
- [x] Double-test clean installation with `helm install`.
- [ ] Extend docs with different installation options (examples): kubectl, Helm, Terraform.


kuuji:
I have most of the controller templated @stackdumper + the crds. But I kept the runners themselves separated. As I want to control them outside of this chart.
I can contribute what I have to your PR if you're cool with that.

But I think I would also like the option to not create runner using the chart.

stackdumper:
> But I kept the runners themselves separated. As I want to control them outside of this chart.
> But I think I would also like the option to not create runner using the chart.

There are two charts here, one for the controller (actions-runner-controller) and one for the runner (actions-runner-deployment). The controller can be deployed without the runner and vice versa, which makes them separated. Doesn't this solve your case?

kuuji:
That is not in your PR @stackdumper. There's only one deployment file for runner and a values for that chart. 
The controller chart is empty.

stackdumper:
@kuuji It's generated automatically by GitHub Actions and deployed to GitHub releases automatically after you create a new GitHub Release from a git tag. You can generate it manually by running `make release`. 

Do you think it should be included in git? It's basically [this](https://github.com/summerwind/actions-runner-controller/releases/latest/download/actions-runner-controller.yaml), and currently it's not in the git history.

kuuji:
Oh so it's not really templated. It basically just installs the entire yaml.

I have a fully templated version that I've been using for some time. But that is not generated, that may be possible to do though.
I'll open a separate PR to get people's eyes on it.

stackdumper:
@kuuji Hm, got it. The problem with non-generated chart is that it's not autoupdated when the source files change. So when a new version of actions-runner-controller comes out, the templated helm chart will need to be manually updated, which is hard for such a big file. What do you think?

kuuji:
We may be able to make it work. I believe some projects are generating their helm charts with kubebuilder/controller-gen.

I'm going to do a little bit of research to see what would be involved.

stackdumper:
Sounds interesting. Let me know if you need any help with this.

charts/actions-runner-controller/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v2
+name: actions-runner-controller
+description: A Helm chart for summerwind/actions-runner-controller
+type: application
+
+# chart version
+version: 0.1.0
+
+# app (controller) version
+appVersion: 0.7.2
```
rowleyaj, line 10: it might be good to pull this from `${VERSION}` in the makefile? maybe using another sed statement to substitute it?

charts/actions-runner-deployment/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v2
+name: actions-runner-deployment
+description: A Helm chart for installing a GitHub Actions runners deployment managed by summerwind/actions-runner-controller 
+type: application
+
+# chart version
+version: 0.1.0
+
+# app (controller) version
+appVersion: 0.7.2
```
rowleyaj, line 10: same as above comment re app version substitution from make file

charts/actions-runner-controller/Chart.yaml
stackdumper: Done. However, since `make release` is executed in GitHub Actions, this will only affect chart archive in GitHub Releases. The version in the git will remain unchanged.

charts/actions-runner-deployment/Chart.yaml
stackdumper: ✅

Makefile
```diff
@@ -68,6 +68,23 @@ release: manifests
 	mkdir -p release
 	kustomize build config/default > release/actions-runner-controller.yaml
 
+	# build helm charts	
+	# copy controller manifest to helm templates
+	cp -v release/actions-runner-controller.yaml charts/actions-runner-controller/templates
```
hfuss, line 73: Consider trying to use `split -p '---'` to split up all the manifests if possible (though no idea how you'd give each file sensible names without having to use `yq` to extract the `metadata.name` and `kind`).

More importantly though, the CRDs should get `cp`ed into the `crds` folder of the chart rather than the templates: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you

Nuru:
Any updates on this? I'm doing a new install and would rather not create new helm charts for it.

Nuru:
Please include the ability to add annotations to the controller Deployment.

callum-tait-pbx:
@stackdumper I'm super excited for this getting merged!

stackdumper:
Replaced by #187. Kudos to @mumoshu 👍

PR #96, opened 2020-09-14
dockerEnabled for Stricter Clusters like OpenShift

hfuss:
Opening a draft while I test this out.

## Motivation

I would like it so my users can make `Runners` without Docker access since on OpenShift that requires a `SecurityContextConstraints`. Other K8s clusters might require a `PodSecurityPolicy` instead FWIW.

## Desired Experience

If users attempt to make a `Runner` with `dockerEnabled: true` and they don't have an SCC for the `ServiceAccount` they specified using `serviceAccountName` it will fail to create a pod and this is reflected in the Runner's status (this already occurs today without the `dockerEnabled` option).

If `dockerEnabled: false` is set, the user will then be able to have `Runners` create `Pods` without having to make an SCC/PSP because the Docker sidecar and volumes are omitted.

### DIFF
--- a/api/v1alpha1/runner_types.go
+++ b/api/v1alpha1/runner_types.go
@@ -77,6 +77,9 @@ type RunnerSpec struct {
 	EphemeralContainers []corev1.EphemeralContainer `json:"ephemeralContainers,omitempty"`
 	// +optional
 	TerminationGracePeriodSeconds *int64 `json:"terminationGracePeriodSeconds,omitempty"`
+
+	// +optional
+	DockerEnabled *bool `json:"dockerEnabled,omitempty"`
 }
 
 // ValidateRepository validates repository field.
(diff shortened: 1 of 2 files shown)


Hi-Fi:
Isn't this same than new setting `dockerdWithinRunnerContainer`?

Warashi:
when set `dockerdWithinRunnerContainer` true, runner container runs with privileged.
this PR maybe useful for users who doesn't need docker and want not to run privileged container.

Warashi:
@hfuss 
Do you have motivation to continue this PR?
If you don't, I will send PR doing same as this.

hfuss:
@Warashi I've had to put this down so please go ahead! Thank you for asking.

Will happily review and QA when it's ready.

mumoshu:
@hfuss #191 has been released as a part of v0.12.0. Would you mind giving it a shot?

Thanks for the PR @Warashi!

PR #104, opened 2020-09-28
chore(deps): bump go to v1.15

chenrui333:
Looks like the repo did not stay with the latest go, thus bumping it to v1.15


mumoshu:
@chenrui333 Thanks for the PR! Sorry for the delay - but would you mind redoing the go version bump? You have a single-line change in go.mod but I'm not sure if that's intended or not.

mumoshu:
Redoing this #160. Thanks for the reminder @chenrui333 ☺️

Issue #89, opened 2020-08-20
Make controller also look at queued jobs.


helder-moreira:
A running workflow may have multiple queued jobs. The controller could add more runners based on this.

As an example: I have a `RunnerDeployment` with autoscaling enabled between 3 and 20 instances. One of the workflows I run, has 6 parallel jobs, of which 3 will be always queued.

I can try to make a PR with this over the next few days if you are open to it.

dlobue:
@helder-moreira FYI I have a PR up that adds this functionality.

helder-moreira:
@dlobue perfect! Thank you 🙏

PR #114, opened 2020-10-07
Prefer autoscaling based on jobs rather than workflows if available

dlobue:
This PR adds the ability to autoscale on jobs in addition to workflows. We fall back to using workflow metrics if job details are not present.

This solves #89


mumoshu (approved):
Awesome "job"! Thanks a lot for your contribution @dlobue 😄

PR #86, opened 2020-08-15
Initial multi-arch image commit

xUnholy:
Adding multi-arch image support for `arm64` and `amd64`. This uses dockers new `buildx` feature, to enable further architectures more work will be required to update the `runner/Dockerfile` file to pull architecture-specific releases.

The Makefile targets really should only be used for local testing and not for release, additional work to appropriately tag the release images may need to be added but for now, I've not added that logic.

Fixes: #86 

Please review @mumoshu @summerwind 🙏 

Signed-off-by: Michael Fornaro <20387402+xUnholy@users.noreply.github.com>

PR #115, opened 2020-10-08
Fixing Docker Build and Push for Runner Image

hfuss:
A new image tag for the runner stopped being published on master merges from changes in #86.

This fixes that in the following ways:
- Tests the GH workflow on PRs w/o pushing the images
- Runner and Docker version are moved from Makefile to GH Actions workflow and are passed in as build args
- GHA workflow runs on PRs, and if the workflow file itself is changed (i.e. version bump) or the runner Docker source changes (excluding the Makefile since thats just for local dev)
- Images are pushed on push (i.e. a merge)


hfuss:
@mumoshu @summerwind @xUnholy hey folks, could I get reviews on this one ASAP so we can get the 2.273.5 image pushed out? Thank you for the time and help!

.github/workflows/build-runner.yml
```diff
@@ -19,16 +31,21 @@ jobs:
...
         working-directory: runner
         run: |
           docker buildx build \
+            --build-arg RUNNER_VERSION=${RUNNER_VERSION} \
+            --build-arg DOCKER_VERSION=${DOCKER_VERSION} \
             --platform linux/amd64,linux/arm64 \
-            --tag summerwind/actions-runner:latest \
-            -f Dockerfile . --push
+            --tag summerwind/actions-runner:v${RUNNER_VERSION} \
+            -f Dockerfile .
```
xUnholy, line 42: Now you're no longer using `--push` you'll need to provide an alternative. To be honest splitting this out will make it much more difficult because fat manifests won't load into docker the same way single platform architectures do.

mumoshu (approved):
LGTM. Merging this to test out. Thanks for your contribution @hfuss and @xUnholy ☺️

mumoshu:
FYI: [The workflow run for the v0.9.1 tag](https://github.com/summerwind/actions-runner-controller/actions/runs/296508676) passed, but [one for push](https://github.com/summerwind/actions-runner-controller/actions/runs/296508676) seems to have failed.

xUnholy:
@mumoshu it won't work in its current form.

xUnholy:
I'm not sure all these changes are necessary and don't give any advantage, we do need the build ARGs but we shouldn't be merging and splitting the tasks.

mumoshu:
@xUnholy Thanks! Honestly I'm not entirely sure what's going on. Perhaps can I just re-merge the tasks and while keeping the added build-args to fix it?

xUnholy:
Yeah you just need to add the build ARGs and undo everything, I'll implement a few things on a new PR tonight or tomorrow ☺️

mumoshu:
@xUnholy Thanks for the confirmation!

Would you mind reviewing #117 instead? It's @jnewland's awesome work to fix and enhances the workflow. It looked great to me, but I'd love your feedback, too 😄

PR #123, opened 2020-10-13
Manifests valid for K8s 1.18 and 1.19

Hi-Fi:
Fixes #113

Fixes #116

PR #125, opened 2020-10-14
Possibility to define resources for DIND container

### LINKED ISSUE #119: Support defining resources also for docker container in runner pod
We have currently all namespaces with default limits set with `LimitRange`. At the `Runner` and `RunnerDeployment` manifest it's possible to define resources needed for runner, but docker gets the defaults which are on purpose quite small.

I see couple of options
1. Provide way to define resources also for docker container (either same as runner or something else)
2. Include dind on runner (kind of relates to #72), so same resources are used

Second option can be just documentation how to create runner suitable for that kind of usage, if ready made image is not in scope.

Hi-Fi:
Fixes #119

### DIFF
--- a/api/v1alpha1/runner_types.go
+++ b/api/v1alpha1/runner_types.go
@@ -39,6 +39,8 @@ type RunnerSpec struct {
 	// +optional
 	Containers []corev1.Container `json:"containers,omitempty"`
 	// +optional
+	DockerdContainerResources corev1.ResourceRequirements `json:"dockerdContainerResources,omitempty"`
+	// +optional
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
 	// +optional
 	VolumeMounts []corev1.VolumeMount `json:"volumeMounts,omitempty"`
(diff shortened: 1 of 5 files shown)


api/v1alpha1/runner_types.go
```diff
@@ -39,6 +39,8 @@ type RunnerSpec struct {
 	// +optional
 	Containers []corev1.Container `json:"containers,omitempty"`
 	// +optional
+	DockerResources corev1.ResourceRequirements `json:"dockerResources,omitempty"`
```
mumoshu, line 42: Thanks for the PR! As this is for configuring dockerd rather than docker, I would personally prefer `DockerdReosurces`. WDYT?

Hi-Fi: Sounds good, fixed. Thought also dindResources, but dockerdResources sound better.

mumoshu (approved):
LGTM. Thank you so much @Hi-Fi!

mumoshu:
@Hi-Fi Ah, there's some conflict. Do you need to rerun manifests generation to incorporate #123?

Hi-Fi:
> @Hi-Fi Ah, there's some conflict. Do you need to rerun manifests generation to incorporate #123?

Fixed PR, and also noticed that there's no need to do fix118 when generating go-code, only when generating manifests.

mumoshu (approved):
LGTM. Thanks for your efforts @Hi-Fi!

PR #127, opened 2020-10-15
Inject Env Vars into Runner defined Container Spec

BrendanGalloway:
The Runner definition allows for the Containers to be completely overridden.  However, if this functionality is used, the env vars containing the github credentials are not added to the containers.  Inject the variables during the step where the Runner spec Container definition overrides the default Pod to fix this.

### DIFF
--- a/controllers/runner_controller.go
+++ b/controllers/runner_controller.go
@@ -382,6 +382,9 @@ func (r *RunnerReconciler) newPod(runner v1alpha1.Runner) (corev1.Pod, error) {
 
 	if len(runner.Spec.Containers) != 0 {
 		pod.Spec.Containers = runner.Spec.Containers
+		for i := 0; i < len(pod.Spec.Containers); i++ {
+			pod.Spec.Containers[i].Env = append(pod.Spec.Containers[i].Env, env...)
+		}
 	}
 
 	if len(runner.Spec.VolumeMounts) != 0 {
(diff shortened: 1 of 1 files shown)


controllers/runner_controller.go
```diff
@@ -382,6 +382,9 @@ func (r *RunnerReconciler) newPod(runner v1alpha1.Runner) (corev1.Pod, error) {
 
 	if len(runner.Spec.Containers) != 0 {
 		pod.Spec.Containers = runner.Spec.Containers
+		for i := 0; i < len(pod.Spec.Containers); i++ {
+			pod.Spec.Containers[i].Env = append(pod.Spec.Containers[i].Env, env...)
```
mumoshu, line 386: I guess this should better be done only when `pod.Spec.Containers[i].Name == containerName`. Otherwise we may end up giving the runner token to unnecessary containers like dind or other user-defined sidecar containers

mumoshu (approved):
LGTM. Thanks for the improvement @BrendanGalloway! This will finally make the `containers` definition actually usable 😄

BrendanGalloway:
@mumoshu Indeed, it seems to be working well for us.  It did take a bit of reverse engineering to set up the custom containers correctly though, there are a number of privilege and naming settings that are assumed to be correct in the default pod and are not merged back in by the controller.  It might be worthwhile documenting these requirements or providing a default definition that includes a containers definition with the correct settings?  I don't mind adding either if you can let me know where you would like it done

PR #136, opened 2020-10-21
Restart if pod ends up succeeded

### LINKED ISSUE #132: Pod with image actions-runner-dind is left to completed status
When run completes, runner exits with RC 0, and pod is left as "Completed" (if no autoscaling is in use).

With autoscaling this is probably handled by creation of the new pods, but when using just `RunnerDeployment` it's possible to run only as many runs as there're replicas created.

When using separate DinD container things seem to be working OK.

HorizontalRunnerAutoscaler doesn't seem to be at least fast help, as completed seem to be counted as one replica.

Hi-Fi:
Fixes #132


mumoshu (approved):
LGTM. Thanks for your efforts @Hi-Fi ☺️

PR #146, opened 2020-10-27
Added support for other than public GitHub URL

Hi-Fi:
Refactoring a bit

### DIFF
--- a/runner/entrypoint.sh
+++ b/runner/entrypoint.sh
@@ -1,5 +1,16 @@
 #!/bin/bash
 
+if [ -z "${GITHUB_URL}" ]; then
+  echo "Working with public GitHub" 1>&2
+  GITHUB_URL="https://github.com/"
+else
+  length=${#GITHUB_URL}
+  last_char=${GITHUB_URL:length-1:1}
+
+  [[ $last_char != "/" ]] && GITHUB_URL="$GITHUB_URL/"; :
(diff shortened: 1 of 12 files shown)


runner/entrypoint.sh
```diff
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+if [ -z "${GITHUB_URL}" ]; then
+  echo "Working with public GitHub" 1>&2
+  GITHUB_URL="https://github.com"
```
mumoshu, line 5: I think we need a trailing `/` here

runner/entrypoint.sh
```diff
@@ -1,5 +1,12 @@
...
+else
+  echo "Github endpoint URL ${GITHUB_URL}"
```
mumoshu, line 7: Would you mind adding additional validation in the controller that ensures GITHUB_URL envvar to have the trailing slash?

runner/entrypoint.sh
Hi-Fi: True, to make runner work with older manager

runner/entrypoint.sh
Hi-Fi: There needs to be some more logic added too, as client's baseURL is by default api.github.com, not github.com.

mumoshu (approved):
LGTM. Thanks a lot for your continued support 👍

PR #152, opened 2020-11-03
Adding GHE URL to transport

### LINKED ISSUE #149: Refresh token at GHE is not working when using GIthub app authentication
GHES with repo and organization level runners tested only with PAT, and apparently there're some issue when using Github App that should be checked (see below).

Original issue related to possibility to register runners to enterprise level at #150

Hi-Fi:
Fixes #149

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"net/url"
 	"strings"
 	"sync"
 	"time"
@@ -48,6 +49,14 @@ func (c *Config) NewClient() (*Client, error) {
 		if err != nil {
 			c.Log.Error(err, "Authentication failed")
 			return nil, fmt.Errorf("authentication failed: %v", err)
+		}
+		if len(c.EnterpriseURL) > 0 {
+			githubAPIURL, err := getEnterpriseApiUrl(c.EnterpriseURL)
+			if err != nil {
+				c.Log.Error(err, "Enterprise URL incorrect")
+				return nil, fmt.Errorf("enterprise url incorrect: %v", err)
+			}
+			tr.BaseURL = githubAPIURL
 		}
 		httpClient = &http.Client{Transport: tr}
 	}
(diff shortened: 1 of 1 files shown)


github/github.go
```diff
@@ -222,3 +231,20 @@ func splitOwnerAndRepo(repo string) (string, string, error) {
...
 	return chunk[0], chunk[1], nil
 }
+
+func getEnterpriseApiUrl(baseURL string) (string, error) {
+	baseEndpoint, err := url.Parse(baseURL)
+	if err != nil {
+		return "", err
+	}
+	if !strings.HasSuffix(baseEndpoint.Path, "/") {
+		baseEndpoint.Path += "/"
```
ZacharyBenamram, line 241: the end URL should be of form: https://<ENTERPRISE_URL>/api/v3

otherwise, the token endpoint is formatted to https://<ENTERPRISE_URL/api/v3//app/installations/<INSTALATION_ID>/access_tokens

recommendation is to remove the trailing / on lines 243 and on lines 246

Hi-Fi: Interesting, as that URL formatting is directly from the client library (https://github.com/google/go-github/blob/76c3c3d7c6e78e8c91e77d2e2578c4e0a7cf96ea/github/github.go#L310-L317).

And at least it doesn't add that extra slash: https://github.com/google/go-github/blob/8e1859fece888f8f56143473c2e857a2ba84dc0c/github/apps.go#L259

ZacharyBenamram: yea - I saw these as well. Let's look to see where that front loaded / comes from

ZacharyBenamram: But this change (along with the slash removal) seems to fix the issue with using a github app to retrieve installation token.

Hi-Fi: Removing last slash when returning URL from that copied function

github/github.go
```diff
@@ -222,3 +231,20 @@ func splitOwnerAndRepo(repo string) (string, string, error) {
...
+	}
+	if !strings.HasSuffix(baseEndpoint.Path, "/api/v3/") &&
+		!strings.HasPrefix(baseEndpoint.Host, "api.") &&
+		!strings.Contains(baseEndpoint.Host, ".api.") {
+		baseEndpoint.Path += "api/v3/"
```
ZacharyBenamram, line 246: `baseEndpoint.Path += "api/v3

github/github.go
```diff
@@ -222,3 +231,20 @@ func splitOwnerAndRepo(repo string) (string, string, error) {
...
+
+func getEnterpriseApiUrl(baseURL string) (string, error) {
+	baseEndpoint, err := url.Parse(baseURL)
+	if err != nil {
+		return "", err
+	}
+	if !strings.HasSuffix(baseEndpoint.Path, "/") {
+		baseEndpoint.Path += "/"
+	}
+	if !strings.HasSuffix(baseEndpoint.Path, "/api/v3/") &&
```
ZacharyBenamram, line 243: `if !strings.HasSuffix(baseEndpoint.Path, "/api/v3") &&`

github/github.go
```diff
@@ -222,3 +231,20 @@ func splitOwnerAndRepo(repo string) (string, string, error) {
 	}
 	return chunk[0], chunk[1], nil
 }
+
+func getEnterpriseApiUrl(baseURL string) (string, error) {
+	baseEndpoint, err := url.Parse(baseURL)
+	if err != nil {
+		return "", err
+	}
+	if !strings.HasSuffix(baseEndpoint.Path, "/") {
```
ZacharyBenamram, line 240: need to remove this if check

github/github.go
ZacharyBenamram: LGTM

mumoshu (approved):
LGTM. Thanks for your support @Hi-Fi @ZacharyBenamram 👍

PR #155, opened 2020-11-04
Stop unnecessary every-1-hour restart

### LINKED ISSUE #143: Runner restarts every hour

mumoshu:
Fixes #143

### DIFF
--- a/api/v1alpha1/runner_types.go
+++ b/api/v1alpha1/runner_types.go
@@ -138,11 +138,6 @@ func (r Runner) IsRegisterable() bool {
 		return false
 	}
 
-	now := metav1.Now()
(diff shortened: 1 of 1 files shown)


Hi-Fi:
I don't think current solution is a good thing, as now we never get new token (which is needed when runner is recreated).

I think `IsRegisterable` is OK, but it just needs to be called only before creating the new pod where token is actually needed.

Hi-Fi:
I have thought this a bit, so if needed I can make PR about this refresh.

Hi-Fi:
I think #159 solves this better.way

mumoshu:
Closing in favor of #159. Thanks for your support @Hi-Fi ☺️

PR #159, opened 2020-11-09
Check token validity only when creating new pod

### LINKED ISSUE #143: Runner restarts every hour

Hi-Fi:
Fixes #143


mumoshu (approved):
LGTM. Thanks a lot for your effort @Hi-Fi!

PR #162, opened 2020-11-10
fixed label argument in entrypoint

### LINKED ISSUE #161: labels no longer working
It seems that since one of the updates the labels are no longer working.

Could it be that this is broken from the GitHub runner itself? The message I see is:

```
Unrecognized command-line input arguments: 'labels k8s-runner'. For usage refer to: .\config.cmd --help or ./config.sh --help
```

reiniertimmer:
Fixes #161 

Putting quotes around "${LABEL_ARG}" will cause it to be passed as a single argument. Removing the quotes will pass them as separate arguments again (--labels x,y,z)

### DIFF
--- a/runner/entrypoint.sh
+++ b/runner/entrypoint.sh
@@ -41,7 +41,7 @@ if [ -z "${RUNNER_REPO}" ] && [ -n "${RUNNER_ORG}" ] && [ -n "${RUNNER_GROUP}" ]
 fi
 
 cd /runner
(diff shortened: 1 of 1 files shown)


reiniertimmer:
@mumoshu / @summerwind this is quite an important one, since labels are no longer working for _any_ runner that is using `latest` or `v0.11.0`

reiniertimmer:
Now that I think about it, it seems that quotes should also be remove from ${RUNNER_GROUP_ARG}, I will add this as well

martinbeentjes (approved):

reiniertimmer:
meanwhile, if you feel like it, there's also a newer release version of the runner which I could incorporate (2.274.1)

mumoshu (approved):
LGTM. Thank you so much for your support @reiniertimmer!

PR #173, opened 2020-11-11
Increase memory limit

Hi-Fi:
We saw controller quite often OOMing, so first help to increase limit a bit.


mumoshu (approved):
LGTM. Thanks as always @Hi-Fi ☺️

PR #176, opened 2020-11-11
Updated Runner to 2.274.1

reiniertimmer:
Right now all my jobs start with downloading the runner to the latest version, which is a bit wasteful :)

Can you release a new docker image with the latest version of the runner? Thanks!


reiniertimmer:
Took a while for me to figure out that the image tag is also specified in the runner workflow. Would it not be better to change the arguments in the docker files to an empty argument? So `ARG RUNNER_VERSION` instead of `ARG RUNNER_VERSION=2.272.0`

reiniertimmer:
I built the docker image with 2.274.1 and it seems to work well (jobs start much quicker since there's no need to download the image)

mumoshu (approved):
LGTM. Thanks for your support @reiniertimmer ☺️ 

> So ARG RUNNER_VERSION instead of ARG RUNNER_VERSION=2.272.0?

Sounds good! The current situation is more than confusing.

Please note that we may need to add `--arg` to:

https://github.com/summerwind/actions-runner-controller/blob/master/Makefile#L97

when RUNNER_VERSION defaults are removed from dockerfiles.

buildx target seems to have it so I think we don't need to take care on it:

https://github.com/summerwind/actions-runner-controller/blob/master/Makefile#L108

Issue #174, opened 2020-11-11
Docker daemon permissions


droidpl:
First of all thanks for this amazing project! 🤟 

I have been trying to spin this up in an EKS cluster and almost everything has worked correctly except when I try to run docker  actions in my workflows. I've reviewed https://github.com/summerwind/actions-runner-controller/issues/138 and doesn't seem to help in my case.

## Spens
I am using the latest 0.11.2 version of the controller
EKS on 1.18 (4xt3.medium)
Default runner image with no customizations

## Tests
The action I am running is a really simple hello world that works on cloud runners. Tested it with DiD option but both fail.

**Without `dockerdWithinRunnerContainer: false`**: 
![Screen Shot 2020-11-11 at 5 53 59 PM](https://user-images.githubusercontent.com/2143931/98840332-f4e6a900-2446-11eb-9727-c5aaca0735c5.png)
```
Run ./.github/actions/hello-docker-action
  with:
    who-to-greet: Mona the Octocat
Building docker image
  Dockerfile for action: '/runner/_work/aws-deployment/aws-deployment/./.github/actions/hello-docker-action/Dockerfile'.
  /usr/local/bin/docker build -t 60e226:b93535436b0153a01b30506bd7f909c2 -f "/runner/_work/aws-deployment/aws-deployment/./.github/actions/hello-docker-action/Dockerfile" "/runner/_work/aws-deployment/aws-deployment/.github/actions/hello-docker-action"
  Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Post http://%2Fvar%2Frun%2Fdocker.sock/v1.40/build?buildargs=%7B%7D&cachefrom=%5B%5D&cgroupparent=&cpuperiod=0&cpuquota=0&cpusetcpus=&cpusetmems=&cpushares=0&dockerfile=Dockerfile&labels=%7B%7D&memory=0&memswap=0&networkmode=default&rm=1&session=3g7trqn404lc92w7lth5a6wb3&shmsize=0&t=60e226%3Ab93535436b0153a01b30506bd7f909c2&target=&ulimits=null&version=1: dial unix /var/run/docker.sock: connect: permission denied
Error: Docker build failed with exit code 1
```

**With `dockerdWithinRunnerContainer: true`**
![Screen Shot 2020-11-11 at 5 53 51 PM](https://user-images.githubusercontent.com/2143931/98840323-f31ce580-2446-11eb-81bc-79fc92f1e4a0.png)
[...]

PR #177, opened 2020-11-12
use tcp DOCKER_HOST instead of sharing docker.sock

Warashi:
docker:dind container creates `/var/run/docker.sock` with root user and root group.
so, docker command in runner container needs root privileges to use docker.sock and docker action fails because lack of permission.

in this PR, use tcp connection between runner and docker container, so runner container doesn't need root privileges to run docker, and can run docker action.

### DIFF
--- a/controllers/runner_controller.go
+++ b/controllers/runner_controller.go
@@ -401,9 +401,9 @@ func (r *RunnerReconciler) newPod(runner v1alpha1.Runner) (corev1.Pod, error) {
 					MountPath: "/runner/_work",
 				},
 			},
-			Env: []corev1.EnvVar {
+			Env: []corev1.EnvVar{
 				{
-					Name: "DOCKER_TLS_CERTDIR",
+					Name:  "DOCKER_TLS_CERTDIR",
 					Value: "",
 				},
 			},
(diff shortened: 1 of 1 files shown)


Warashi:
this fixes #174

Warashi:
Is there any difficulties to merge this PR?
If there is, I will try to solve, please tell me.

mumoshu (approved):
Thank you so much for your contribution @Warashi ☺️ 

I believe that this way should be much better than stumbling to setup consistent group IDs across containers. Let's go ahead.

If possible, would you mind contributing to setup TLS between the runner and dind container, by setting `DOCKER_TLS_CERTDIR` to a non-empty, shared directory? (Probably you have expertise in it than me, but it seems like explained in the `TLS` section of https://hub.docker.com/_/docker/

You can also find the context around enabling TLS in my comment at https://github.com/summerwind/actions-runner-controller/issues/174#issuecomment-727666073

Warashi:
Thank you to merge this 😄 

> If possible, would you mind contributing to setup TLS between the runner and dind container, by setting DOCKER_TLS_CERTDIR to a non-empty, shared directory? 

OK I try. Please give me time.

Warashi:
Opened #192 to use TLS between the runner and dind container.

PR #179, opened 2020-11-12
Enable matrix build for runner image

reiniertimmer:
If you're interested, this would run the docker build of the runner image in parallel, using the build matrix


mumoshu (approved):
Nice! Thanks for the enhancement @reiniertimmer ☺️

PR #180, opened 2020-11-12
Add read-only permissions to actions in application template for being able to autoscale

emaincourt:
Hi,

Thanks for the great work you've done with this project. It's definitely been of great help to integrate GitHub Actions with Kubernetes.

I noticed in the documentation for creating a GitHub application to authorize the controller that no `actions` scope was set. However, without `read-only` permission, I get 403 when the controller fetches pending runs to autoscale. Please let me know if I'm wrong.

This PR adds that scope to the default application template you link in the documentation.

Cheers


mumoshu (approved):
Ahh, good catch! This does make sense. Thanks a lot for the fix @emaincourt ☺️

PR #182, opened 2020-11-13
docs/improvements

callum-tait-pbx:
**Changes**
- Reordered where the documented permissions go
- Added a common errors section

**Reasons**
1.  The same permissions apply regardless of which auth method is implementated so I didn't think having them documented at the end in the PAT section made sense
2. Added a common errors section, I wasn't really that sure about the formatting of if they should be broken into their on README.md, up to you really as to if this is fine for now

### DIFF
--- a/README.md
+++ b/README.md
@@ -433,6 +433,19 @@ spec:
   image: YOUR_CUSTOM_DOCKER_IMAGE
 ```
 
+## Common Errors
+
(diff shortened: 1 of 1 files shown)


mumoshu (approved):
Awesome! Thank you so much for your contribution @callum-tait-pbx ☺️

PR #185, opened 2020-11-13
parametrized working directory

reiniertimmer:
The default working directory on the runner is `/runner/_work`, which is fine for most cases.

However, I ran into the issue when using multiple consecutive jobs where I am mixing ubuntu-latest and the summerwind runners, that restoring cache and artifacts may not always work as expected, because the working directory is different on `ubuntu-latest` (`/home/runner/work`). In this case it may be convenient to change the working directory.

In order to be compatible with older versions, I added a `workDir` parameter to the runner spec. If not specified, this will use the old `/runner/_work` directory like normal. When `workDir` is specified, the runner is configured using `--work ${RUNNER_WORKDIR}`.

### Example

I generated an image v2.274.1-wd with the updated entrypoint for testing. This supports the (optional) `workDir` parameter as in the example below.

```yaml
kind: Runner
metadata:
  name: my-runner
spec:
  image: reiniert/actions-runner-dind:v2.274.1-wd
  organization: my-org
  workDir: /home/runner/work
```

### Compatibility

Since this change involves both the controller as in the runner docker image, I have analyzed the compatibility modes between them.

Controller | Runner Image | Working
---|---|---
new | new | ✔️ 
old | new | ✔️ 
new | old | ❌/ ✔️  

So in short, the image can be updated without issues. Using the _new_ controller with an _old_ image will still work, though for the default image, the `mountPath: /your/custom/path` will be set whereas the runner would still work in the old location. Since the image is more frequently (and automatically) updated, I feel that this is not such a big problem, would you agree?

### Testing

Change was tested with both the regular as well as the DinD image. Results:

image | Spec.WorkDir | working directory reported by workflow
---|---|---
default | - | `/runner/_work`
default | `/home/runner/work` | `/home/runner/work`
DinD | - | `/runner/_work`
DinD | `/home/runner/work` | `/home/runner/work`

### Acceptance tests

I tested this with the following runners (DinD/normal with/without custom workdirs)

```yaml
# DinD runner with default working directory
kind: Runner
metadata:
  name: dind
spec:
  image: "reiniert/actions-runner-dind:v2.274.2"
  repository: reiniertimmer/actions-runner-controller
  labels:
    - dind
---
# DinD runner with customized working directory
kind: Runner
metadata:
  name: dind-wd
spec:
  image: "reiniert/actions-runner-dind:v2.274.2"
  repository: reiniertimmer/actions-runner-controller
  workDir: "/home/runner/work"
  labels:
    - dind-wd
---
# Default runner with default working directory
kind: Runner
metadata:
  name: normal
spec:
  image: "reiniert/actions-runner:v2.274.2"
  dockerdWithinRunnerContainer: false
  repository: reiniertimmer/actions-runner-controller
  labels:
    - normal
---
# Default runner with custom working directory
kind: Runner
metadata:
  name: normal-wd
spec:
  image: "reiniert/actions-runner:v2.274.2"
  dockerdWithinRunnerContainer: false
  repository: reiniertimmer/actions-runner-controller
  workDir: "/home/runner/work"
  labels:
    - normal-wd
```

Then I ran the following acceptance test workflow:

```yaml
on: push
name: Validate workdirs
jobs:
  build:
    runs-on: ${{ matrix.runs-on }}
    name: ${{ matrix.runs-on }}
    strategy:
      matrix:
        include:
          - runs-on: dind
            work-dir: /runner/_work
          - runs-on: dind-wd
            work-dir: /home/runner/work
          - runs-on: normal
            work-dir: /runner/_work
          - runs-on: normal-wd
            work-dir: /home/runner/work
    steps:
      - name: Check working directory
        run: |
          [ "$RUNNER_WORKSPACE" == "${{ matrix.work-dir }}/${{ github.event.repository.name }}" ]
```

This workflow runs a matrix test to validate the `$RUNNER_WORKSPACE` environment variable, which defaults to `/runner/_work/$REPO_NAME` but can be customized using the `workDir` parameter in the runner spec.

See the result on: https://github.com/reiniertimmer/actions-runner-controller/actions/runs/375208748/workflow

### DIFF
--- a/config/crd/bases/actions.summerwind.dev_horizontalrunnerautoscalers.yaml
+++ b/config/crd/bases/actions.summerwind.dev_horizontalrunnerautoscalers.yaml
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.2.4
+    controller-gen.kubebuilder.io/version: v0.3.0
   creationTimestamp: null
   name: horizontalrunnerautoscalers.actions.summerwind.dev
 spec:
--- a/config/crd/bases/actions.summerwind.dev_horizontalrunnerautoscalers.yaml
+++ b/config/crd/bases/actions.summerwind.dev_horizontalrunnerautoscalers.yaml
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.3.0
+    controller-gen.kubebuilder.io/version: v0.2.4
   creationTimestamp: null
   name: horizontalrunnerautoscalers.actions.summerwind.dev
 spec:
--- a/config/crd/bases/actions.summerwind.dev_runnerdeployments.yaml
+++ b/config/crd/bases/actions.summerwind.dev_runnerdeployments.yaml
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.2.4
+    controller-gen.kubebuilder.io/version: v0.3.0
   creationTimestamp: null
   name: runnerdeployments.actions.summerwind.dev
 spec:
@@ -407,12 +407,20 @@ spec:
                       properties:
                         limits:
                           additionalProperties:
-                            type: string
+                            anyOf:
+                              - type: integer
+                              - type: string
+                            pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                            x-kubernetes-int-or-string: true
                           description: 'Limits describes the maximum amount of compute resources allowed. More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/'
                           type: object
                         requests:
                           additionalProperties:
-                            type: string
+                            anyOf:
+                              - type: integer
+                              - type: string
+                            pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                            x-kubernetes-int-or-string: true
                           description: 'Requests describes the minimum amount of compute resources required. If Requests is omitted for a container, it defaults to Limits if that is explicitly specified, otherwise to an implementation-defined value. More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/'
                           type: object
                       type: object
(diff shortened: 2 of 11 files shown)


reiniertimmer:
I'll still have to do some additional testing to see if my test case with cache/artifact save & restore works as expected now

config/crd/bases/actions.summerwind.dev_horizontalrunnerautoscalers.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.3.0
+    controller-gen.kubebuilder.io/version: v0.2.4
```
mumoshu, line 7: nit: You'd better use a newer version of kubebuilder (0.3.0)

config/crd/bases/actions.summerwind.dev_runnerdeployments.yaml
```diff
@@ -404,20 +404,12 @@ spec:
                       properties:
                         limits:
                           additionalProperties:
-                            anyOf:
-                              - type: integer
-                              - type: string
-                            pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
-                            x-kubernetes-int-or-string: true
+                            type: string
```
mumoshu, line 407: Perhaps this regression(?) disappear once you upgrade your `controller-gen` binary. 0.3.0 should be automatically installed via [our make target](https://github.com/summerwind/actions-runner-controller/blob/eb002cfed45b8598b1ff825e2fe7020118f4346c/Makefile#L148-L165), after you `rm $GOBIN/bin/controller-gen` or `rm ~/go/bin/controller-gen`.

mumoshu:
CI failure seems to be due to controller-gen version mismatch

mumoshu:
@reiniertimmer Thanks for your effort! This seems great so far.

If possible, a basic example workflow definition yaml to test this scenario manually would be much appreciated, so that I can possible incorporate it into #168

reiniertimmer:
> CI failure seems to be due to controller-gen version mismatch

I'll look into it

config/crd/bases/actions.summerwind.dev_horizontalrunnerautoscalers.yaml
reiniertimmer: I wasn't sure what the issue was here, but Ive updated the version and now it seems to be fine

config/crd/bases/actions.summerwind.dev_runnerdeployments.yaml
reiniertimmer: I was on an old version, it did not update automatically (because it could still find a binary). Once I removed it, the proper controller-gen version was installed automatically. So, works!

reiniertimmer:
@mumoshu I've managed to run the manifest generator with the correct version. However, the CI still seems to be breaking somehow. Could you (or someone) clarify this?

mumoshu:
@reiniertimmer Thanks for your effort. I think you need to rerun `make manifests` and then git-add/commit changes under `charts/actions-runner-controller/crds/` as well

reiniertimmer:
> 
> 
> @reiniertimmer Thanks for your effort. I think you need to rerun `make manifests` and then git-add/commit changes under `charts/actions-runner-controller/crds/` as well

Thanks, I already did that, but my branch did not include the latest master branch.

Since a PR runs on a merge from master & branch, the test was failing. Once I found out that, the fix was fortunately easy 😄

reiniertimmer:
@mumoshu I updated the readme and ran some acceptance tests (and updated the PR description accordingly).

Is this something you can work with?

mumoshu (approved):
LGTM. Thanks a lot for the enhancement @reiniertimmer ☺️

PR #188, opened 2020-11-14
refactor/sync-period

callum-tait-pbx:
As discussed in https://github.com/summerwind/actions-runner-controller/issues/184 adding an explicit setting for the --sync-period argument and docs to go with it


mumoshu (approved):
LGTM. Thanks a lot for the fix @callum-tait-pbx ☺️

PR #191, opened 2020-11-16
add dockerEnabled option

Warashi:
Add dockerEnabled option for users who does not need docker and want not to run privileged container.
if `dockerEnabled == false`, dind container not run, and there are no privileged container.

Do the same as closed #96


mumoshu (approved):
LGTM. Thanks a lot for your continued effort @Warashi!

Note: I initially thought if the new flag is same as DockerdWithinRunnerContainer, but it isn't. DockerWithinRunnerContainer also add priviledges to the `runner` container, which DockerEnabled doesn't. https://github.com/summerwind/actions-runner-controller/blob/e613219a8980c88977f8231bedbe691b5caa3d9c/controllers/runner_controller.go#L368

callum-tait-pbx:
The new configuration option needs documenting still

Warashi:
oh, I forgot to document this feature.

Warashi:
Opened #193

PR #193, opened 2020-11-16
add dockerEnabled document

Warashi:
In #191 , I forgot to document new feature, dockerEnabled.
this PR documents the feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,6 +321,8 @@ spec:
         requests:
           cpu: "2.0"
           memory: "4Gi"
+      # If set to false, there are no privileged container but you cannot use docker. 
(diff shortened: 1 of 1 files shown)


callum-tait-pbx:
Is this worded right? `# If set to false, there are no privileged container but you cannot use docker.`. weird wordering. I would expect `and you cannot use docker` or `but you can use docker`.  @Warashi

Warashi:
Sorry for my poor English. I fixed as suggested.

mumoshu (approved):
@Warashi No need to say sorry 😄  Thanks a lot for the improvement!

PR #214, opened 2020-11-26
GetRegistrationToken: Remove 10 minute buffer to token expiration

ZacharyBenamram:
Theres been a few outstanding issues related to runner pods not deleting due to token updates blocking the reconcile function as well as too many calls being made to the github api (https://github.com/summerwind/actions-runner-controller/issues/77 and https://github.com/summerwind/actions-runner-controller/issues/206). 

I'm not sure of the original intention behind the buffer, but it can safely be removed all together and reduce the number of retries on the runner queue. 

Below is a graph of the total number of reconcile retries made before and after the change with 2 runner pods:
Screen Shot 2020-11-26 at 2 43 54 PM

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -84,7 +84,7 @@ func (c *Client) GetRegistrationToken(ctx context.Context, org, repo, name strin
 	key := getRegistrationKey(org, repo)
 	rt, ok := c.regTokens[key]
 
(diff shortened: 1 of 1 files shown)


mumoshu (approved):
@ZacharyBenamram I just read https://github.com/summerwind/actions-runner-controller/issues/77#issuecomment-733507143 - Impressed by your insight! Thanks for your efforts 🙏

Issue #195, opened 2020-11-16
Publish helm chart to a registry


ahmad-hamade:
Is there any plan to publish the helm chart to some sort of registry?

Maybe to [artifacthub](https://artifacthub.io/)?

mumoshu:
I'm looking for a solution that makes the process super easy. It would be best if it's matter of just passing `helm packate` result into another single command.

callum-tait-pbx:
It would be really great if we could perhaps make use of Helms chart releaser to package and release the chart onto github pages. This would be simple and self managing. I'll put together a PR? Once it's in pages people (like myself :D) can hook onto the releases allowing them to be sync with internal feeds.

https://github.com/marketplace/actions/helm-chart-releaser

pierluigilenoci:
@mumoshu any update?

ahmad-hamade:
There were 2 releases published so far for helm chart but I can't find a way to add helm repo as I'm getting the following error:

```
helm repo add runner https://github.com/summerwind/actions-runner-controller
Error: looks like "https://github.com/summerwind/actions-runner-controller" is not a valid chart repository or cannot be reached: failed to fetch https://github.com/summerwind/actions-runner-controller/index.yaml : 404 Not Found
```

@mumoshu Do you have any update on this?

donovanmuller:
@ahmad-hamade use:

```console
$ helm repo add runner https://summerwind.github.io/actions-runner-controller/
"runner" has been added to your repositories

$ helm search repo actions-runner-controller
NAME                            	CHART VERSION	APP VERSION	DESCRIPTION
runner/actions-runner-controller	0.2.0        	0.16.0     	A Kubernetes controller that operates self-host...
```

ahmad-hamade:
Thank you @donovanmuller. I was using the wrong repo URL.

I will be closing this issue now.

PR #219, opened 2020-12-01
feat: include Helm chart in pipeline and release

callum-tait-pbx:
Creating a draft PR to get initial comments. @mumoshu 

**Problem**
Currently, the Helm chart in the repo was manually created and is not part of the formal release. The static nature means changing templates 2 places and it also creates issues for me around automation as I don't want to have to monitor the repository for chart changes, nor do I want to come up with an complicated sync job to work around this. Attaching the built chart to the release would allow for a much simplier sync job for my internal feed.

**Solution**
Have the Helm chart produced as part of the CI and attached to the release so it is much easier to apply automation to as an external user of the solution. The need for this was highlighted in #195 

**Limitation**
- The work assumes all merges are done via a squash and merge operation as only the last git commit is checked to see if any files were changed within the `charts/` and `config/crd/bases` folder.
- The helm package is simply attached to the release. The more standard approach for releasing charts in github actions https://github.com/helm/charts-repo-actions-demo/blob/master/.github/workflows/release.yaml would require a lot more work as it does not support appending to an already existing release and automatically creates the release.
- The chart should only be built when there are changes to the charts folder itself as well as the CRDs. Ideally the entire chart should be defined once (including templates etc) however the rest of the solution doesn't change much so as an initial PR this limitation is OK and doesn't make it worse.

**Questions**
- @mumoshu the existing recipes just copies the CRD bases to the helm chart, what about the patches? What are they? Are they needed? Any suggestions on how to handle including them?

**Notes**
https://medium.com/@mattiaperi/create-a-public-helm-chart-repository-with-github-pages-49b180dbb417
https://jamiemagee.co.uk/blog/how-to-host-your-helm-chart-repository-on-github/

There are some interesting posts on how you can use GitHubs pages feature to use the repository itself as a Helm chart feed. To do this the pipeline would need to, in addition to the changes in this PR, update an `index.yaml` file and commit that into the history. This file would be stored under `Docs/` and github pages would be configured to serve the `Docs/` folder, any thoughts on this would be welcomed.

I have learnt I hate Makefiles


mumoshu:
> @mumoshu the existing recipes just copies the CRD bases to the helm chart, what about the patches?

If you're talking about kustomize patches, we won't need to take care on them, as we cant reuse them in Helm and all the equivalent customization should be possible via chart values and chart templates!

PR #220, opened 2020-12-02
feat: add helm package to release assets when changes occur

callum-tait-pbx:
Creating a draft PR to get initial comments. @mumoshu

**Problem**
Currently, the Helm chart in the repo was manually created and is not part of the formal release. The static nature means changing templates 2 places and it also creates issues for me around automation as I don't want to have to monitor the repository for chart changes, nor do I want to come up with an complicated sync job to work around this. Attaching the built chart to the release would allow for a much simplier sync job for my internal feed.

**Solution**
Have the Helm chart produced as part of the CI and attached to the release so it is much easier to apply automation to as an external user of the solution. The need for this was highlighted in #195

Helm linting is on on the test workflow and helm charts are included in the publish if they are found to have changed beteen the previous tag and the latest tag

**Limitations**

- The work assumes that all tags get published as a release as it gets the latest tag at the time (which will by the end of the workflow be the previous tag) and compares the changes between that tag and HEAD to see if any files were changed within the charts/ and config/crd/bases folder.
- The helm package is simply attached to the release. The more standard approach for releasing charts in github actions https://github.com/helm/charts-repo-actions-demo/blob/master/.github/workflows/release.yaml would require a lot more work as it does not support appending to an already existing release and automatically creates the release.
- The chart should only be built when there are changes to the charts folder itself as well as the CRDs. Ideally the entire chart should be defined once (including templates etc) however the rest of the solution doesn't change much so as an initial PR this limitation is OK and doesn't make it worse.

**Notes**
https://medium.com/@mattiaperi/create-a-public-helm-chart-repository-with-github-pages-49b180dbb417
https://jamiemagee.co.uk/blog/how-to-host-your-helm-chart-repository-on-github/

There are some interesting posts on how you can use GitHubs pages feature to use the repository itself as a Helm chart feed. To do this the pipeline would need to, in addition to the changes in this PR, update an index.yaml file and commit that into the history. This file would be stored under `Docs/` and github pages would be configured to serve the `Docs/` folder, any thoughts on this would be welcomed.

I have learnt I hate Makefiles


callum-tait-pbx:
I actually didn't mean to do those last 2 commits in thsi PR, I had the wrong VS code open 🤦 . I'll keep them seen as they are commited now. Would love a review when you get 10 @mumoshu

Makefile
```diff
@@ -34,9 +37,12 @@ endif
 all: manager
 
 # Run tests
-test: generate fmt vet manifests
+test: create-release-dir generate fmt vet manifests helm-lint
```
mumoshu, line 40:
```suggestion
test: generate fmt vet manifests helm-lint
```

I think we don't need `create-release-dir` dependency here.

If it's actually needed, I'd rather suggest move it to where it is depended. Like if `manifests` depends on `create-release-dir `, add `create-release-dir ` as `manifests`'s dependency.

Makefile
```diff
@@ -147,7 +164,7 @@ acceptance/tests:
 	acceptance/checks.sh
 
 # Upload release file to GitHub.
-github-release: release
+github-release: create-release-dir release
```
mumoshu, line 167:
```suggestion
github-release: release
```

`release` seems to depend on `create-release-dir` so I think we don't need to re-add the dependency here

charts/actions-runner-controller/values.yaml
```diff
@@ -38,6 +38,18 @@ securityContext: {}
...
+# Values need to be base64'ed in values.yaml
+githubAuthenticationSchema:
+  app:
+    enabled: false
+    github_app_id: ""
+    github_app_installation_id: ""
+    github_app_private_key: ""
+  token:
+    enabled: false
+    github_token: ""
```
mumoshu, line 51:
```suggestion
# Set `enabled: true` and either specify github_app_* or github_token to let the chart
# create the K8s secret required by the controller to authenticate against GitHub API.
# All the values must NOT be base64-encoded, as the chart template base64-encodes values for you.
authSecret:
  enabled: false
  github_app_id: ""
  github_app_installation_id: ""
  github_app_private_key: ""
  github_token: ""
```

I slightly prefer modeling this around the fact that we just create a single K8s secret containing various credentials.

That would allow us to easily enhance this secret/chart in the future for e.g. automated docker-login to the dockerhub to avoid pull rate limit.

mumoshu: In addition to that, `manager_secret.yaml` can be simplified like this:

```
{{- if or .Values.authSecret.enabled }}
apiVersion: v1
kind: Secret
metadata:
  name: controller-manager
type: Opaque
stringData:
{{ toYaml .Values.authSecret | indent 2 }}
{{- end }} 
```

// I'm aware of the redundant `enabled: true` added to the stringData but had no quick idea to remove that 😃

Makefile
```diff
@@ -3,10 +3,13 @@ VERSION ?= latest
 # From https://github.com/VictoriaMetrics/operator/pull/44
 YAML_DROP=$(YQ) delete --inplace
 YAML_DROP_PREFIX=spec.validation.openAPIV3Schema.properties.spec.properties
-
+ 
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
+# This is ultimately the previous tag however at the time of running the Makefile it would be the latest
+GITHUB_LATEST_TAG := $(strip $(shell git tag --sort=committerdate | tail -1))
```
mumoshu, line 11: At the time of running `release.yml` workflow, I think we already have the latest version tag. Probably we'd better use `tail -n 2 | head -n 1`?

callum-tait-pbx: ah, maybe this is my lack of Makefile knowledge, are the global variables not ran before any targets are ran? So we run `make github-release` and the global variables are populated first? As a result they are ran before the `ghr ${VERSION} release/` step happens meaning the new release hasn't be created yet?

charts/actions-runner-controller/values.yaml
callum-tait-pbx: yeh, this makes more sense than the way I was doing it, I'll make the change

callum-tait-pbx: Patched, resolve if happy

callum-tait-pbx:
Moved the secret work into https://github.com/summerwind/actions-runner-controller/pull/240, I'll revert the secret commits on this branch

Makefile
mumoshu: AFAIK our usage of GitHub actions workflow ensures that the job is triggered only after we manually tagged the new release.

So, the global variable is populated first, earlier than the tasks. But it still `tail -1` would return the tag just created manually, where you want seems like the previous one of that.

pierluigilenoci:
@callum-tait-pbx @mumoshu any update?

callum-tait-pbx:
tbh at this point I was waiting to see what happens with https://github.com/summerwind/actions-runner-controller/pull/257. I went down the Makefile route to just fall in line with the convention of the repo. I'm not actually a fan of going that direction at all so I would probably go with the https://github.com/summerwind/actions-runner-controller/pull/257 PR unless @mumoshu is keen on sticking with Makefiles, I was waiting for his comments tbh. The only thing I would changed about https://github.com/summerwind/actions-runner-controller/pull/257 is the removal of the CRDs held in the chart and instead jsut copy them from the config as atm we have to maintain 2 copies of the same CRDs for no real reason.

callum-tait-pbx:
This is no longer needed now that https://github.com/summerwind/actions-runner-controller/pull/257 has been merged, closing.

PR #223, opened 2020-12-04
Horizontal Pod Autoscaler - Busy Runner Scheme

ZacharyBenamram:
Providing an alternative to the current hpa scheme. The TotalNumberOfQueuedAndInProgressWorkflowRuns scheme requires that the hpa be aware of all the repositories that are currently managed by actions - this doesnt scale well in larger organizations where multiple teams are enrolling into actions - they must be aware that the hpa spec be updated. 

The PercentageRunnersBusy scheme is a simple autoscaling scheme that retrieves all action runners from github, finds the subset of runners that exist in the hpa namespace by comparing it against the runner resources, and then evaluates the scale up / scale down condition.

This scheme also addresses the issue where an organization has various environments + multiple hpas - where each hpa should be responsible for scaling only the runners the shared namespace.
 
Screen Shot 2020-12-04 at 3 46 23 PM

### DIFF
--- a/controllers/autoscaling.go
+++ b/controllers/autoscaling.go
@@ -10,6 +10,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	defaultScaleUpThreshold   = 0.8
+	defaultScaleDownThreshold = 0.3
+	defaultScaleUpFactor      = 1.3
+	defaultScaleDownFactor    = 0.7
+)
+
 func (r *HorizontalRunnerAutoscalerReconciler) determineDesiredReplicas(rd v1alpha1.RunnerDeployment, hra v1alpha1.HorizontalRunnerAutoscaler) (*int, error) {
 	if hra.Spec.MinReplicas == nil {
 		return nil, fmt.Errorf("horizontalrunnerautoscaler %s/%s is missing minReplicas", hra.Namespace, hra.Name)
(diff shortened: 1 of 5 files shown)


controllers/autoscaling.go
```diff
@@ -135,3 +142,68 @@ func (r *HorizontalRunnerAutoscalerReconciler) determineDesiredReplicas(rd v1alp
...
+	numRunnersBusy := 0
+	for _, runner := range runners {
+		if _, ok := runnerMap[*runner.Name]; ok && runner.GetBusy() {
+			numRunnersBusy++
+		}
+	}
+
+	var desiredReplicas int
+	fractionBusy := float64(numRunnersBusy) / float64(numRunners)
+	if fractionBusy >= 0.8 {
```
ZacharyBenamram, line 177: it might make sense if the scale up limit + scale down limit (0.8, 0.3) lives in the hpa spec - as well as the scale up / scale down factor (1.3 + 0.7)

ZacharyBenamram:
https://github.com/summerwind/actions-runner-controller/pull/213 to cross link another active hpa PR

ahmad-hamade:
Hello @ZacharyBenamram, I have a small question and I appreciate it if you can help me to understand more about this.

Assume I have 3 runners (running in separate EKS clusters) that are added to a specific Github repository, How does the controller know which runner deployment to scale up/down?

I think it supposed to be based on the runner labels maybe? but I guess it will always scale all the 3 runners (due to a workflow that is in the queue pending).

ZacharyBenamram:
@ahmad-hamade this autoscaling scheme will scale the runners that you have defined in your runner-deployment. If you are running 3 runners all in separate EKS clusters - i am also assuming you are also running 3 separate controllers. If that is the case, each horizontal runner autoscaler that you have defined in each cluster will be responsible for scaling that clusters runners.

ahmad-hamade:
Thank you @ZacharyBenamram 👍

controllers/autoscaling.go
```diff
@@ -135,3 +150,103 @@ func (r *HorizontalRunnerAutoscalerReconciler) determineDesiredReplicas(rd v1alp
...
+	for _, runner := range runners {
+		if _, ok := runnerMap[*runner.Name]; ok && runner.GetBusy() {
+			numRunnersBusy++
+		}
+	}
+
+	var desiredReplicas int
+	fractionBusy := float64(numRunnersBusy) / float64(numRunners)
+	if fractionBusy >= scaleUpThreshold {
+		scaleUpReplicas := int(float64(numRunners)*scaleUpFactor + 0.5)
```
mumoshu, line 221:
```suggestion
		scaleUpReplicas := int(float64(numRunners)*scaleUpFactor)
```

Can we remove the magic number 0.5 out of this calculation? I think the defaultScaleUpFactor of `1.3` is enough, and you can optionally add that to the default value so that it becomes `defaultScaleUpFactor      = 1.8`.

mumoshu (approved):
@ZacharyBenamram Awesome work! Thank you so much for your contribution ☺️ 

I had [a comment regarding the defaultScaleUpFactor](https://github.com/summerwind/actions-runner-controller/pull/223/files#r541803753) so it would be great if you could address that in comment or another pull request.

PR #227, opened 2020-12-07
Don't compare runner connetion token at restart need check

### LINKED ISSUE #143: Runner restarts every hour

Hi-Fi:
Fixes #143


mumoshu (approved):
LGTM. Super glad to see we can finally fix this! Thank you so much for your continued efforts @Hi-Fi

mumoshu:
Please watch out for https://github.com/summerwind/actions-runner-controller/pull/226 and wish I don't break anything merging that into this change 😉

PR #232, opened 2020-12-10
feat: adding helm repo config for dind sidecar

callum-tait-pbx:


charts/actions-runner-controller/values.yaml
```diff
@@ -8,9 +8,10 @@ syncPeriod: 10m
 
 image:
   repository: summerwind/actions-runner-controller
+  # Overrides the manager image tag whose default is the chart appVersion if the tag key is commented out
+  tag: "latest"
+  dindSidecarRepositoryAndTag: "docker:dind"
```
damacus, line 13: I wonder if we should pin this to a particular so people know which version of Docker they're getting?

callum-tait-pbx: Depends on the maintainer I guess, I didn't adjust it as I was mainly hoping to get the customation aspect in the repo. The the end user can apply version pinning as they see fit keeping maintenence lower for the actual repo itself. I know  @mumoshu is keen to keep the maintenence overhead low and so he may prefer for the default to be get latest.

For my perspective as long as the end user can customise the image then it's good enough tbh.

mumoshu: I'm fine with this default :+1:
A nice-to-have would be to add a recommendation in our README to use the specific tag for prod env tho.

Also in the future we may migrate to specific tags in the default values, but that would require more man power to maintain, as we need to track new tags and test it, cut a new helm chart release with the new tag as the default, etc.

mumoshu (approved):
LGTM. Thanks for the addition @callum-tait-pbx ☺️

PR #233, opened 2020-12-11
Sets the docker image to use 20.10.0-dind

damacus:
As this is not yet configurable and imagePullPolicy is not set, new versions of the image do not get pulled.

### DIFF
--- a/main.go
+++ b/main.go
@@ -36,7 +36,7 @@ import (
 
 const (
 	defaultRunnerImage = "summerwind/actions-runner:latest"
-	defaultDockerImage = "docker:dind"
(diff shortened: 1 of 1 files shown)


mumoshu:
@damacus Ah sorry I missed this... FWIW I guess we'd better make it configurable by extending RunnerSpec.

Perhaps `DockerImage` and `DockerImagePullPolicy` should be consistent with existing fields for runners.

https://github.com/summerwind/actions-runner-controller/blob/0783ffe98901c6ec5be9ac8acaa92ee7dbe3bdd3/api/v1alpha1/runner_types.go#L57-L60

PR #235, opened 2020-12-14
Autoscaling: Percentage runners busy - remove magic number used for round up

ZacharyBenamram:
Removing 0.5 used to round up any scale-up calculations to the next highest integer. This can also be accomplished by re-setting the scale-up factor.

### DIFF
--- a/controllers/autoscaling.go
+++ b/controllers/autoscaling.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


controllers/autoscaling.go
```diff
@@ -218,7 +218,7 @@ func (r *HorizontalRunnerAutoscalerReconciler) calculateReplicasByPercentageRunn
 	var desiredReplicas int
 	fractionBusy := float64(numRunnersBusy) / float64(numRunners)
 	if fractionBusy >= scaleUpThreshold {
-		scaleUpReplicas := int(float64(numRunners)*scaleUpFactor + 0.5)
+		scaleUpReplicas := int(float64(numRunners) * scaleUpFactor)
```
mumoshu, line 221: > Removing 0.5 used to round up any scale-up calculations to the next highest integer. This can also be accomplished by re-setting the scale-up factor.

Ah! That makes sense. Can we use `math.Ceil(float64(numRunners) * scaleUpFactor)` instead?

mumoshu (approved):
@ZacharyBenamram Thanks for your support! LGTM 👍

PR #237, opened 2020-12-16
Adding readme for runner busy hpa scheme

ZacharyBenamram:
adding readme for runner busy hpa scheme

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,11 +267,7 @@ spec:
     - summerwind/actions-runner-controller
 ```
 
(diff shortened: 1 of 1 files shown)


callum-tait-pbx:
Does this mean you don't need to maintain a list of repository names with this schema? You can create a RunnerDeployment against an organisation and use a hpa with this schema negating the need to maintain a list? If so I think that is worth pointing out explicitly.

mumoshu (approved):
LGTM. Thank you so much for your continued effort @ZacharyBenamram ☺️

PR #254, opened 2021-01-13
feat/helm: add manager secret to Helm chart

callum-tait-pbx:
Added the secret to Helm, worked on our cluster

**Notes**
`toString` is required to ensure we can base64 encode all secrets, incuding booleans. It doesn't however seem to work as I would expect, the  `github_app_id` and `github_app_installation_id` values had to be quoted in your `values.yaml` or you get the error `Error: Client creation failed. authentication failed: could not read private key: open : no such file or directory`.

Fixes https://github.com/summerwind/actions-runner-controller/issues/241


mumoshu (approved):
LGTM. Thanks for the addition ☺️

PR #259, opened 2021-01-18
Adding --name-space parameter in GitHub Enterprise Server config example

jonico:
* when setting a GitHub Enterprise server URL without a namespace, an error occurs: "error: the server doesn't have a resource type "controller-manager"
* setting default namespace "actions-runner-system" makes the example work out of the box


mumoshu (approved):
Good catch. Thanks for fixing it!

PR #260, opened 2021-01-18
Update custom container example

jonico:
* use latest instead of outdated version
* use sudo for package install (required)
* use sudo for package meta data removal (required)


mumoshu (approved):
Thanks for fixing it ☺️

PR #261, opened 2021-01-18
GitHub runner groups have to be created first

jonico:
* in contrast to runner labels, GitHub runner groups are not automatically created


mumoshu (approved):
LGTM. Thanks for the improvement @jonico ☺️

PR #267, opened 2021-01-21
minReplicas <= desiredReplicas <= maxReplicas

jonico:
* ensure that minReplicas <= desiredReplicas <= maxReplicas no matter what
* before this change, if the number of runners was much larger than the max number, the applied scale down factor might still result in a desired value > maxReplicas
* if for resource constraints in the cluster, runners would be permanently restarted, the number of runners could go up more than the reverse scale down factor until the next reconciliation round, resulting in a situation where the number of runners climbs up even though it should actually go down
* by checking whether the desiredReplicas is always <= maxReplicas, infinite scaling up loops can be prevented


jonico:
Test error seems to be related to the fact that I created a PR for the code, as git diff only complains about changes in the code (not the manifest files)

mumoshu (approved):
This makes sense a lot. Thanks for sharing your awesome work 👍

PR #273, opened 2021-01-25
Actually use 'dockerdContainerResources' to set resources on the dind…

### LINKED ISSUE #242: dockerdContainerResources doesn't set resource limits for docker container
I've been following the Readme, in additional tweaks section to beef up the docker containers resources(to solve some customers OOM issues).  I'm using a forked version of the summerwind runner container(as well as the official one to test) and latest version of the controller(version 0.16 at the time of this writing).  Setting the resource limits on the Runner container work fine,  but nothing for the docker container.

```
kind: RunnerDeployment
metadata:
  namespace: actions-runner-system
spec:
  replicas: 1
  template:
    spec:
      organization: myOrg
      dockerEnabled: true     #(true|false) works fine.
      dockerdWithinRunnerContainer: false #(true|false) works fine.
      dockerdContainerResources:  # Does nothing, no limits get set.
        limits:
          cpu: "4.0"
          memory: "8Gi"
        requests:
          cpu: "2.0"
          memory: "4Gi"
```

thrawny:
… container

Fixes #242

### DIFF
--- a/controllers/runner_controller.go
+++ b/controllers/runner_controller.go
@@ -504,6 +504,7 @@ func (r *RunnerReconciler) newPod(runner v1alpha1.Runner) (corev1.Pod, error) {
 			SecurityContext: &corev1.SecurityContext{
 				Privileged: &privileged,
 			},
+			Resources: runner.Spec.DockerdContainerResources,
 		})
 
 	}


mumoshu (approved):
Awesome! Thanks for fixing it @thrawny ☺️

martinm82:
Hi all, we are encountering an issue where setting resource limits are not taken into account when using the DinD approach. 

I am wondering whether the expected behaviour got broken by the following change:

https://github.com/actions-runner-controller/actions-runner-controller/commit/6f130c2db5e572e667455b647dd8c6f81dfe8d41

The original change did not had a condition when to use the resource limit whereas the commit above adds a condition in case the `DockerdWithinRunnerContainer` is null or false.

mumoshu:
@martinm82 Hey! Which setup are you referring, runner container + the dind sidecar, or a hybrid container of runner + dind?

mumoshu:
AFAICS, the changes made in #696 is valid, because DockerdContainerResources makes sense only when you configured the runner to have the dind sidecar.

martinm82:
I am referring to runner + dind and had the impression that for this case the resources would as well be supported.

mumoshu:
@martinm82 Thanks! But I'm confused. #696 should have nothing to do with the example introduced in https://github.com/actions-runner-controller/actions-runner-controller/issues/277#issuecomment-769272816.
If I'm not missing anything, it should be https://github.com/actions-runner-controller/actions-runner-controller/pull/1383 if it's broken.
Probably you could instead use something like the below as a workaround that works today?

```
spec:
  template:
    spec:
      containers:
      - name: runner
        resources:
          # your resource requests/limits here
```

PR #275, opened 2021-01-27
docs: fix install installation method

clementlecorre:
Hi, 

The link doesn't work, I think this file doesn't exist anymore.... I replaced it by the installation via helm.

Thanks


mumoshu:
@clementlecorre Ah thanks for reporting! The manifests url must be `https://github.com/summerwind/actions-runner-controller/releases/v0.16.1/download/actions-runner-controller.yaml`, i.e. `v0.16.1` rather than `latest`, as the latest gh release can be one for our helm charts now.

As you've suggested, adding the instructions for helm usage is nice. Let me add the manifests way back to your branch and merge this.

PR #282, opened 2021-01-31
feat: HorizontalRunnerAutoscaler Webhook server

mumoshu:
This introduces a Webhook server that responds GitHub `check_run`, `pull_request`, and `push` events by scaling up matched HorizontalRunnerAutoscaler by 1 replica. This allows you to immediately add "resource slack" for future GitHub Actions job runs, without waiting next sync period to add insufficient runners.

HorizontalRunnerAutoscaler.Spec has been extended to make this behavior configurable.

## Example 1: Scale up on each `check_run` event

> Note: This should work almost like https://github.com/philips-labs/terraform-aws-github-runner

To scale up replicas of the runners for `example/myrepo` by 1 for 5 minutes on each `check_run`, you write manifests like the below:

```yaml
kind: RunnerDeployment:
metadata:
   name: myrunners
spec:
  repository: example/myrepo
---
spec:
  scaleTargetRef:
    name: myrunners
  scaleUpTrigggers:
  - githubEvent:
      checkRun:
        types: ["created"]
        status: "queued"
    amount: 1
    duration: "5m"
```

## Example 2: Scale on each `pull_request` event against `develop` or `main` branches

```yaml
kind: RunnerDeployment:
metadata:
   name: myrunners
spec:
  repository: example/myrepo
---
spec:
  scaleTargetRef:
    name: myrunners
  scaleUpTrigggers:
  - githubEvent:
      pullRequest:
        types: ["synchronize"]
        branches: ["main", "develop"]
    amount: 1
    duration: "5m"
```

See ["activity types"](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request) for the list of valid values for `scaleUpTriggers[].githubEvent.pullRequest.types`.

## Implementation

First of all, this feature is highly inspired by https://github.com/philips-labs/terraform-aws-github-runner. I appreciate authors and maintainers of terraform-aws-github-runner for sharing their awesome ideas and work as opensource.

The biggest difference would be that terraform-aws-github-runner can manage one set of runners per deployment, where actions-runner-controller with this feature can manage as many sets of runners as you declare with HorizontalRunnerAutoscaler and RunnerDeployment pairs.

On each GitHub event received, the webhook server searches for repository-wide and organizational runners from the cluster and searches for the single target to scale up.

> Note that searches are executed using controller-runtime's internal search index, so it doesn't end up DoS attacking your K8s API server.

Then the webhook server tries to match `HorizontalRunnerAutoscaler.Spec.ScaleUpTriggers[].GitHubEvent.[CheckRun|Push|PullRequest]` against the event and if it finds only one HRA, it considers that as the scale target.

If none or two or more targets are found for repository-wide runners, it does the same on organizational runners.

The webhook server scales up the HorizontalRunnerAutoscaler by adding an array entry to `HRA.Spec.CapacityReservations[]` which is also added in this PR. The HRA controller watches the change and reconciles the HRA resource by updating the scale target(RunnerDeployment)'s desired replicas.

The `CapacityReservations` item (we call it "capacity reservation") is composed of `Name`, `ExpirationTime`, and `Replicas` fields. Each item is uniquely identified by `Name`. Each capacity reservation denotes that you want to "temporarily" add the number of `Replicas` "in addition" to the desired number of replicas computed by querying GitHub API. "Temporarily" because it is valid only until `ExpirationTime` is met.


controllers/horizontalrunnerautoscaler_controller.go
```diff
@@ -117,7 +141,37 @@ func (r *HorizontalRunnerAutoscalerReconciler) Reconcile(req ctrl.Request) (ctrl
...
+			updated = hra.DeepCopy()
+		}
+
+		var cacheEntries []v1alpha1.CacheEntry
+
+		for _, ent := range updated.Status.CacheEntries {
+			if ent.ExpirationTime.Before(&metav1.Time{Time: now}) {
+				cacheEntries = append(cacheEntries, ent)
+			}
+		}
```
jonico, line 157: @mumoshu: I am having trouble understanding the intent of this loop. It seems to go through all expired keys and adds them to an array but this array is not used afterwards any more.

controllers/horizontalrunnerautoscaler_controller.go
```diff
@@ -117,7 +141,37 @@ func (r *HorizontalRunnerAutoscalerReconciler) Reconcile(req ctrl.Request) (ctrl
...
+			cacheDuration = r.CacheDuration
+		} else {
+			cacheDuration = 10 * time.Minute
+		}
+
+		updated.Status.CacheEntries = append(updated.Status.CacheEntries, v1alpha1.CacheEntry{
+			Key:            v1alpha1.CacheEntryKeyDesiredReplicas,
+			Value:          *replicas,
+			ExpirationTime: metav1.Time{Time: time.Now().Add(cacheDuration)},
+		})
```
jonico, line 171: if I understand this code correctly, it keeps adding to the status of the horizontal runner autoscaler until it finally exceeds the limits of the etcd storage backend, see #346 . Was your intent to remove the expired cache entries ☝️ ?

jonico (commented):
@mumoshu: After all my auto scalers started to show the accumulation / etcd storage extended problem (#346 ), I looked into this code and think you may have forgotten to delete updated entries from the cache so that it just adds up

controllers/horizontalrunnerautoscaler_controller.go
mumoshu: @jonico Ah! Good catch. Definitely, my intent was to remove the expired entries

Nuru:
@mumoshu Please add some documentation on how to set up the webhook on the GitHub side:
- org vs repo webhook
- which events to configure
- setting "Content type" to `application/json`
- Configuring a Secret (if that is supported)

racker7:
Hi, your doc https://github.com/actions-runner-controller/actions-runner-controller#webhook-driven-scaling
mentions "workflow_job" and points here.
However, its not recognized and its not mentioned on this page.
Has "workflow_job" and/or "workflowJob" webhook event type been implemented, and if not can it be added please?

```
$ kubectl -n myns apply -f test/deploy-hra-webhook.yaml                                                                      
error: error validating "test/deploy-hra-webhook.yaml": error validating data: ValidationError(HorizontalRunnerAutoscaler.spec.scaleUpTriggers[0].githubEvent): unknown field "workflowJob" in dev.summerwind.actions.v1alpha1.HorizontalRunnerAutoscaler.spec.scaleUpTriggers.githubEvent;
```

This also fails with "workflow_job", or "pullRequest", but it succeeds with "checkRun" and "push".

PR #284, opened 2021-02-01
bug-fix: patched dir owned by runner

alex-mozejko:
in relation to:

https://github.com/summerwind/actions-runner-controller/issues/283

### DIFF
--- a/runner/Dockerfile
+++ b/runner/Dockerfile
@@ -83,7 +83,7 @@ RUN echo AGENT_TOOLSDIRECTORY=/opt/hostedtoolcache > .env \
   && chmod g+rwx /opt/hostedtoolcache
 
 COPY entrypoint.sh /
-COPY patched $RUNNER_ASSETS_DIR/patched
+COPY --chown=runner:docker patched $RUNNER_ASSETS_DIR/patched
 
 USER runner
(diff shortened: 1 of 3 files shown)


alex-mozejko:
commit e719724 added so that images are always built using the latest possible actions/runner release. 

could be nice to build this daily so no need to keep bumping the version

runner/Dockerfile
```diff
@@ -1,7 +1,6 @@
 FROM ubuntu:18.04
 
 ARG TARGETPLATFORM
-ARG RUNNER_VERSION=2.274.2
```
mumoshu, line 4: Thanks for the fix! But could we keep this build-arg as this is required to inject the runner version from within our github actions workflow?

mumoshu: > could be nice to build this daily so no need to keep bumping the version

Sorry I missed your comment. I got your intention, but doing so would prevent us from tagging container images with their RUNNER_VERSION.

I'd certainly investigate a way to automate it, probably reusing your snippet for getting the latestest available runner version. But it should be done in a separate PR!

mumoshu (approved):
LGTM on the `--chown` addition. Also, please allow me to replicate the same fix onto the dockerd-within-runner-container variant of your Dockerfile too!

Thanks for your contribution, @alex-mozejko ☺️

PR #286, opened 2021-02-01
Added github.sha to DockerHub push

ap0phi5:
Resolves https://github.com/summerwind/actions-runner-controller/issues/285.

Adds a reference to `GITHUB_RUN_NUMBER` so that tags are generated dynamically. https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables

### DIFF
--- a/.github/workflows/build-runner.yml
+++ b/.github/workflows/build-runner.yml
@@ -60,5 +60,5 @@ jobs:
             RUNNER_VERSION=${{ env.RUNNER_VERSION }}
             DOCKER_VERSION=${{ env.DOCKER_VERSION }}
           tags: |
-            ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:v${{ env.RUNNER_VERSION }}
+            ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:v${{ env.RUNNER_VERSION }}-${{ github.run_number }}
             ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:latest
--- a/.github/workflows/build-runner.yml
+++ b/.github/workflows/build-runner.yml
@@ -61,5 +61,5 @@ jobs:
             DOCKER_VERSION=${{ env.DOCKER_VERSION }}
           tags: |
             ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:v${{ env.RUNNER_VERSION }}
-            ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:v${{ env.RUNNER_VERSION }}-${{ github.sha }}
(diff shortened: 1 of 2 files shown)


ap0phi5:
There - the resultant docker tags would look like this:
`summerwind/actions-runner:v2.275.1-104`
`summerwind/actions-runner-dind:v2.275.1-104`
@mumoshu You still would need to modify 2.275.1 to the file, but what do you think...?

callum-tait-pbx:
Would it not be better to suffix the tag with the git hash at that point in time so there is a relation between the code running and the continer itself? It's relation to the workflow run version isn't super revealing no?

ap0phi5:
Good shout - I did think about that, though the difference there is that `github.run_number` will create ordered output, as opposed to the hash.

.github/workflows/build-runner.yml
```diff
@@ -60,5 +60,5 @@ jobs:
             RUNNER_VERSION=${{ env.RUNNER_VERSION }}
             DOCKER_VERSION=${{ env.DOCKER_VERSION }}
           tags: |
-            ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:v${{ env.RUNNER_VERSION }}
+            ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:v${{ env.RUNNER_VERSION }}-${{ github.run_number }}
```
mumoshu, line 63: In addition to the new immutable tag, could you also keep the older mutable one, too?

ap0phi5:
Alright- I've taken your comments onboard @callum-tait-pbx @mumoshu and made the two updates. In retrospect, ordering on dockerhub is by means of timestamp anyway so I don't see that as being a problem; ive switched to the SHA for the immutable tag.
Output is...

```summerwind/actions-runner:v2.276.1
summerwind/actions-runner:v2.276.1-70a717086ae2686df369d0dc091ec0acb9e89012
summerwind/actions-runner:latest```

cool?

.github/workflows/build-runner.yml
ap0phi5: Resolved, see comment below. 👍

callum-tait-pbx:
Probably best to use the the git short SHA instead?

mumoshu:
@ap0phi5 Thanks for your efforts! This looks perfect now. Did you accidentally mark this PR as draft, of is there any remaining task to be done before merging?

ap0phi5:
> @ap0phi5 Thanks for your efforts! This looks perfect now. Did you accidentally mark this PR as draft, of is there any remaining task to be done before merging?

Hi @mumoshu - I'm just sorting the short form of the SHA - using :: notation seems not to work.

ap0phi5:
Alright, looks sound.
Tag outputs will look like:
 summerwind/actions-runner:v2.276.1
 summerwind/actions-runner:v2.276.1-1ed3238
 summerwind/actions-runner:latest

mumoshu (approved):
Awesome! Thank you so much for your efforts @ap0phi5 ☺️

PR #294, opened 2021-02-06
chore: bumping Helm chart version

callum-tait-pbx:
Previous Helm chart release failed in master as we didn't bump the chart after making changes that impacted the Helm chart

### DIFF
--- a/charts/actions-runner-controller/Chart.yaml
+++ b/charts/actions-runner-controller/Chart.yaml
@@ -15,12 +15,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.2.1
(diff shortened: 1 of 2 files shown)


charts/actions-runner-controller/Chart.yaml
```diff
@@ -15,12 +15,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.2.1
+version: 0.3.1
```
mumoshu, line 18: Out of curiosity, what would be the reason you chose 0.3.1 rather than 0.3.0? 😃

mumoshu (approved):
LGTM. Thanks for the fix!

callum-tait-pbx:
You're right! That was a mistake, oh well, I'll bump it correctly next time 🙃

PR #296, opened 2021-02-07
chore: bumping helm chart sematically

callum-tait-pbx:
Chart verison needs bumping again, I did it right this time 😅

### DIFF
--- a/charts/actions-runner-controller/Chart.yaml
+++ b/charts/actions-runner-controller/Chart.yaml
@@ -15,12 +15,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.3.1
+version: 0.4.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 0.17.0
+appVersion: 0.18.0
 
 home: https://github.com/summerwind/actions-runner-controller
 
--- a/charts/actions-runner-controller/Chart.yaml
+++ b/charts/actions-runner-controller/Chart.yaml
@@ -17,11 +17,6 @@ type: application
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
 version: 0.4.0
 
-# This is the version number of the application being deployed. This version number should be
-# incremented each time you make changes to the application. Versions are not expected to
-# follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 0.18.0
-
 home: https://github.com/summerwind/actions-runner-controller
 
 sources:


callum-tait-pbx:
I've removed the app version config as it isn't really a thing with this chart. The underlying container version defines what app version is being ran, this can be configured in the chart and so is independant of the chart version itself. 


Lints OK outside of the deprecating errors, these are highlighted as needing bumping on this issue and should probably be looked at sooner rather than later, this is a Helm problem though https://github.com/summerwind/actions-runner-controller/issues/144:
```shell
==> Linting actions-runner-controller/
[INFO] Chart.yaml: icon is recommended
[WARNING] templates/webhook_configs.yaml: admissionregistration.k8s.io/v1beta1 MutatingWebhookConfiguration is deprecated in v1.16+, unavailable in v1.22+; use admissionregistration.k8s.io/v1 MutatingWebhookConfiguration
[WARNING] templates/webhook_configs.yaml: admissionregistration.k8s.io/v1beta1 ValidatingWebhookConfiguration is deprecated in v1.16+, unavailable in v1.22+; use admissionregistration.k8s.io/v1 ValidatingWebhookConfiguration
```

callum-tait-pbx:
@mumoshu  Just noticed you use the `.Chart.AppVersion` in your webhook work. Could you just use the charts `values.yaml` as the source for the default value? The AppVersion in the chart is super misleading as the app version isn't really tied to the chart at all so is probably best removed entirely to avoid confusion?

It's also super confusing as to what the app version actually is at any given time. Is it the controller container version? Is it the kube-proxy container version? Is it the dind container verison? It's a bit meaningless for this project with regards to Helm as it doesn't tell you anything about what version of the project is actually being ran because the version can be changed without deploying a new chart by deploying a new config.

callum-tait-pbx:
@mumoshu any thoughts on this in relation to https://github.com/summerwind/actions-runner-controller/issues/302 and https://github.com/summerwind/actions-runner-controller/issues/144

mumoshu:
@callum-tait-pbx Hey! I was thinking that `appVersion` could be used as the minimum controller version that is supposed to work with that version of the chart. Without it, we have no way to say which chart is supported to work with which version of the controller at all, which might make user support harder.

`appVersion` can be a few versions behind the latest controller release, but it will still serve the purpose.

Thoughts?

callum-tait-pbx:
I'm totally fine with keeping the `appVersion` if you want, I removed it because the default comment that is included if you scafold a chart file is:

```yaml
# This is the version number of the application being deployed. This version number should be
# incremented each time you make changes to the application. Versions are not expected to
# follow Semantic Versioning. They should reflect the version the application is using.
```

This obviously isn't true for this project as the `values.yaml` is defining what version of the application is deployed. Haoppy to re-add it if you think it's needed and jsut include a comment that says somethig like

```yaml
# This is the minimum controller version that this version of the chart supports.
# Increment only when the minimum version has changed.
```

This just might be confusing to anyone who is used to Helm as they won't see that comment if they are deploying from this repository and so will assume the `appVersion` is the normal definition making `helm list` commands confusing and probably wrong.

Let me know if you want it re-added, happy to do so. I would however shy away from using it as the source version for any aspect of the templating and instead rely on the `values.yaml` file for default values. So I would remove the ` default (cat "v" .Chart.AppVersion` entries in the templates. If you disagree I'm happy to leave them in, if not let me know and I'll include the removal of references to the AppVersion as part of this PR?

mumoshu (approved):
Makes sense! Let's go ahead without appVersion. In the future, we may revisit once anyone asks about which chart version supports which version of controller - but I believe we have no strong reason to do so right now 😃 

Thanks for the detailed feedback and the fix, @callum-tait-pbx!

PR #306, opened 2021-02-13
ci: create workflow for runner releases

callum-tait-pbx:
@mumoshu would something like this resolve https://github.com/summerwind/actions-runner-controller/issues/302 ? Do you have any thoughts on if we need to provide mutable and immutable canary builds, tbh I didn't think so but I wasn't too sure so kept them in.

### DIFF
--- a/.github/workflows/build-and-release-runners.yml
+++ b/.github/workflows/build-and-release-runners.yml
@@ -1,3 +1,5 @@
+name: Build and Release Runners
+
 on:
   pull_request:
     branches:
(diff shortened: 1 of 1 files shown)


mumoshu:
@callum-tait-pbx Whoa, good job! Yes, the mutable canary tag seems like exactly what I had in my mind. I don't have concrete use-case for immutable ones, but having them isn't bad.

Apparently this is still marked as draft, but can we proceed by merging this and fix it later?
If there's anything you wanted to do before proceeding, I'll hold on.

callum-tait-pbx:
Converted from draft @mumoshu

mumoshu (approved):
LGTM. Thanks as always for your contribution @callum-tait-pbx ☺️

PR #307, opened 2021-02-14
Introduce pod deletion timeout and forcefully delete stuck pods preventing reconciler from doings its jobs when a node crashed or became unresponsive

jonico:
* if a k8s node becomes unresponsive, the kube controller will soft delete all pods after the eviction time (default 5 mins)
* as long as the node stays unresponsive, the pod will never leave the last status and hence the runner controller will assume that everything is fine with the pod and will not try to create new pods
* this can result in a situation where the runner replica set thinks that those nodes are still running and ready (although not busy but starting up back soon) and the horizontal autoscaler not scheduling any further runners / pods, resulting in a broken runner deployment until the runnerreplicaset is deleted or the node comes back online
* introducing a pod deletion timeout (1 minute) after which the runner controller will try to reboot the runner and create a pod on a working node
* use forceful deletion and requeue for pods that have been stuck for more than one minute in terminating state
* gracefully handling race conditions if pod gets finally forcefully deleted within the (next) reconciliation loop

### DIFF
--- a/controllers/runner_controller.go
+++ b/controllers/runner_controller.go
@@ -185,7 +185,39 @@ func (r *RunnerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		}
 
 		if !pod.ObjectMeta.DeletionTimestamp.IsZero() {
-			return ctrl.Result{}, err
+			deletionTimeout := 1 * time.Minute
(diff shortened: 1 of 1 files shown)


mumoshu (approved):
LGTM. Impressive job @jonico!

PR #308, opened 2021-02-14
Pod eviction policy examples to not wait 400 seconds until runner controller creates new pods for crashed nodes

jonico:
* ... otherwise it will take 40 seconds (until a node is detected as unreachable) + 5 minutes (until pods are evicted from unreachable/crashed nodes)
* pods stuck in "Terminating" status on unreachable nodes will only be freed once #307 gets merged (with a timeout of another minute)

### DIFF
--- a/README.md
+++ b/README.md
@@ -529,6 +529,14 @@ spec:
         requests:
           cpu: "2.0"
           memory: "4Gi"
+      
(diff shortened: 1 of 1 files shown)


mumoshu (approved):
LGTM. Thanks @jonico ☺️

PR #309, opened 2021-02-14
Fix rate limit and runner not registered detection logic and action

jonico:
* in the changes introduced by #297   - conditions to detect not found runners or rate limit exceptions have never been triggered because `errors.Is` compares all members of a struct to return true which never happened
* switched to type check instead of exact value check and use references instead of copying values

* in changes introduced by #297 runners have been deregistered all 10 minutes (de-registration timeout), no matter whether they have failed to register or not because the if statement checking notRegistered was using double negation in if statement which lead to deregistration in all cases
* changed if statement to only deregister if actually needed and cleaned up error messages and detection logic in replicaset and runner controllers

### DIFF
--- a/controllers/runner_controller.go
+++ b/controllers/runner_controller.go
@@ -212,16 +212,18 @@ func (r *RunnerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 			return ctrl.Result{}, err
 		}
 
-		var notRegistered bool
+		notRegistered := false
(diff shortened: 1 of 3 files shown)


mumoshu (approved):
Awesome! I had a hard time testing/debugging the latest changes myself which delayed releasing them so this was SUPER helpful :heart: Thanks again for your help, @jonico :bow:

Issue #281, opened 2021-01-30
controller manager not running on arm64


sefose:
I tried to run the controller manager on my Raspberry Pi 4 cluster.

The kube-rbac-proxy ist always crashing with:
`standard_init_linux.go:219: exec user process caused: exec format error`

My Workaround so far:
I change the image from "gcr.io/kubebuilder/kube-rbac-proxy:v0.4.1" to "carlosedp/kube-rbac-proxy:v0.5.0" and it seems to work.

jonico:
@sefose: Just a drive-by comment but I think it is super cool to create a Kubernetes cluster on Raspberry Pi and have GitHub Actions running on it - in fact, this is my passion project for the next weekends. Finally a way to show Kubernetes self-healing capabilities but pulling some actual plugs and some great visualization opportunities for matrix builds by displaying the runner status on a connected LED matrix :D

As you are already further down the road - have you encountered any other gotchas but the kube-rbac-proxy? Any changes needed for the Action runner base images? Any issues building the Go code? 

Looking forward to hearing about your experiences and will also share mine once available 😊

sefose:
@jonico: I had a small fullstack project for educational purposes and tried to use the pi cluster as much as possible. I ran into images not available for arm64 very often and got shipwrecked when I needed a helm chart for a postgresql database.

Regarding the action runner - I got it running to a green light in Github, but I was not able to use Docker with it. As far as I remember, there where some problems with connecting to the docker socket.
I was a bit short of time then so in the end I started a runner without the cluster and the build times where so lengthy that I switched back to the ones provided by Github.

PR #310, opened 2021-02-15
Use ARM64 compatible kube-rbac-proxy from upstream (https://github.com/brancz/kube-rbac-proxy/)

jonico:
* as pointed out in #281 the currently used image for the 
kube-rbac-proxy - gcr.io/kubebuilder/kube-rbac-proxy:v0.4.1" - does not 
have an ARM64 image
* hence, trying to use the standard deployment manifest / helm char will 
fail on ARM64 systems
* replaced image with quay.io/brancz/kube-rbac-proxy:v0.8.0 which is the 
latest version from the upstream maintainer 
(https://github.com/brancz/kube-rbac-proxy/blob/master/Makefile#L13)
* successfully tested on both AMD64 and ARM64 clusters

### DIFF
--- a/charts/actions-runner-controller/values.yaml
+++ b/charts/actions-runner-controller/values.yaml
@@ -28,8 +28,8 @@ image:
 
 kube_rbac_proxy:
   image:
-    repository: gcr.io/kubebuilder/kube-rbac-proxy
-    tag: v0.4.1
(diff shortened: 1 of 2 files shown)


mumoshu (approved):
LGTM. Thanks for your continued effort @jonico ☺️

PR #311, opened 2021-02-15
Make "runner not found" log message less scary

jonico:
* the reconciliation loop is often much faster than the runner startup,  so changing runner not found related messages to INFO and also add the possibility that the runner just needs more time
* less stack traces in the logs make it easier to follow what is happening during up-and down-scaling


mumoshu (approved):
Good idea!

PR #290, opened 2021-02-02
Add support for enterprise runners

zetaab:
waiting for PR https://github.com/google/go-github/pull/1789 and after that I can verify this

fixes #289

PR #314, opened 2021-02-16
Fix enterprise runners misusing cached token

mumoshu:
Follow-up for #290

github_test.go seems to have been failing since #290 and I've found a potential bug in it while fixing tests. This fixes the bug and tests.

cc/ @zetaab I believe enterprise runners have been mistakenly using other runners token in mixed setup(enterprise and repository/organizational runner co-existing). This fixes that.


zetaab:
Great! Thanks

PR #318, opened 2021-02-16
fix: duplicate name attribute in workflow

callum-tait-pbx:
workflow failing due to duplicate workflow `Name` attribute


mumoshu (approved):
LGTM. Thanks!

PR #323, opened 2021-02-17
Fix wip.yml workflow for building  controller canary tags

### LINKED ISSUE #302: Use `latest` docker image tags for new releases only
#300 made me realize that some people use `latest` tag for the controller image in production deployments. I believe changing our workflow to update `latest` tag only on new releases makes things a bit more reliable for them.

Currently `latest` is automatically updated on each new commit in the main branch. As I usually do some additional manual tests before releasing a new version of the controller, updating `latest` only on new releases result in the controller deployments get updated only with manually tested releases.

From a maintainer's perspective, reports like #300 is SUPER HELPFUL as I can notice/confirm that there's indeed issue(s) that I missed while reviewing or limited manual tests I often do before merging PRs.

So, I'd like to introduce another tag, `canary` for one that is updated on every commit.

mumoshu:
In #306 we seem to have accidentally updated a wrong workflow, which was for runner builds. This updates the one for the controller.

Resolves #302

### DIFF
--- a/.github/workflows/build-and-release-runners.yml
+++ b/.github/workflows/build-and-release-runners.yml
@@ -16,8 +16,6 @@ on:
       - runner/dindrunner.Dockerfile
       - runner/entrypoint.sh
       - .github/workflows/build-runner.yml
-  release:
-    types: [published]
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


callum-tait-pbx:
Apologies, I don't know how I didn't realise this when I did the original change 😬

mumoshu:
@callum-tait-pbx Haha, no - I should have noticed this while reviewing 😆

PR #325, opened 2021-02-17
fix: add trunate to ensure service kinds have valid names

callum-tait-pbx:
Fixes https://github.com/summerwind/actions-runner-controller/issues/324

**Tests**
Installed the controller on minikube with the change and no `fullnameOverride` or `nameOverride` specificed in the `values.yaml` as well as a really long string specified for `fullnameOverride`. No errors in the `kubectl logs` in the case of `actions-actions-runner-controller-66984955f9-fg495` but I got a weird can't find the pod error with the weird pod name deployment. The pod was running fine though 🤷‍♂️ . I put it down to the ridiclously long name that isn't realistic in the real world.

```shell
NAME                                                 READY   STATUS    RESTARTS   AGE
actions-actions-runner-controller-66984955f9-fg495   2/2     Running   0          84s
```

```shell
NAME                                                              READY   STATUS    RESTARTS   AGE
asdjbksakbdfjksabfdjkbfdkjsbfdskbfdskbfkdsbfuidsgfuidsgfdugkpdf   2/2     Running   0          11s
```

### DIFF
--- a/charts/actions-runner-controller/Chart.yaml
+++ b/charts/actions-runner-controller/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.5.0
+version: 0.5.1
 
 home: https://github.com/summerwind/actions-runner-controller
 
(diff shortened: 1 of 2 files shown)


mumoshu (approved):
Awesome! Thank you so much for the fix ☺️

PR #326, opened 2021-02-17
Also log into DockerHub for release event

jonico:
* so far, only push events would trigger the DockerHub login step
* hence, attempts to release would fail because of a permission problem (tested locally)
* adding OR condition to also login in case a release got published

### DIFF
--- a/.github/workflows/build-and-release-runners.yml
+++ b/.github/workflows/build-and-release-runners.yml
@@ -52,7 +52,7 @@ jobs:
 
       - name: Login to DockerHub
         uses: docker/login-action@v1
-        if: ${{ github.event_name == 'push' }}
+        if: ${{ github.event_name == 'push' || github.event_name == 'release' }}
         with:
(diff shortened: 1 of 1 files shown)


mumoshu (approved):
LGTM. Thanks!

> hence, attempts to release would fail because of a permission problem (tested locally)

Out of curiosity, how are you testing the workflow locally? `nektos/act`, or any other way?

jonico:
> Out of curiosity, how are you testing the workflow locally? `nektos/act`, or any other way?

@mumoshu: I  have a mirror of your project connected to a GitHub Enterprise Server 2.22, a GCP cluster (amd64), some AWS runners and a local Raspberry PI 4 Kubernetes cluster (ARM64) and an ARM32 Raspberry Pi 3 runner.

Fun fact: I use some LEDs connected to my Raspberry 3 to visualize the Pod status on the local Kubernetes cluster which is running your actions controller, so that I can visualize the scaling up and down + error states:

image

PR #337, opened 2021-02-19
Fix user used for dind runner group permissions

audip:
Fix https://github.com/summerwind/actions-runner-controller/issues/336

See that issue for the description


mumoshu (approved):
LGTM. Thanks for the fix @audip!

PR #341, opened 2021-02-20
Handle runners with invalid tokens gracefully

jonico:
### Problem

* if a runner pod starts up with an invalid token (crashes, timeouts, etc), it will go in an 
infinite retry loop, appearing as RUNNING from the outside
* normally, this error situation is detected because no corresponding 
runner objects exists in GitHub and the pod will get removed after 
registration timeout
* if the GitHub runner object already existed before - e.g. because a 
finalizer was not properly run as part of a partial Kubernetes crash, 
the runner will always stay in a running mode, even updating the 
registration token will not kill the problematic pod
* consequently, a huge number of the runner pool could end up being "zombie" runners that appear offline at GitHub but appear running in the runner controller

### Solution

* introducing RunnerOffline exception that can be handled in runner 
controller and replicaset controller
* as runners are offline when a pod is completed and marked for restart, 
only do additional restart checks if no restart was already decided, 
making code a bit cleaner and saving GitHub API calls after each job 
completion

### DIFF
--- a/controllers/runner_controller.go
+++ b/controllers/runner_controller.go
@@ -244,60 +244,85 @@ func (r *RunnerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 			return ctrl.Result{}, err
 		}
 
-		notRegistered := false
+		// all checks done below only decide whether a restart is needed
+		// if a restart was already decided before, there is no need for the checks
+		// saving API calls and scary log messages
+		if !restart {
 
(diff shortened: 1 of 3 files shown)


mumoshu (approved):
Great! This would greatly improve the reliability of our setup. Thank you so much for your continued support @jonico!

reiniertimmer:
This is great! I was just wondering about the same thing (some runners just don't come online) while the pod itself is just running, but continuously retrying.

I was about to report an issue about it, and then I found this. Made my day 👍🏻

PR #343, opened 2021-02-22
Update helm chart

kimxogus:
- Set missing `.spec.replica` in deployments
- Implement missing ingress for github webhook server
- Add `authSecret.existingSecret`
- Unset default `image.tag` as `values.yaml` is default value for chart, uncommenting `image.tag` will never make app version to be tag


callum-tait-pbx:
Just from a quick glance:

1. This isn't valid yaml in the `values.yaml`, tag needs a default value or should be commented out, why not `latest`?
```yaml  
  tag:
  dindSidecarRepositoryAndTag: "docker:dind"
  pullPolicy: IfNotPresent
```

2. The `enabled: false` attribute has been deleted in the values.yaml under `authSecret` but is still referenced and is needed. If we are going to add support for externally defined secrets (which I'm not a big fan of personally but that is up to the maintainer not me) this should probably be changed to something like `createManagerSecret` maybe and `existingSecret` should become `managerSecretName` with a default value of `controller-manager` maybe? Then `managerSecretName` can be general purpose allowing it to be valid for if the secret is being defined within the chart, or it referencing a secret externally defined secret instead?:

```yaml
authSecret:
  createManagerSecret: false
  managerSecretName: "controller-manager"
  ### GitHub Apps Configuration
  #github_app_id: ""
  #github_app_installation_id: ""
  #github_app_private_key: |
  ### GitHub PAT Configuration
  #github_token: ""
```

Previously the secret was implemented to be general purpose so it was done in a loop rather than referencing the individual secrets.

What do you think @mumoshu @kimxogus ?

kimxogus:
@callum-tait-pbx
1. When you add repo and install following README.md(not `helm install ./charts/actions-runner-controller` in directory), user defined `values.yaml` will override default `values.yaml`, which means `latest` will be always default and appVersion-based default value will never be applied, which means the comment `# Overrides the manager image tag whose default is the chart appVersion if the tag key is commented out` won't work as expected, So I removed that default value. I actually prefer to set explicit default versioned tag like `v0.17.0`.

2. Using externally defined secrets(just above https://github.com/summerwind/actions-runner-controller#using-personal-access-token) is already in install guide in README, this is necessary anyway. I left checking `.Values.authSecret.enabled` for backward compatibility, but as `existingSecret` style is already generally used since https://github.com/helm/charts used this style, I recommend to use this way.
What if `createManagerSecret` is `false` and there's no `managerSecretName` defined? This chart will do nothing and I believe this will lead users to deploy an in-complete app.
I think setting default name in `values.yaml` is a good idea. I'll apply that. Thanks.

callum-tait-pbx:
The docs on the `chart.yaml` is wrong is seems. I actually removed the `appVersion` attribute, see https://github.com/summerwind/actions-runner-controller/pull/296 for details why. The `values.yaml` should be getting used as the default, if the code references the `appVersion` it should be cleaned up as that isn't there anymore in the chart and the bundles `values.yaml` should be used instead for default values. I agree setting the version to a pinned value is better for a real deployment but having `latest` as the default makes sense so someone can just get going. I would expect them to configure their `values.yaml` for production if deploying the solution which includes version pinning in their values file.

kimxogus:
Oh, I couldn't see that `appVersion` is not defined. Thanks for pointing this out :) It looks weird to me because there's no clue about target(or tested) app version is chart is written for. I agree `appVersion` is confusing sometime, but I think this has meaning that this provides minimum information about which version this chart targets or tested on.

I'm not a fan of `latest`, this might lead your app deployed with mismatched version and config. Generally that would be ok, but there's no guarantee that your app won't crash.
What if there's a major version bump with breaking change? App probably crash unless you catch new release and handle it ahead. I'm fine with using major or minor version alias tag(like `v0.17`), but not with `latest`.

> If it can't find one then it will error and I think it already provides a useful error in that circumstance.

Do you mean [this part](https://github.com/summerwind/actions-runner-controller/blob/master/charts/actions-runner-controller/templates/deployment.yaml#L103)?

callum-tait-pbx:
> but I think this has meaning that this provides minimum information about which version this chart targets or tested on.

Realistically this is an open source project with loose release versioning maintained by private individuals, when new charts are released the chart is not going to be tested against the stated `appVersion` and so I think it's a bit misleading using it to define the minimum app version this chart supports. That said, as you saw from my comments on https://github.com/summerwind/actions-runner-controller/pull/296 I'm not especially bothered if we really want to include an `appVersion`, it just seemed odd to me for this project specifically as `helm list` will be wrong most of the time.

If we do want to include it again I would just include a comment about it in the `chart.yaml` to say something like, so it is clear when it needs incrementing:

```
# This is the minimum summerwind/actions-runner-controller verison that this chart works with.
# Bump this when a new minimum version is needed for this chart version.
```

If we are going to go down the `appVersion` route then including `| default (cat "v" .Chart.AppVersion | replace " " "")` is probably the sensible way of setting the controller version used then.

> Do you mean this part?

yeh, I would just have the chart set it to the defined `managerSecretName` value with a default of `controller-manger`. This way the key is generic and covers both use cases without complicated logic. If the secret doesn't exist on the cluster then you will get a very easy to understand error.

kimxogus:
I think `enabled` is confusing, so I made `create` and `name` variables just like `rbac` or `serviceAccount`s. When `create` is `true`, secret will be created by chart, and use an external secret with the `name` otherwise. What about this?

callum-tait-pbx:
1. @mumoshu up to you if you want default values to be derived from the `values.yaml` or you'd rather do it in the templates themselves.

2. `enabled` is what you typically see in Helm charts for a flag to determine if something is created or not, some examples (but you'll find them all over github):

- https://github.com/jetstack/cert-manager/blob/efbd1a0095e104177f7dea9c0966d9d8b98d5f6f/deploy/charts/cert-manager/templates/cainjector-psp-clusterrole.yaml
- https://github.com/prometheus-community/helm-charts/blob/3ab7f9b04e1ab5abd6c0b35268368422436ae3a6/charts/kube-prometheus-stack/ci/01-provision-crds-values.yaml
- https://jfrog.com/blog/helm-charts-best-practices/

I'm not sure what is typically used in a circumstance like this where we are supporting creating the secret within the chart as well as actively supporting the secret externally created 🤷‍♂️, if someone who has used Helm more than me could comment that would be wonderful!

I agree that `create` sounds better in this scenario, especially with the secret no longer necessarily being create by the chart but I do think `covention over configuration` is the right approach too so if `enabled` is accepted as the word for conditionally creating something then we should use that.

4. If we are moving from a generic auth secret implementation (a loop) in the template which was what the maintainer @mumoshu wanted when the work was originally done. Then you could probably make it more reliable. At the moment if the various github ids are provided as an int the secret will fail https://github.com/summerwind/actions-runner-controller/issues/241. This wasn't fixable because `enabled` is a boolean type. Now that you are targeting individual secrets you could probably make it be able to work with the value in the `values.yaml` is either an int or a string without it breaking.

3. @mumoshu I can't comment on all the webhook stuff, I haven't had a chance to use it yet so I have no idea

kimxogus:
2. I see the examples you mentioned are little bit different to this. Those `enabled` are like black and white cases. When `enabled` is true, you use those resources or not at all otherwise(e.g. in `cert-manager` chart, when you set `cainjector.enabled` false, then you never use cainjector whether it's external or not). But for external secrets, secret is still used and also referenced in `deployment`, I see this is more similar to service account use cases in many public charts. When `create` is false, chart don't create service account, but reference external service account when it's name is defined in `values.yaml`
- https://github.com/jetstack/cert-manager/blob/efbd1a0095e104177f7dea9c0966d9d8b98d5f6f/deploy/charts/cert-manager/templates/_helpers.tpl#L37
- https://github.com/prometheus-community/helm-charts/blob/3ab7f9b04e1ab5abd6c0b35268368422436ae3a6/charts/kube-prometheus-stack/templates/_helpers.tpl#L59

mumoshu:
## create vs enable

Let's go with `create`.

Not saying `enable` is inappropriate, I can definitely agree `create` is a bit more appropriate in this case.

Some more context:

- To me, `create` and `enabled` both works. The former seems to be occasionally used for toggling whether the chart should manage the resource, or the user brings one external to the chart. An example is `serviceAccount.create` which is seen in many charts. However, I saw some charts using `someResource.enabled` for toggling resource creation. So both works to me.
- I also agree that `enabled` is commonly/mostly used for toggling a feature at once.

mumoshu:
## appVersion

I had been thinking about this for a while, and I'd say we'd better NOT use it for now.

See the original definition of `appVersion`:

https://v2.helm.sh/docs/developing_charts/#the-chart-yaml-file

It does say `appVersion: The version of the app that this contains (optional). This needn't be SemVer.` and our chart clearly doesn't "contain" or being tied to any specific version of our app(actions-runner-controller). So it is more than confusing alone. `helm list` showing the App Version regardless of the value of `image.tag` makes it a bit more confusing. At this point, I think we have no big motivation to use it.

That said, the below comment should better be reattached to the comment for `image.tag` in our values.yaml:

```
# This is the minimum summerwind/actions-runner-controller verison that this chart works with.
# Bump this when a new minimum version is needed for this chart version.
appVersion: 0.17.0
```

mumoshu:
> I'm not a fan of latest, this might lead your app deployed with mismatched version and config. Generally that would be ok, but there's no guarantee that your app won't crash.

That's really good point - I agree. Considering this and the above, how about making the default `image.tag` point to a specific semver tag instead?

mumoshu:
Other than the above three points, LGTM. Thank you so much for your support @kimxogus and @callum-tait-pbx!

kimxogus:
Thanks for reviews @callum-tait-pbx and @mumoshu !
I removed `appVersion`. Also I forgot to mention that I [removed `githubWebhookServer.image`](https://github.com/summerwind/actions-runner-controller/pull/343/commits/3b221dc5f8f65be288491dfbc6446f2548bb52f3) because both controller and webhook server uses the same image.

mumoshu:
> Also I forgot to mention that I removed githubWebhookServer.image because both controller and webhook server uses the same image.

Okay- To clarify, the original intention was to build a separate image for the webhook server and allow using that. But I'll just re-add it on demand.

mumoshu (approved):
LGTM. Thank you very much for your contribution @kimxogus and @callum-tait-pbx!

charts/actions-runner-controller/templates/deployment.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
   labels:
     {{- include "actions-runner-controller.labels" . | nindent 4 }}
 spec:
+  replicas: {{ .Values.replicaCount }}
```
mumoshu, line 9: This should be set when and only when replicaCount is defined. Otherwise, HRA is unable to auto-scale the replicas value. I'll add a commit myself to make the change.

PR #345, opened 2021-02-23
Avoid zombie runners that missed token expiration by a bit

jonico:
* if a new runner pod was just scheduled to start up right before a 
registration token expired, it will not get a new registration token and go in 
an infinite update loop (until #341 kicks in)
* if registration tokens got updated a little bit before they actually 
expired, just starting up pods will way more likely get a working token

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -85,7 +85,10 @@ func (c *Client) GetRegistrationToken(ctx context.Context, enterprise, org, repo
 	key := getRegistrationKey(org, repo, enterprise)
 	rt, ok := c.regTokens[key]
 
-	if ok && rt.GetExpiresAt().After(time.Now()) {
+	// we like to give runners a chance that are just starting up and may miss the expiration date by a bit
+	runnerStartupTimeout := 3 * time.Minute
+
+	if ok && rt.GetExpiresAt().After(time.Now().Add(runnerStartupTimeout)) {
 		return rt, nil
 	}
 


mumoshu:
I'm investigating the CI failure as it seems to be consistently failing, despite that it doesn't seem related to the changes made in this PR.

jonico:
@mumoshu: I ran the tests in my Codespaces and sometimes that test passed and sometimes it did not. When I reverted my commit and tested again, I was still seeing the very same test problem, so it seems to be a shaky test.

jonico:
@mumoshu: https://github.com/summerwind/actions-runner-controller/runs/1966732471#step:4:175 shows that one of the runners got already deleted as it was not showing up as busy in GitHub. Are we mocking the API methods to indicate that the runners are online?

jonico:
seems, the replicasetcontroller already thinks that 3 replicas are good enough: https://github.com/summerwind/actions-runner-controller/runs/1966732471#step:4:170

mumoshu:
@jonico Thanks! That can be the important point.

Our GitHub API mock used in the integration test doesn't support removing already registered runners, hence creating and registering 4 runners and deleting 1 runner still result in the API mock returning 4 runners. I'll fix it soon.

I'm not sure why the replicaset controller decided to create 4 runners in the first place for this test case though.

mumoshu (approved):
LGTM. Thanks for your support as always @jonico :relaxed:

PR #349, opened 2021-02-24
Log correct runner that was deleted

jonico:
* index variable i is referring to busy runners, not to all runners of replica set
* hence, wrong runner name to be deleted was shown
* printing correct runner name in event


mumoshu (approved):
Good catch!! Appreciate your efforts @jonico

PR #355, opened 2021-02-25
Manage runner with label

tapih:
This PR fixes https://github.com/summerwind/actions-runner-controller/issues/330 .

- `RunnerDeployment` manages `RunnerReplicaSet` with label selector
- `RunnerReplicaSet` manages `Runner`s with label selector
- `HorizontalRunnerAutoscler` lists `Runner`s by labels when `PercentageRunnersBusy` is specified

I manually checked that `HorizontalRunnerAutoscler` works properly on the `PercentageRunnersBusy` mode when multiple `RunnerDeployment` exist in the same namespace.

### DIFF
--- a/controllers/autoscaling.go
+++ b/controllers/autoscaling.go
@@ -275,6 +275,8 @@ func (r *HorizontalRunnerAutoscalerReconciler) calculateReplicasByPercentageRunn
 			return nil, err
 		}
 	}
+	r.Log.V(0).Info("debug", "runnerList", runnerList, "selector", selector)
+
 	runnerMap := make(map[string]struct{})
 	for _, items := range runnerList.Items {
 		runnerMap[items.Name] = struct{}{}
--- a/controllers/autoscaling.go
+++ b/controllers/autoscaling.go
@@ -10,6 +10,8 @@ import (
 	"time"
 
 	"github.com/summerwind/actions-runner-controller/api/v1alpha1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -257,10 +259,21 @@ func (r *HorizontalRunnerAutoscalerReconciler) calculateReplicasByPercentageRunn
 		scaleDownFactor = sdf
 	}
 
+	selector, err := metav1.LabelSelectorAsSelector(rd.Spec.Selector)
+	if err != nil {
+		return nil, err
+	}
 	// return the list of runners in namespace. Horizontal Runner Autoscaler should only be responsible for scaling resources in its own ns.
 	var runnerList v1alpha1.RunnerList
-	if err := r.List(ctx, &runnerList, client.InNamespace(rd.Namespace)); err != nil {
-		return nil, err
+	if err := r.List(
+		ctx,
+		&runnerList,
+		client.InNamespace(rd.Namespace),
+		client.MatchingLabelsSelector{Selector: selector},
+	); err != nil {
+		if !kerrors.IsNotFound(err) {
+			return nil, err
+		}
 	}
 	runnerMap := make(map[string]struct{})
 	for _, items := range runnerList.Items {
--- a/controllers/autoscaling.go
+++ b/controllers/autoscaling.go
@@ -275,7 +275,6 @@ func (r *HorizontalRunnerAutoscalerReconciler) calculateReplicasByPercentageRunn
 			return nil, err
 		}
 	}
-	r.Log.V(0).Info("debug", "runnerList", runnerList, "selector", selector)
 
 	runnerMap := make(map[string]struct{})
 	for _, items := range runnerList.Items {
--- a/controllers/runnerdeployment_controller.go
+++ b/controllers/runnerdeployment_controller.go
@@ -322,12 +322,17 @@ func CloneSelectorAndAddLabel(selector *metav1.LabelSelector, labelKey, labelVal
 }
 
 func (r *RunnerDeploymentReconciler) newRunnerReplicaSet(rd v1alpha1.RunnerDeployment) (*v1alpha1.RunnerReplicaSet, error) {
+	return newRunnerReplicaSet(&rd, r.CommonRunnerLabels, r.Scheme)
+}
+
+func newRunnerReplicaSet(rd *v1alpha1.RunnerDeployment, commonRunnerLabels []string, scheme *runtime.Scheme) (*v1alpha1.RunnerReplicaSet, error) {
 	newRSTemplate := *rd.Spec.Template.DeepCopy()
+
 	templateHash := ComputeHash(&newRSTemplate)
 	// Add template hash label to selector.
 	labels := CloneAndAddLabel(rd.Spec.Template.Labels, LabelKeyRunnerTemplateHash, templateHash)
 
-	for _, l := range r.CommonRunnerLabels {
+	for _, l := range commonRunnerLabels {
 		newRSTemplate.Spec.Labels = append(newRSTemplate.Spec.Labels, l)
 	}
 
@@ -353,7 +358,7 @@ func (r *RunnerDeploymentReconciler) newRunnerReplicaSet(rd v1alpha1.RunnerDeplo
 		},
 	}
 
-	if err := ctrl.SetControllerReference(&rd, &rs, r.Scheme); err != nil {
+	if err := ctrl.SetControllerReference(rd, &rs, scheme); err != nil {
 		return &rs, err
 	}
 
--- a/controllers/runnerdeployment_controller.go
+++ b/controllers/runnerdeployment_controller.go
@@ -145,11 +145,14 @@ func (r *RunnerDeploymentReconciler) Reconcile(req ctrl.Request) (ctrl.Result, e
 	}
 
 	if !reflect.DeepEqual(newestSet.Spec.Selector, desiredRS.Spec.Selector) {
+		updateSet := newestSet.DeepCopy()
+		updateSet.Spec = *desiredRS.Spec.DeepCopy()
+
 		// A selector update change doesn't trigger replicaset replacement,
 		// but we still need to update the existing replicaset with it.
 		// Otherwise selector-based runner query will never work on replicasets created before the controller v0.17.0
 		// See https://github.com/summerwind/actions-runner-controller/pull/355#discussion_r585379259
-		if err := r.Client.Update(ctx, desiredRS); err != nil {
+		if err := r.Client.Update(ctx, updateSet); err != nil {
 			log.Error(err, "Failed to update runnerreplicaset resource")
 
 			return ctrl.Result{}, err
--- a/controllers/runnerdeployment_controller.go
+++ b/controllers/runnerdeployment_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"hash/fnv"
+	"reflect"
 	"sort"
 	"time"
 
@@ -140,6 +141,24 @@ func (r *RunnerDeploymentReconciler) Reconcile(req ctrl.Request) (ctrl.Result, e
 
 		// We requeue in order to clean up old runner replica sets later.
 		// Otherwise, they aren't cleaned up until the next re-sync interval.
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+	}
+
+	if !reflect.DeepEqual(newestSet.Spec.Selector, desiredRS.Spec.Selector) {
+		// A selector update change doesn't trigger replicaset replacement,
+		// but we still need to update the existing replicaset with it.
+		// Otherwise selector-based runner query will never work on replicasets created before the controller v0.17.0
+		// See https://github.com/summerwind/actions-runner-controller/pull/355#discussion_r585379259
+		if err := r.Client.Update(ctx, desiredRS); err != nil {
+			log.Error(err, "Failed to update runnerreplicaset resource")
+
+			return ctrl.Result{}, err
+		}
+
+		// At this point, we are already sure that there's no need to create a new replicaset thanks
+		// as the runner template hash is not changed.
+		// But we still need to requeue for the (possibly rare) cases that there are still old replicasets that needs
+		// to be cleaned up.
 		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
 	}
 
--- a/controllers/runnerdeployment_controller.go
+++ b/controllers/runnerdeployment_controller.go
@@ -158,8 +158,9 @@ func (r *RunnerDeploymentReconciler) Reconcile(req ctrl.Request) (ctrl.Result, e
 			return ctrl.Result{}, err
 		}
 
-		// At this point, we are already sure that there's no need to create a new replicaset thanks
+		// At this point, we are already sure that there's no need to create a new replicaset
 		// as the runner template hash is not changed.
+		//
 		// But we still need to requeue for the (possibly rare) cases that there are still old replicasets that needs
 		// to be cleaned up.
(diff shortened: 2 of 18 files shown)


controllers/runnerdeployment_controller.go
```diff
@@ -270,6 +316,12 @@ func (r *RunnerDeploymentReconciler) newRunnerReplicaSet(rd v1alpha1.RunnerDeplo
 
 	newRSTemplate.Labels = labels
 
+	if rd.Spec.Selector == nil {
```
mumoshu, line 319: This seems like a breaking change and I'd like to make it backward-compatible as much as possible.

Sorry if I'm missing something or I have any misconception - but can we use `labels` to compose the default rd.Spec.Selector here? As far as remember, `labels` is inherited to RunnerReplicaSet.ObjectMeta.Labels and and Runner.ObjectMeta.Labels, which might be enough for resolving the original issue.

As far as it is backward compatible. I think I'm fine to add `selector` and `matchLabels`, `matchExpressions`. IMHO it isn't strictly needed for the original issue, though. WDYT?

controllers/runnerdeployment_controller_test.go
```diff
@@ -189,20 +217,24 @@ var _ = Context("Inside of a new namespace", func() {
 
 				Eventually(
 					func() int {
-						err := k8sClient.List(ctx, &runnerSets, client.InNamespace(ns.Name))
+						selector, err := metav1.LabelSelectorAsSelector(rd.Spec.Selector)
 						if err != nil {
-							logf.Log.Error(err, "list runner sets")
+							logf.Log.Error(err, "failed to get RunnerDeployment selector")
+							return -1
```
mumoshu, line 223: Can we instead change the signature of `func() int` to `func (int, error)` and return an error here? (I can do it after merging anyway so please not hesitate to defer it to me. Of course doing it yourself is more than welcome :)

controllers/runnerdeployment_controller.go
mumoshu: I've tried to express the above as code in 62fcd5f. Does it make sense to you?

tapih: [62fcd5f](https://github.com/summerwind/actions-runner-controller/commit/62fcd5f3de285f2c1384406fd0efbbfad3f01a92) basically looks good, but it's not completely backward-compatiblem, IMO.

Let's say users upgrades the controller while `RunnerReplicaSet`s already exist.
The new controller cannot find the existing `RunnerReplicaSet`s, because it looks for `RunnerReplicaSet`s by labels but the existing `RunnerReplicaSet`s  are not labeled.
I guess that new `RunnerReplicaSet`s will be created after the controller is upgraded.

I found 2 ways to avoid that.

- The controller lists `RunnerReplicaSet`s both with `labels` and `ownerReference` and makes a set of both.
- The controller lists `RunnerReplicaSet`s first, and if they are not labeled, it labels `RunnerReplicaSet`s first.
  Then, it lists `RunnerReplicaSet`s both with `labels` again.

These are temporal. and you can delete them when you are forced to break backward compatibility.

mumoshu: @tapih Thanks! Could you double-check my idea?

> The new controller cannot find the existing RunnerReplicaSets, because it looks for RunnerReplicaSets by labels but the existing RunnerReplicaSets are not labeled.
>I guess that new RunnerReplicaSets will be created after the controller is upgraded.

Would this be about the updated controller results in recreating [RunnerReplicaSet due to the replicasets' template hash change](https://github.com/summerwind/actions-runner-controller/blob/master/controllers/runnerdeployment_controller.go#L134) by the addition of the selector? As long as the controller update doesn't result in requiring the user to manually update K8s resources other than the controller's, it looks fine to me. WDYT?

My motivation behind https://github.com/summerwind/actions-runner-controller/commit/62fcd5f3de285f2c1384406fd0efbbfad3f01a92 was to keep backward-compatibility in the sense the sense of the RunnerDeployment API / avoid [forcing users to manually update their RunnerDeployment resources to include Spec.Template.Selector](https://github.com/summerwind/actions-runner-controller/commit/bbfb0b6bc137d077cf2ba159864bed148884ea5e#diff-4b55da611104d855c66d352a05727ec60b647333a87e0401e825bd5d1cb0e489R316).

tapih: `desiredRS` would change due to the addition of the selector, but [these lines](https://github.com/tapih/actions-runner-controller/blob/8de84f9e3f62fbc90d6fd580f5fb9c1501cc99d9/controllers/runnerdeployment_controller.go#L82-L85) in the current code does not get the existing `RunnerReplicaSet`s in the first place because it now matches `RunnerReplicaSet`s with labels but `RunnerReplicaSet`s are not labeled before the controller is updated, IIUC.
I understand your motivation here and agree with you, so we should do what I suggested [here](https://github.com/summerwind/actions-runner-controller/pull/355#discussion_r583381120) to avoid this problem, .

What do you think about it? Thank you.

mumoshu: @tapih Thanks for confirming!

What if we reverted that line to https://github.com/summerwind/actions-runner-controller/blob/master/controllers/runnerdeployment_controller.go#L79 so that we can keep using controller-runtime's index for searching RunnerReplicaSets owned by the RunnerDeployment?

That would still make the updated controller automatically replacing RunnerReplicaSets due to the selector changes, with neither additional K8s API call nor user intervention. WDYT?

tapih: @mumoshu 

Yes, it seems to work correctly enough, keeping the backward compatibility.
I also considered what happens on `RunnerReplicaSet` and `Runner`, and thought that it also works correctly because
the old `RunnerReplicaSet`s are deleted [here](https://github.com/summerwind/actions-runner-controller/blob/master/controllers/runnerdeployment_controller.go#L178-L191) and the existing `Runner`s are deleted eventually by the controller, referencing `ownerReference`.

The Kubernetes native `Deployment` manages `ReplicaSet` with labels, so I still prefer to do the same thing for `RunnerDeployment` at a certain point, but your idea is reasonable for now with consideringg the compatibility

controllers/runnerdeployment_controller_test.go
mumoshu: Confirmed. Thanks!

controllers/runnerdeployment_controller.go
mumoshu: @tapih Thanks again for the confirmation and the update 9f8220c!

I took some more time to think about this. Let me confirm our direction:

### HRA controller

The HRA controller should use the label selector for querying runners from HRA for autoscaling. For that, we'd also need to propagate labels from RunnerDeployment to Runners via RunnerReplicaSet. Currently, it lists the all runners in a namespace for calculation, which is wrong.

### RunnerDeployment controller

The RunnerDeployment controller should keep using the owner-reference index for querying replica sets. By doing that the controller can reliably delete the old RS and create new RS on controller update.

### RunnerReplicaSet controller

Similarly, RunnerReplicaSet controller should use the owner-reference index for querying runners. Currently, it lists all the runners in a namespace which is definitely wrong. This change allows the controller to keep tracking all the runners before/after the controller update, while extra support for multiple RunnerReplicaSet per namespace.

### Future work

The new RunnerReplicaSet and Runner will get more labels after the controller update.

In the future, the controller may use those labels for querying runners and pods. This should make our controller more like K8s deployment controller(ref https://github.com/summerwind/actions-runner-controller/pull/355#discussion_r584473993)

---

I believe that the label propagation and the RunnerDeployment controller changes are already done thanks to you. I'll update the RunnerReplicaSet controller to use the owner-reference index.

mumoshu: desiredRS changes due to the addition of the selector, but apparently it doesn't trigger the replacement of the RS managed by the deployment, as it is triggered only when the runner template hash changes. The hash is calculated from RunnerDeployment.Spec.Template, not RunnerDeployment.Spec.Selector or the [default selector populated by the controller](https://github.com/summerwind/actions-runner-controller/blob/9f8220ced3b373d6fd64def36370494f24e05acf/controllers/runnerdeployment_controller.go#L316).

Probably this is the correct behaviour considering how the standard K8s deployment controller works. https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#updating-a-deployment

But doesn't it break https://github.com/summerwind/actions-runner-controller/blob/9f8220ced3b373d6fd64def36370494f24e05acf/controllers/runnerreplicaset_controller.go on the replicaset controller, as the selector is not populated on the existing replicaset after the controller update, and the repiicaset is not automatically replaced by the updated controller?

tapih: Thank you for wrapping up!

One thing I would like to confirm is that you canceled the `Similarly...` line, and that means `RunnerReplicaSet` manages `Runner` with labels? If so, I'm gonna make another small fix for HRA to work when no selector is given.

mumoshu: I'm not entirely sure yet but I think we need 9d75e30

mumoshu: I'm not entirely sure yet but I think we need 9d75e30. Would you agree @tapih?

mumoshu: The reason I canceled `Similarly...` line was that it was my total misconception. The replicaset controller was not using the owner-reference index but it was filtering runners by the owner reference on its own.

> If so, I'm gonna make another small fix for HRA to work when no selector is given.

Would you think that it's needed even after https://github.com/summerwind/actions-runner-controller/commit/9d75e30c9cd39b9f46aca5afd3b56caec1a99d27?

mumoshu: > RunnerReplicaSet manages Runner with labels?

Thanks to your change, [RunnerReplicaSet already seems to manage Runner with labels](https://github.com/summerwind/actions-runner-controller/blob/9d75e30c9cd39b9f46aca5afd3b56caec1a99d27/controllers/runnerreplicaset_controller.go#L81). After 9d75e30, the controller even works correctly for RunnerReplicaSet with/without Spec.Selector.

> If so, I'm gonna make another small fix for HRA to work when no selector is given.

Are you willing to add something like the owner-reference check similar to https://github.com/summerwind/actions-runner-controller/commit/9d75e30c9cd39b9f46aca5afd3b56caec1a99d27 the below?

https://github.com/summerwind/actions-runner-controller/blob/9d75e30c9cd39b9f46aca5afd3b56caec1a99d27/controllers/runnerreplicaset_controller.go#L77-L81

mumoshu: @tapih Just a friendly reminder for the case that you missed my comment 😃

mumoshu:
It seems like we get `validation failure list:\nspec.selector in body must be of type object: \"null\""` errors on Spec.Selector for existing RunnerReplicaSet resources and RunnerReplicaSet update errors due to some bug. Added 2e0eab0 for the fix.

controllers/runnerdeployment_controller.go
tapih: Sorry, I took a day off yesterday...

tapih: I took this way to fix the HRA bug because I thought we should basically use labels to have `RunnerDeployment` manage `Runner`s to make the logic as symmetric as possible with k8s's `Deployment`
At the same time, I undestood that you didn't want to break the compatibility, so I made some patches not to break the compatibility afterwards.

However, now the code is getting more and more complex, and I'm being strongly concerned that this would make it unnecessaliry harder to maintain this project later.
I'm really sorry for taking your time so much, but I would like to suggest taking the following steps:

1. Fix the HRA bug without using labels
2. Merge the PR which has the `RunnerDeployment` manage `Runner`s with labels when you are forced to break the compatibility later

I'm willing to make PRs again. What do you think about it?

mumoshu: @tapih Just to be clear, I was actually inclined to add `selector` to RunnerDeployment and RunnerReplicaSet, despite that it requires some work to make it backward-compatible.

Not only it helps us shape RunnerDeplomyet as much as similar as K8s Deployment, I found that `selector` itself can be a good/common language that allows the third-party(user) to easily find a specific set of runners managed by a runnerreplicaset/runnerdeployment.

Without `selector`, one had to read actions-runner-controller code to see what information can be used to correlate a runner deployment/replicaset to the corresponding runner(s).

So we now seem to have another benefit that we weren't aware of, for adding `selector`.

In addition to that-

> Fix the HRA bug without using labels

(I may have been pretty confused, but) if you literally meant to NOT use labels, I'm not sure how we can achieve that. If you meant to not use selectors, I'm still unsure how easy it is.

HRA needs to find the latest runners managed by the targeted RunnerDeployment anyway.

Just following owner-references from HRA, RD, RS, and Runners would be not enough.

For HRA-RD, RS-R it's fine, as there might be only 1-to-1 relationship among them. However for RD, there may be two or more RS. They have different `runner-template-hash` label values and that label has been already there for RunnerReplicaSet, so using it for filtering RunnerReplicaSet seems natural.

We also need to determine which `runner-template-hash` value is one for the latest runner spec. Instead of relying on `Selector`, we can probably modify the HRA controller to call the same hash computing function on the RD.Spec.Template as the RD controller for that. But doing so would make a tight coupling between RD and HRA controllers. Wouldn't it be easier to add/use `Selector` then, in terms of design/implementation complexity?

WDYT?

You may have thought if it's really up to me to decide which direction we'd go, but I'd really appreciate your feedback!

controllers/autoscaling.go
```diff
@@ -257,11 +259,23 @@ func (r *HorizontalRunnerAutoscalerReconciler) calculateReplicasByPercentageRunn
...
+	}
 	// return the list of runners in namespace. Horizontal Runner Autoscaler should only be responsible for scaling resources in its own ns.
 	var runnerList v1alpha1.RunnerList
-	if err := r.List(ctx, &runnerList, client.InNamespace(rd.Namespace)); err != nil {
-		return nil, err
+	if err := r.List(
+		ctx,
+		&runnerList,
+		client.InNamespace(rd.Namespace),
+		client.MatchingLabelsSelector{Selector: selector},
```
mumoshu, line 272: As I have roughly mentioned in https://github.com/summerwind/actions-runner-controller/pull/355#discussion_r587191728, I thought this might be incorrect.

We don't include `runner-template-hash` into account here, so at least this can result in runners from multiple RunnerReplicaSet to match. It would double `numRunners` while replacing RunnerReplicaSet due to RD.Spec.Template change.

mumoshu: Probably we could just use `rd.Spec.Replicas` (that defaults to 1 if missing) instead of `numRunners := len(runnerList.Items)`, while keeping this line as-is.

mumoshu: Well, it might be out-of-scope- the above issue should have been there before this PR.

mumoshu:
Actually, I'm unsure if we really need `matchExpressions`. Can we just remove it from the spec, so that the amount of change becomes a bit smaller?

mumoshu (approved):
@tapih Let me merge this anyway so that everyone can participate in testing!

To be clear, I'm pretty confident this is working as expected.

But still wondering if we can remove `matchExpression` for simplicity.

I'll shortly fix https://github.com/summerwind/actions-runner-controller/pull/355#discussion_r587199428. It's clearly out of the scope of this PR.

Other than those, I'm not aware of any other enhancements we need on top of this right now.

If you have any comments/opinions/feedbacks/etc, it's more than welcome!

controllers/runnerdeployment_controller.go
```diff
@@ -143,6 +144,28 @@ func (r *RunnerDeploymentReconciler) Reconcile(req ctrl.Request) (ctrl.Result, e
 		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
 	}
 
+	if !reflect.DeepEqual(newestSet.Spec.Selector, desiredRS.Spec.Selector) {
```
mumoshu, line 147: AFAIK, and to be clear, `Selector` is immutable in the standard K8s deployment. But we'd prefer making this mutable for our use-case

PR #372, opened 2021-03-05
chore: bumping helm chart

callum-tait-pbx:
https://github.com/summerwind/actions-runner-controller/pull/355 made changes to the CRDs but didn't bump the version


mumoshu (approved):
@callum-tait-pbx Ah, good catch! Thanks for your support!

PR #373, opened 2021-03-05
fix typo in examples

robwhitby:
hey, we copied an example and it tripped us up for a while!

thanks for all the great work on this project.

Rob


mumoshu (approved):
Good catch! Thanks for your contribution @robwhitby ☺️

PR #376, opened 2021-03-06
Set runner group for runners with enterprise scope

jonico:
* so far, runner group parameter is only set for runners with org scope
* now set group for enterprise runners as well
* removed null check for org scope as either org or enterprise will be set


mumoshu (approved):
LGTM. Thanks for your support as always, @jonico!

PR #379, opened 2021-03-08
docs: fix typos and add PR link

callum-tait-pbx:
minor changes to docs I noticed

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,11 +355,11 @@ The `HorizontalRunnerAutoscaler` will pole GitHub based on the configuration syn
 **Helm Config :** `syncPeriod`
 
 **Benefits of this metric**
-1. Allows for multiple controllers to be deployed as each controller deployed is responsible for scaling their own runner pods on a per namespace basis.
+1. Allows for multiple controllers to be deployed as each controller deployed is responsible for scaling their own runner pods on a per namespace basis. [#223](https://github.com/summerwind/actions-runner-controller/pull/223)
 2. Supports named repositories server side the same as the `TotalNumberOfQueuedAndInProgressWorkflowRuns` metric [#313](https://github.com/summerwind/actions-runner-controller/pull/313)
 3. Supports github organisation wide scaling without maintaining an explicit list of repositories, this is especially useful for those that are working at a larger scale. [#223](https://github.com/summerwind/actions-runner-controller/pull/223)
 4. Like all scaling metrics, you can manage workflow allocation to the RunnerDeployment through the use of [Github labels](#runner-labels)
-5. Supports scaling runner count on both a percentage increase / descrease basis as well as on a fixed runner count basis [#223](https://github.com/summerwind/actions-runner-controller/pull/223) [#315](https://github.com/summerwind/actions-runner-controller/pull/315)
+5. Supports scaling desired runner count on both a percentage increase / decrease basis as well as on a fixed increase / decrease count basis [#223](https://github.com/summerwind/actions-runner-controller/pull/223) [#315](https://github.com/summerwind/actions-runner-controller/pull/315)
 
 **Drawbacks of this metric**
 1. May not scale quick enough for some users needs as we are scaling up and down based on indicative information rather than a direct count of the workflow queue depth
--- a/README.md
+++ b/README.md
@@ -303,7 +303,7 @@ The scale out performance is controlled via the manager containers startup `--sy
 **Benefits of this metric**
 1. Supports named repositories allowing you to restrict the runner to a specified set of repositories server side.
 2. Scales the runner count based on the actual queue depth of the jobs meaning a more 1:1 scaling of runners to queued jobs.
-4. Like all scaling metrics, you can manage workflow allocation to the RunnerDeployment through the use of [Github labels](#runner-labels).
+3. Like all scaling metrics, you can manage workflow allocation to the RunnerDeployment through the use of [Github labels](#runner-labels).
 
 **Drawbacks of this metric**
 1. Repositories must be named within the scaling metric, maintaining a list of repositories may not be viable in larger environments or self-serve environments.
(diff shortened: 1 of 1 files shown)


callum-tait-pbx:
@mumoshu I think the positive:

```
 1. Allows for multiple controllers to be deployed as each controller deployed is responsible for scaling their own runner pods on a per namespace basis. [#223](https://github.com/summerwind/actions-runner-controller/pull/223)
```

Mentioned for the `PercentageRunnersBusy` isn't correct. It came from the below snipped taken from the https://github.com/summerwind/actions-runner-controller/pull/223:

```
@ahmad-hamade this autoscaling scheme will scale the runners that your runner-deployment creates on your behalf, or the runners you've created Runner resources for. If you are running 3 runners all in separate EKS clusters - i am also assuming you are also running 3 separate controllers. If that is the case, each horizontal runner autoscaler that you have defined in each cluster will be responsible for scaling that clusters runners - which is a benefit over the previous implementation, where a single horizontal runner autoscaler is scaling across all runners, since it cannot distinguish pods running in separate clusters.
```

I don't quite get what the author is getting at. Are you able to figure it out so the benefit in the README can be updated to be correct? I'm struggling to grasp how this is a benefit unique to the `PercentageRunnersBusy` schema

mumoshu:
@callum-tait-pbx Before #355, `TotalNumberOfQueuedAndInProgressWorkflowRuns` was unable to distinguish runners from different namespaces. So probably it was theoretically correct to say that `PercentageRunnersBusy` was more appropriate for a multiple actions-runner-controller deployment across namespaces.

However,

> Allows for multiple controllers to be deployed as each controller deployed is responsible for scaling their own runner pods on a per namespace basis.

It wasn't working like that until #380.

mumoshu:
I've reread the doc and I would say that this part of the doc:

> 1. Allows for multiple controllers to be deployed as each controller deployed is responsible for scaling their own runner pods on a per namespace basis. [#223](https://github.com/summerwind/actions-runner-controller/pull/223)

looks obsolete and can be removed now.

After #355 and #380, both scaling metrics(strategies) should work er eithper namespace or per cluster basis without any problem.

callum-tait-pbx:
@mumoshu that is probably worded better

README.md
```diff
@@ -302,12 +302,12 @@ The scale out performance is controlled via the manager containers startup `--sy
...
 **Benefits of this metric**
 1. Supports named repositories allowing you to restrict the runner to a specified set of repositories server side.
-2. Scales quickly (within the bounds of the syncPeriod) as it will spin up the number of runners based on the depth of the workflow queue
+2. Scales the runner count based on the actual queue depth of the jobs meaning a more 1:1 scaling of runners to queued jobs.
 3. Like all scaling metrics, you can manage workflow allocation to the RunnerDeployment through the use of [Github labels](#runner-labels).
 
 **Drawbacks of this metric**
 1. Repositories must be named within the scaling metric, maintaining a list of repositories may not be viable in larger environments or self-serve environments.
-2. May not scale quick enough for some users needs
+2. May not scale quick enough for some users needs. This metric is pull based and so the queue depth is polled as configured by the sync period, as a result scaling performance is bound by this sync period meaning there is a lag to scaling activity.
```
mumoshu, line 310: Great explanation 💡

README.md
```diff
@@ -355,14 +355,14 @@ The `HorizontalRunnerAutoscaler` will pole GitHub based on the configuration syn
...
+2. Supports GitHub organisation wide scaling without maintaining an explicit list of repositories, this is especially useful for those that are working at a larger scale. [#223](https://github.com/summerwind/actions-runner-controller/pull/223)
+3. Like all scaling metrics, you can manage workflow allocation to the RunnerDeployment through the use of [Github labels](#runner-labels)
+4. Supports scaling desired runner count on both a percentage increase / decrease basis as well as on a fixed increase / decrease count basis [#223](https://github.com/summerwind/actions-runner-controller/pull/223) [#315](https://github.com/summerwind/actions-runner-controller/pull/315)
 
 **Drawbacks of this metric**
-1. May not scale quick enough for some users needs as we are scaling up and down based on indicative information rather than a direct count of the workflow queue depth
+1. May not scale quick enough for some users needs. This metric is pull based and so the number of busy runners are polled as configured by the sync period, as a result scaling performance is bound by this sync period meaning there is a lag to scaling activity.
```
mumoshu, line 364: Probably you could try combining two. Webhook-based scaling will quickly add (possibly more than necessary) runners on demand and PercentageRunnersBusy will periodically "correct" the number of runners depending on runner statuses.

mumoshu (approved):
LGTM. Thank you so much for your continuous effort @callum-tait-pbx ☺️

README.md
callum-tait-pbx: Does the webhook scaling work inconjuction with the regular metrics? I saw you removed the need to define a metric when using webhooks, can you still define a metric and it will compliment the webhook scaling?

mumoshu: > Does the webhook scaling work inconjuction with the regular metrics?

Yes

> I saw you removed the need to define a metric when using webhooks

Yes. Actually, it works only in conjunction with regular metrics before/after I've removed the need to explicitly define metrics. It wasn't working when repository runners + TotalNumberOfQueuedAndInProgressWorkers combination so I fixed it in #381. Other combinations should have worked and will keep working.

> can you still define a metric and it will compliment the webhook scaling?

Exactly! In other words, again, the webhook-based scaling works in combination with either TotalNumberOfQueuedAndInProgressWorkers or PercentageRunnersBusy. When you omited HRA.Spec.Metrics it defaults to TotalNumberOfQueuedAndInProgressWorkers. So you can say that it is still working in conjunction with the regular metrics.

mumoshu: I've changed my mind and will soon change the default `Metrics[]` when ScaleUpTriggers isn't empty, so that ScaleUpTrigger can be actually used standalone. But the general idea described above should still stand.

PR #385, opened 2021-03-10
Add ability to set the MTU size of the docker in docker container

bkimbrough88:
I have encountered a problem where the Max Transmission Unit (MTU) in the docker in docker container is set higher than my ethernet interface's MTU. This is causing packets to get dropped and my runs to hang indefinitely. To fix this, I need to be able to set the MTU to a value lower than my ethernet interface's. 

Here is the output from ifconfig that clued me into what my problem actually was, as you can see, `docker0` has a higher MTU value than `eth0` and that is known to cause packet loss.
```
/ # ifconfig
docker0   Link encap:Ethernet  HWaddr 02:42:D5:CF:F0:57  
          inet addr:172.17.0.1  Bcast:172.17.255.255  Mask:255.255.0.0
          UP BROADCAST MULTICAST  MTU:1500  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0 
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

eth0      Link encap:Ethernet  HWaddr 0A:58:AC:16:02:B5  
          inet addr:172.22.2.181  Bcast:172.22.3.255  Mask:255.255.254.0
          inet6 addr: fe80::dc73:2fff:fe16:65ef/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1450  Metric:1
          RX packets:86735 errors:0 dropped:0 overruns:0 frame:0
          TX packets:51169 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0 
          RX bytes:86094899 (82.1 MiB)  TX bytes:22133196 (21.1 MiB)

lo        Link encap:Local Loopback  
          inet addr:127.0.0.1  Mask:255.0.0.0
          inet6 addr: ::1/128 Scope:Host
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)
```


callum-tait-pbx:
Out of interest do you know why your MTU is 1450? From my knowledge 1500 is super duper standard unless you are doing something like jumbo frames

bkimbrough88:
This decision was made before my time, but from what I am told, it is because of OpenShift's SDN was having packet loss do to largely the same issue, so they jacked the MTU size for our cluster's nodes down to 1450. I get this isn't a common use case, but I imagine it will crop up from time to time.

controllers/runner_controller.go
```diff
@@ -530,6 +530,10 @@ func (r *RunnerReconciler) newPod(runner v1alpha1.Runner) (corev1.Pod, error) {
 		},
 	}
 
+	if dockerdInRunner {
+		pod.Spec.Containers[0].Command = []string{"startup.sh", fmt.Sprintf("--mtu %d", *runner.Spec.DockerMTU)}
```
mumoshu, line 534: It looks like to fail when DockerMTU is nil. Probably we'd better enclose this inside a if block?

```suggestion
		if mtu := runner.Spec.DockerMTU; mtu != nil {
			pod.Spec.Containers[0].Command = []string{"startup.sh", fmt.Sprintf("--mtu %d", *runner.Spec.DockerMTU)}
		}
```

controllers/runner_controller.go
```diff
@@ -600,6 +604,7 @@ func (r *RunnerReconciler) newPod(runner v1alpha1.Runner) (corev1.Pod, error) {
 					MountPath: "/certs/client",
 				},
 			},
+			Args: []string{fmt.Sprintf("--mtu %d", *runner.Spec.DockerMTU)},
```
mumoshu, line 607: Would it be easier to just set `MTU` via an envvar, not over `--mtu` flag?

```
Env: []corev1.EnvVar{
				{
					Name:  "MTU",
					Value: fmt.Sprintf("%d", *Zrunner.Spec.DockerMTU),
				},
			},
```

mumoshu:
@callum-tait-pbx @bkimbrough88 I can imagine where this feature become handy- I've experienced a similar issue running drone in some on-prem environment and had to use mtu of 1450. I've also found that there are a few folks experiencing similar issues in https://discourse.drone.io/t/docker-mtu-problem/1207

callum-tait-pbx:
Probably should update the additional tweaks section of the docs to include an example maybe?

mumoshu:
@callum-tait-pbx Yeah that would be great 👍

mumoshu (approved):
LGTM. Thanks a lot for your support @bkimbrough88 and @callum-tait-pbx ☺️ 

It would be great if you could also add some doc in our README. But that's another story!

PR #391, opened 2021-03-14
Disable metrics-based autoscaling by default when scaleUpTriggers are enabled

mumoshu:
Relates to https://github.com/summerwind/actions-runner-controller/pull/379#discussion_r592813661
Relates to https://github.com/summerwind/actions-runner-controller/issues/377#issuecomment-793266609

When you defined HRA.Spec.ScaleUpTriggers[] but HRA.Spec.Metrics[], the HRA controller will now enable ScaleUpTriggers alone and insteaed of automatically enabling TotalNumberOfQueuedAndInProgressWorkflowRuns. This allows you to use ScaleUpTriggers alone, so that the autoscaling is done without calling GitHub API at all, which should grealy decrease the change of GitHub API calls get rate-limited.

### DIFF
--- a/controllers/autoscaling.go
+++ b/controllers/autoscaling.go
@@ -71,7 +71,13 @@ func (r *HorizontalRunnerAutoscalerReconciler) determineDesiredReplicas(rd v1alp
 	}
 
 	metrics := hra.Spec.Metrics
-	if len(metrics) == 0 || metrics[0].Type == v1alpha1.AutoscalingMetricTypeTotalNumberOfQueuedAndInProgressWorkflowRuns {
+	if len(metrics) == 0 {
+		if len(hra.Spec.ScaleUpTriggers) == 0 {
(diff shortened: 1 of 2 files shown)


Puneeth-n:
@mumoshu how does scale down happen when we use Github Webhook based scaling and no metrics?

mumoshu:
@Puneeth-n Each scaleUpTrigger expires after the configured [duration](https://github.com/summerwind/actions-runner-controller/blob/master/api/v1alpha1/horizontalrunnerautoscaler_types.go#L62) passed. Let's say `duration` was `5m`(5 minutes), RD's desired replicas increases by the configured replica(s) for 5 minutes. After 5 minutes the RD's desired replicas decreases by 1.

You'll also find https://github.com/summerwind/actions-runner-controller#faster-autoscaling-with-github-webhook useful!

Puneeth-n:
@mumoshu Thanks for clarifying. I didn't know it would scale down as well after the "duration"

PR #394, opened 2021-03-16
Fix "pole" typo in README

ba32107:
I think these should be "poll".


mumoshu (approved):
@ba32107 Good catch! Thanks for fixing ☺️

PR #398, opened 2021-03-18
Fix `status.lastRegistrationCheckTime in body must be of type string: \"null\"` error

mumoshu:
Follow-up for #392

### DIFF
--- a/api/v1alpha1/horizontalrunnerautoscaler_types.go
+++ b/api/v1alpha1/horizontalrunnerautoscaler_types.go
@@ -156,6 +156,7 @@ type HorizontalRunnerAutoscalerStatus struct {
 	DesiredReplicas *int `json:"desiredReplicas,omitempty"`
 
 	// +optional
+	// +nullable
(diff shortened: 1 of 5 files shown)


int128:
FYI: I have deployed 9b871567b1213eee40b22ae6ea05341328cf8b2b and got the following error:

```
2021-03-18T00:19:02.734Z	ERROR	controllers.Runner	Failed to update runner status	{"runner": "NAME", "error": "Runner.actions.summerwind.dev \"NAME\" is invalid: status.lastRegistrationCheckTime: Invalid value: \"null\": status.lastRegistrationCheckTime in body must be of type string: \"null\""}
...
github.com/summerwind/actions-runner-controller/controllers.(*RunnerReconciler).updateRegistrationToken
	/workspace/controllers/runner_controller.go:467
```

I hope this will fix it.

mumoshu:
Turns out this happens when the registration token is being renewed here

https://github.com/summerwind/actions-runner-controller/blob/7a7086e7aa14848ec19fe3a9ed6b1f97277397f2/controllers/runner_controller.go#L466

This in conjunction with the wrong CRD scheme resulted in the error. In other places those errors doesn't occur as we use `Patch` instead of `Update` there.

PR #412, opened 2021-03-22
chore: adding Helm app version back

callum-tait-pbx:
Fixes https://github.com/summerwind/actions-runner-controller/issues/402


callum-tait-pbx:
If you do decide to merge this I've bumped the `appVersion` to latest, no point releasing the chart again on an older appVersion at this point

mumoshu (approved):
Thanks! We'd need to automate version bumping somehow but let's get the ball rolling 😄

PR #253, opened 2021-01-13
Fix self-update failuers due to /runner/externals mount

mumoshu:
Please follow the steps described in https://github.com/summerwind/actions-runner-controller#developing to build the container images for testing.

Fixes #252

PR #419, opened 2021-03-25
Fix dindWithinRunnerContainer not to crash-loop runner pods

mumoshu:
Apparently #253 broke dindWithinRunnerContainer completely due to the difference in how /runner volume is set up.

### DIFF
--- a/controllers/runner_controller.go
+++ b/controllers/runner_controller.go
@@ -634,45 +634,58 @@ func (r *RunnerReconciler) newPod(runner v1alpha1.Runner) (corev1.Pod, error) {
 		}...)
 	}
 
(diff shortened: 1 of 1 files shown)


int128:
I got the following error when "Runner with DinD" is enabled:

```
/runner should be an emptyDir mount. Please fix the pod spec.
```

For now, we can workaround this issue by manually setting the volume as follows:

```yaml
      image: summerwind/actions-runner-dind
      volumeMounts:
      - mountPath: /runner
        name: runner
      volumes:
      - emptyDir: {}
        name: runner
```

mumoshu:
@int128 Thanks for sharing!

PR #420, opened 2021-03-25
Quote .parameters in entrypoint.sh to prevent argument split

FloThinksPi:
Prevents arguments from being split when e.g. the RUNNER_GROUP_ARG Variable contains spaces (which is legit. One can create such groups in github)
Ive seen that all workers with non space groupnames can register successfully, while all workers with groups that contains spaces will not register.
Furthermore i suppose also other chars can be used here to inject arbitary commands in an unsuported way via e.g. pipe symbol.
Quoting the vars should prevent that and allow for Group names with bash reseved caracters.
Could not test it though but when i saw this i was quite sure the behaviour i observed is caused by the unqoted variables.
I am quite new to this project, any idea how i could test this properly? Do i need to build and upload an own runner image and test with that ?

### DIFF
--- a/runner/entrypoint.sh
+++ b/runner/entrypoint.sh
@@ -29,21 +29,13 @@ else
   exit 1
 fi
 
-if [ -n "${RUNNER_WORKDIR}" ]; then
-  WORKDIR_ARG="--work ${RUNNER_WORKDIR}"
-fi
-
-if [ -n "${RUNNER_LABELS}" ]; then
-  LABEL_ARG="--labels ${RUNNER_LABELS}"
-fi
-
 if [ -z "${RUNNER_TOKEN}" ]; then
   echo "RUNNER_TOKEN must be set" 1>&2
   exit 1
 fi
 
 if [ -z "${RUNNER_REPO}" ] && [ -n "${RUNNER_GROUP}" ];then
-  RUNNER_GROUP_ARG="--runnergroup ${RUNNER_GROUP}"
+  RUNNER_GROUPS=${RUNNER_GROUP}
 fi
 
 # Hack due to https://github.com/summerwind/actions-runner-controller/issues/252#issuecomment-758338483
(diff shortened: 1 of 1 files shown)


mumoshu:
@FloThinksPi Hey! Thanks a lot for the PR ☺️ 

> Do i need to build and upload an own runner image and test with that ?

Absolutely. Our [developer guide](https://github.com/summerwind/actions-runner-controller#developing) would help it a bit but generally `NAME=$DOCKER_USER/actions-runner TAG=canary make -C runner docker-build docker-push`  can build and push the `$DOCKER_USER/actions-runner:canary` image to Docker Hub. You can let the controller use that custom runner image via `--runner-image` flag or by setting `RunnerDeployment.Spec.Template.Spec.Image`.

FloThinksPi:
I Tested it with a self build image. Thanks for the commands these worked straight out of the box !
stock:
```
+ ./config.sh --unattended --replace --name test-grop-with-space-runners-9fj8q-8hc49 --url https://github.tools.sap/cloudfoundry --token REDACTED --runnergroup Test Group With Space --labels Ubuntu_20.04
....
Could not find any self-hosted runner group named "Test".
```
With change:
```
./config.sh --unattended --replace --name test-grop-with-space-runners-6xk72-sfldw --url https://github.tools.sap/cloudfoundry --token REDACTED --runnergroup 'Test Group With Space' --labels 'Ubuntu,Seconds,Test Label With Space' --work /runner/_work
```

Also Labels can now contain spaces.
I Twisted my head around char escape in bash but this was the only reliable solution i came up with(splittin the args in 2 variables and escaping them properly where used)

I also tried escaping them inside a var e.g.
```
 Var="--group \"${groups}\""
```
But then bash escaped it to `.config .... --group '"groups"'. ` Also tried `${var@Q}` directives etc but in the end splitting the vars seems easiest and most simple solution.

runner/entrypoint.sh
```diff
@@ -56,7 +59,14 @@ sudo chown -R runner:docker /runner
 mv /runnertmp/* /runner/
 
 cd /runner
-./config.sh --unattended --replace --name "${RUNNER_NAME}" --url "${GITHUB_URL}${ATTACH}" --token "${RUNNER_TOKEN}" ${RUNNER_GROUP_ARG} ${LABEL_ARG} ${WORKDIR_ARG}
+./config.sh --unattended --replace \
+  --name "${RUNNER_NAME}" \
+  --url "${GITHUB_URL}${ATTACH}" \
+  --token "${RUNNER_TOKEN}" \
+  "${RUNNER_GROUP_ARG}" "${RUNNER_GROUP_ARG_VALUE}" \
```
mumoshu, line 66: Does entrypoint.sh or config.sh complain when `RUNNER_GROUP_ARG_VALUE` is not defined?

If not, I guess we can just rewrite these like:

```suggestion
  --runnergroup "${RUNNER_GROUP}" \
```

FloThinksPi: Tested it, it also works since .config uses defaults on empty arguments.
Groups get assigned to default.
Also default tags.
Workdir seems also to be default.
Changed it accordingly 👍🏽

mumoshu:
Confirmed it does work for `--labels` as well!

```
kind: RunnerDeployment
metadata:
  name: example-runnerdeploy
spec:
  template:
    spec:
      labels:
      - "mylabel 1"
      - "mylabel 2"
      #snip
```

Without this change:

image

With this change:

image

mumoshu (approved):
LGTM. Thanks a lot for your contribution @FloThinksPi ☺️

PR #430, opened 2021-03-31
docs: adding creation requirement for labels

callum-tait-pbx:
Fixes https://github.com/summerwind/actions-runner-controller/issues/429

### DIFF
--- a/README.md
+++ b/README.md
@@ -617,6 +617,8 @@ spec:
 
 ### Runner labels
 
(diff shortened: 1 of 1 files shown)


callum-tait-pbx:
May not be needed, going to do some testing.

callum-tait-pbx:
Couldn't replicate this again, must have been something weird in my environment. I suspect it might be something like if you have never made a label before via github.com then it won't work, once you've made a label once then subsequent labels of any value will apply going forward. 🤷

PR #432, opened 2021-03-31
fix: remove dockerd as args in runner controller

cbrand:
This fixes https://github.com/summerwind/actions-runner-controller/issues/431

Succesfully tested on a with the default runner image. It enables creation of the necessary docker certificates.

### DIFF
--- a/controllers/runner_controller.go
+++ b/controllers/runner_controller.go
@@ -703,7 +703,6 @@ func (r *RunnerReconciler) newPod(runner v1alpha1.Runner) (corev1.Pod, error) {
 		pod.Spec.Containers = append(pod.Spec.Containers, corev1.Container{
 			Name:  "docker",
 			Image: r.DockerImage,
-			Args:  []string{"dockerd"},
 			VolumeMounts: []corev1.VolumeMount{
 				{
 					Name:      "work",


controllers/runner_controller.go
```diff
@@ -703,7 +703,6 @@ func (r *RunnerReconciler) newPod(runner v1alpha1.Runner) (corev1.Pod, error) {
 		pod.Spec.Containers = append(pod.Spec.Containers, corev1.Container{
 			Name:  "docker",
 			Image: r.DockerImage,
-			Args:  []string{"dockerd"},
```
mumoshu, line 706: Thanks! I'm wondering if removing this breaks dockerMTU handling at L733. Let me check

mumoshu: Nevermind. This looks OK seeing

https://github.com/docker-library/docker/blob/b1d2628005e12e79079c025c3653cba248d6f264/19.03/dind/dockerd-entrypoint.sh#L94

vs

https://github.com/docker-library/docker/blob/b1d2628005e12e79079c025c3653cba248d6f264/19.03/dind/dockerd-entrypoint.sh#L137

With `Args:  []string{"dockerd"},` we go to the latter where default options that includes docker certs dirs are not set.

Without that we go to the former path where default options exists as before.

mumoshu (approved):
LGTM. Thank you so much for fixing it @cbrand!

PR #436, opened 2021-04-01
feat: allow to configure runner volume size limit

sp-manuel-jurado:
What?
- Allow to configure runner volume size limit

Why?
- I would be interested in this feature, to be able to set a limit size on the volume of the runner and avoid affecting other resources of the same cluster

Notes:
- I am not very familiar with the code. If there is a better way or I am posing it in the wrong way, please let me know and I'll try to improve this pr if you are interested in this feature. For me, this feature would be of great help

Thanks in advance


api/v1alpha1/runner_types.go
```diff
@@ -94,6 +94,9 @@ type RunnerSpec struct {
 	DockerEnabled *bool `json:"dockerEnabled,omitempty"`
 	// +optional
 	DockerMTU *int64 `json:"dockerMTU,omitempty"`
+
+	// +optional
+	RunnerVolumeSizeLimit string `json:"runnerVolumeSizeLimit,omitempty"`
```
mumoshu, line 99: Anything related to the runner container in the runner pod is currently not prefixed(like `Image`, `Env`, `EnvFrom`) so just for consistency, you could just use `VolumeSizeLimit`.

Also, I thought controller-runtime(although it's often said to be a part of kubebuilder) had support for Quantity field so you could directly use `*resource.Quantity` instead of `string`.

```suggestion
	VolumeSizeLimit *resource.Quantity `json:"volumeSizeLimit,omitempty"`
```

sp-manuel-jurado:
Hi @mumoshu I already fixed the code review.

But I can see, tests failing (volumeSizeLimit anyOf => string|integer):
```
--- a/config/crd/bases/actions.summerwind.dev_runners.yaml
+++ b/config/crd/bases/actions.summerwind.dev_runners.yaml
@@ -760,6 +760,12 @@ spec:
                   - name
                 type: object
               type: array
+            volumeSizeLimit:
+              anyOf:
+                - type: integer
+                - type: string
+              pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+              x-kubernetes-int-or-string: true
             volumes:
               items:
                 description: Volume represents a named volume in a pod that may be accessed by any container in the pod.
Error: Process completed with exit code 1.
```

Am I doing something the wrong way? (or need to modify something else...)

mumoshu:
@sp-manuel-jurado Thanks! I think you need to (re)run `make manifests` (or `make release`? I forgot) to regenerate the CRD templates included in the Helm chart and the CRD manifests, and commit them.

sp-manuel-jurado:
> @sp-manuel-jurado Thanks! I think you need to (re)run `make manifests` (or `make release`? I forgot) to regenerate the CRD templates included in the Helm chart and the CRD manifests, and commit them.

@mumoshu done ✅

mumoshu (approved):
@sp-manuel-jurado LGTM. Thanks for your contribution!

PR #437, opened 2021-04-01
Fix helm webhook ingress error: `port name or number is required`

javipolo:
Use the first exposed port for the service


mumoshu (approved):
LGTM. Thanks for the fix @javipolo!

peimanja:
can we please bump the chart number? I need this fix :)

PR #438, opened 2021-04-01
chore: bumping runner to ubuntu 20.04

callum-tait-pbx:
Fixes https://github.com/summerwind/actions-runner-controller/issues/422

I've done a build and deployed it our test cluster. I've ran the integration workflow I'm building out to validate upgrades. Everything that should pass passed, I've only just started it so it only tests a few things and it won't test everything as it's an integration test for my environment e.g. we don't use Ruby so it won't test that

The failed item isn't relevant as it relies on some stuff we add to our custom images so it's not surprising it failed

@mumoshu If there is anything in particular you want tested I can add it to the test suite and run a test.

image

Ubuntu version of runner

image


mumoshu:
@callum-tait-pbx Hey! Thanks for your support as always.

First- Would you mind contributing your integration test workflow? 😃

I was wondering if we could have two or more integration test workflows like that in our repository. We can then instruct everyone to use it for their own integration test workflow to run before upgrading actions-runner-controller, so that they are unlikely to break their production workflows.

Second- can we also leave existing ubuntu 18.04 dockerfiles and keep pushing the images? That way, users can still rollback to 18.04 whenever they like, as you could do on managed runners by specifying `ubuntu-18.04` instead of `ubuntu-latest`.

callum-tait-pbx:
> I was wondering if we could have two or more integration test workflows like that in our repository. We can then instruct everyone to use it for their own integration test workflow to run before upgrading actions-runner-controller, so that they are unlikely to break their production workflows.

Cool, that's a good idea to include a reference test workflow. I'll flesh out and commit a slightly more generic one tomorrow as it's 23:30 my time now and I am sleepy :D

> Second- can we also leave existing ubuntu 18.04 dockerfiles and keep pushing the images? That way, users can still rollback to 18.04 whenever they like, as you could do on managed runners by specifying ubuntu-18.04 instead of ubuntu-latest.

Makes sense, done 👍 I don't like the Dockerfile names, I'd rather it be `Dockerfile.%unique_bit%` so all the Dockerfiles are grouped together but I didn't fancy missing a reference somewhere and breaking things

callum-tait-pbx:
euh, it was bothering me too, much I've been brave and renamed them 💪 💪 💪

callum-tait-pbx:
> Second- can we also leave existing ubuntu 18.04 dockerfiles and keep pushing the images? That way, users can still rollback to 18.04 whenever they like, as you could do on managed runners by specifying ubuntu-18.04 instead of ubuntu-latest.

My apologies, I think I miss-read this last night, you want to push to DockerHub both a 18.04 and a 20.04 image? I can make the change for this, any thoughts on tagging though? If we are going to push multiple OS versions of the runner we should probably include the OS in the images tag?

actions-runner:%OS%-%RUNNER_VERSION%
actions-runner:ubuntu-18.04-v2.277.1
actions-runner:ubuntu-20.04-v2.277.1

We could just add the OS tag to just 18.04 but when happens when ubuntu-20.04 gets replaced? I suppose realistically no one is going to be running a super old runner version (or they shouldn't be) so the may not cause a problem? What do you think @mumoshu ? I'm mainly thinking of the situtation where the OS untagged runner on 1 release goes from OS version foo and then the next verison bar. We could move to adopting the same setup as GitHub to alleviate this?

actions-runner:ubuntu-18.04-v2.277.1
actions-runner:ubuntu-latest-v2.277.1

The other aspect is that github have on their public roadmap the goal of offering an expanded selection of machiens. This may include other OS versions (maybe not though, it may just be around machine sizes, hardware accleration options etc). If they do introduce more OS options then it may not be as clear as it is today that the `actions-runner` is Ubuntu based without investigating the machine / Docekrfile

https://github.com/github/roadmap/issues/161
https://github.com/github/roadmap/projects/1?card_filter_query=label%3Aactions `Q3 2021 – Jul-Sep`

runner/Makefile
```diff
@@ -22,29 +22,28 @@ else
...
+docker-buildx-ubuntu:
 	export DOCKER_CLI_EXPERIMENTAL=enabled
 	@if ! docker buildx ls | grep -q container-builder; then\
 		docker buildx create --platform ${PLATFORMS} --name container-builder --use;\
 	fi
 	docker buildx build --platform ${PLATFORMS} \
 		--build-arg RUNNER_VERSION=${RUNNER_VERSION} \
 		--build-arg DOCKER_VERSION=${DOCKER_VERSION} \
-		-t "${NAME}:latest" \
+		-t "${NAME}:ubuntu-latest" \
```
mumoshu, line 41: Although I don't particularly like using `latest` in production, I think this should better remain `latest` for backward compatibility and docker convention to use `latest` as... `latest`. 
```suggestion
		-t "${NAME}:latest" \
```

acceptance/pipelines/eks-integration-tests.yaml
```diff
@@ -0,0 +1,36 @@
+name: EKS Integration Tests
```
mumoshu, line 1: 😍

.github/workflows/build-and-release-runners.yml
```diff
@@ -66,6 +72,6 @@ jobs:
             RUNNER_VERSION=${{ env.RUNNER_VERSION }}
             DOCKER_VERSION=${{ env.DOCKER_VERSION }}
           tags: |
-            ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:v${{ env.RUNNER_VERSION }}
-            ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:v${{ env.RUNNER_VERSION }}-${{ steps.vars.outputs.sha_short }}
-            ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:latest
+            ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:ubuntu-${{ matrix.os-version }}-v${{ env.RUNNER_VERSION }}
+            ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:ubuntu-${{ matrix.os-version }}-v${{ env.RUNNER_VERSION }}-${{ steps.vars.outputs.sha_short }}
+            ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:ubuntu-latest
```
mumoshu, line 77: Thanks for the input!!

I took a few days to this about what tagging strategy work best for us. Following conventions seen in popular docker images like https://hub.docker.com/_/docker and https://hub.docker.com/_/golang, I got to think we'd better use...:

```suggestion
            ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:v${{ env.RUNNER_VERSION }}-ubuntu-${{ matrix.os-version }}
            ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:v${{ env.RUNNER_VERSION }}-${{ steps.vars.outputs.sha_short }}-ubuntu-${{ matrix.os-version }}
            ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:ubuntu-${{ matrix.os-version }}
```

Also, I think including `${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:ubuntu-latest`, a.k.a` laetst` here could result in the `ubuntu-latest` tag unstable(can randomly change between be 18.04 or 20.04, where we would expect it to consistently be 20.04), due to that it runs on both `os-version: 20.04` and `os-version: 18.04` for `name: actions-runner`.

mumoshu: I haven't tested it yet myself but probably we need a dedicated step for building `ubuntu-latest` a.k.a` `latest`, like below? Note that I added `if: ${{ matrix.os-version == '20.04' }}` so that it only runs once per workflow run.

```
      - name: Build and Push (latest)
        uses: docker/build-push-action@v2
        with:
          context: ./runner
          file: ./runner/${{ matrix.dockerfile }}
          platforms: linux/amd64,linux/arm64
          push: ${{ github.event_name != 'pull_request' }}
          if: ${{ matrix.os-version == '20.04' }}
          build-args: |
            RUNNER_VERSION=${{ env.RUNNER_VERSION }}
            DOCKER_VERSION=${{ env.DOCKER_VERSION }}
          tags: |
            ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:v${{ env.RUNNER_VERSION }}
            ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:latest
```

ToMe25:
First of all thanks for this PR, i intended to do something similar later, once i figured some things out, but now i don't have to :smiley:
But also it should now be possible to replace dockerfile line 40(`&& cd /usr/bin && ln -sf python3 python \`) with the `python-is-python3` package like in the dindrunner.
This is purely a consistency change, but i think it would be good for consistency reasons anyways.

.github/workflows/build-and-release-runners.yml
callum-tait-pbx: Alright! So in conclusion tagging wise we are going with this format:

`actions-runner:v2.277.1-ubuntu-18.04`
`actions-runner:v2.277.1-ubuntu-20.04`
`actions-runner:latest` <--- this is going to be considered ubuntu regardless of if more OS types get introduced. Ubuntu is considered our default runner and so latest is implicitly OS type `ubuntu`. This also ensures we keep backwards compatibility too and no weirdness with docker standards.

I think you are saying we won't bother with an `ubuntu-latest` right, keep regular `latest` instead for the above reasons?

mumoshu: Exactly! Thanks for summarizing

mumoshu:
@ToMe25 Thanks for chiming in!
Honestly speaking, I have no clue what `python-is-python3`does- can it be a drop-in replacement to `python3 python`?

callum-tait-pbx:
https://github.com/summerwind/actions-runner-controller/pull/361/files looks like it was introduced to allow the `shell: python` syntax to work

I'll try streamlining the Python packages and subsequently test:
- Setting up and validating a few python versions with action/setup-python
- Running that python shell snippet (I'll add this to the intergation test pipelines too)

ToMe25:
The `python-is-python3` package basically only contains the python symlink to python3 that is manually created later in the dockerfile.
Whether this is any better then creating the link manually is up for debate, however i think it would make sense to handle this the same way as in the dindrunner image, which uses the `python-is-python3` package.
I would be fine with removing the package there and creating the link manually there too tho.

callum-tait-pbx:
**Test Steps Taken**
1. ran `make` on the runner folder, this produced:
```
summerwind/actions-runner-dind   latest            b09cefc23007   7 seconds ago    940MB
summerwind/actions-runner        latest            f29b56dd3e83   37 seconds ago   768MB
```

2. Pushed and tested both images in our  est cluster and ran our work in progress test suite on each, same result each time:

image

_The failed test is because it requires https://github.com/awslabs/amazon-ecr-credential-helper to be included in the runner and configured appropriately_

@mumoshu this is ready for review again

callum-tait-pbx:
noticed the docker-buildx stuff still have `ubuntu-latest` so aligned that with the `latest` tag as discussed

Mic92 (approved):
Worked out for us! Thanks for doing this.

mumoshu (approved):
LGTM! Thank you so much for all your efforts ☺️

PR #441, opened 2021-04-06
Feat/harden actions runner controller

anandg112:
This PR hardens the actions runner controller with some best practices - running as non-root, using default `seccomp` profile and removing `sudo` from entrypoint script for the runner.

### DIFF
--- a/charts/actions-runner-controller/templates/deployment.yaml
+++ b/charts/actions-runner-controller/templates/deployment.yaml
@@ -5,6 +5,8 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "actions-runner-controller.labels" . | nindent 4 }}
+  annotations:
+    seccomp.security.alpha.kubernetes.io/pod: runtime/default
 spec:
   replicas: {{ .Values.replicaCount }}
   selector:
--- a/charts/actions-runner-controller/values.yaml
+++ b/charts/actions-runner-controller/values.yaml
@@ -23,7 +23,7 @@ authSecret:
 image:
   repository: summerwind/actions-runner-controller
   tag: "v0.17.0"
-  dindSidecarRepositoryAndTag: "docker:dind"
+  dindSidecarRepositoryAndTag: "docker:20.10-dind-rootless"
   pullPolicy: IfNotPresent
 
 kube_rbac_proxy:
@@ -47,17 +47,15 @@ serviceAccount:
 podAnnotations: {}
 
 podSecurityContext:
-  {}
-  # fsGroup: 2000
+  fsGroup: 1000
 
 securityContext:
-  {}
   # capabilities:
   #   drop:
   #   - ALL
   # readOnlyRootFilesystem: true
   # runAsNonRoot: true
-  # runAsUser: 1000
+  runAsUser: 1000
 
 service:
   type: ClusterIP
--- a/runner/entrypoint.sh
+++ b/runner/entrypoint.sh
@@ -44,7 +44,7 @@ if [ ! -d /runner ]; then
   exit 1
 fi
 
-sudo chown -R runner:docker /runner
+chown -R runner:docker /runner
 mv /runnertmp/* /runner/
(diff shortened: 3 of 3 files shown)


runner/entrypoint.sh
```diff
@@ -44,7 +44,7 @@ if [ ! -d /runner ]; then
   exit 1
 fi
 
-sudo chown -R runner:docker /runner
+chown -R runner:docker /runner
```
callum-tait-pbx, line 47: Have you tested this? Are you able to attach a screenshot to the PR showing this still works without sudo?

charts/actions-runner-controller/values.yaml
```diff
@@ -47,17 +47,15 @@ serviceAccount:
...
 
 securityContext:
-  {}
   # capabilities:
   #   drop:
   #   - ALL
   # readOnlyRootFilesystem: true
   # runAsNonRoot: true
-  # runAsUser: 1000
+  runAsUser: 1000
```
callum-tait-pbx, line 58: Does this break EKS IRSA integration? Are you able to still read the bootstrapped web identity token file? Which user is 1000 again?

charts/actions-runner-controller/templates/deployment.yaml
```diff
@@ -5,6 +5,8 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "actions-runner-controller.labels" . | nindent 4 }}
+  annotations:
+    seccomp.security.alpha.kubernetes.io/pod: runtime/default
```
callum-tait-pbx, line 9: I know nothing about this, the controller and kube-proxy are based on a distroless images:
https://github.com/summerwind/actions-runner-controller/blob/master/Dockerfile
https://github.com/brancz/kube-rbac-proxy/blob/master/Dockerfile
does this actually do anything in the context of a distroless image?

anandg112: Yes, it provides sane defaults on which syscalls are allowed https://kubernetes.io/docs/tutorials/clusters/seccomp/#create-pod-that-uses-the-container-runtime-default-seccomp-profile

charts/actions-runner-controller/values.yaml
anandg112: It's a user with non-root privileges, linux starts creating normal users at UID 1000. You are right though, this may not do much as distroless images are already being used and run as non-root.

runner/entrypoint.sh
anandg112: Let me know how to test this and I will do so

mumoshu: It would be great if you could try deploying actions-runner-controller with the modified chart and the runner image and see if a simple GitHub Actions workflow runs through!

FYI, to build a custom runner image, you can run:

```
NAME=$DOCKER_USER/actions-runner-controller \
  make docker-build docker-push
```

Also https://github.com/summerwind/actions-runner-controller#developing

charts/actions-runner-controller/templates/deployment.yaml
callum-tait-pbx: Cool, needs testing. Is there a reason you have implemented it the < v.1.19 way instead of the => 1.19 way?

The doc says it should be implemented like this for V1.19 or later:

```yaml
spec:
  securityContext:
    seccompProfile:
      type: RuntimeDefault
```

Mic92:
In my own image I do the following for now:

```
# entrypoint requires sudo, but we want to remove before starting action runner
RUN sudo sed -i -e '/exec/i sudo SUDO_FORCE_REMOVE=yes apt remove -y sudo/' /entrypoint.sh
```

It basically removes sudo before starting the github action.

mumoshu:
@anandg112 Hey! Please let me adopt/revisit this once we've figured out how to automate more comprehensive testing in https://github.com/actions-runner-controller/actions-runner-controller/issues/564.

PR #447, opened 2021-04-10
feat: allow autoscaler to scale down to 0

cbrand:
This is a tiny change which allows horizontalrunnerautoscalers to be able to scale down a deployment to "0".

We utilize on our Google Cloud instance the autoscaler with webhooks and this scaler to dynamically scale down and up when a build is requested. The deployments are run on dedicated servers which are dynamically scaled up and down when needed.

In this scenario and if no build runs at all in our account we want to be able to scale down to "0".

Tested on a google cloud cluster, the change seems to work as expected.

### DIFF
--- a/controllers/horizontalrunnerautoscaler_controller.go
+++ b/controllers/horizontalrunnerautoscaler_controller.go
@@ -19,8 +19,9 @@ package controllers
 import (
 	"context"
 	"fmt"
-	corev1 "k8s.io/api/core/v1"
 	"time"
+
+	corev1 "k8s.io/api/core/v1"
 
 	"github.com/summerwind/actions-runner-controller/github"
 	"k8s.io/apimachinery/pkg/types"
@@ -186,7 +187,7 @@ func (r *HorizontalRunnerAutoscalerReconciler) SetupWithManager(mgr ctrl.Manager
 
 func (r *HorizontalRunnerAutoscalerReconciler) computeReplicasWithCache(log logr.Logger, now time.Time, rd v1alpha1.RunnerDeployment, hra v1alpha1.HorizontalRunnerAutoscaler) (int, int, *int, error) {
 	minReplicas := defaultReplicas
-	if hra.Spec.MinReplicas != nil && *hra.Spec.MinReplicas > 0 {
+	if hra.Spec.MinReplicas != nil && *hra.Spec.MinReplicas >= 0 {
 		minReplicas = *hra.Spec.MinReplicas
 	}
 


mumoshu:
@cbrand Hey! Just a quick question but how does scale-from-zero work? AFAIK, GitHub Actions immediately failed a queued workflow run when there's no available runners to assign it. That effectively blocked us from automatically scaling from zero to non-zero on a new workflow run.

cbrand:
@mumoshu Hey. I actually did observe a different behavior. I have the github webhooks installed and it calls the service as expected. This autoscales the workers and github actions waits till the pod scaled up and then executes the actions.

Just retested it and it works with the scale down from 0 when utilizing webhooks for these jobs. I require this scheduler to verify-ably scale down as the webhook downscaling doesn't seem to work as expected for me.

mumoshu:
@cbrand Hey! Thanks for the info. To be extra sure, you've verified scale-up-from-zero to work with either PercentageRunnersBusy or TotalNumberOfQueuedAndInProgressWorkflowRuns, and you've also verified that scale-up-from-zero with the webhook-based autoscaling to work, right?

> I require this scheduler to verify-ably scale down as the webhook downscaling doesn't seem to work as expected for me.

Sry but I couldn't fully understand this. Did you see any other behavior than the webhook autoscaler "expires the reserved capacity after the duration you've configured"(`5m` in the below example)?

```
  scaleUpTriggers:
  - githubEvent:
      checkRun:
        types: ["created"]
        status: "queued"
    amount: 1
    duration: "5m"
```

To be clear, the above configurations means that you want to add `1` replica only for `5 minutes` after the autoscaler received the webhook event.

cbrand:
@mumoshu I only tested it with github-webhooks and just also tried it with `PercentageRunnersBusy` which doesn't scale up from `0` which intuitively makes sense.

`TotalNumberOfQueuedAndInProgressWorkflowRuns` is working as expected.

I am not quite sure how to handle the `PercentageRunnersBusy` case though. You could make an argument that this is a miss configuration and should not scale up at all, but you could also say that the software should anticipate this and always handle a scale up to at least one. Number 2 I implemented in a separate PR as it is a relatively quick fix: https://github.com/summerwind/actions-runner-controller/pull/449 but am not sure if it should behave like this or not. Please check which behavior you find more logical.

I actually couldn't reproduce my scale down to 0 issues with any autoscaler anymore, so the downscaling issue I was talking about seems to be not valid and is working as expected..

mumoshu:
@cbrand Hey! Thanks a lot for the clarifications. I'll test this myself and will merge asap.

In the meantime, the only one thing I'm unsure about now is that if the scale-up-from-zero scenario is working only because you somehow had `offline` runners remaining registered on your GitHub repository/organization/etc.

When I first read about how GitHub queues actions jobs, it was said that it is able to queue jobs(without immediately failing due to now runners) only when you had one or more runners, regardless of what statuses runners are in.

callum-tait-pbx:
> In the meantime, the only one thing I'm unsure about now is that if the scale-up-from-zero scenario is working only because you somehow had offline runners remaining registered on your GitHub repository/organization/etc.

The docs line up with your logic @mumoshu  https://docs.github.com/en/actions/hosting-your-own-runners/using-self-hosted-runners-in-a-workflow#routing-precedence-for-self-hosted-runners

There needs to be at least 1 runner registered in any state for a workflow to not immediately fail. We should only support a 0 runner count for webhooks, both the pull methods will require at least 1 runner to be registered at all times.

mumoshu:
@callum-tait-pbx Thanks for the info! That looks like what I read before.

Also, I've manually (re)verified that scaling-from-zero works only when you have at least 1 offline worker(s).

FWIW, it failed without leaving any useful error message about missing runners this time.

For the workflow run:

image

For the specific job:

image

I initially thought that the scale-from-zero would work with the webhook-based autoscaling, but apparently it doesn't.

The webhook-based autoscaling only adds the missing pod after a minute or so that is delay due to the container startup and the time taken to complete `./configure.sh` for runner registration. That said, trying to queue a job against a webhook-autoscaling-enabled runnerdeployment would immediately fail due to runners being still zero when the GitHub Actions tries to find the runner.

Going forward, we'll need a way to somehow reserve offline runners only to allow GitHub Actions to queue jobs. I've read https://github.com/actions/runner and it turned out we need to either reverse-engineer it and use e.g. https://github.com/microsoft/azure-devops-go-api which seems compatible with the internal API used by runners or do some serious shell scripting to call `configure.sh` to setup offline runners.

mumoshu:
Note that when we go to the azure-devops-go-api way, we'd need to reimplement most of https://github.com/actions/runner/blob/fc4027b3f1ea66e26c334a2142da25a34f10ec8c/src/Runner.Listener/Configuration/ConfigurationManager.cs#L67-L327 with it in actions-runner-controller.

cbrand:
@mumoshu Thanks for figuring this out! I am kind of confused why that never happened up until now in the setup we run. I try to double check on the weekend if there were offline runners (cannot it during the week as people rely on the service there).

mumoshu:
@cbrand Hey! I've managed to get #465 working. With that change in place, you always get one `offline` runner per runner deployment for free. In combination with changes introduced via this PR, it should fully support the scale-from/to-zero scenario.

I'm pretty out of my time today, but it would be great if you could help testing the mix of #465 and this!

mumoshu:
@cbrand #465 has been merged. Let's merge this so that `TotalNumberOfQueuedAndInProgressWorkflowRuns` would reliably work in a scale-from-zero scenario.

I've also read https://github.com/actions-runner-controller/actions-runner-controller/pull/449 but still not sure what to do with PercentageRunnersBusy.

I think we have a few potential solutions.

The first is to let HRA controller use either PercentageRunnersBusy or TotalNumberOfQueuedAndInProgressWorkflowRuns that gives you larger desired replicas.

I thought HorizontalPodAutoscaler had support for multiple metrics and used the latest recommended replicas. So probably this looks more like a standard approach.

The second is to let the user configure webhook-based autoscaling, too. As long as the PercentageRunnersBusy process comes up until the temporary replicas added by the webhook-based autoscaling expires, this should work.

cbrand:
@mumoshu Thank you so much for implementing the required functionality for making this possible!

I sadly was quite busy the last few weeks. I test out the feature this week to see how it reacts on our cluster infrastructure. It looks very good and by now I also saw one time in the cluster that scale down from zero didn't work as github actions didn't had a registered runner. I guess this has mainly been not happening as there are "legacy" runners still installed as the workers are usually spinned up and down quite frequently.

PR #448, opened 2021-04-10
Make acceptance use local docker image

ToMe25:
This Pull Request makes the acceptance build target load the local docker image to the kind cluster.
This is saves both time and bandwidth, and should thus be preferable over pushing it and then pulling it 4 times.
This PR also adds a small delay in acceptance/checks.sh and increases the ready timeout in the same file, because while everything else was working most of the time with the default timeouts, this consistently failed, even with an average download speed of 40 mbit.

I would also like to locally cache the other docker images used by make acceptance on the host, as downloading them 4 times seems like a waste of time and bandwidth.
However this would likely require changing their ImagePullPolicy to IfNotPreset, so i wanted to ask here whether that would have a chance of getting merged.
So would it have a chance?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,7 +126,7 @@ release/clean:
 	rm -rf release
 
 .PHONY: acceptance
-acceptance: release/clean docker-build docker-push release
+acceptance: release/clean docker-build release
 	ACCEPTANCE_TEST_SECRET_TYPE=token make acceptance/kind acceptance/setup acceptance/tests acceptance/teardown
 	ACCEPTANCE_TEST_SECRET_TYPE=app make acceptance/kind acceptance/setup acceptance/tests acceptance/teardown
 	ACCEPTANCE_TEST_DEPLOYMENT_TOOL=helm ACCEPTANCE_TEST_SECRET_TYPE=token make acceptance/kind acceptance/setup acceptance/tests acceptance/teardown
@@ -134,6 +134,7 @@ acceptance: release/clean docker-build docker-push release
 
 acceptance/kind:
 	kind create cluster --name acceptance
+	kind load docker-image ${NAME}:${VERSION} --name acceptance
 	kubectl cluster-info --context kind-acceptance
 
 acceptance/setup:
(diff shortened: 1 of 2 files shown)


mumoshu (approved):
LGTM. Thanks for the improvement!

> However this would likely require changing their ImagePullPolicy to IfNotPreset

That would be a nice addition to https://github.com/summerwind/actions-runner-controller/blob/325c2cc385ce312909203e6620326ff943503556/acceptance/deploy.sh#L29

(I already have some local modifications in that file so I won't say the change is required, but it would definitely worth being added

Makefile
```diff
@@ -126,14 +126,15 @@ release/clean:
...
-acceptance: release/clean docker-build docker-push release
+acceptance: release/clean docker-build release
 	ACCEPTANCE_TEST_SECRET_TYPE=token make acceptance/kind acceptance/setup acceptance/tests acceptance/teardown
 	ACCEPTANCE_TEST_SECRET_TYPE=app make acceptance/kind acceptance/setup acceptance/tests acceptance/teardown
 	ACCEPTANCE_TEST_DEPLOYMENT_TOOL=helm ACCEPTANCE_TEST_SECRET_TYPE=token make acceptance/kind acceptance/setup acceptance/tests acceptance/teardown
 	ACCEPTANCE_TEST_DEPLOYMENT_TOOL=helm ACCEPTANCE_TEST_SECRET_TYPE=app make acceptance/kind acceptance/setup acceptance/tests acceptance/teardown
 
 acceptance/kind:
 	kind create cluster --name acceptance
+	kind load docker-image ${NAME}:${VERSION} --name acceptance
```
mumoshu, line 137: I have not tried myself yet but probably- `kind load`ing the runner image as well would further speed up the test run?

ToMe25: @mumoshu That is one of the other images i was talking about to cache on the host :smiley: 
So i would probably download them to the host docker once, and then kind load on the four clusters.
I also think it would make sense to have the change you told me in deploy.sh should be done in the repo, if this is an issue for everyone.
Should i just add that to my next PR?

mumoshu: @ToMe25 Ah gotcha! Yeah, adding that to your next PR would be very much appreciated 👍

PR #449, opened 2021-04-11
feat: Allow Percentage runner busy minimum 1

cbrand:
This is a PR which is built on top of https://github.com/summerwind/actions-runner-controller/pull/447 and ensures the `PercentageRunnersBusy` to always return 1 when scaling if it is lower than 0. Please refer to the original PR for further details.


mumoshu:
@cbrand Hey! Would mind checking #522 and #524? In nutshell, I'm going to enable scale-from-zero for PercentageRunnersBusy when it has a secondary metric of TotalInProgoressAndQueuedWorkflowRuns.

Probably we can still make it a validation error when `minReplicas` is set to 0 while the primary metric being PercentageRunnersBusy but there's no secondary metric or webhook-based autoscaling defined. But that's another story.

PR #450, opened 2021-04-12
docs: correct and simplify example

callum-tait-pbx:
Fixes https://github.com/summerwind/actions-runner-controller/issues/446 (sort of)

Also removed the alternatives section and this is covered really by the [awesome-runners](https://github.com/jonico/awesome-runners) tag at this point, when this section was added originally awesome runners wasn't a thing

Wasn't sure if there was a reason the Enterprise example was with the dind runner but I've removed it as I suspect not.


mumoshu (approved):
LGTM. Thanks a lot for your continued effort @callum-tait-pbx :+1:

Issue #265, opened 2021-01-20
hostAliases for runner pods


reallydontask:
Is there any way of getting hostAliases into the runner pods?

Essentially, what I'd like to achieve is to be able to modify the /etc/hosts file so that the runners can connect to jumpboxes with their current names

I've tried the following but to no avail, see below

Any suggestions?

```
apiVersion: actions.summerwind.dev/v1alpha1
kind: RunnerDeployment
metadata:
  name: ansible
  namespace: actions-runner-system
spec:
  replicas: 1
  template:   
    spec:  
      hostAliases:
       - ip: 10.10.1.1
         hostnames:
         - jumpbox    
      repository: ministryofjustice/ansible
      image: ministryofjustice/runner
      labels:
        - runner
```

mumoshu:
This has been implemented in #456. Thanks for your contribution, @reallydontask!

PR #456, opened 2021-04-14
feat: allow runners to configure hostAliases

reallydontask:
This PR will allow the configuration of hostAliases on the runner pods.

Our ansible code is deployed across multiple environments and relies on each environment having a jumpbox CNAME DNS record, so that the deployment is always from jumpbox, thus having this is crucial.

Hopefully this will also be useful to other users too.

I should add that I raised #265, which this would solve

Thanks in advance


mumoshu (approved):
LGTM. Thanks for the addition @reallydontask ☺️

PR #459, opened 2021-04-17
docs: removing superfluous title reference

callum-tait-pbx:
Missed this in https://github.com/summerwind/actions-runner-controller/pull/450


mumoshu (approved):
Good catch! Thanks for the fix @callum-tait-pbx ☺️

PR #460, opened 2021-04-17
docs: cleaning up Enterprise and fsGroup docs

callum-tait-pbx:
Cleaned up the Enterprise documenation so it's a bit clearer and aligns with the docs better. I don't have a GitHub Enterprise Server environment but I think I haven't said anything wrong. The docs weren't great on this as Actions has had GitHub Enterprise support since the beginning but they read like the enterprise stuff relates a enterprise cloud environment when really it is specifically for the server edition.

The other change I've done is change fsGroup. There isn't a group with the GID `1447` so this doesn't make sense although it does work. The `runner` user is in the following groups and the groups and those groups have GIDs of:

```shell
runner@default-runner-deployment-798vd-bx6dz:/$ groups
runner sudo docker

root:x:0:
sudo:x:27:runner
runner:x:1000:
docker:x:1001:runner
```

I think the `fsGroup` should be set to 1000 as a result? I'm making that change to my environment Monday so I can confirm if it all works then but I don't see why it wouldn't.

Please read through @mumoshu when you get a chance to review the PR just to double check I haven't made a typo or mistake with all the changes I was making.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ ToC:
 
 - [Motivation](#motivation)
 - [Installation](#installation)
-  - [GitHub Enterprise support](#github-enterprise-support)
+  - [GitHub Enterprise Server Support](#github-enterprise-server-support)
 - [Setting up authentication with GitHub API](#setting-up-authentication-with-github-api)
   - [Deploying using GitHub App Authentication](#deploying-using-github-app-authentication)
   - [Deploying using PAT Authentication](#deploying-using-pat-authentication)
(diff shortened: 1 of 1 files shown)


callum-tait-pbx:
@zetaab I'm just cleaning up the docs for the Enterprise Server support feature.

Does the `enterprise` attribute the work the same way as the `organisation` attribute or is there a slight difference? Is it a case of you point a kind e.g. a `runner` or a `runnerdeployment` at your enterprise and then use runner groups to manage org wise runners to individual organisations within that organisation or is literally the same as an `organisation` on github enterprise cloud? The docs aren't super clear and it's a bit abstract to my brain.

zetaab:
Basically yes, with enterprise you can make whole github enterprise server wide runners. I do not have experience about github enterprise cloud.

callum-tait-pbx:
How does it work for repository level runners? Do you need to specify the enterprise attribute as well as the specific repositories?

zetaab:
Autoscaling does not work for enterprise runners afaik

callum-tait-pbx:
So as it stands you can just deploy sets of runners with the `RunnerDeployment` kind which a static amount of runners as defined by the replicas attribute? You can't then couple those RunnerDeployments with a HRA for autoscaling? How can you target individual sets of runnerdeployments, does it suppoer labels? Can you use the `runner` kind to assign a runner to an individual repository?

zetaab:
Enterprise level runners needs to be approved to repositories in organisation level in github. We use currently runnerdeployment with static amount of replicas. Runnerdeployment supports labels and those works also in enterprise level runners. HRA is not supported.

callum-tait-pbx:
Cool, last bit is do you have repository specific runners in GitHub Enterprise Server? Does this implementation support them if so or does it work only via regsitering runners against the Enterprise level only and then allocate those runner groups to repositories?

zetaab:
With github enterprise server we can also use normal organisation/repo level runners that people do use in github.com. The only difference is enterprise level runners. There is no such thing like register enterprise level runner and use only in one repo. Instead we should in that case just use repo specified runner (like in github.com)

callum-tait-pbx:
Ah, so from you've said and reading https://docs.github.com/en/github-ae@latest/admin/user-management am I right in saying Enterprise runners are a layer above your organisations. You are able to have runners avaliable to your enterprise which could mean multiple organisations and whatever set of repositories you allocate the runner groups to?

You can still make use of the other kinds in this project, enterprise runners ar ejust a unique type of runner specific to the server edition of github enterprise?

callum-tait-pbx:
I can confirm changing the fsGroup to 1000 hasn't broken assume in my environment.

Updated docs to reflect @zetaab 's comments

image

mumoshu (approved):
Whoa! This is A LOT easier to read now. Thank you so much for your contribution @callum-tait-pbx and @zetaab ☺️

PR #462, opened 2021-04-17
ci: fix latest tag push logic

callum-tait-pbx:
Fixes https://github.com/summerwind/actions-runner-controller/issues/461

We we always referencing `Dockerfile` as our source Dockerfile for the latest tag build / push for each item in the matrix. As a result the dind latest image was the non-dind image.


mumoshu (approved):
Ah, gotcha! Thanks for fixing it 🙏

PR #469, opened 2021-04-21
docs: adding scaling configuration detail

callum-tait-pbx:
Taken from the comments on https://github.com/summerwind/actions-runner-controller/issues/206. Including the `replicas` attribute in conjuction with the `minReplicas` and `maxReplicas` attributes causes issues apparently


mumoshu (approved):
Nice! Thanks as always for your effort @callum-tait-pbx ❤️

PR #471, opened 2021-04-21
docs: slightly better wording about support

callum-tait-pbx:
I think this is right right? The maintainers don't have an Enterprise env cloud _OR_ server edition right?


mumoshu (approved):
Absolutely! Thanks for clarifying that ☺️

PR #472, opened 2021-04-21
docs: improved details on authentication

callum-tait-pbx:
Didn't realise you can configure your rate limiting settings withi GitHub Enterprise Server so I thought I would put that detail in


mumoshu (approved):
Wow, that's good to know! Thanks for the info 👍

PR #473, opened 2021-04-22
Locally build runner image instead of pulling it

### LINKED ISSUE #464: acceptance helm test uses old actions-runner-controller image
Running make acceptance runs two kubectl and two helm tests, and while the kubectl tests use the newly built docker image, the helm ones seem to always use v0.17.0.
Running `kubectl describe pod -n actions-runner-system actions-runner-controller-7b5686c456-knwgp` returns this: https://pastebin.com/Vqayd2LH
This line is what causes me to assume it uses this version: `Normal   Pulling      55s                kubelet            Pulling image "summerwind/actions-runner-controller:v0.17.0"`.
Pulling this version here seems to indicate it doesn't use the newly built docker image, but instead version 0.17.0.

ToMe25:
This PR changes `make acceptance` to locally build the actions-runner image, instead of pulling it from dockerhub.
This is done by adding the building of the runner image to the docker-build target.
This should make it easier to test changes to the runner.
The name of the actions runner image defaults to `DOCKER_USER/actions-runner`, where `DOCKER_USER` is either specified as a environment variable, or inferred from `NAME`.
The docker-push target also pushes the runner image.

This PR also Fixes #464.

Unlike my previous PRs this one does make the acceptance build target take longer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,10 +107,7 @@ generate: controller-gen
 # Build the docker image
 docker-build: test
 	docker build . -t ${NAME}:${VERSION}
-
-# Push the docker image
-docker-push:
-	docker push ${NAME}:${VERSION}
+	docker build runner -t summerwind/actions-runner:${VERSION} --build-arg TARGETPLATFORM=$(shell arch)
 
 docker-buildx:
 	export DOCKER_CLI_EXPERIMENTAL=enabled
@@ -124,6 +121,18 @@ docker-buildx:
 		-f Dockerfile \
 		. ${PUSH_ARG}
 
+# Pull the docker images for acceptance
+docker-pull: docker-build
+        docker pull quay.io/brancz/kube-rbac-proxy:v0.8.0
+        docker pull docker:dind
+        docker pull quay.io/jetstack/cert-manager-controller:v1.0.4
+        docker pull quay.io/jetstack/cert-manager-cainjector:v1.0.4
+        docker pull quay.io/jetstack/cert-manager-webhook:v1.0.4
+
+# Push the docker image
+docker-push:
+        docker push ${NAME}:${VERSION}
+
 # Generate the release manifest file
 release: manifests
 	cd config/manager && kustomize edit set image controller=${NAME}:${VERSION}
@@ -135,8 +144,7 @@ release/clean:
 	rm -rf release
 
 .PHONY: acceptance
-acceptance: release/clean docker-build release
-	make acceptance/pull
+acceptance: release/clean docker-pull release
 	ACCEPTANCE_TEST_SECRET_TYPE=token make acceptance/kind acceptance/setup acceptance/tests acceptance/teardown
 	ACCEPTANCE_TEST_SECRET_TYPE=app make acceptance/kind acceptance/setup acceptance/tests acceptance/teardown
 	ACCEPTANCE_TEST_DEPLOYMENT_TOOL=helm ACCEPTANCE_TEST_SECRET_TYPE=token make acceptance/kind acceptance/setup acceptance/tests acceptance/teardown
@@ -146,20 +154,12 @@ acceptance/kind:
 	kind create cluster --name acceptance
 	kind load docker-image ${NAME}:${VERSION} --name acceptance
 	kind load docker-image quay.io/brancz/kube-rbac-proxy:v0.8.0 --name acceptance
-	kind load docker-image summerwind/actions-runner:latest --name acceptance
+	kind load docker-image summerwind/actions-runner:${VERSION} --name acceptance
 	kind load docker-image docker:dind --name acceptance
 	kind load docker-image quay.io/jetstack/cert-manager-controller:v1.0.4 --name acceptance
 	kind load docker-image quay.io/jetstack/cert-manager-cainjector:v1.0.4 --name acceptance
 	kind load docker-image quay.io/jetstack/cert-manager-webhook:v1.0.4 --name acceptance
 	kubectl cluster-info --context kind-acceptance
-
-acceptance/pull:
-	docker pull quay.io/brancz/kube-rbac-proxy:v0.8.0
-	docker pull summerwind/actions-runner:latest
-	docker pull docker:dind
-	docker pull quay.io/jetstack/cert-manager-controller:v1.0.4
-	docker pull quay.io/jetstack/cert-manager-cainjector:v1.0.4
-	docker pull quay.io/jetstack/cert-manager-webhook:v1.0.4
 
 acceptance/setup:
 	kubectl apply --validate=false -f https://github.com/jetstack/cert-manager/releases/download/v1.0.4/cert-manager.yaml	#kubectl create namespace actions-runner-system
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,12 @@
-ifdef USERNAME
-	NAME ?= ${USERNAME}/actions-runner-controller
+ifdef DOCKER_USER
+	NAME ?= ${DOCKER_USER}/actions-runner-controller
 else
 	NAME ?= summerwind/actions-runner-controller
 endif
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -124,6 +121,18 @@ docker-buildx:
 		-f Dockerfile \
 		. ${PUSH_ARG}
 
+# Pull the docker images for acceptance
+docker-pull: docker-build
```
mumoshu, line 125: Can we use a bit different style for anything related to dependencies, like `docker-pull-deps`?

`docker-whatever` was used to build/push the controller image until now so I thought it would be nice to make a distinction

ToMe25:
I can just change it back to acceptance/pull, i just thought it makes sense to name it docker-something because it's purely a docker target.But first I'll have to fix some other issue.Am 23.04.2021 02:44 schrieb Yusuke Kuoka ***@***.***>:
@mumoshu commented on this pull request.



In Makefile:
> @@ -124,6 +121,18 @@ docker-buildx:
 		-f Dockerfile \
 		. ${PUSH_ARG}
 
+# Pull the docker images for acceptance
+docker-pull: docker-build

Can we use a bit different style for anything related to dependencies, like docker-pull-deps?
docker-whatever was used to build/push the controller image until now so I thought it would be nice to make a distinction

Makefile
ToMe25: They also build and push the runner image with this PR, that's fine, right?

ToMe25: I renamed it back to acceptance/pull.

ToMe25:
I have no idea why the CI Test fails, but i think its just because it took too long.
I have this exact failure locally sometimes, but its not consistent, and didn't get more common with this PR.
Unfortunately increasing the timeout just makes it fail on the next thing instead, which is why i didn't do that.

If you have any idea how to work around that, or fix it please let me know.

Also due to the seemingly random nature of this failure i can't debug it, i'm sorry.

ToMe25:
I just realized that the controller image should probably not be called ${NAME}, but instead ${USERNAME}/actions-runner-controller.
I already realized that once yesterday, but then forgot.
I unfortunately don't have the time to implement this today, but i will hopefully get to it tomorrow.
You can decide whether or not that should be a part of this PR, as this isn't something this PR broke, but a separate improvement.

mumoshu:
> I just realized that the controller image should probably not be called ${NAME}, but instead ${USERNAME}/actions-runner-controller.

Why? I thought `NAME` had been used to refer the controller image

ToMe25:
It is currently used for two things that aren't necessarily the same.
One is the repo to which the runner gets added, and the other one is the name of the controller image

ToMe25:
NAME being used for both may actually be the fault of this PR.

I'm sorry, i misremembered what NAME was originally used for.

I thought it was already used for the repo somewhere, but i think that may be my fault.

What should i change the repo name to then?
A new env var that defaults to NAME?

mumoshu:
> One is the repo to which the runner gets added

@ToMe25 Thanks for confirming! I thought it was hard-coded before.

> What should i change the repo name to then?
> A new env var that defaults to NAME?

Maybe you can just introduce another envvar just to represent the target repository that you add runners?
It doesn't need to be defaulted to $NAME, as I don't like polluting the GitHub repository that coincidentally has same name as $NAME. Again I believe $NAME has been used to represent runner container image, not a GitHub repo.

Makefile
```diff
@@ -1,5 +1,11 @@
-NAME ?= summerwind/actions-runner-controller
+ifdef USERNAME
+	NAME ?= ${USERNAME}/actions-runner-controller
+else
+	NAME ?= summerwind/actions-runner-controller
+endif
 VERSION ?= latest
+USERNAME ?= $(shell echo ${NAME} | cut -d / -f1)
```
mumoshu, line 7: Can we just introduce an envvar to represent the docker repo and let the user set it before running tests?

```suggestion
DOCKER_REPO ?= summerwind
NAME ?= ${DOCKER_REPO}/actions-runner-controller
VERSION ?= latest
```

README.md
```diff
@@ -762,12 +762,12 @@ the acceptance test:
 # This sets `VERSION` envvar to some appropriate value
 . hack/make-env.sh
 
-NAME=$DOCKER_USER/actions-runner-controller \
+USERNAME=$DOCKER_USER \
```
mumoshu, line 765: This can just be, without indirection:

```suggestion
DOCKER_REPO=*** \
```

README.md
```diff
@@ -782,7 +782,7 @@ If you prefer to test in a non-kind cluster, you can instead run:
 
 ```shell script
 KUBECONFIG=path/to/kubeconfig \
-NAME=$DOCKER_USER/actions-runner-controller \
+USERNAME=$DOCKER_USER \
```
mumoshu, line 785:
```suggestion
DOCKER_REPO=*** \
```

acceptance/testdata/runnerdeploy.yaml
```diff
@@ -6,7 +6,10 @@ spec:
 #  replicas: 1
   template:
     spec:
-      repository: mumoshu/actions-runner-controller-ci
+      repository: ${NAME}
```
mumoshu, line 9: Why are there envvar-like expressions in this file now? Are they supposed to be expanded by anything? (I imagine if there's any small script that involves e.g. sed to replace these expressions but I couldn't fine one in a quick look

Makefile
ToMe25: This change breaks backwards compat for no reason.
Its not much more code to keep it compatible with the old env var, so i will keep it like it was.
If you want o can rename USERNAME to DOCKER_REPO tho.

acceptance/testdata/runnerdeploy.yaml
ToMe25: This is now called by being piped into a envsubst which expands them and then to kubectl, that way there doesn't have to be an expanded forever changing copy on the disk.

ToMe25:
> It doesn't need to be defaulted to $NAME, as I don't like polluting the GitHub repository that coincidentally has same name as $NAME. Again I believe $NAME has been used to represent runner container image, not a GitHub repo.

While it has represented the container image, for most people the repo to test the runners in will probably be $USERNAME/actions-runner-controller, which can easily be defaulted to.
Especially because acceptance doesn't allow actually testing them anyways, but just tests whether they can be added.
But yes, i can add another env var for that.

Makefile
ToMe25: Afaik a repository on dockerhub is more what NAME is, so i will rename USERNAME to DOCKER_USER instead of DOCKER_REPO, unless i get feedback on this.

ToMe25:
@mumoshu I have now renamed USERNAME to DOCKER_USER, and changed the github test repo to its own var(TEST_REPO).
I am still using DOCKER_USER as the default user for the github repo tho.
I hope this isn't an issue, but if it is a can add another var for the github user and default it to the docker one.
This is because most people will have the same username on both platforms.

Makefile
```diff
@@ -1,5 +1,13 @@
-NAME ?= summerwind/actions-runner-controller
+ifdef DOCKER_USER
+	NAME ?= ${DOCKER_USER}/actions-runner-controller
+else
+	NAME ?= summerwind/actions-runner-controller
+endif
+DOCKER_USER ?= $(shell echo ${NAME} | cut -d / -f1)
+RUNNER_NAME ?= ${DOCKER_USER}/actions-runner
 VERSION ?= latest
+TEST_REPO ?= ${DOCKER_USER}/actions-runner-controller
```
mumoshu, line 9: I think this should usually be another github repository for testing various github actions workflows and jobs, rather than a relatively big repo containing all the actions-runner-controller code.

mumoshu (approved):
LGTM. Thanks a lot for your contribution @ToMe25 ☺️

PR #477, opened 2021-04-23
docs: better enterprise runner documentation

callum-tait-pbx:
Adding the discoveries to the documentation made here https://github.com/summerwind/actions-runner-controller/issues/470


mumoshu (approved):
@callum-tait-pbx Thanks! I'm still reading the original issue and is on the way to installing its content onto my brain but it has already been a good learning 🙏

PR #485, opened 2021-04-26
docs: updating helm install command

callum-tait-pbx:
Fixes https://github.com/summerwind/actions-runner-controller/issues/483 somewhat

To be honest the install instructions probably need re-ordering and rewriting a bit. The structure of the install steps were created pre-helm chart and GitHub Enterprise Server support and so it doesn't really work as well as it did originally.

The install process should probably be split into 2 sections:

- Kustomize
- Helm


mumoshu (approved):
LGTM. Thanks!

PR #486, opened 2021-04-26
chore: bump runner

callum-tait-pbx:
A new runner version is out so time to bump


mumoshu (approved):
LGTM. Thanks for the update ☺️

PR #503, opened 2021-05-02
Update README with more detailed test instructions

mumoshu:
- You can now use `make acceptance/run` to run only a specific acceptance test case
- Add note about Ubuntu 20.04 users / snap-provided docker
- Add instruction to run Ginkgo tests
- Extract acceptance/load from acceptance/kind
- Make `acceptance/pull` not depend on `docker-build`, so that you can do `make docker-build acceptance/load` for faster image reload
- You can now set `SYNC_PERIOD=$DURATION make acceptance/deploy` to modify the sync-period when installing with the helm chart


mumoshu:
@ToMe25 Probably you'll be interested in this 😃

ToMe25:
I'm sure this will be useful, however i have something in another project i need to take care of before working with this again.

PR #524, opened 2021-05-05
edge: Enable scaling from zero with PercentageRunnersBusy

### LINKED ISSUE #522: edge: Make scale-from-zero even more reliable
Extracted from https://github.com/actions-runner-controller/actions-runner-controller/pull/447#issuecomment-830766521

We've added support for scale-to/from-zero scenarios to the controller recently.

The biggest known issue so far is that it doesn't scale-from-zero well with `PercentageRunnersBusy`.

You can definitely combine the webhook-based scaling just for scaling from zero. But the webhook-based autoscaling can't be used in an environment that disallows exposing anything to the internet or to the lb accessible from the internet.

A possible solution would be to enable having two or more `HRA.Spec.Metrics[]` items, so that the controller can choose the largest desired replica numbers from one of `Metrics[]` items.

For example, having both `PercentageRunnersBusy` and `TotalInProgressAndQueuedWorkflowRuns` when you're trying to scale-from-zero would use the number of `TotalInProgressAndQueuedWorkflowRuns`, as it can safely become greater than 1 even if you had only offline runners(#465).

The only downside of doing this would be it's harder to use correctly, especially when you need to scale-to/from-zero organizational runners, and maybe enterprise runners. More concretely, `TotalInProgressAndQueuedWorkflowRuns` forces you to encode all the repository names that uses the organization runners into HRA.Spec.Metrics[].RepositoryNames`.

Another potential solution would be to let `PercentageRunnersBusy` somehow detect pending check-runs that requires the runner from the targeted RunnerDeployment.

We usually correlate check-runs received that the webhook-based autoscaler to  HRA and RunnerDeployment by check-runs' `names`. Doing something similar but without webhook would theoretically work.

mumoshu:
`PercentageRunnersBusy`, in combination with a secondary `TotalInProgressAndQueuedWorkflowRuns` metric, enables scale-from-zero for PercentageRunnersBusy.

Please see the new `Autoscaling to/from 0` section in the updated documentation about how it works.

Resolves #522

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,9 +17,10 @@ ToC:
   - [Organization Runners](#organization-runners)
   - [Enterprise Runners](#enterprise-runners)
   - [Runner Deployments](#runnerdeployments)
-    - [Note on scaling to/from 0](#note-on-scaling-to-from-zero)
-    - [Autoscaling](#autoscaling)
-      - [Faster Autoscaling with GitHub Webhook](#faster-autoscaling-with-github-webhook)
+    - [Note on scaling to/from 0](#note-on-scaling-tofrom-0)
+  - [Autoscaling](#autoscaling)
+    - [Faster Autoscaling with GitHub Webhook](#faster-autoscaling-with-github-webhook)
+    - [Autoscaling to/from 0](#autoscaling-tofrom-0)
   - [Runner with DinD](#runner-with-dind)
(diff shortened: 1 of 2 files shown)


mumoshu:
@callum-tait-pbx I've added some documentation around this new feature. Would you mind reviewing it? ☺️

PR #511, opened 2021-05-03
chart: Allow to disabling kube-rbac-proxy and expose metrics

yfried:
Fixes #454

PR #527, opened 2021-05-05
chart: Add service monitor and remove kube_rbac_proxy leftovers

yfried:
Remove rbac-proxy leftovers from #511 and add serviceMonitor support.

Bump chart minor version


yfried:
@mumoshu Hi, can this be released sooner than the new controller?

callum-tait-pbx:
Pretty sure you can't do that. The CRDs have been updated and so the chart will fall over if someone deploys the new version but against with the old app code.

neillturner:
Thanks for the quick work on this issue. -) 

I've been testing the metrics with the change chart: Allow to disabling kube-rbac-proxy and expose metrics #511 and found a couple of problems that i don't think are fully addressed here. 
I had to change auth_proxy_service.yaml to 
apiVersion: v1
kind: Service
metadata:
  labels:
  name: {{ include "actions-runner-controller.authProxyServiceName" . }}
spec:
  ports:
  {{- if .Values.kube_rbac_proxy.enabled }}
  - name: https
    port: 8443
    targetPort: https
  {{- else }}
  - name: http
    port: 8080
    targetPort: http
  {{- end }}  
  selector:
    {{- include "actions-runner-controller.selectorLabels" . | nindent 4 }}

and change deployment.yaml to 
ports:
        - containerPort: 9443
          name: webhook-server
          protocol: TCP
        {{- if not .Values.kube_rbac_proxy.enabled }}
        - containerPort: 8080
          name: http
          protocol: TCP 
        {{- end }}

just making the above changes will fix Allow to disabling kube-rbac-proxy and expose metrics #511 and happy to put a PR in if that's helpful.
you can see my corrected helm chart at https://neillturner.github.io/helm-charts 
chart: actions-runner-controller
      version: "0.11.5"
see the index at https://neillturner.github.io/helm-charts/index.yaml
this works for just have port 8080 as the metrics port. 

I appreciate you trying to make the metrics port configurable and im not expert at writing helm charts but i think to do that 
values.yaml needs to be: 

metrics:
  serviceMonitor: false
  port: 8080

so it has a default value of 8080 

also in deployment.yaml 

{{- if not .Values.kube_rbac_proxy.enabled }}
        - containerPort: 8443
          name: {{ .Values.metrics.port }}
          protocol: TCP
        {{- end }}
needs to be 
      - containerPort: {{ .Values.metrics.port }}
          name: http
          protocol: TCP 
I don't fully understand the service monitor configuration but assume you understand that better then me. 

happy test out your changes. -) 

Hope that's helpful
Neill Turner

yfried:
@neillturner I think this PR should fix the port issues you've mentioned. It switches the metrics port from `8080` to `8443`.
Your ref says `9443` which is definitely a typo in my previous PR.
The port number isn't configurable. Only the port name so its use is clearer

neillturner:
> @neillturner I think this PR should fix the port issues you've mentioned. It switches the metrics port from `8080` to `8443`.
> Your ref says `9443` which is definitely a typo in my previous PR.
> The port number isn't configurable. Only the port name so its use is clearer
> 
> I've verified that the serviceMonitor works for me using the following values:
> 
> ```yaml
>           kube_rbac_proxy:
>             enabled: false
>           metrics:
>             serviceMonitor: true
> ```

I prefer to have metric point at 8080 as we have network rules to allow that access on kubernetes

yfried:
@neillturner I agree that configuring the metrics port is a nice addition.
Can you please add a PR against my own branch to extend this functionality?

neillturner:
> @neillturner I agree that configuring the metrics port is a nice addition.
> Can you please add a PR against my own branch to extend this functionality?

ok - give me a day or 2 to develop and test. -)

neillturner:
when i tested this with 
metrics:
      serviceMonitor: true 
i get error  no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1

So this doesn't look to me a very reliable change

charts/actions-runner-controller/templates/auth_proxy_role_binding.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.kube_rbac_proxy.enabled }}
```
mumoshu, line 1: Woud you mind renaming the key `kube_rbac_proxy` to `metricsProxy` if you're willing to rename `authProxyServiceName` to `metricsServiceName`?

yfried:
> when i tested this with
> 
> ```
>    kube_rbac_proxy:
>       enabled: false
>     metrics:
>       serviceMonitor: true
> ```
> 
> i get error no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1
> 
> So this doesn't look to me a very reliable change
> seems others have this problem
> https://stackoverflow.com/questions/51095556/servicemonitor-not-found-in-monitoring-coreos-com-v1

Yes, you can only use the serviceMonitor CR if you have the prometheus-operator's CRDs. This is why the `serviceMonitor` flag is disabled by default

yfried:
@mumoshu done. checkout the new values tree

neillturner:
> > when i tested this with
> > ```
> >    kube_rbac_proxy:
> >       enabled: false
> >     metrics:
> >       serviceMonitor: true
> > ```
> > 
> > 
> > i get error no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1
> > So this doesn't look to me a very reliable change
> > seems others have this problem
> > https://stackoverflow.com/questions/51095556/servicemonitor-not-found-in-monitoring-coreos-com-v1
> 
> Yes, you can only use the serviceMonitor CR if you have the prometheus-operator's CRDs. This is why the `serviceMonitor` flag is disabled by default

but i do have prometheus server installed. 
i used 
      helm repo: https://prometheus-community.github.io/helm-charts
      chart: prometheus
      version: "13.2.1"
but i'm not using the full prometheus stack
     chart: kube-prometheus-stack
as i'm forwarding to AWS managed prometheus -)

yfried:
@neillturner serviceMonitor is used by prometheus-operator.
I'm guessing you are looking for a different way. i.e. annotating the service/pod/deployment with some `scrape: true` annotation to tell prometheus to scrape this service. Maybe add this to the chart?

As for controlling the port number - WDYT about [my latest commit](https://github.com/actions-runner-controller/actions-runner-controller/pull/527/commits/35cfa286ff60e5cd87502c74fe3185f54bee377d)? Does it answer your need?

neillturner:
> @neillturner serviceMonitor is used by prometheus-operator.
> I'm guessing you are looking for a different way. i.e. annotating the service/pod/deployment with some `scrape: true` annotation to tell prometheus to scrape this service. Maybe add this to the chart?
> 
> As for controlling the port number - WDYT about [my latest commit](https://github.com/actions-runner-controller/actions-runner-controller/pull/527/commits/35cfa286ff60e5cd87502c74fe3185f54bee377d)? Does it answer your need?

yes that good that metric port is configurable but i still can't use this chart as i'm installing prometheus with

neillturner:
> > @neillturner serviceMonitor is used by prometheus-operator.
> > I'm guessing you are looking for a different way. i.e. annotating the service/pod/deployment with some `scrape: true` annotation to tell prometheus to scrape this service. Maybe add this to the chart?
> > As for controlling the port number - WDYT about [my latest commit](https://github.com/actions-runner-controller/actions-runner-controller/pull/527/commits/35cfa286ff60e5cd87502c74fe3185f54bee377d)? Does it answer your need?
> 
> yes that good that metric port is configurable but i still can't use this chart as i'm installing prometheus with
https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus which doesn't install the custom resouces like 
https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack. which is new name for prometheus operator.

charts/actions-runner-controller/values.yaml
```diff
@@ -63,6 +57,16 @@ service:
   type: ClusterIP
   port: 443
 
+metrics:
+  serviceMonitor: false
+  port: metrics-port
+  portNumber: 8443
```
mumoshu, line 63:
```suggestion
  port: 8443
```

I think we can just use the fixed name and use `port` for the port number instead, so that it is more consistent with our own other port-related settings like `service.port` and common charts like https://github.com/helm/charts/blob/7e45e678e39b88590fe877f159516f85f3fd3f38/stable/fluent-bit/values.yaml#L36

neillturner:
it would be good to change the first line of serviceMonitor.yaml
from 
  {{- if .Values.metrics.serviceMonitor }}
to 
{{- if and ( .Capabilities.APIVersions.Has "monitoring.coreos.com/v1" ) ( .Values.metrics.serviceMonitor ) }}

that way the service monitor won't be created if the crd it relies on is not defined. otherwise the chart breaks if the full prometheus operator is not already installed in k8s. 
hope that helpful. It would enable me to use the updated chart as i don't install the full prometheus stack. -)

yfried:
> it would be good to change the first line of serviceMonitor.yaml
> from
> {{- if .Values.metrics.serviceMonitor }}
> to
> {{- if and ( .Capabilities.APIVersions.Has "monitoring.coreos.com/v1" ) ( .Values.metrics.serviceMonitor ) }}
> 
> that way the service monitor won't be created if the crd it relies on is not defined. otherwise the chart breaks if the full prometheus operator is not already installed in k8s.
> hope that helpful. It would enable me to use the updated chart as i don't install the full prometheus stack. -)

@neillturner the serviceMonitor will not be created unless explicitly asked for (`--set metrics.serviceMonitor=true`).
Do we have to defend it using the `.Capabilities` as well?

charts/actions-runner-controller/values.yaml
yfried: done

neillturner:
> > it would be good to change the first line of serviceMonitor.yaml
> > from
> > {{- if .Values.metrics.serviceMonitor }}
> > to
> > {{- if and ( .Capabilities.APIVersions.Has "monitoring.coreos.com/v1" ) ( .Values.metrics.serviceMonitor ) }}
> > that way the service monitor won't be created if the crd it relies on is not defined. otherwise the chart breaks if the full prometheus operator is not already installed in k8s.
> > hope that helpful. It would enable me to use the updated chart as i don't install the full prometheus stack. -)
> 
> @neillturner the serviceMonitor will not be created unless explicitly asked for (`--set metrics.serviceMonitor=true`).
> Do we have to defend it using the `.Capabilities` as well?

Only issue is  Values.metrics.serviceMonitor needs to be enabled to define metrics-port in githubwebhook.service.yaml. 
Not sure of implications of that. 
Its good programming practice to  defend it using the `.Capabilities` as it stops it breaking -)

neillturner:
> > > it would be good to change the first line of serviceMonitor.yaml
> > > from
> > > {{- if .Values.metrics.serviceMonitor }}
> > > to
> > > {{- if and ( .Capabilities.APIVersions.Has "monitoring.coreos.com/v1" ) ( .Values.metrics.serviceMonitor ) }}
> > > that way the service monitor won't be created if the crd it relies on is not defined. otherwise the chart breaks if the full prometheus operator is not already installed in k8s.
> > > hope that helpful. It would enable me to use the updated chart as i don't install the full prometheus stack. -)
> > 
> > 
> > @neillturner the serviceMonitor will not be created unless explicitly asked for (`--set metrics.serviceMonitor=true`).
> > Do we have to defend it using the `.Capabilities` as well?
> 
> Only issue is Values.metrics.serviceMonitor needs to be enabled to define metrics-port in githubwebhook.service.yaml.
> Not sure of implications of that.
> Its good programming practice to defend it using the `.Capabilities` as it stops it breaking -)

i tested this chart and it works well 
just specified: 
```
    metrics:
      serviceMonitor: false
      port: 8080
      proxy:
        enabled: false
```
and can access metrics (when installed in the actions-runner-system namespace) from 
http://actions-runner-controller-metrics-service.actions-runner-system:8080/metrics

Thanks 
for all your work -)
look forward to merge and deployment -)

yfried:
@mumoshu can we plz merge this?

mumoshu:
@yfried Hey! Could you hold on several days? This triggers a new chart release once merged. And I'd like to pend the next chart release until we release the controller v0.19.0, because there's a lot of changes in the chart that makes sense only with the next controller release.

yfried:
@mumoshu Sure, I'll wait.
Just wanted to make sure this PR is good to go.

toast-gear:
LGTM, I've reverted the Chart version bump so we can get this merged in preparation for a chart release. We are currently working on this and hope to release a new chart soon. The current process causes issues as the solution needs some specific release notes attached with releases.

PR #529, opened 2021-05-06
chart: Allow to disabling kube-rbac-proxy and expose metrics

neillturner:
corrections to PR
chart: Allow to disabling kube-rbac-proxy and expose metrics #511

this chart will allow setting in the values.yaml 
```
        kube_rbac_proxy:
            enabled: false
```
then metrics will be accessible via 
http://actions-runner-controller-mojanalytics-metrics-service.actions-runner-system:8080/metrics
assumming the actions-runnner-controller is deployed at namespace actions-runner-system

this is an alternative to
chart: Add service monitor and remove kube_rbac_proxy leftovers #527
which i could not get to work. 

sorry. just thought i would submit this as an alternative. you folks decide what you want to do. I'm just trying to get a metrics endpoint to work -)

### DIFF
--- a/charts/actions-runner-controller/templates/auth_proxy_service.yaml
+++ b/charts/actions-runner-controller/templates/auth_proxy_service.yaml
@@ -7,8 +7,14 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   ports:
+  {{- if .Values.kube_rbac_proxy.enabled }}
   - name: https
     port: 8443
     targetPort: https
+  {{- else }}
+  - name: http
+    port: 8080
+    targetPort: http
+  {{- end }}
   selector:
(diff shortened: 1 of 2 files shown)


yfried:
Can you open this against my branch instead of master? So we don't step on each other's work

neillturner:
> Can you open this against my branch instead of master? So we don't step on each other's work

but its incompatiable with your change. its an alternative as i couldn't get your change to work on my EKS 1.9 cluster. 

I'm not trying to tell you folks how to change your product just sharing my experiences so you can figure out best way to change it -)

neillturner:
tested chart changes in 
 chart: Add service monitor and remove kube_rbac_proxy leftovers #527
and my pr is no longer necessary 
thanks for all your work

PR #536, opened 2021-05-10
Add pip3 and cmake to default runner image

### LINKED ISSUE #533: acceptance helm test failure
Since a few commits back, i don't know exactly how many, the acceptance helm tests fail.
I tried this with a fresh clone of the master branch, but it fails with the same error.
I do not know what exactly fails, but this seems to be the only hint the log gives:
```
Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed calling webhook "mutate.runnerdeployment.actions.summerwind.dev": Post "https://actions-runner-controller-webhook.actions-runner-system.svc:443/mutate-actions-summerwind-dev-v1alpha1-runnerdeployment?timeout=30s": dial tcp 10.96.135.248:443: connect: connection refused
```
Here is a full `make acceptance` log: https://pastebin.com/TnikkGh7
I do not know for sure whether this is because of a timeout set too low, but most of them caused inconsistencies where it sometimes works and sometimes doesn't, while this always fails.

ToMe25:
Pip is used for alot of things in github runners, and using the python setup action to get it is both needlessly slow, and doesn't work on arm.
This whole PR is more of a suggestion, implemented as a PR, then anything else, so feel free to close this PR if this isn't intended for the default image.
CMake is only added for convenience, and I would understand if it isn't wanted, in that case just let me know and i will remove it.
This PR adds pip by installing pip3, adding pip installed things to the path, and creating a symlink to make pip use pip3.
This pr also fixes the helm acceptance tests.
Fix #533
This PR also sorts the packets to be installed alphabetically, as this seems to be how it has been until recently.
I don't think using ENV for the path is ideal, but it works and is from what i could find pretty much the best solution there is.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -192,7 +192,7 @@ acceptance/teardown:
 	kind delete cluster --name acceptance
 
 acceptance/deploy:
-	NAME=${NAME} DOCKER_USER=${DOCKER_USER} VERSION=${VERSION} RUNNER_NAME=${RUNNER_NAME} RUNNER_TAG=${RUNNER_TAG} TEST_REPO=${TEST_REPO} acceptance/deploy.sh
+	NAME=${NAME} DOCKER_USER=${DOCKER_USER} VERSION=${VERSION} RUNNER_NAME=${RUNNER_NAME} RUNNER_TAG=${RUNNER_TAG} TEST_REPO=${TEST_REPO} SYNC_PERIOD=${SYNC_PERIOD} acceptance/deploy.sh
 
 acceptance/tests:
 	acceptance/checks.sh
--- a/runner/Dockerfile
+++ b/runner/Dockerfile
@@ -40,6 +40,7 @@ RUN apt update -y \
     wget \
     zip \
     zstd \
+    && ln -sf /usr/bin/pip3 /usr/bin/pip \
     && rm -rf /var/lib/apt/lists/*
 
 RUN export ARCH=$(echo ${TARGETPLATFORM} | cut -d / -f2) \
--- a/runner/Dockerfile
+++ b/runner/Dockerfile
@@ -13,8 +13,9 @@ RUN apt update -y \
     && apt update -y \
     && apt install -y --no-install-recommends \
     build-essential \
-    curl \
     ca-certificates \
+    cmake \
+    curl \
     dnsutils \
     ftp \
     git \
@@ -26,6 +27,8 @@ RUN apt update -y \
     netcat \
     openssh-client \
     parallel \
+    python3-pip \
+    python-is-python3 \
     rsync \
     shellcheck \
     sudo \
@@ -37,12 +40,13 @@ RUN apt update -y \
     wget \
     zip \
     zstd \
-    python-is-python3 \
     && rm -rf /var/lib/apt/lists/*
 
 RUN export ARCH=$(echo ${TARGETPLATFORM} | cut -d / -f2) \
   && curl -L -o /usr/local/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v1.2.2/dumb-init_1.2.2_${ARCH} \
   && chmod +x /usr/local/bin/dumb-init
+
+ENV PATH="${HOME}/.local/bin:${PATH}"
 
 # Docker download supports arm64 as aarch64 & amd64 as x86_64
 RUN set -vx; \
--- a/runner/Dockerfile.dindrunner
+++ b/runner/Dockerfile.dindrunner
@@ -37,6 +37,7 @@ RUN apt update -y \
     wget \
     zip \
     zstd \
+    && ln -sf /usr/bin/pip3 /usr/bin/pip \
     && rm -rf /var/lib/apt/list/*
 
 ENV PATH="${HOME}/.local/bin:${PATH}"
--- a/runner/Dockerfile.dindrunner
+++ b/runner/Dockerfile.dindrunner
@@ -8,12 +8,14 @@ RUN apt update -y \
     && apt install -y --no-install-recommends \
     software-properties-common \
     build-essential \
-    curl \
     ca-certificates \
+    cmake \
+    curl \
     dnsutils \
     ftp \
     git \
     iproute2 \
+    iptables \
     iputils-ping \
     jq \
     libunwind8 \
@@ -21,9 +23,12 @@ RUN apt update -y \
     netcat \
     openssh-client \
     parallel \
+    python3-pip
+    python-is-python3 \
     rsync \
     shellcheck \
     sudo \
+    supervisor \
     telnet \
     time \
     tzdata \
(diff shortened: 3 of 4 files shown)


ToMe25:
I hope these tests failing isn't my fault, but i don't think so.
Also this is fully tested.
Please note that adding a '+' to the branch name was stupid, so now the version has to be manually changed after running `hack/make-env.sh`, or the test will fail.

callum-tait-pbx:
The decision rests with the maintainers of which I am not one, however I do have a vested interest in this project being maintained as I have implemented it at work so I have strong views, the maintainers are more than welcome to disagree with me.

> Pip is used for alot of things in github runners, and using the python setup action to get it is both needlessly slow, and doesn't work on arm.

That's all great and everything but the problem with this is that every tool in existance can be said to be used by a lot of things. Every time a tool is added into this project that means that is another tool to bump, another tool to test whenever changes happen, more maintenance, more overhead. There isn't a company behind this project, it's just people willing to use their free time to create and maintain something cool. Whilst it would be great to bundle every tool into the base runner image like GitHub does, this just isn't possible as the runners will be completely unmaintainable if that was done. You could argue that we should only include "core" tools but who is to say what a core tool is? Furthermore 1 new tool does not mean 1 new test case, it means many new test cases.

Given this project is for deploying runners on a k8s cluster it not too much imo to put the onus on the end users for providing appropriate runners for their setup. If someone is able to deploy this project they will have a docker capability where they work, they will have access to registery where they work in some form. Producing a custom runner should not be a big deal and so I don't think indroducing more code to maintain in this project is appropriate nor helpful given all this.

@mumoshu you're the maintainer so please do what you think is best. I'm just a dude with a vested interest in this project staying maintained long term and so I hold strong views for that purpose.

ToMe25:
I can see this for CMake, this is why I explicitly stated if it isn't wanted I'm fine with just removing it.
However, I think for pip which is more of a package manager then a development tool, and which I've seen in a lot non python projects too, not just python, it is worth including.
I do think this should only include core tools, however I also think pip is pretty much a core tool.
If a maintainer say no, i will accept that, but i do think adding pip is worth the extra maintaining effort, which ideally shouldn't be much.

callum-tait-pbx:
> However, I think for pip which is more of a package manager then a development tool, and which I've seen in a lot non python projects too, not just python, it is worth including.

Whilst true, and it isn't the kind of tool that needs bumping every time there is a new release, it can mostly be bumped whenever, the expectation is being set that if there are problems with it it will get supported in the base runner and fixed in the base runner. If pip does get included we now need to test in the project what happens when you use use N number of python versions with the (actions/setup-python)[https://github.com/actions/setup-python] action across 2 OSs (Ubuntu 20.04 and 18.04), that or just accept runners may break. Ideally we also need to keep the version aligned with whatever GitHub put on their runners so be get consistent behaviour between self-hosted and github hosted runners.

Do you use custom runners at all in your environment? If not what is stopping you from producing your own?

@mumoshu if you do want to include it I just think clear lines to be added into the documentation around what the project is willing to include and what they aren't. Someone will always want to include a new tool and eventually the runners will end up with lots of tools of which no one really knows why they are there and the runners will become a bigger headache to manage.

ToMe25:
> Do you use custom runners at all in your environment? If not what is stopping you from producing your own?

I do not use custom runners, because i currently only use self hosted runners for very few things.
Whats stopping me is the need to do so.
It is currently easier to have the workflow install stuff, then to build custom runners with more software pre installed.
I am currently not using many self hosted runners, because i did have not yet finished the main things i want to use them for, and for which I set them up.

What i primarily intend to use self hosted runners for, is to provide arm builds for projects that aren't in a state in which i want to publish them in yet.

What is also stopping me from using custom runners, is that for a bunch of reasons the devices hosting the runners can't build their own containers atm, and i haven't set anything up for them to use local containers from a different device.

mumoshu:
> if you do want to include it I just think clear lines to be added into the documentation around what the project is willing to include and what they aren't. Someone will always want to include a new tool and eventually the runners will end up with lots of tools of which no one really knows why they are there and the runners will become a bigger headache to manage. The project shouldn't be aiming to outsource people producing their own runners imo (again feel free to disagree, I'm just viewing this from an enterprise perspective which is what the project should target imo)

@callum-tait-pbx I basically agree with your perspective.

What has been difficult for me though was clearly stating `what the project is willing to include and what they aren't`. 

My vague idea is that anything that can be installed by running e.g. `curl | bash` or an os package manager like `apt` or `setup-WHATEVER` actions like [setup-python](https://github.com/actions/setup-python) on the default runner image should NOT be included. Does that sound good enough to you?

@ToMe25 Hey! Thanks for your help.

Even if we ended up not including this change in our current default image, I'd still like to explore other possibilities.

What if we had other github repositories under this organization for maintaining bigger images for specific purposes? I imagine repositories like `https://github.com/actions-runner-controller/actions-runner-python`, `https://github.com/actions-runner-controller/actions-runner-aws`, etc.

How's your project built? `Go`, `C`,`Rust`, or else? Where are you trying to deploy from your workflows, Docker Hub, another K8s cluster, some PaaS, or else? I think having a detailed information of your project would help us come up with a better place to start maintaining a bigger image.

ToMe25:
> What if we had other github repositories under this organization for maintaining bigger images for specific purposes?

In general that sounds like a decent idea to me, however they would have to have a significant change against the base runner to be worth the trouble.
I don't think something like the python runner example would be worth that, because python itself is part of the base runner and installing the additional things useful for python dev in the workflow is easy enough.

Also i think this will cause issues with people wanting an image with two of the specializations, and not wanting to build their own runner if both individually exist.

And i think that stuff like pip would be an issue with this, because it is python but also used for meson for example.

In short i like the idea, but doubt its worth the trouble, except for things like maybe an AWS/GCE specific one or something.

> How's your project built?

I currently only have one project using self hosted runners, because i wanted to have them ready for when I release some other projects.
This project is `C++` based, and built using meson.
Its also tiny because i basically used it as a test for a bunch of things.
I use github hosted runners for Windows and Linux x86_64 builds, and self hosted for Linux Arm builds.
I technically also have github hosted MacOS builds, but static linking doesn't work there and i can't verify these builds work at all.

> Where are you trying to deploy from your workflows

Currently only to GitHub, long term maybe also Docker Hub, if i finish something that i want to publish there.
If that project just dies, just GitHub.

@mumoshu Should i make a new PR for the makefile fix in this PR then?

mumoshu:
> @mumoshu Should i make a new PR for the makefile fix in this PR then?

Yep that would be nicer 😄

ToMe25:
@mumoshu I #542 now proposes the helm Fix.
I can't run acceptance rn, but i will test it later.
I think it should be fine tho, because i tested the change for this PR.

neillturner:
i think the best idea is for your runner to have a minimum of software in it and have references to other runners folks have built: 
 we have a runner at 
  https://github.com/ministryofjustice/docker-github-actions-runner
and another good runner is: 
    https://github.com/myoung34/docker-github-actions-runner

toast-gear:
@ToMe25 `pip` is now part of the runner, I noticed a few other bits that needed doing so I raised a new PR myself to do it all together. Thank you for your contribution in fixing the tests as raising this requirement! We haven't bundle `cmake` and aren't willing to include it in the base runner, you will need to install it as part of your pipeline or create a custom runner.

ToMe25:
@toast-gear Thanks for adding pip, I am aware cmake might not fit, thats why I said its an optional part of this PR :)

PR #542, opened 2021-05-11
Fix #533(Acceptance helm failure)

### LINKED ISSUE #533: acceptance helm test failure

ToMe25:
# Changes:
 * Fix the acceptance helm test failing because deploy.sh doesn't have access to the `SYNC_PERIOD` env var.


Makefile
```diff
@@ -195,7 +195,7 @@ acceptance/teardown:
 
 acceptance/deploy:
 	NAME=${NAME} DOCKER_USER=${DOCKER_USER} VERSION=${VERSION} RUNNER_NAME=${RUNNER_NAME} RUNNER_TAG=${RUNNER_TAG} TEST_REPO=${TEST_REPO} \
-	TEST_ORG=${TEST_ORG} TEST_ORG_REPO=${TEST_ORG_REPO} \
+	TEST_ORG=${TEST_ORG} TEST_ORG_REPO=${TEST_ORG_REPO} SYNC_PERIOD={SYNC_PERIOD} \
```
mumoshu, line 198:
```suggestion
	TEST_ORG=${TEST_ORG} TEST_ORG_REPO=${TEST_ORG_REPO} SYNC_PERIOD=${SYNC_PERIOD} \
```

Missing `$`? 😃

ToMe25:
I have tested this since then, and it works :)

mumoshu:
@ToMe25 Thanks for confirming ☺️

PR #573, opened 2021-05-24
feat: adding pip to base image

toast-gear:
Adding pip as requested by https://github.com/actions-runner-controller/actions-runner-controller/pull/536

Binaries are searched in PATH order, as a result including this shouldn't break anything as the `actions/setup-python` action bolts onto the front of the PATH. The Python env configured by the action should always be resolved over the built in python and pip installs due to this, see below for evidence.

I have also removed the `python-is-python3` package as we are now symlinking pip manually. I thought it was more consistent if we were symlinking both the interpreter and pip rather than letting a package do it for interpreter and manually doing it for pip.

image

image

### DIFF
--- a/runner/Dockerfile
+++ b/runner/Dockerfile
@@ -26,6 +26,7 @@ RUN apt update -y \
     netcat \
     openssh-client \
     parallel \
+    python3-pip \
     rsync \
     shellcheck \
     sudo \
@@ -37,7 +38,6 @@ RUN apt update -y \
     wget \
     zip \
     zstd \
-    python3-pip \
     && ln -sf /usr/bin/python3 /usr/bin/python \
     && ln -sf /usr/bin/pip3 /usr/bin/pip \
     && rm -rf /var/lib/apt/lists/*
--- a/runner/Dockerfile
+++ b/runner/Dockerfile
@@ -43,8 +43,8 @@ RUN apt update -y \
     && rm -rf /var/lib/apt/lists/*
 
 RUN export ARCH=$(echo ${TARGETPLATFORM} | cut -d / -f2) \
-  && curl -L -o /usr/local/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v1.2.2/dumb-init_1.2.2_${ARCH} \
-  && chmod +x /usr/local/bin/dumb-init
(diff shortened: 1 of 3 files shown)


mumoshu (approved):
LGTM. Thanks a lot for your efforts! Glad to learn some correct way to configure these python/pip things 🙏

ToMe25:
@toast-gear this PR seems to just not add things installed with pip to the path, is there a reason for that?
#536 did that with `ENV PATH="${HOME}/.local/bin:${PATH}"`, which isn't a great solution but should reliably work afaik.
`ENV PATH="${PATH}:${HOME}/.local/bin"` should also work.

toast-gear:
`pip` doesn't install packages in `~/.local/bin`? Adding `${HOME}/.local/bin` to the PATH wouldn't achieve this?

image

ToMe25:
I don't know about this, I just know I tested it with a few random pip things and they all worked with the PATH change I did above, while they didn't without, so I assumed thats how everything worked?
I also looked some things up back then, and I think I found two paths for pip stuff, not four.
I did find this `~/.local/lib/python3.8/site-packages` path as a relevant one, however after some looking into this I decided it probably wasn't needed, tho I can't remember why.

toast-gear:
1. I think packages get installed in the `User Site Directory` location by default which is python version specific
2. You are also able to install into `User Script Directory` however I'm struggling to actually do this? Have you got an example?

ToMe25:
An example would be meson, I am using this as an example because I use it in the project I basically use as a test for self hosted runners atm.
I don't really use self hosted runners much yet, because the things I wanted it for aren't ready yet, and some not even started.
Anyways when running `pip install meson`, that should install meson, however running meson after that tells me meson isn't on the path.
However with the path set it works on a self hosted runner.

toast-gear:
Did you test on GitHub Actions? It's super interesting that the HOME var seems to be getting set to `/github/home`, we don't actively set it in the Dockerfile

ToMe25:
I'm sorry, I seem to have missed the notification for this.
> Did you test on GitHub Actions?

Yes, I did. Both that, and on a local kind cluster.
For the local kind cluster, I let acceptance set it up, and then ran the pip commands using kubectl.

PR #580, opened 2021-05-27
feat: allow specifying runtime class in runner spec

inahga:
This allows using the `runtimeClassName` directive in the runner's spec.

Our use case for this is Kata Containers, which use `runtimeClassName` in a pod spec as an indicator that the pod should run inside a Kata container. This allows us a greater degree of pod isolation.

I tried my best to follow a similar PR https://github.com/actions-runner-controller/actions-runner-controller/pull/436. I had a bit of trouble running integration tests locally, but I was able to test this on a live cluster. Let me know if anything looks off.


mumoshu (approved):
LGTM. Thanks for your contribution @inahga ☺️

mumoshu:
Waiting for the CI workflow to pass.

inahga:
Glad that error wasn't my fault it was pretty confusing 🤣 

Thanks for reviewing so fast!

toast-gear:
Yes could we get it added to the docs please, additionally there needs to be a description included for the key in the CRD @inahga

inahga:
Sorry for the delay, it was US holiday this weekend. README and CRDs have been updated!

For future pioneers, to update the CRD we need to comment the field in the API definition, like we would for `godoc`. Then run `make manifests`.

mumoshu (approved):
Awesome! Thanks for your contribution @inahga ☺️

Issue #581, opened 2021-05-28
[Helm Chart] Labels for pods


cdobinsky:
**Is your feature request related to a problem? Please describe.**
Currently you can only set labels for the deployment but not for the pods. Since we use the labels to get the logs, we would like to add pod labels as well.

**Describe the solution you'd like**
Having a value for example podLabels where you can set labels for the pods.

mumoshu:
@cdobinsky hey! just set runnerdeployment.spec.template.metadata.labels and that should be inherited down to runnerreplicasets, runners, and pods.

cdobinsky:
Yes, but I use the helm chart and there is no value which sets something at runnerdeployment.spec.template.metadata.labels.

mumoshu:
@cdobinsky Ah gotcha! Thanks for confirming. Would you mind contributing something like `podLabels`, so that it is consistent with `podAnnotations`?

cdobinsky:
Sure, I opened a pr. I wasn't sure about the Chart Version, I hope thats fine.

PR #583, opened 2021-05-28
chart: add podLabels to helm chart

cdobinsky:
This fixes: #581


charts/actions-runner-controller/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.11.0
+version: 0.11.1
```
```suggestion
version: 0.11.0
```

A chart version change triggers a new chart release automatically on merge. That is usually ok but we're currently pending the next chart release until we cut a new controller release, as the unreleased chart contains a lot of changes that only makes sense with the next controller release. Sry for inconvenience!

mumoshu (approved):
LGTM. Thanks a lot for your contribution @cdobinsky ☺️

PR #585, opened 2021-05-28
fix: fixing mising pip PATH

toast-gear:
Conversation is on #573

**Notes**

- We aren't setting the `HOME` var in the Dockerfile. When using `summerwind/actions-runner:latest` on Actions, `HOME` seems to get set as `/github/home` for some reason. `HOME` is a common var used and assumed to be the home directory of the user running the process so we should be setting this in the Dockerfile to ensure consistency. Interestingly when I run up the runner image locally with docker, `HOME` appears to be set correctly so it must be something Actions is doing in the background, I assume if we have it explicitly set they will honour it:

current behaviour:

```yaml
name: On Manual 1 Public Runner

on:
  workflow_dispatch:

jobs:
    my_job:
      runs-on: ['self-hosted', 'Linux']
      container: 
        image: summerwind/actions-runner:latest
      steps:
      - run: |
          echo $HOME
```

image

- Adding the User Script Directory to the end of the PATH https://www.python.org/dev/peps/pep-0370/

after change:

```shell
$ docker run -it runner:test sh
$ echo $HOME
/home/runner
$ echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/home/runner/.local/bin
```

### DIFF
--- a/runner/Dockerfile
+++ b/runner/Dockerfile
@@ -89,9 +89,8 @@ RUN echo AGENT_TOOLSDIRECTORY=/opt/hostedtoolcache > .env \
 COPY entrypoint.sh /
 COPY --chown=runner:docker patched $RUNNER_ASSETS_DIR/patched
 
-# Add the Python "User Site Directory" and "User Script Directory" to the PATH
-# The "User Site Directory" is python version specific so it will need changing with Python bumps
-ENV PATH="${PATH}:${HOME}/.local/bin:${HOME}/.local/lib/python3.8/site-packages"
+# Add the Python "User Script Directory" to the PATH
+ENV PATH="${PATH}:${HOME}/.local/bin"
 
 USER runner
 ENTRYPOINT ["/usr/local/bin/dumb-init", "--"]
--- a/runner/Dockerfile
+++ b/runner/Dockerfile
@@ -62,6 +62,7 @@ RUN set -vx; \
     && echo "%sudo   ALL=(ALL:ALL) NOPASSWD:ALL" > /etc/sudoers
 
 ENV RUNNER_ASSETS_DIR=/runnertmp
+ENV HOME=/home/runner
 
 # Runner download supports amd64 as x64. Externalstmp is needed for making mount points work inside DinD.
 #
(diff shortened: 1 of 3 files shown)


callum-tait-pbx:
Interesting, looks like it won't pay attention to the HOME variable regardless https://github.com/actions/runner/issues/863

Oh well, no harm in leaving it in there I guess

mumoshu (approved):
Ah, so setting `HOME=/home/runner` in our Dockerfile makes it compatible with runner's hard-coded HOME setting of `/home/runner`. LGTM. Thanks a lot for your efforts @callum-tait-pbx ☺️

callum-tait-pbx:
For posterity's sake:

`/github/home` gets created and the `HOME` environment variable set to it if you are using the `container` key in a pipeline, e.g.:

```yaml
on:
  workflow_dispatch:

jobs:
    my_job:
      runs-on: ['self-hosted', 'Linux']
      container: 
        image: summerwind/actions-runner:latest
      steps:
      - run: |
          echo $HOME
```

`/github/home` is a volume that gets mounted so you can share a workspace between jobs. As a result this is a scenario we would never run into as you would never run the actions-runner container like this beyond wanting to test something quickly.

dnshio:
Hi, this is happening when I want to run `pytest` inside my own container. The fact that `HOME` gets set to `/home/github` is causing python to not find packages that exist in my user's home dir (`/home/ubuntu/.local`). 

This is the step: 

```
  test:
    name: Run tests (pact)
    runs-on: "non-prod-amd64"
    needs: build_and_push
    timeout-minutes: 10
    container: url-to-ecr:${{ github.sha }}
    defaults:
      run:
        working-directory: /home/ubuntu/app
    steps:
      - name: Pact contracts
        run: |
          pytest /home/ubuntu/app/tests/contract --asyncio-mode=auto -vvvx
```

I get the following error: 

```
Traceback (most recent call last):
  File "/home/ubuntu/.local/bin/pytest", line 5, in <module>
    from pytest import console_main
ModuleNotFoundError: No module named 'pytest'
```

This does not happen if I run `HOME=/home/ubuntu pytest /home/ubuntu/app/tests/contract --asyncio-mode=auto -vvvx` 

How do I make it so that the GH runner can execute my tests inside my own container?

mumoshu:
@dnshio Thanks for reporting! It does sound problematic. Perhaps actions/runnner changed the behavior to override the HOME present in the runner image since then? We need to investigate this further.
In the meantime, would you mind submitting a new bug report with the information, so that it is more trackable?

PR #586, opened 2021-05-28
ci: only run latest tag job on push / release

toast-gear:
We don't need to run the latest tag job on pull request activity? I don't think we even care about release activity do we? I've kept it as we had that logic in there originally but I think that can be removed too so the latest tag job only runs on pushes? (pull request merges create a push event) @mumoshu

### DIFF
--- a/.github/workflows/build-and-release-runners.yml
+++ b/.github/workflows/build-and-release-runners.yml
@@ -76,7 +76,7 @@ jobs:
             ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:v${{ env.RUNNER_VERSION }}-ubuntu-${{ matrix.os-version }}-${{ steps.vars.outputs.sha_short }}
 
   latest-tags:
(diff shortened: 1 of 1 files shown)


mumoshu:
> I think that can be removed too so the latest tag job only runs on pushes?

@toast-gear Hey! I believe we do rely on builds running on every pull request sync so that the contributor can see if docker-builds work, before merging.

A real-world example would be https://github.com/actions-runner-controller/actions-runner-controller/pull/585

CleanShot 2021-05-29 at 10 03 27@2x

mumoshu:
Talked with @callum-tait-pbx on slack and it turned our `Build actions-runner-(dind-)ubuntu-20.04` and Build actions-runner-ubuntu-18.04` jobs already do the job and the latest jobs are just redundant and therefore can be safely removed.

mumoshu (approved):
LGTM. Thanks a lot for the clean-up! 🙏

PR #587, opened 2021-05-30
Add support for extraEnv to set arbit. vars

ekarlso:


mumoshu (changes requested):
@ekarlso Hey! Thanks for the PR. But could you please clarify why you would need this, or why everyone wants to use this? Without that, it's so hard for us to decide either if this should be in this project, or in your own fork.

Issue #591, opened 2021-06-01
Runners and Istio


js-timbirkett:
Hi 👋 

We deploy most of our services onto an Istio service mesh. We'd like to put the runners on "the mesh" so that they can take advantage of `MUTUAL_TLS` and reach services without needing to expose the service through ingressgateway for integration and e2e tests.

The trouble is that I don't think the Istio sidecar comes up before the runner tries to execute `entrypoint.sh` this causes the good old:
```
│ runner Starting Runner listener with startup type: service                                                                                                                                                          │
│ runner Started listener process                                                                                                                                                                                     │
│ runner An error occurred: Not configured                                                                                                                                                                            │
│ runner Runner listener exited with error code 2                                                                                                                                                                     │
│ runner Runner listener exit with retryable error, re-launch runner in 5 seconds.
```

loop of death.

Shelling into the pod and running `./entrypoint.sh` registers the runner as you'd expect at startup and running the runner with Istio sidecar injection also works. 

There are a couple of things that could be done in the (top of the) entrypoint script:
- Allow an env variable configured sleep
- Add a wait-for-ready style loop

Is anyone else running this on Istio?

mumoshu:
@js-timbirkett Hey! Thanks for bringing this up. Honestly speaking I have no much experience with Istio and things like this is why I'm still taking me away from using any kind of sidecar-based service mesh :) I was looking forward to https://github.com/kubernetes/enhancements/issues/753 but it postponed...

I think I'm fine with the first option, sleeping. But I'm interested in the second option, too.

How would the second work? How do you usually `wait-for-ready` when using Istio? Any idea integrating it in a generic way into actions-runner-controller?

js-timbirkett:
Hey @mumoshu - thanks for the quick reply. Back in the days of `docker-compose` you'd often end up having dependencies like databases that you'd want to be available before starting your application.

What you'd end up doing is adding something like:
- https://github.com/ufoscout/docker-compose-wait
- https://github.com/nickjj/wait-until
- https://github.com/vishnubob/wait-for-it

These scripts pretty much run a while loop and attempt to connect to a host and port with a smol sleep.

Istio takes a few hundred milliseconds before network traffic can flow through the sidecar proxy which is fine for java apps (long startup times), or apps that crash and restart when failing to connect to a thing.

mabushey:
I'm using the runner with Istio  1.10.0 just fine. Add this to your istio operator config:
```
  meshConfig:
[...]

Issue #470, opened 2021-04-21
enterprise autoscaling issues, indefinitely queued jobs within workflows


kathleenfrench:
hey - first off, thanks for all of your work on this 😄  - it's a great project!

apologies in advance for the lengthy note -- just hoping to provide as much helpful info as possible upfront. nevertheless, a (poor) attempt at the cliffsnotes version:

> when a workflow with more jobs than currently-running runner pods is kicked off, _sometimes_ the only jobs that **ever** run - even if a scale up is later triggered (though that scale-up is never triggered within the `scaleUp` period) - are those that are initially picked up. _however_, sometimes the remaining queued jobs _do_ run, but only after the first jobs complete. there are no errors in the controller logs in either instance.

**scenario**
> 1 workflow with 4 jobs
> 10s `syncPeriod`

`minReplicas`: 2
`maxReplicas`: 20
`scaleUpThreshold`: 0.5
`scaleUpFactor`: 3.0

**expected behavior**
- 2 jobs immediately picked up by the 2 listening pods
- trigger a scale up after the 10s `syncPeriod` completes as > `0.5` of the running pods are now 'busy' 
- this will result in `6` runner pods
- the remaining 2 jobs in the workflow would get picked up by the newly created runner pods and run alongside the 2 already running

**actual behavior**
- 2 jobs immediately picked up by the 2 listening pods
- no scale up after 10s `syncPeriod` - controller logs show:
```
2021-04-21T18:12:14.019Z        DEBUG   actions-runner-controller.horizontalrunnerautoscaler    Calculated desired replicas of 2      {"horizontalrunnerautoscaler": "github-actions/ghe-runner-deployment-autoscaler", "suggested": 1, "reserved": 0, "min": 2, "cached": 1, "max": 20}
```
- once a job completes its pod terminates, a new pod is created, and a queued job runs. at this point, no scale up has been triggered yet
- scale up to 6 pods seems to coincide subsequent to the 3rd job starting on a newly created pod (when the total number of pods is still 2) **but** none of the remaining queued jobs seem to get picked up by these pods created as part of that scale up, but instead wait for _new_ pods using the same deployment from the initial jobs that did run

<details>
<summary>corresponding scale up logs (click me)</summary>

```
[...]

PR #592, opened 2021-06-01
feat: add STARTUP_DELAY to entrypoint.sh

js-timbirkett:
As a starter for #591 

This is the absolute simplest way that I can think of to allow Istio to get itself sorted out before registering the runner with Github.

Any discussion or ideas would be appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -957,6 +957,12 @@ This is because the `istio-proxy` has not completed configuring itself when the
 configuration script tries to communicate with the network.
 
 **Solution**<br />
+
+> This feature is experimental and will be dropped once maintainers think that
+> everyone has already migrated to ues Istio's `holdApplicationUntilProxyStarts` ([istio/istio#11130](https://github.com/istio/istio/issues/11130)).
+>
+> Please read the discussion in #592 for more information.
+
 You can add a delay to the entrypoint script by setting the `STARTUP_DELAY` environment
 variable. This will cause the script to sleep `STARTUP_DELAY` seconds.
 
--- a/README.md
+++ b/README.md
@@ -938,6 +938,55 @@ Your base64'ed PAT token has a new line at the end, it needs to be created witho
 * `echo -n $TOKEN | base64`
 * Create the secret as described in the docs using the shell and documeneted flags
 
+#### Runner coming up before network available
+
+If you're running your action runners on a service mesh like Istio, you might
+have problems with runner configuration accompanied by logs like:
+
+```
+....
+runner Starting Runner listener with startup type: service
+runner Started listener process
+runner An error occurred: Not configured
+runner Runner listener exited with error code 2
+runner Runner listener exit with retryable error, re-launch runner in 5 seconds.
+....
+```
+
+This is because the `istio-proxy` has not completed configuring itself when the
+configuration script tries to communicate with the network.
+
+**Solution**<br />
+You can add a delay to the entrypoint script by setting the `STARTUP_DELAY` environment
+variable. This will cause the script to sleep `STARTUP_DELAY` seconds.
+
+*Example `Runner` with a 2 second startup delay:*
+```yaml
+apiVersion: actions.summerwind.dev/v1alpha1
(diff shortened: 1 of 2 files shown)


toast-gear:
I don't mind introducing an arbitrary wait option, it could be useful in various scenarios. Should this be done via an API change though? It would be overkill for this specific use case however done all runner config via the API so far. Whilst it would be simpler to just check for the arbitrary environment variable and leave the API out of it, it would be a marked change to how we have done everything so far. I'm a fan of convention over configuration, @mumoshu what are you thoughts on if this should be an API change?

js-timbirkett:
Thanks @toast-gear  - could you elaborate on "API Change"? do you mean by adding some sort of `startupDelay` to the Runner and RunnerDeployment specs?

js-timbirkett:
In terms of Istio... there's a long-running thread: https://github.com/istio/istio/issues/11130 - apparently they've fixed, or attempted to fix it in version 1.7 with a: `holdApplicationUntilProxyStarts` configuration option which injects the sidecar as the first container in the pod (and probably breaks anything expecting `containers[0]` to be the primary application container.

At the moment, I'm stuck on Istio 1.5 until all of our product teams redeploy all of their applications so this fairly simple change might be "just enough" and may become redundant in the future in my case.

callum-tait-pbx:
> Thanks @toast-gear - could you elaborate on "API Change"? do you mean by adding some sort of startupDelay to the Runner and RunnerDeployment specs?

Yup

>  configuration option which injects the sidecar as the first container in the pod (and probably breaks anything expecting containers[0] to be the primary application container.

This would break many things in this project O_O

js-timbirkett:
Thanks @callum-tait-pbx - I can also confirm that I'm running with this change in a custom built image. I'm also running scale to 0 on an `actions-runner-controller` built from `master` - because I enjoy the pain... I'll feedback any bugs or observations. This is a pretty good star.

callum-tait-pbx:
If you experimenting with the scale from 0 (🙇  for that!) you should be aware that this should improve somewhat soon as GitHub are making some backend changes so we don't need to create a fake registration only runner https://github.com/actions-runner-controller/actions-runner-controller/issues/470

mumoshu:
> Should this be done via an API change though?

@callum-tait-pbx That's a good point and I took some time to think about that.

Well, probably we'd better mark it as a temporary feature that may change or go away?

As @js-timbirkett has kindly shared and summarized, recent versions of Istio seem to have the startup-delay container injector thing. So this feature is necessary only util everyone migrates to newer versions of Istio.

An API change persists longer and we can't easily remove already added field from e.g. Runner spec. So I'd like not to make API change for an experimental feature.

Thoughts?

> > configuration option which injects the sidecar as the first container in the pod (and probably breaks anything expecting containers[0] to be the primary application container.
>
> This would break many things in this project O_O

I thought so too at first. But apparently, it doesn't break at all?

We refer to it as "the first container" only when we compose the runner pod spec. The mutating webhook of Istio comes in after that. And our controller looks for the container not by index but by name after the pod creation:

https://github.com/actions-runner-controller/actions-runner-controller/blob/cb60c1ec3baed8690d4546dacfdfdc564891e852/controllers/runner_controller.go#L260-L262

@js-timbirkett It would be awesome if you could try newer Istio with the holdApplicationUntilProxyStarts thing later! In theory. actions-runner-controller should just work fine with it.

js-timbirkett:
Rebased on to master and pushed. See if that solves the broken workflows or not 🤔

mumoshu:
I think any of the maintainers must approve it to run for the first contributor. I just approved so it will be run shortly

mumoshu:
Added some note about this being experimental. The build seems to be failing due to that some shell variables not available to the workflow. Perhaps GitHub changed something on their end that we need to tweak the config somehow to fix? 🤔

mumoshu (approved):
Thanks a lot for your contribution @js-timbirkett ☺️

Issue #262, opened 2021-01-19
Pod annotations support


jlsan92:
Hi!, this is more a question than a "bug report". 

I see that there is support for "EKS IAM role for service accounts" which is pretty nice but how do we enable IAM authentication using other solutions such as https://github.com/jtblin/kube2iam? This one asks for setting pod annotations to match IAM role permissions. It is possible to set up Pod annotations for the Runner's resources?

mumoshu:
@jlsan92 Hey! Apparently it's mostly impossible today. You're using RunnerDeployment, not Runner, right?

Runner annotations are already [inherited to runner pods](https://github.com/summerwind/actions-runner-controller/blob/ace95d72aba5156b6f402f54743466b3dd47d7a7/controllers/runner_controller.go#L407), but there seems no way to set Runner annotations via RunnerDeployment.

https://github.com/summerwind/actions-runner-controller/blob/a63860029a85ac35c2839c19d3bcf3b5294377e5/controllers/runnerdeployment_controller.go#L272

sp-ludovic-ivain:
Got the same problem, would love to have support for annotations in RunnerDeployment 🙏

mumoshu:
@jlsan92 @sp-ludovic-ivain Well I was pretty confused- I believe you can just use `RunnerDeployment.Spec.Template.Metadata.Annotations` for setting custom runner pod annotations, as I've commented in another issue https://github.com/summerwind/actions-runner-controller/issues/299#issuecomment-775566247

Would you mind giving it a shot?

sp-ludovic-ivain:
Hi @mumoshu, yes, I was doing it wrong, putting the annotations into the RunnerDeployment's metadata...

Setting it as you said into ` RunnerDeployment.Spec.Template.Metadata.Annotations` is working fine

Thanks!

PR #600, opened 2021-06-03
docs: add an annotation example in Additional Tweaks

pdemagny:
Hi,

First of all, thanks a lot for actions-runner-controller, it is simple to get started, quite efficient, and fits perfectly for my use case !

Here is an humble contribution that would help for #262 and hopefully avoid a few people some head-scratching ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -722,10 +722,12 @@ metadata:
 spec:
   replicas: 2
   template:
+    metadata:
+      annotations:
+        cluster-autoscaler.kubernetes.io/safe-to-evict: "true"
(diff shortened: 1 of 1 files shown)


mumoshu (commented):
LGTM. Thanks a lot for your contribution @pdemagny!

// Just curious, but were you able to successfully let cluster-scaler bring down the node running runners by using this annotation?
// Just asking because I'm gathering more samples of how everyone's using various configurations 😉

mumoshu (approved):

Issue #614, opened 2021-06-08
kube webhook issue: webhooks[0].clientConfig.caBundle should not have a default empty value because it breaks ArgoCD sync


jstewart612:

jstewart612:
![Untitled](https://user-images.githubusercontent.com/32647598/121218530-d1ef6c80-c850-11eb-8140-772a065e536f.png)

mumoshu:
@jstewart612 How can we fix that? I don't know much about ArgoCD

jstewart612:
@mumoshu I have put up a PR for a suggested fix in https://github.com/actions-runner-controller/actions-runner-controller/pull/617

peimanja:
having the same issue. appreciate the fix

PR #617, opened 2021-06-09
Don't set default caBundle for webhooks

jstewart612:
Fixes https://github.com/actions-runner-controller/actions-runner-controller/issues/614

### DIFF
--- a/charts/actions-runner-controller/Chart.yaml
+++ b/charts/actions-runner-controller/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.12.0
+version: 0.12.1
 
 # Used as the default manager tag value when no tag property is provided in the values.yaml
 appVersion: 0.19.0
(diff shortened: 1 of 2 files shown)


jstewart612:
@mumoshu Helm chart syncing functionalities do three-way merges of generated YAML.  When you specify a value for caBundle, it clobbers the value that the controller puts in there after it's setup.  This makes CD systems like ArgoCD and Flux unable to keep the Helm chart in sync because it enters a constant update loop for this value between rendered chart and the on-cluster value which the controller is setting.  This breaks the loop.

mumoshu:
@jstewart612 Thanks for the PR! This looks good, except Helm 2. I know Helm 2 is already deprecated and perhaps not maintained anymore, but I want to prepare before being asked. That said..

> Helm chart syncing functionalities do three-way merges of generated YAML

What would happen when you run `helm upgrade` on the updated chart with helm 2, which doesn't support three-way merge? Override the caBundle field with an empty value?

jstewart612:
Definitely not maintained anymore: https://helm.sh/blog/helm-v2-deprecation-timeline/

I don't know what happens, but feel it more than sufficient to no longer support that which is deprecated.

mumoshu (approved):
LGTM. Thanks for the fix @jstewart612 ☺️

PR #625, opened 2021-06-11
Instruct ServiceMonitor to connect using https for controller

jstewart612:
The controller metrics endpoint serves over https using a self-signed cert by default in this chart so correct the ServiceMonitor to reflect.


jstewart612:
Okay it only does this if metrics.proxy.enabled is set to true and if so then you need this and mTLS handling.  This needs more work.

mumoshu:
@jstewart612 Hey! Thanks. How can we move forward? Should I just enclose the below part within a `{{ if }} {{ end }}` block that is enabled only when you enabled metrics proxy?

```
      scheme: https
      tlsConfig:
        insecureSkipVerify: true
```

jstewart612:
@mumoshu yes but an additional problem is that when https is enabled, it appears to function using mTLS so I hadn't yet researched the schema of ServiceMonitor yet to find the mTLS options to specify.

mumoshu:
@jstewart612 Hey! Thanks, that makes sense.

I've googled about the matter a bit and found a potential solution at https://github.com/operator-framework/operator-sdk/issues/4764#issuecomment-820344706.

Unfortunately, I have other works I have to prioritize higher now. I would appreciate it very much if you could test it and include the changes into this PR if it worked 🙏

charts/actions-runner-controller/templates/controller.metrics.serviceMonitor.yaml
```diff
@@ -9,6 +9,9 @@ spec:
   endpoints:
     - path: /metrics
       port: metrics-port
+      scheme: https
```
mumoshu, line 12: @jstewart612 Hey. Awesome @AlexDHoffer has figured out that we can add `bearerTokenFile` to the endpoint to make it work!

```suggestion
      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
      scheme: https
```

Would you mind confirming it and wrap this change into a `if end` block to make it configurable? Once done I'll quickly merge this

jstewart612:
@mumoshu added.  Should be good to go now.  Works for me as well as the user that reported that.

charts/actions-runner-controller/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.12.2
+version: 0.12.3
```
```suggestion
version: 0.12.4
```

mumoshu (approved):
LGTM. Thanks a lot for your contribution @jstewart612 (and @AlexDHoffer for support) ☺️

PR #641, opened 2021-06-21
We are exiting if it's a registration-only runner

shubhindia:


mumoshu (approved):
LGTM. Thanks for the fix!

PR #645, opened 2021-06-22
Clean up import list

### LINKED ISSUE #644: Clean up import list
Hi,

Not a bug per say, and sorry if this was mentioned before, I've noticed that some imports are pointing to `summerwind` rather than `actions-runner-controller`, as I understand it, it's some left overs from the rename.

A few examples:
https://github.com/actions-runner-controller/actions-runner-controller/blob/master/cmd/githubwebhookserver/main.go#L28-L29

https://github.com/actions-runner-controller/actions-runner-controller/blob/master/main.go#L27-L29

and basically everywhere. I presume it still works because of the github redirection after the project was moved,
Just thought I'd file an issue to mention this, and if that makes sense I could file a quick PR to search/replace `summerwind` into `actions-runner-controller` as a first contribution to this great project?

mumoshu:
Resolves #644


charts/actions-runner-controller/README.md
```diff
@@ -21,7 +21,7 @@ _Default values are the defaults set in the charts values.yaml, some properties
...
 | `authSecret.github_app_private_key`                      | The multiline string of your GitHub App's private key. **This can't be set at the same time as `authSecret.github_token`** |                                                                      |
 | `authSecret.github_token`                                | Your chosen GitHub PAT token. **This can't be set at the same time as the `authSecret.github_app_*`**                      |                                                                      |
-| `image.repository`                                       | The "repository/image" of the controller container                                                                         | summerwind/actions-runner-controller                                 |
+| `image.repository`                                       | The "repository/image" of the controller container                                                                         | actions-runner-controller/actions-runner-controller                                 |
```
sledigabel, line 24: I do not think that one needs changing? This is the container image that will run the controller right?

charts/actions-runner-controller/values.yaml
```diff
@@ -26,7 +26,7 @@ authSecret:
   #github_token: ""
 
 image:
-  repository: summerwind/actions-runner-controller
+  repository: actions-runner-controller/actions-runner-controller
```
sledigabel, line 29: Same as above, not sure that one needs changing?

mumoshu: @sledigabel Ah, sorry I missed your comments! Thanks- I'm fixing the two issues in #663

Issue #164, opened 2020-11-10
Docs: Developer Docs?


callum-tait-pbx:
We are considering adopting this within my organisation, I would love to take a more active part in imporving the solution. Would it be possible to create some developer docs for instructions on setting up a dev env etc?

mumoshu:
@callum-tait-pbx Hey! Thanks for considering that. What a coincidence I was going to add some instruction on how to run acceptance tests #168

What else would you like me to write? A quick list of questions and unclear steps you'd like me to answer would help.

callum-tait-pbx:
https://github.com/summerwind/actions-runner-controller/pull/168 is an excellent addition. I am mostly unsure about getting setup to dev really. I've only ever done local dev really and so I'm unsure how to start really. I might need a bit more guidance than you can reasonably expect I guess but a seperate readme that's a bit more detailed for getting setup to dev might be helpful?

mumoshu:
@callum-tait-pbx Hey! I've enhanced our `contributing` documentation recently. It now looks like https://github.com/actions-runner-controller/actions-runner-controller#contributing.

What else would you think nice to have? Maybe a step to install development dependencies, like Go, VS Code, `make`, etc?

Also, I usually use macOS and Ubuntu for my development machines. What do you use?

PR #648, opened 2021-06-22
docs: contributing docs

toast-gear:
Resolves https://github.com/actions-runner-controller/actions-runner-controller/issues/164

Mostly a lift and shift from the root `README.md` to the `CONTRIBUTING.md` as it's better placed there as it is not relevant to most users. Ideally we should move the stuff around what is installed on the runner into `CONTRIBUTING.md` too but that can be done separately


mumoshu (approved):
LGTM. Thanks for your effort @toast-gear ☺️

PR #649, opened 2021-06-22
Add GitHub Enterprise URL to chart

cbui:
I needed to be able to provide a GitHub Enterprise URL in the chart.


mumoshu (approved):
LGTM. Thanks a lot for your contribution @cbui ☺️

mumoshu:
@callum-tait-pbx Do you think we can cut a new patch release for that after we merge this?

It will include the [runnerset CRD](https://github.com/actions-runner-controller/actions-runner-controller/blob/master/charts/actions-runner-controller/crds/actions.summerwind.dev_runnersets.yaml) which won't be used because we don't have a controller release for the RunnerSet support yet. Probably there's no issue having an unused CRD though?

toast-gear:
Yeh I don't see that being a problem, 2 changes first though if you don't mind:

1. Fix the minor whitespace difference with the right column `|`
2. Could we amend the description to refer to this as the `GitHub Enterprise Server` URL please. It can potentially be a bit confusing without specifying `Server` because GitHub Enterprise Cloud is a thing. We've had tickets before with people mixing up the 2 products.

Once those 2 things are resolved, I'm happy for a new release to go ahead.

mumoshu:
@toast-gear The latter sounds good! Not sure what you mean by the former point though. To me it seems to have correct indentation?

CleanShot 2021-06-26 at 20 57 20@2x

toast-gear:
https://raw.githubusercontent.com/actions-runner-controller/actions-runner-controller/fcf13251cf1e661a2753d3706103d74ef0fc5891/charts/actions-runner-controller/README.md yeh it looked OK in the diff but when I viewed it raw it was misaligned, I haven't checked out the branch locally but I assume it will reflect what you can see if you view the raw file.

mumoshu:
@toast-gear Thanks!

Gotcha- I do see several lines with misaligned vertical bars in my editor using a monospaced font and a lot more in the raw github page.

Perhaps the latter seems to be due to our browsers aren't using a monospaced font for the pages with text/plain content types?

I believe we'd better have another github issue for fixing it to look ok on our editor.

mumoshu:
@cbui Thanks for the follow-up commits!

Issue #659, opened 2021-06-26
Update runners to use `RUNNER_TOOL_CACHE` instead of `AGENT_TOOLSDIRECTORY`


toast-gear:
**Is your feature request related to a problem? Please describe.**
N/A

**Describe the solution you'd like**
https://github.com/actions/runner/blob/main/src/Runner.Common/HostContext.cs#L245 all other environment variables are deprecated according to a comment in the code:

```
// TODO: Coallesce to just check RUNNER_TOOL_CACHE when images stabilize
path = Environment.GetEnvironmentVariable("RUNNER_TOOL_CACHE") ?? Environment.GetEnvironmentVariable("RUNNER_TOOLSDIRECTORY") ?? Environment.GetEnvironmentVariable("AGENT_TOOLSDIRECTORY") ?? Environment.GetEnvironmentVariable(Constants.Variables.Agent.ToolsDirectory);

```

**Describe alternatives you've considered**
N/A

**Additional context**
https://github.com/actions-runner-controller/actions-runner-controller/issues/655

toast-gear:
https://github.com/actions/runner/pull/232 looks like they reverted referencing `RUNNER_TOOL_CACHE` exclusively which makes me concerned about this issue

PR #660, opened 2021-06-26
chore: updating to use non-deprecated env var

toast-gear:
Fixes: https://github.com/actions-runner-controller/actions-runner-controller/issues/659


mumoshu (approved):
LGTM. Thanks for the update!

Issue #664, opened 2021-06-28
GithubWebhookServer Service Annotations Field


rlandesman:
**Is your feature request related to a problem? Please describe.**
For my use case of the helm chart I would like to add annotations to the githubWebhookServer Service, i.e here: https://github.com/actions-runner-controller/actions-runner-controller/blob/master/charts/actions-runner-controller/templates/githubwebhook.service.yaml#L4-L8

**Describe the solution you'd like**
Implementation of the annotation field

toast-gear:
The raised PR has been merged which included a chart bump, closing.

PR #665, opened 2021-06-28
Add annotations to githubWebhookServer Service in Helm Chart

rlandesman:
This PR aims to resolve the issue I opened here: https://github.com/actions-runner-controller/actions-runner-controller/issues/664 by adding annotations to the server's service. Beyond general applications I use these annotations within my own projects to configure various LB values.


charts/actions-runner-controller/templates/githubwebhook.service.yaml
```diff
@@ -6,6 +6,10 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "actions-runner-controller.labels" . | nindent 4 }}
+{{- if .Values.githubWebhookServer.service.annotations }}
+  annotations:
+{{ toYaml .Values.githubWebhookServer.service.annotations | nindent 4 }}
```
mumoshu, line 11: Hey! Thanks for the PR. I thought this had to have 4 spaces if you use `nindent 4`?

```suggestion
    {{ toYaml .Values.githubWebhookServer.service.annotations | nindent 4 }}
```

Otherwise, `annotations: {"a":"a","b":"b"}` might have wrong indentation like:

```
  annotations:
a: a
    b: b
```

rlandesman: Hey @mumoshu, thanks for replying so quickly (and I appreciate the support on a seemingly small issue).

I ran some tests and it looks like the templating of go removes the indentations before adding the `nindent` regardless of the original indentation level. I went searching a bit and it looks like other oss projects do the same (i.e k8s dashboard here: https://github.com/kubernetes/dashboard/blob/master/aio/deploy/helm-chart/kubernetes-dashboard/templates/deployment.yaml#L19-L22). Let me know what you think, I'd be happy to make the edits.

sledigabel: Generally you'd start at the beginning of the line with `indent`, `nindent` allows you add the indentation after a new line (so the current indent is _reset_ so to speak) and it looks a lot more yaml-y to the trained eye. I *think* the empty line would get parsed as valid yaml and fixed and helm templating time.

sledigabel:
This is super useful for EKS and automatically generating ELBs 👍

sledigabel:
@rlandesman you might want to bump the chart version too as part of this PR to release a new one?

sledigabel:
@rlandesman I just saw that the chart was bumped to v0.12.4 and merged 5h ago by #625.
I'll update #666 to bump to v0.12.5 and you can bump to v0.12.6, does that sound good? Alternatively, @mumoshu can advice on the best course of action here.

rlandesman:
Thanks for the help here everyone!

First off, I bumped the chart to `v0.12.6`, hope that works.

Second, I read through more documentation on `nindent` and played around with some examples. I also tested this code in this PR with dummy data `{a: a, b: b}` and it correctly templates out with `helm template`.

charts/actions-runner-controller/templates/githubwebhook.service.yaml
mumoshu: @rlandesman @sledigabel Wow thanks a lot! I've been misunderstanding how it works. 

https://github.com/kubernetes/dashboard/blob/master/aio/deploy/helm-chart/kubernetes-dashboard/templates/deployment.yaml#L19-L22 uses `indent 4` instead of `nindent 4` which is used here, but I learned that both works.

I'd still appreciate it if you could have 4 spaces like:

```
  annotations:
    {{ toYaml .Values.githubWebhookServer.service.annotations | nindent 4 }}
```

so that it is consistent with other parts of the chart.

I once thought it might be even better to write it like:

```
  annotations: {{ toYaml .Values.githubWebhookServer.service.annotations | nindent 4 }}
```

But this turns out not to compose well when you want to add more fields to be considered:

```
  annotations: {{ toYaml .Values.githubWebhookServer.service.annotations | nindent 4 }}
    {{ toYaml $otherCommonAnnotations | nindent 4 }}
```

rlandesman: Sure thing, changed it 👍🏼

mumoshu (approved):
LGTM. Thanks a lot for your contribution @rlandesman (and reviewing @sledigabel) ☺️

PR #669, opened 2021-06-29
chore: update helm create secret defaults to false

toast-gear:
There's no reason to create a non-working secret by default. If someone wants to deploy the secrets via the chart they will need to do some config regardless so they might as well also set the create flag

Will need a release but this can be release along with another chart change, no need for a unique release just for this change


mumoshu (approved):
LGTM. Thanks for the fix!

PR #670, opened 2021-06-29
Add additional labels to serviceMonitor

kirillbilchenko:
Add a way to add additional labels for service monitor. Could be helpful in case if you are using unified labels to scrape the metrics in k8s


charts/actions-runner-controller/templates/controller.metrics.serviceMonitor.yaml
```diff
@@ -4,6 +4,9 @@ kind: ServiceMonitor
 metadata:
   labels:
     {{- include "actions-runner-controller.labels" . | nindent 4 }}
+  {{- with .Values.metrics.serviceMonitorLabels }}
+    {{- toYaml . | nindent 8 }}
+  {{- end }}
```
mumoshu, line 9: Hey! Thanks for the PR. Just confirming, but how did you test it?

I think this doesn't work as it removes all the preceding spaces until the end of line 6, resulting in something like the below for e.g. `serviceMonitorLabels: {"foo": "FOO", "bar": "BAR"}`:

```
metadata:
  labels:
    release: whatever
    chart: actions-runner-controllerfoo: FOO
        bar: BAR
```

I guess it should better be:

```suggestion
  {{- with .Values.metrics.serviceMonitorLabels }}
    {{ toYaml . | nindent 4 }}
  {{- end }}
```

so that the prev example becomes:

```
metadata:
  labels:
    release: whatever
    chart: actions-runner-controller
    foo: FOO
    bar: BAR
```

Would you mind giving it a shot once again?

mumoshu: Well I've been misunderstanding how nindent works. So `nindent` seems to add an empty line at the beginning so in the current form the above example would become:

```
metadata:
  labels:
    release: whatever
    chart: actions-runner-controller
        foo: FOO
        bar: BAR
```

Then perhaps this should be:

```suggestion
  {{- with .Values.metrics.serviceMonitorLabels }}
    {{- toYaml . | nindent 4 }}
  {{- end }}
```

kirillbilchenko: @mumoshu fixed, yes, I missed the number of spaces.

mumoshu:
@kirillbilchenko Thanks a lot for your contribution! I'm releasing 0.12.7 for this.

PR #671, opened 2021-06-30
README.md: add example for monitoring the github runner

Mic92:


peimanja:
Does it makes sense for controller to also get these metrics and expose them? or is that considered out of scope?

Mic92:
I would use them if they were available. But I also see value in asking github directly.

toast-gear:
Thanks for the work however we aren't going to merge this PR unfortunately. The PR isn't very related to the project really beyond the fact it monitors an API we happen to consume. There are millions of tools and if we try to document all of them the documentation will very quickly become out of date, incorrect and unusable and so it just isn't possible.

> Does it makes sense for controller to also get these metrics and expose them? or is that considered out of scope?

It's out of scope for now, this really is something GitHub need to resolve with their Actions offering. There is an issue about observability which they really need to look into https://github.com/actions/runner/issues/1116. Please thumbs up and / or comment on this issue as it's a major shortcoming of the product offering atm.

Mic92:
I trust telegraf to be more stable than your project to be honest.

Mic92:
Drop it here: https://github.com/actions/runner/discussions/1184

toast-gear:
https://github.com/Spendesk/github-actions-exporter for posterity's sake I thought I'd added this here. I just found this project which tackles the observability issues around runners as well as the workflows themselves. Haven't had a chance to test it yet but it looks promising.

PR #684, opened 2021-07-07
Update README.md to fix scaleUp/Down examples

lucas-pate:
The examples for scaleDownAdjustment and scaleUpAdjustment were incorrect. The first letter should not have been capitalized, and the value should have been an integer instead of a string.

Links to the CRDs:
[scaleDownAdjustment](https://github.com/actions-runner-controller/actions-runner-controller/blob/f19e7ea8a8bce21bf66d2ee9eaaabb245c7992fa/charts/actions-runner-controller/crds/actions.summerwind.dev_horizontalrunnerautoscalers.yaml#L84)
[scaleUpAdjustment](https://github.com/actions-runner-controller/actions-runner-controller/blob/f19e7ea8a8bce21bf66d2ee9eaaabb245c7992fa/charts/actions-runner-controller/crds/actions.summerwind.dev_horizontalrunnerautoscalers.yaml#L99)

### DIFF
--- a/README.md
+++ b/README.md
@@ -469,8 +469,8 @@ spec:
   - type: PercentageRunnersBusy
     scaleUpThreshold: '0.75'    # The percentage of busy runners at which the number of desired runners are re-evaluated to scale up
     scaleDownThreshold: '0.3'   # The percentage of busy runners at which the number of desired runners are re-evaluated to scale down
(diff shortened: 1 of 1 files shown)


mumoshu (approved):
@lucas-pate Ahh good catch! Thanks a lot for fixing this ☺️

Issue #656, opened 2021-06-25
Service monitor reports "server returned HTTP status 400 Bad Request" on latest


AlexDHoffer:
**Describe the bug**

Set to latest, my Prometheus Operator installation cannot seem to scrape the service monitor. It reports the following:

![image](https://user-images.githubusercontent.com/19369483/123486067-371cbf00-d5c0-11eb-9dd1-8e6ace066bd2.png)

The endpoint it is looking at is "metrics-port". Kind of curious if this is an HTTP/HTTPS issue?

**Checks**

- [ x] My actions-runner-controller version (v0.x.y) does support the feature
- [ ] I'm using an unreleased version of the controller I built from HEAD of the default branch

**To Reproduce**

Run the latest Helm chart. Make sure CRDs are updated to this version. Make sure runner pods are using latest image. Then, enable the service monitor with a label that your Prometheus Operator can key off. Then, look at your targets in your Prometheus UI.

**Expected behavior**
Successful flowing of metrics.

**Environment (please complete the following information):**
 - Controller Version [e.g. 0.18.2] summerwind/actions-runner-controller:v0.19.0
 - Deployment Method [e.g. Helm and Kustomize ] Helm
 - Helm Chart Version [e.g. 0.11.0, if applicable] 0.12.2

**Additional context**

Running on AKS. Prometheus stack version is kube-prometheus-stack-9.3.4 deployed via Helm.

AlexDHoffer:
Note: disabling the metrics proxy fixed the issue here. Any ideas how I can adjust so that I can enable the metrics proxy here and it will work?

mumoshu:
@AlexDHoffer Hey! Thanks for reporting. I think this is the exact same issue as I'm discussing for a potential fix at #625 with @jstewart612.

Would you mind checking https://github.com/actions-runner-controller/actions-runner-controller/pull/625#issuecomment-867229742 and see if the potential solution work?

AlexDHoffer:
@mumoshu Thanks for the heads up! I'll give it a shot Monday and report back. Hopefully it is OK if I leave this open til then.

AlexDHoffer:
@mumoshu This config borrowed from the posted link indeed allows you to scrape via the proxy:

````
endpoints:
  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
    path: /metrics
    port: https
    scheme: https
    tlsConfig:
      insecureSkipVerify: true

[...]

PR #685, opened 2021-07-08
Use https connection when metrics enabled for githubwebhook server

rlandesman:
This PR matches #625 and adds necessary RBAC permissions to fix the 401 issue first reported [here](https://github.com/actions-runner-controller/actions-runner-controller/issues/656). I've tested this locally and Prometheus seems to be able to read from the servicemonitor as expected now.

### DIFF
--- a/charts/actions-runner-controller/Chart.yaml
+++ b/charts/actions-runner-controller/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.12.7
+version: 0.12.8
 
 # Used as the default manager tag value when no tag property is provided in the values.yaml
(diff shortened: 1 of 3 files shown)


charts/actions-runner-controller/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.12.7
+version: 0.12.8
```
```suggestion
version: 0.12.7
```

Thanks! This almost LGTM, but would you mind reverting the version bump?
The rationale is that the current unreleased version of chart includes changes that requires a new controller release, and the new controller release will take a week or so to be shipped. We don't want to a chart that requires unreleased version of the controller.

mumoshu (approved):
Thanks for your contribution @rlandesman ☺️

roeelandesman:
Thanks for the help and merge here @mumoshu!

PR #700, opened 2021-07-15
Document volumeStorageMedium and volumeSizeLimit

rofafor:


README.md
```diff
@@ -830,6 +830,14 @@ spec:
...
         - mountPath: /home/runner/work/repo
           name: repo
+      # Optional storage medium type of runner volume mount.
+      # More info: https://kubernetes.io/docs/concepts/storage/volumes/#emptydir
+      # "" (default) = Node's default medium
+      # Memory = RAM-backed filesystem (tmpfs)
+      volumeStorageMedium: ""
+      # Total amount of local storage required for runner volume mount.
+      # The default limit is undefined.
+      volumeSizeLimit: 4Gi
```
mumoshu, line 840:
```suggestion
      # Optional storage medium type of runner volume mount.
      # More info: https://kubernetes.io/docs/concepts/storage/volumes/#emptydir
      # "" (default) = Node's default medium
      # Memory = RAM-backed filesystem (tmpfs)
      # See https://github.com/actions-runner-controller/actions-runner-controller/pull/674 for when to use this.
      volumeStorageMedium: ""
      # Total amount of local storage required for runner volume mount.
      # The default limit is undefined.
      # See https://github.com/actions-runner-controller/actions-runner-controller/pull/674 for when to use this.
      volumeSizeLimit: 4Gi
```

@rofafor Probably some hint for "why you need this" would be helpful for users, like the above suggestion? WDYT?

rofafor: That PR didn't reveal any major info about using storage medium, so linking it doesn't make much sense (IMO).  Anyway, I've rephrased these. Better now?

mumoshu (approved):
This looks much better now! Thanks for your contribution @rofafor ☺️

Issue #722, opened 2021-08-11
action-runner has crashed loop with latest version.


h-r-k-matsumoto:
**Describe the bug**
Suddenly runner pod has crashed loop after 2021-08-11.
I think `summerwind/actions-runner:latest` is not built correctly.

**Checks**

- [x] My actions-runner-controller version (v0.x.y) does support the feature
- [ ] I'm using an unreleased version of the controller I built from HEAD of the default branch

**To Reproduce**

First o all , `summerwind/actions-runner:latest` cannot be started.

```bash
$ docker images |findstr  summerwind/actions-runner
summerwind/actions-runner                                                     latest         42c0cefce236   3 hours ago     788MB
$ docker run -it --rm summerwind/actions-runner:latest
standard_init_linux.go:228: exec user process caused: exec format error
$
```

I think `dump-init` doesnot install correctly.
```bash
$ docker run -it --rm --entrypoint '' summerwind/actions-runner:latest /bin/sh
$ cat  /usr/local/bin/dumb-init
Not Found$
```

**Expected behavior**
It can be started normally.

**Screenshots**
Nothing.

**Environment (please complete the following information):**
 - Controller Version: 0.18.2
 - Deployment Method: Kustomize 

**Additional context**
Nothing

h-r-k-matsumoto:
There is the code below.
https://github.com/actions-runner-controller/actions-runner-controller/blob/master/runner/Dockerfile#L51
```bash
    && curl -L -o /usr/local/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v${DUMB_INIT_VERSION}/dumb-init_${DUMB_INIT_VERSION}_${ARCH} \
```
This is run as follows:

```bash
  && curl -L -o /usr/local/bin/dumb-init  https://github.com/Yelp/dumb-init/releases/download/v1.2.5/dumb-init_1.2.5_amd64
```

However `dumb-init_1.2.5_amd64` file is not exists in [Yelp/dumb-init/releases/tag/v1.2.5](https://github.com/Yelp/dumb-init/releases/tag/v1.2.5)

h-r-k-matsumoto:
A Condition with downloading dumb-init is wrong.
https://github.com/Yelp/dumb-init/releases/tag/v1.2.3
> x86_64 (previously amd64)
> aarch64 (previously arm64)
> ppc64le (previously ppc64el)

if amd64, $ARCH should be set `x86_64` .

[...]

PR #710, opened 2021-08-02
chore: bump dumb-init

callum-tait-pbx:
- bump dumb-init to 1.2.5 (latest) https://github.com/Yelp/dumb-init
- align Dockerfile's to be consistent
- align runner Makefile with root Makefile

my integration tests passed, we need to sort out something proper for these changes long term though, bump the docker version next

image

PR #723, opened 2021-08-11
fix: arch type with downloading dumb-init.

h-r-k-matsumoto:
I'd like to fix a problem #722.

This problem is caused by #710.

### DIFF
--- a/runner/Dockerfile
+++ b/runner/Dockerfile
@@ -48,7 +48,7 @@ RUN apt update -y \
 RUN export ARCH=$(echo ${TARGETPLATFORM} | cut -d / -f2) \
     && if [ "$ARCH" = "arm64" ]; then export ARCH=aarch64 ; fi \
     && if [ "$ARCH" = "amd64" ] || [ "$ARCH" = "i386" ]; then export ARCH=x86_64 ; fi \
-    && curl -L -o /usr/local/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v${DUMB_INIT_VERSION}/dumb-init_${DUMB_INIT_VERSION}_${ARCH} \
+    && curl -f -L -o /usr/local/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v${DUMB_INIT_VERSION}/dumb-init_${DUMB_INIT_VERSION}_${ARCH} \
     && chmod +x /usr/local/bin/dumb-init
 
 # Docker download supports arm64 as aarch64 & amd64 / i386 as x86_64
--- a/runner/Dockerfile
+++ b/runner/Dockerfile
@@ -47,7 +47,7 @@ RUN apt update -y \
 # arch command on OS X reports "i386" for Intel CPUs regardless of bitness
 RUN export ARCH=$(echo ${TARGETPLATFORM} | cut -d / -f2) \
     && if [ "$ARCH" = "arm64" ]; then export ARCH=aarch64 ; fi \
-    && if [ "$ARCH" = "i386" ]; then export ARCH=x86_64 ; fi \
+    && if [ "$ARCH" = "amd64" ] || [ "$ARCH" = "i386" ]; then export ARCH=x86_64 ; fi \
     && curl -L -o /usr/local/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v${DUMB_INIT_VERSION}/dumb-init_${DUMB_INIT_VERSION}_${ARCH} \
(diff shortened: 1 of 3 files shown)


runner/Dockerfile
```diff
@@ -47,7 +47,7 @@ RUN apt update -y \
 # arch command on OS X reports "i386" for Intel CPUs regardless of bitness
 RUN export ARCH=$(echo ${TARGETPLATFORM} | cut -d / -f2) \
     && if [ "$ARCH" = "arm64" ]; then export ARCH=aarch64 ; fi \
-    && if [ "$ARCH" = "i386" ]; then export ARCH=x86_64 ; fi \
+    && if [ "$ARCH" = "amd64" ] || [ "$ARCH" = "i386" ]; then export ARCH=x86_64 ; fi \
```
ammmze, line 50: We probably need to do this over [here](https://github.com/actions-runner-controller/actions-runner-controller/blob/a9421edd465b327ed6379209087f88119649189e/runner/Dockerfile.dindrunner#L111) too, right?

h-r-k-matsumoto: Yes, you're right. 
I'm going to fix it, too :)

h-r-k-matsumoto: I've fixed it!
cba8691bebb0b891fa11f1c4b61fdbb101e0b261

runner/Dockerfile
```diff
@@ -47,7 +47,7 @@ RUN apt update -y \
...
     && curl -L -o /usr/local/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v${DUMB_INIT_VERSION}/dumb-init_${DUMB_INIT_VERSION}_${ARCH} \
```
toast-gear, line 51: add the `-f` flag so curl doesn't silently error if it gets a 404

runner/Dockerfile.dindrunner
```diff
@@ -108,7 +108,7 @@ RUN chmod +x /usr/local/bin/startup.sh /usr/local/bin/entrypoint.sh /usr/local/b
 # arch command on OS X reports "i386" for Intel CPUs regardless of bitness
 RUN export ARCH=$(echo ${TARGETPLATFORM} | cut -d / -f2) \
     && if [ "$ARCH" = "arm64" ]; then export ARCH=aarch64 ; fi \
-    && if [ "$ARCH" = "i386" ]; then export ARCH=x86_64 ; fi \
+    && if [ "$ARCH" = "amd64" ] || [ "$ARCH" = "i386" ]; then export ARCH=x86_64 ; fi \
     && curl -L -o /usr/local/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v${DUMB_INIT_VERSION}/dumb-init_${DUMB_INIT_VERSION}_${ARCH} \
```
toast-gear, line 112: add the `-f` flag so curl doesn't silently error if it gets a 404

runner/Dockerfile.ubuntu.1804
```diff
@@ -47,7 +47,7 @@ RUN apt update -y \
 # arch command on OS X reports "i386" for Intel CPUs regardless of bitness
 RUN export ARCH=$(echo ${TARGETPLATFORM} | cut -d / -f2) \
     && if [ "$ARCH" = "arm64" ]; then export ARCH=aarch64 ; fi \
-    && if [ "$ARCH" = "i386" ]; then export ARCH=x86_64 ; fi \
+    && if [ "$ARCH" = "amd64" ] || [ "$ARCH" = "i386" ]; then export ARCH=x86_64 ; fi \
     && curl -L -o /usr/local/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v${DUMB_INIT_VERSION}/dumb-init_${DUMB_INIT_VERSION}_${ARCH} \
```
toast-gear, line 51: add the `-f` flag so curl doesn't silently error if it gets a 404

toast-gear (changes requested):
In addition, could you add the `-f` to all curl commands in each Dockerfile please

runner/Dockerfile.dindrunner
h-r-k-matsumoto: I've added -f option.

runner/Dockerfile.ubuntu.1804
h-r-k-matsumoto: I've added -f option.

hammadzz:
once this PR is merged is the latest image tag now containing the fix?

toast-gear:
@hammadzz yup, should be

PR #724, opened 2021-08-11
ci: set username statically

toast-gear:
GitHub a while ago changed how secrets work, secrets are not accessible from forks now (also impacts dependabot). The docker username isn't a secret so I've just statically set it.

Additionally we had the runner version out of sync between latest and our immutable tags. I've moved all of the env stuff to the workflow level so we use the same runner software, docker version and username for both builds as they shouldn't ever be out of sync


mumoshu (approved):
LGTM. Thanks for fixing it!

PR #629, opened 2021-06-12
feat: RunnerSet backed by StatefulSet

mumoshu:
TL;DR; RunnerSet is a more feature-rich, flexible, easy to configure, and maintainable alternative to RunnerDeployment.

- It can be said feature-rich as it supports #612 for using persistent volumes for caching.
- It is flexible as it supports all the pod template settings from StatefulSet API while supporting all the runner-related settings from Runner API
- It is easy to configure and maintainable as the pod-related and container-related settings are now inherited from StatefulSet/Pod Template API. We don't need to maintain and rely on our own variants included in Runner Spec.

---

A runnerset can manage a set of "stateful" runners by combining a statefulset and an admission webhook. A statefulset is a standard Kubernetes construct that manages a set of pods and a pool of persistent volumes. We use that to manage runner pods, while using the admission webhook mutates each pod to have required environment variables and registration tokens.

It is considered to be a complete replacement to the former method of deploying a set of runners, RunnerDeployment, which also creates pods with the required environment variables and registration tokens.

## Differences between RunnerSet and RunnerDeployment

The only and big functional difference between RunnerSet and RunnerDeployment is that the former has support for `volumeClaimTemplates`, which allows actions-runner-controller to manage a pool of dynamically provisioned persistent volumes. This should be useful to make certain types of actions workflows faster by utilizing per-pod-identity cache, like docker layer caches in `/var/lib/docker` persistent across pod restarts.

The basic usage of RunnerSet is very similar to that of RunnerDeployment.

This RunnerDeployment:

```yaml
# runnerdeployment.yaml
apiVersion: actions.summerwind.dev/v1alpha1
kind: RunnerDeployment
metadata:
  name: example-runnerdeploy
spec:
  replicas: 2
  template:
    spec:
      repository: mumoshu/actions-runner-controller-ci
      env: []
```

can be rewritten to:

```yaml
# runnerset.yaml
apiVersion: actions.summerwind.dev/v1alpha1
kind: RunnerSet
metadata:
  name: example
spec:
  # NOTE: RunnerSet supports non-ephemeral runners only today
  ephemeral: false
  replicas: 2
  repository: mumoshu/actions-runner-controller-ci
  # Other mandatory fields from StatefulSet
  selector:
    matchLabels:
      app: example
  serviceName: example
  template:
    metadata:
      labels:
        app: example
```

[...]

PR #727, opened 2021-08-16
Update documentation about epehemral runners and RunnerSet

mumoshu:
Follow-up for #721 and #629


consideRatio:
LGTM for what its worth!

mumoshu:
@consideRatio Thanks for reviewing ☺️

PR #729, opened 2021-08-16
feat: bump runner version

dhruv-4:
Bumping the runner version


toast-gear:
Hi @dhruv-4  appreciate the PR however we have a few things to bump so we are going to do them together in this PR https://github.com/actions-runner-controller/actions-runner-controller/pull/730

Issue #715, opened 2021-08-05
Set ImageOS environment variable in runner images


strideynet:
**Is your feature request related to a problem? Please describe.**

Several off the shelf actions expect the ImageOS environment variable to be populated (e.g ruby/setup):

```
Run ruby/setup-ruby@v1
  with:
    ruby-version: 2.6
    bundler: default
    bundler-cache: false
    working-directory: .
    cache-version: 0
Error: The environment variable ImageOS must be set
```

**Describe the solution you'd like**

We should set a closest approximation of the ImageOS environment variable in the docker images (e.g `ubuntu20`)

**Describe alternatives you've considered**

You can set this as part of the Runner spec.

callum-tait-pbx:
https://github.com/actions/virtual-environments/blob/b7f276c003aea42575b52247bdb2183e355fca2f/images/linux/scripts/installers/configure-environment.sh 

We should probably set it the same way GitHub does for their ubuntu instances i.e. whatever value the below produces:

```
tee -a /etc/environment
```

consideRatio:
This PR is fixed by #730 according to the PR

toast-gear:
No issues reported since release, closing as resolved.

Issue #451, opened 2021-04-12
Bumping the docker client in runner image


Puneeth-n:
Any plans of bumping the runner's docker client version to something like `20.10.5`?

https://github.com/summerwind/actions-runner-controller/blob/b42b8406a2b6d606ef9b531867720039e1de1b42/runner/Dockerfile#L5

callum-tait-pbx:
It does need doing. We can look at bumping it once the runner image itself is on ubuntu 20.04 https://github.com/summerwind/actions-runner-controller/pull/438

callum-tait-pbx:
@mumoshu could we get the `chore` label on this, we will want to bump the docker version, probably worth targeting after the new scale from 0 features are stable.

consideRatio:
This PR is fixed by #730 according to the PR

toast-gear:
No issues reported since release, closing as resolved.

PR #730, opened 2021-08-16
chore: bump docker and runner version and add imageos env var

toast-gear:
fixes:
https://github.com/actions-runner-controller/actions-runner-controller/issues/715
https://github.com/actions-runner-controller/actions-runner-controller/issues/451

I've added in the missing fail `-f` switches too to the remaining curl commands 

We don't use docker massively so I've done limited testing, will try to test a bit more, my testing is with the dind runner atm

tests:
image


Puneeth-n:
@toast-gear @mumoshu  any ETA on when this would be merged? We are stuck with a bunch of runners that haven't terminated and come up again

PR #735, opened 2021-08-19
Update README.md

Tarasovych:


README.md
```diff
@@ -100,6 +100,7 @@ _Note: Links are provided further down to create an app for your logged in user
 
 * Actions (read)
 * Administration (read / write)
+* Checks (read) (if you are going to use [Faster Autoscaling with GitHub Webhook](#faster-autoscaling-with-github-webhook))
```
mumoshu, line 103: Oh, really? I was in an impression that we are only relying on webhook payload of check run events so there's no need to call any GitHub API to access checks, hence no need to add the permission.

README.md
```diff
@@ -111,6 +112,9 @@ _Note: Links are provided further down to create an app for your logged in user
 **Organization Permissions**
 * Self-hosted runners (read / write)
 
+**Subscribe to events**
+* Check run (if you are going to use [Faster Autoscaling with GitHub Webhook](#faster-autoscaling-with-github-webhook))
```
mumoshu, line 116: Ah so you need to configure a subscription to let your github apps' webhook to receive check run events? (I have not yet tested it with GitHub App myself so just confirming

README.md
Tarasovych: > we are only relying on webhook payload

aren't we?

README.md
Tarasovych: I haven't managed to setup the App in other way. From the very beginning I hadn't any App subsriptions, autoscaling wasn't working

README.md
mumoshu: What do you mean? AFAIK, here's the only place in where we rely on check runs. You'll see there's no API calls related to check runs?

https://github.com/actions-runner-controller/actions-runner-controller/blob/f858e2e43273a861b06b546c6311fe68fb6e1ce5/controllers/horizontal_runner_autoscaler_webhook_on_check_run.go#L9

Tarasovych: It's impossible to subscribe to "Check run" events at all if you don't grant read access to Checks

mumoshu: @Tarasovych Oh, that's good to know! Thanks for your clarification ☺️

README.md
```diff
@@ -56,7 +56,7 @@ kubectl apply -f https://github.com/actions-runner-controller/actions-runner-con
 
 **Helm Deployment:**
 
-__**Note: For all configuration options for the Helm chart see the chart's [README](./charts/actions-runner-controller/README.md)
+__**Note: For all configuration options for the Helm chart see the chart's [README](./charts/actions-runner-controller/README.md)**__
```
consideRatio, line 59: `__` and `**` are alternative syntax of the same thing, and using `**` is according to [this](https://www.markdownguide.org/basic-syntax/#bold) the best practice for compatibility purposes.

```suggestion
**Note: For all configuration options for the Helm chart see the chart's [README](./charts/actions-runner-controller/README.md)**
```

README.md
```diff
@@ -352,7 +356,7 @@ This, in combination with a correctly configured HorizontalRunnerAutoscaler, all
...
 
-__**NOTE: Once `workflow_job` webhook events are released on GitHub, the webhook-based autoscaling is the preferred way of autoscaling, because it is easy to configure and has the ability to accurately detect which runners to scale. See [Example 3: Scale on each `workflow_job` event](#example-3-scale-on-each-workflow_job-event)**___
+__**NOTE: Once `workflow_job` webhook events are released on GitHub, the webhook-based autoscaling is the preferred way of autoscaling, because it is easy to configure and has the ability to accurately detect which runners to scale. See [Example 3: Scale on each `workflow_job` event](#example-3-scale-on-each-workflow_job-event)**__
```
consideRatio, line 359:
```suggestion
**NOTE: Once `workflow_job` webhook events are released on GitHub, the webhook-based autoscaling is the preferred way of autoscaling, because it is easy to configure and has the ability to accurately detect which runners to scale. See [Example 3: Scale on each `workflow_job` event](#example-3-scale-on-each-workflow_job-event)**
```

mumoshu:
@Tarasovych @consideRatio Thanks for your contributions!

mumoshu (approved):

PR #738, opened 2021-08-20
Update actions.summerwind.dev_horizontalrunnerautoscalers.yaml

Tarasovych:
According to https://github.com/actions-runner-controller/actions-runner-controller/blob/master/api/v1alpha1/horizontalrunnerautoscaler_types.go#L118

### DIFF
--- a/charts/actions-runner-controller/crds/actions.summerwind.dev_horizontalrunnerautoscalers.yaml
+++ b/charts/actions-runner-controller/crds/actions.summerwind.dev_horizontalrunnerautoscalers.yaml
@@ -83,16 +83,19 @@ spec:
                       items:
                         type: string
                       type: array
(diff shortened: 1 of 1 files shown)


mumoshu (approved):
@Tarasovych Hey!

Thanks for the PR. But this needs to be auto-generated if it needs to be really `nullable`.

The magic comment to enable setting `nullable: true` via autogeneration is `// +nullable` as far as I know.

I'd suggest reading https://book.kubebuilder.io/reference/markers/crd-validation.html for more information about that.

I'd like to ask you a thing though- why this needs to be nullable? If it needs to be nullable you'd also make it a pointer type in Go code, too. But you haven't changed Go code. So I'm quite unsure what's your goal is.

PR #763, opened 2021-08-25
Fix a few small YAML typos

pje:
- Remove two extra colons that were making the yaml invalid 🕵️
- Add `yaml` tags to the markdown blocks 🧹

Thank you for the excellent and thorough documentation!


toast-gear:
LGTM

Issue #761, opened 2021-08-24
Unable to pass dnsConfig to self-hosted autoscaling runners


sammcj:
I'm trying to set `dnsConfig` on the self-hosted autoscaling runner deployments.

Apologies if I'm simply trying to set it in the wrong place - I'm new to helm and very new to Actions.

**Checks**

- [?] My actions-runner-controller version (v0.12.7) does support the feature
- [-] I'm using an unreleased version of the controller I built from HEAD of the default branch

**To Reproduce**

I followed [actions-runner-controller.md](https://github.com/actions-runner-controller/actions-runner-controller#autoscaling), using helm for the controller and deploying the runners via a simple kubernetes deployment manifest.

Which errors as it can't find RunnerDeployment.spec.template.spec.dnsConfig:

```
error: error validating "runnerdeployment.yaml": error validating data: ValidationError(RunnerDeployment.spec.template.spec): unknown field "dnsConfig" in dev.summerwind.actions.v1alpha1.RunnerDeployment.spec.template.spec; if you choose to ignore these errors, turn validation off with --validate=false
```

1. Deploy the self-hosted runner controller using the provided helm chart.
2. Create a deployment.yaml similar to those supplied in [actions-runner-controller.md](https://github.com/actions-runner-controller/actions-runner-controller#autoscaling).
3. Add `dnsConfig` to the deployment.yaml and apply.
4. Try applying the deployment.
5. Experience error above as the CRD doesn't accept `RunnerDeployment.spec.template.spec.dnsConfig`

**Expected behavior**

The following k8s deployment manifest (or similar) to work:

```yaml
apiVersion: actions.summerwind.dev/v1alpha1
kind: RunnerDeployment
metadata:
  name: gha-runner-deployment
spec:
  template:
    spec:
      dnsConfig:
        options:
        - name: ndots     
          value: "1"
```

and from that `dnsConfig` should be passed to runner pods which configures /etc/resolv.conf inside containers with the provided options.

**Environment:**
 - Controller Version - 0.19.0
 - Deployment Method - Helm
 - Helm Chart Version - v0.12.7

**Additional Context:**

The reason I'm needing to configure this is because it seems with DIND (or really Docker in ContainerD) some base images (such as alpine) incorrectly get `ndots:5` configured in `/etc/resolv.conf` which causes DNS lookup failures inside the container.

[...]

PR #764, opened 2021-08-25
Add dnsConfig to runner deployments

sammcj:
Add ability to set dnsConfig for runner deployments.

As per https://github.com/actions-runner-controller/actions-runner-controller/issues/761#issuecomment-904404176

Note: I've never touched any golang or helm so just sort of fumbled my way through this - I'm sure I've missed several things.


sammcj:
The test failure doesn't sound related:

```
Summarizing 1 Failure:

[Fail] INTEGRATION: Inside of a new namespace when no existing resources exist [It] should create and scale user's repository runners only on pull_request event 
/home/runner/work/actions-runner-controller/actions-runner-controller/controllers/integration_test.go:850

Ran 12 of 12 Specs in 34.532 seconds
FAIL! -- 11 Passed | 1 Failed | 0 Pending | 0 Skipped
```

@mumoshu any chance you could point me in the right direction?

mumoshu:
That might be due to test flakiness. Let me rerun it.

sammcj:
Looks like it worked, do you think it looks like everything that's required?

mumoshu:
Yep, let's go ahead! Thanks for your contribution ☺️

mumoshu (approved):

guyjacksongit (commented):
nice

billimek:
@sammcj has this worked as expected for you?  I'm struggling at the moment and wondering if there is a configuration nuance that I'm missing.

With the following `RunnerDeployment` config snippet,

```yaml
kind: RunnerDeployment
metadata:
  name: myorg-ephemeral-dns
  namespace: actions
spec:
  replicas: 1
  template:
    spec:
      organization: myorg
      labels:
        - myorg-onprem-dns
      dnsConfig:
        - options:
          - name: "ndots"
            value: "1"
```

... The `Runners` get created with the dnsConfig settings, but the associated pods do not.  I've experimented with different indenting and even other dnsConfig type settings, but nothing seems to translate to a change to the pod spec configuration or the actual `/etc/resolv.conf` file being changed.

billimek:
I think we may be missing something like this in `controllers/runner_controller.go`, based on comparing this change with a [similar change](https://github.com/actions-runner-controller/actions-runner-controller/pull/456)

```go
	if len(runner.Spec.DnsConfig ) != 0 {
		pod.Spec.DnsConfig = runner.Spec.DnsConfig 
	}
```

Going to try to test locally and if that's the case submit a new PR.

billimek:
https://github.com/actions-runner-controller/actions-runner-controller/pull/1227 raised for the above, FYI

PR #774, opened 2021-08-30
Update README.md

Tarasovych:


mumoshu (approved):
Ahhh, good catch! Thanks for fixing it 👍

PR #779, opened 2021-08-31
Renovate/k8s.io client go 12.x

lgyorgyi:


mumoshu:
Your branch seems to be messed up with a lot of old commits! Perhaps you'd better restart from the latest branch, if this isn't a total mistake and you have something to do really.

Issue #455, opened 2021-04-14
Document the  --watch-namespace feature


callum-tait-pbx:
**Relevant Issues / Commits**
https://github.com/summerwind/actions-runner-controller/commit/1b8a656051c8a8a187fe1b34f20b92f169485ae6
https://github.com/summerwind/actions-runner-controller/issues/377
https://github.com/summerwind/actions-runner-controller/issues/442

callum-tait-pbx:
@mumoshu this is legit issue. Please add the `documentation` label and remove the stale label so the bot doens't kill it.

shichengripple001:
where is the example? the document is very hard to understand..

PR #786, opened 2021-09-02
docs: watch namespace feature

toast-gear:
Fixes: https://github.com/actions-runner-controller/actions-runner-controller/issues/455

I wasn't sure if the controller needs to be installed in the same namespace as the runners themselves due to how garbage collection works https://kubernetes.io/docs/concepts/architecture/garbage-collection/#owners-dependents

```
Cross-namespace owner references are disallowed by design. Namespaced dependents can specify cluster-scoped 
or namespaced owners. A namespaced owner must exist in the same namespace as the dependent. If it does not, 
the owner reference is treated as absent, and the dependent is subject to deletion once all owners are verified 
absent.
```

I've said same namespace installtion is required, if this is not a requirement please comment on the PR and I'll update the docs to reflect this in this PR @mumoshu

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+ # actions-runner-controller
+
 [![awesome-runners](https://img.shields.io/badge/listed%20on-awesome--runners-blue.svg)](https://github.com/jonico/awesome-runners)
 
 This controller operates self-hosted runners for GitHub Actions on your Kubernetes cluster.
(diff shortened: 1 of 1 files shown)


mumoshu:
> Cross-namespace owner references are disallowed by design

I thought `owner` in this context to represent "a parent K8s object", not a K8s controller like actions-runner-controller's controller-manager.

K8s garbage collection works only on a tree of dependent K8s objects, like deletion of a RunnerDeployment results in terminating RunnerReplicaSet, and then Runner, and finally runner pods.

> the controller needs to be installed in the same namespace as the runners

So, I think this isn't the case.

Just to be extra sure- did you actually observed that actions-runner-controller needs to be installed onto the same namespace as runners when you enabled `watch-namespace`? If so, I may be missing something.

mumoshu (approved):
Awesome work as always! Thanks @callum-tait-pbx ☺️

Issue #782, opened 2021-09-01
flag provided but not defined: --watch-namespace


rxa313:
When installing via helm, I have the `singleNamespace: true` and `watchNamespace: "gitactions-xxx"` populated. I check in rancher the status of the manager and I'm seeing the error that the flag is provided but not defined.

I'm running v0.12.0 of the runner controller.

I checked the source code of the v0.12.0 and I can see the --watch-namespace feature, it's just not working as expected when installing. To be sure, I cleared all of my namespaces, crds, and all our of rancher and tried to do a clean install of the controller.

 - Controller Version: v0.12.0
 - Deployment Method: Helm

the watchNamespace feature is very important in running multiple organizations with different installation ID's. There is no documentation on how to get this working properly..

mumoshu:
@rxa313 Hey! The controller version v0.12.0 is way too outdated that it doesn't support `--watch-namespace`

https://github.com/actions-runner-controller/actions-runner-controller/blob/v0.12.0/main.go#L73-L82

Could you try e.g. v0.19.0 which does support it?

https://github.com/actions-runner-controller/actions-runner-controller/blob/c424d1afee693d0f2297df53d663f8a8137b9e51/main.go#L98

rxa313:
Hi @mumoshu , upgrading to version v0.19.0 did indeed fix that issue. 

I did reply to this discussion here [#505](https://github.com/actions-runner-controller/actions-runner-controller/discussions/505#discussioncomment-1255499)

How can I deploy the controller to multiple namespaces in one cluster? I believe that is how it's supposed to work when running multiple orgs? I just see an error that the ClusterRole is already assigned to a namespace (actions-runner-system)

Thanks for your help.

mumoshu:
@rxa313 Could you share the exact error message you've seen?

Generally speaking, I think each installation of actions-runner-controller needs to have a dedicated serviceaccount in its namespace (say, actions-runner-system-foo and actions-runner-system-bar) and each serviceaccount needs to have a corresponding RoleBinding to bind the ClusterRole to each serviceaccount.

If running `helm install -n actions-runner-system-foo --set watchNamespace=foo ...` and `helm install -n actions-runner-system-bar --set watchNamespace-bar ...` doesn't work, maybe our chart is not supporting the use-case correctly?

If you could share exact errors you've seen and exact commands you've run, probably I can provide a better guidance (or idea)

toast-gear:
[...]

PR #794, opened 2021-09-07
docs: adding requirement for unique secret name

toast-gear:
Waiting on https://github.com/actions-runner-controller/actions-runner-controller/issues/782 to be resolved before merging


toast-gear:
Looks like #782 is more or l eft resolved, the fix to the cert is transparent and so doesn't need mentioning.

Issue #801, opened 2021-09-10
`containers` key on RunnerDeployment causes error


tyrken:
**Describe the bug**
I was trying to use the quasi-documented `containers` key on the RunnerDeployment's template-spec to add a lifecycle postStart script, but kept getting an error about the name missing show up, when it was set in my definition (see below in **To Reproduce**):

```
... ERROR	controller-runtime.manager.controller.runner-controller	Reconciler error	{"reconciler group": "actions.summerwind.dev", "reconciler kind": "Runner", "name": "test-runners-zltbw-cmx8d", "namespace": "gha", "error": "Runner.actions.summerwind.dev \"test-runners-zltbw-cmx8d\" is invalid: spec.containers.name: Required value"}```
```

**Checks**
- [x] My actions-runner-controller version (v0.x.y) does support the feature - Tested v0.19.0 originally via helm chart
- [x] I'm using an unreleased version of the controller I built from HEAD of the default branch - Also present in canary DockerHub image & self-built master image

**To Reproduce**
1. Install v0.19.0 or master build of action-runner-controller
2. Install resource below

```
apiVersion: actions.summerwind.dev/v1alpha1
kind: RunnerDeployment
metadata:
  name: test-runners
  namespace: gha
spec:
  replicas: 2
  template:
    spec:
      organization: my-org
      labels:
        - test
      # Use Containers block for full control
      containers:
        - name: runner
          # Example customisation
          lifecycle:
            postStart:
              exec:
                command:
                  - /bin/sh
                  - -c
                  - sleep 50
```

**Expected behavior**
The customisations inside the `containers` key apply to the final created pods.

**Screenshots**
After a little exploration, I noticed the YAML output from the K8s masters was actually saying the same thing, note the array of a single empty object in the containers field below:

```
$ kubectl get runnerdeployment -o yaml
apiVersion: v1
items:
- apiVersion: actions.summerwind.dev/v1alpha1
  kind: RunnerDeployment
  metadata:
   # snipped
  spec:
    replicas: 2
    selector: null
    template:
      metadata: {}
      spec:
        #### NOTE EMPTY OBJECT BELOW
        containers:
        - {}
        dockerdContainerResources: {}
        image: ""
        labels:
        - test
        organization: my-org
        resources: {}
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

**Environment (please complete the following information):**
 - Controller Version: 0.19.0 and master
 - Deployment Method: Helm
[...]

PR #802, opened 2021-09-10
Correct default container image logic

tyrken:
This might (not yet tested so in DRAFT) fix the minor issue at the end of #801, whereby an `image` inside the `containers` key was being reset to the default image when I had no top-level `image` key.


mumoshu (commented):
I was never intended to support specifying image inside the containers for RunnerDeployment API. (We do for RunnerSet API though) But if this works, I'd be happy to accept 😄

tyrken:
This passes acceptance tests along with the fix in #810

mumoshu (approved):
LGTM. Thanks for the fix @tyrken ☺️

PR #803, opened 2021-09-10
Revert manifest workaround for v1.18

tyrken:
As suggested in #801, this undoes the earlier workaround in #123 to allow the newer generator (I haven't updated any deps but hope the current generator version is new enough) to create complete CRDs.

I have verified the result successfully applied on K8s 1.20 and fixes my main issue in #801, but haven't gotten any end-to-end tests running locally yet...


tyrken:
This passes acceptance tests along with the fix in #810

mumoshu (approved):
LGTM. Thanks @tyrken ☺️

PR #808, opened 2021-09-13
Adding unit test for ephemeral feature flag

sledigabel:
This was something that was missing in #707.
Adding a new test to make sure the ephemeral feature flag from upstream
is set up correctly by the script.


mumoshu (approved):
Great job @sledigabel!!

PR #810, opened 2021-09-13
Complete CRDs for acceptance testing

tyrken:
I'm not sure if this has any other affect, but I couldn't get acceptance tests to pass until completing the set of CRDs it deploys via this change.


mumoshu (approved):
LGTM. Thanks for all your efforts @tyrken!

PR #806, opened 2021-09-11
Build tweaks

rofafor:
The first commit `effb083` is for #782 required for running multiple controllers simultaneously.

The next two commits are just for nerdctl/containerd (0.11.1) compatibility.

```
> cat /usr/local/bin/docker
#/bin/sh
lima nerdctl "$@"
```

The last commit is for unique leader election id in order to run multiple controllers in a single namespace. This lacks documentation update for README.

PR #815, opened 2021-09-14
Update docs and values.yaml for leader election customization

mumoshu:
Follow-up for #806

### DIFF
--- a/charts/actions-runner-controller/values.yaml
+++ b/charts/actions-runner-controller/values.yaml
@@ -8,11 +8,10 @@ replicaCount: 1
 
 syncPeriod: 10m
 
-controller:
-  enableLeaderElection: true
-  # Specifies the controller id for leader election.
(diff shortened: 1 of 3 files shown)


charts/actions-runner-controller/values.yaml
```diff
@@ -8,11 +8,10 @@ replicaCount: 1
 
 syncPeriod: 10m
 
-controller:
-  enableLeaderElection: true
-  # Specifies the controller id for leader election.
-  # Must be unique if more than one controller installed.
-  #leaderElectionId: "actions-runner-controller"
+enableLeaderElection: true
```
mumoshu, line 11: @rofafor Hey! Thanks again for #806. I'm moving this to the top-level for consistency with existing keys like `syncPeriod`.

toast-gear (approved):

PR #820, opened 2021-09-15
Add GitHub runner metrics

s4nji:
It would be nice if the controller provides these metrics, as controller already fetches runner information from GitHub periodically via [`ListRunners`](https://github.com/actions-runner-controller/actions-runner-controller/blob/58d2591f09f70bf271c165c755563a19450c30ee/github/github.go#L152) in [reconcile](https://github.com/actions-runner-controller/actions-runner-controller/blob/58d2591f09f70bf271c165c755563a19450c30ee/controllers/runner_controller.go#L353) and for [`PercentageRunnersBusy`](https://github.com/actions-runner-controller/actions-runner-controller/blob/58d2591f09f70bf271c165c755563a19450c30ee/controllers/runner_controller.go#L353) autoscaler, so it's not necessary to set up another system to periodically pull this data from GitHub (e.g. with [tchelovilar/github-org-runner-exporter](https://github.com/tchelovilar/github-org-runner-exporter))

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -173,6 +173,8 @@ func (c *Client) ListRunners(ctx context.Context, enterprise, org, repo string) 
 		opts.Page = res.NextPage
 	}
 
+	metrics.SetRunnerStatus(runners, enterprise, org, repo)
+
 	return runners, nil
 }
 
--- a/github/metrics/runner.go
+++ b/github/metrics/runner.go
@@ -1,0 +1,73 @@
+// Package metrics provides monitoring of the GitHub related metrics.
+//
+// This depends on the metrics exporter of kubebuilder.
+// See https://book.kubebuilder.io/reference/metrics.html for details.
+package metrics
+
+import (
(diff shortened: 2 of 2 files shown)


peimanja:
Can we also add runners labels as metric label here? we run different runners and would be nice to have visibility over each group

s4nji:
@peimanja I agree, having the runner labels in the metric is definitely useful. It is now included via 98e31462a2a1b494ea3ce946746cb8f8c1bbbc54.

I'm slightly concerned about the metric cardinality due to the many possible combination of the labels, but I think with [this approach](https://github.com/actions-runner-controller/actions-runner-controller/pull/820/files#diff-99d7b8d140285c9fc438f16bea42686ed59494a9916de43a2005b6c060dc808cR68-R77) it should be fine; you'd need to add (and combine) a _lot_ of runner labels to hit untenable (>10000?) levels that would bring down prometheus.

PR #824, opened 2021-09-15
feat: Added option for secret annotation

maxtacu:


charts/actions-runner-controller/values.yaml
```diff
@@ -21,6 +21,9 @@ syncPeriod: 10m
 authSecret:
   create: false
   name: "controller-manager"
+  ## Set specific secret annotations
+  # annotations: |
```
mumoshu, line 25: I think you must omit `|` otherwise the whole `annotations` becomes a YAML string and it doesn't work correctly when piped to `toYaml`, right?

charts/actions-runner-controller/templates/manager_secrets.yaml
```diff
@@ -4,6 +4,12 @@ kind: Secret
 metadata:
   name: {{ include "actions-runner-controller.secretName" . }}
   namespace: {{ .Release.Namespace }}
+  {{- if .Values.authSecret.annotations }}
+  annotations:
```
mumoshu, line 8: I thought you can just do it like `annotations: {{ toYaml .Values.authSecret.annotations | nindent 4 }}` as nindent prepends a new line before the output

charts/actions-runner-controller/templates/ci-secret.yaml
```diff
@@ -7,4 +7,8 @@ data:
 kind: Secret
 metadata:
   name: controller-manager
+  {{- if .Values.authSecret.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
```
mumoshu, line 13: I believe you forgot to add `{{ with }}` here so `.` in L12 points to the root template data rather than the annotations object

ANd this can be slimmed down to one line as https://github.com/actions-runner-controller/actions-runner-controller/pull/824/files#r710589956

charts/actions-runner-controller/ns.json
```diff
@@ -0,0 +1,23 @@
+{
```
mumoshu, line 1: Is this file necessary? 🤔

charts/actions-runner-controller/README.md
```diff
@@ -44,6 +44,7 @@ _Default values are the defaults set in the charts values.yaml, some properties
 | `serviceAccount.name`                                    | Set the name of the service account                                                                                        |                                                                      |
 | `securityContext`                                        | Set the security context for each container in the controller pod                                                          |                                                                      |
 | `podSecurityContext`                                     | Set the security context to controller pod                                                                                 |                                                                      |
+| `secretAnnotations`                                           | Set `controller-manager` secret additional annotations                                                                                 |
```
mumoshu, line 47: Did you mean `authSecret.annotations`?

charts/actions-runner-controller/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.12.8
+version: 0.12.9
```
mumoshu, line 18: Could you revert this change so that we can control when to cut a new chart release?

charts/actions-runner-controller/templates/ci-secret.yaml
maxtacu: updated. but need to test this one. will do it tomorrow

charts/actions-runner-controller/templates/manager_secrets.yaml
maxtacu: same here about testing. check please. I will test it later or let me know if you tested the deployment on your side

PR #828, opened 2021-09-16
Fix: Have githubwebhook monitor a single namespace

sledigabel:
When using `scope.singleNamespace: true` in Helm, expected behaviour is
that the github webhook server behaves the same way as the controller.

The current behaviour is that the webhook server monitors all the
namespaces.

### DIFF
--- a/charts/actions-runner-controller/templates/githubwebhook.deployment.yaml
+++ b/charts/actions-runner-controller/templates/githubwebhook.deployment.yaml
@@ -42,6 +42,9 @@ spec:
         {{- if .Values.githubWebhookServer.logLevel }}
         - "--log-level={{ .Values.githubWebhookServer.logLevel }}"
         {{- end }}
+        {{- if .Values.scope.singleNamespace }}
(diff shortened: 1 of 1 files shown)


toast-gear:
Could you update the docs as well please https://github.com/actions-runner-controller/actions-runner-controller/blob/master/charts/actions-runner-controller/README.md:

1. The docs need updating as these settings talk about impacting the controller only, this change will mean it impacts the controller and webhook server
2. Probably worth amending `scope.watchNamespace` to say that if this is not provided the release namespace is used by default

sledigabel:
Good shout @callum-tait-pbx, didn't notice that in the doc.

sledigabel:
@mumoshu not sure if you got the chance to review this?

toast-gear:
hi @sledigabel, this PR isn't lost, we want to resolve some outstanding issues from the most recent release before merging in more features

sledigabel:
@callum-tait-pbx thanks for the quick reply!

PR #834, opened 2021-09-22
Support installing without cert-manager

### LINKED ISSUE #531: Can we install this without cert-manager
We already have our self-signed certificates or ou certifications. Is it possible to use them instead of cert manager?

jsok:
Enables users to supply their own certificate for the admission
webhooks.

Closes #531


toast-gear:
@jsok `openssl` is probably better given the project implementation is fairly technology agnostic, a link will be fine

jsok:
I've fixed some conflicts and added a link to what seems like a trustworthy and also stable set of documentation.
Realistically I think anyone who is going to be doing this already has a mechanism or their own googling skills to work it out 🤷 .

PR #837, opened 2021-09-22
Correct test to append docker container

### LINKED ISSUE #835: Upgrade v0.20.0 and dockerenable = false
Deployed the new v0.20.0 version but if I add the flag dockerEnabled: false it fails with the error in the controller log:

`Reconciler error	{"reconciler group": "actions.summerwind.dev", "reconciler kind": "Runner", "name": "github-runner-bn4xw-55gx9", "namespace": "github-runner", "error": "Pod \"github-runner-bn4xw-55gx9\" is invalid: spec.containers[1].image: Required value"}`

We use the following example deployment:

```
kind: RunnerDeployment
metadata:
  name: github-runner
  namespace: github-runner
spec:
  replicas: 1
  template:
    spec:
      organization: org-IT
      ephemeral: true
      dockerEnabled: false
      resources:
        limits:
          memory: "2Gi"
        requests:
          memory: "1Gi"
      image: privateacr.azurecr.io/github/actions-runner:v2.283.1-ubuntu-20.04
      env:
      - name: NODE_EXTRA_CA_CERTS
        value: /usr/local/share/ca-certificates/rootca.pem
```

If I revert back to 0.19.0 it works as expected (no dind is injected)

tyrken:
This updates the do-we-append-a-docker-container decision to more like elsewhere, e.g. https://github.com/actions-runner-controller/actions-runner-controller/blob/24602ff3ee05303e019b395b5d6805f2ebedaa14/controllers/runner_controller.go#L1040

This appears to fix #835

### DIFF
--- a/controllers/runner_controller.go
+++ b/controllers/runner_controller.go
@@ -619,7 +619,7 @@ func (r *RunnerReconciler) newPod(runner v1alpha1.Runner) (corev1.Pod, error) {
 			Resources:       runner.Spec.Resources,
 		})
 
-		if runner.Spec.DockerdWithinRunnerContainer == nil || !*runner.Spec.DockerdWithinRunnerContainer {
+		if (runner.Spec.DockerEnabled == nil || *runner.Spec.DockerEnabled) && (runner.Spec.DockerdWithinRunnerContainer == nil || !*runner.Spec.DockerdWithinRunnerContainer) {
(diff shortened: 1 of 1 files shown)


yuri-1987:
I can confirm that this fix solved the issue I have complained about in #801 
thanks

mumoshu (approved):
This was totally my mistake! Thanks a lot for spotting it and your patience ☺️

Issue #823, opened 2021-09-15
Unable to apply actions-runner-controller.yaml v0.20.0 


pje:
**Describe the bug**

When I run this:

```sh
$ kubectl apply -f https://github.com/actions-runner-controller/actions-runner-controller/releases/download/v0.20.0/actions-runner-controller.yaml
```

I get this error:

```
namespace/actions-runner-system unchanged
customresourcedefinition.apiextensions.k8s.io/horizontalrunnerautoscalers.actions.summerwind.dev configured
customresourcedefinition.apiextensions.k8s.io/runnersets.actions.summerwind.dev configured
role.rbac.authorization.k8s.io/leader-election-role unchanged
clusterrole.rbac.authorization.k8s.io/manager-role configured
clusterrole.rbac.authorization.k8s.io/proxy-role unchanged
rolebinding.rbac.authorization.k8s.io/leader-election-rolebinding unchanged
clusterrolebinding.rbac.authorization.k8s.io/manager-rolebinding unchanged
clusterrolebinding.rbac.authorization.k8s.io/proxy-rolebinding unchanged
service/controller-manager-metrics-service unchanged
service/webhook-service unchanged
deployment.apps/controller-manager unchanged
certificate.cert-manager.io/serving-cert unchanged
issuer.cert-manager.io/selfsigned-issuer unchanged
mutatingwebhookconfiguration.admissionregistration.k8s.io/mutating-webhook-configuration configured
validatingwebhookconfiguration.admissionregistration.k8s.io/validating-webhook-configuration configured
Error from server (Invalid): error when applying patch:
...[truncated]...
to:
Resource: "apiextensions.k8s.io/v1, Resource=customresourcedefinitions", GroupVersionKind: "apiextensions.k8s.io/v1, Kind=CustomResourceDefinition"
Name: "runners.actions.summerwind.dev", Namespace: ""
for: "https://github.com/actions-runner-controller/actions-runner-controller/releases/download/v0.20.0/actions-runner-controller.yaml": CustomResourceDefinition.apiextensions.k8s.io "runners.actions.summerwind.dev" is invalid: spec.preserveUnknownFields: Invalid value: true: must be false in order to use defaults in the schema
```

**Checks**

- [x] My actions-runner-controller version (v0.18.2) does support the feature
- [ ] I'm using an unreleased version of the controller I built from HEAD of the default branch

**To Reproduce**

[...]

PR #843, opened 2021-09-23
Add preserveUnknown=false to crds

tyrken:
This `yq` hack forces the preserveUnknown value to false, as proposed in #823.

When upgrading from v0.19.0 or earlier with it's `apiextensions.k8s.io/v1beta1` type CRDs, the field defaults to `true`, but for `apiextensions.k8s.io/v1` it now defaults to `false`.  So if left unset upgraders will get it `true` and hit the error when combined with the defaults now in the CRD, whereas fresh installs will start with it `false`.

Whilst I could have gone the other way, back to `true` with an additional `yq` invocation to remove the defaults, I'm following what feels a more strict approach of fully defining the CRDs.  Hopefully everything still works...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,8 +86,11 @@ deploy: manifests
 # Generate manifests e.g. CRD, RBAC etc.
 manifests: manifests-gen-crds chart-crds
 
-manifests-gen-crds: controller-gen
+manifests-gen-crds: controller-gen yq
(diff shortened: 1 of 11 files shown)


mumoshu (approved):
LGTM. Thanks a lot for your contribution @tyrken ☺️

PR #844, opened 2021-09-24
docs: updating to reflect the latest release

toast-gear:


README.md
```diff
@@ -569,7 +563,8 @@ GitHub sending PING events to the Webhook server - create or update your `Horizo
 by learning the following configuration examples.
 
 - [Example 1: Scale up on each `check_run` event](#example-1-scale-up-on-each-check_run-event)
-- [Example 2: Scale on each `pull_request` event against `develop` or `main` branches](#example-2-scale-on-each-pull_request-event-against-develop-or-main-branches)
+- [Example 2: Scale on each `pull_request` event against a given set of branches](#example-2-scale-on-each-pull_request-event-against-a-given-set-of-branches)
+- [Example 3: Scale on each `workflow_job` event](#example-3-scale-on-each-workflow_job-event)
```
mumoshu, line 567: Maybe we'd better recommend using `workflow_job` first, because I believe it should work in every use-case in practice. But not sure where and how to recommend it. Perhaps write some words of recommendation around L377? 🤔

toast-gear: re-ordered this so workflow_job is first as well as suggesting this is the recommended setup, I renamed the section as this is now the recommended approach. The docs need further work in the scaling sections tbh now but this will be fine for now.

Issue #809, opened 2021-09-13
Helm chart: update template for github_app_installation_id


aliusmiles:
**Is your feature request related to a problem? Please describe.**
I was deploying controller using helm chart and GitHub App and wrote github_app_installation_id in values.yaml without quotes. Got error message:

> Error: Client creation failed. authentication failed: using private key of size 0 (...): could not parse private key: Invalid Key: Key must be PEM encoded PKCS1 or PKCS8 private key

After some happy debugging time I've realized that helm chart produces incorrect id value in resulting secret: number gets converted to non integer format. As soon as I wrap github_app_installation_id in quotes in values.yaml - everything works.

Here's how to reproduce it:

```yaml
# chart/templates/cm.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
data:
  test: {{ .Values.test | toString }}
  test: {{ .Values.test | int | toString }}
  
# chart/values.yaml
test: 12345678
```

```sh
➜  helm template test chart
---
# Source: chart/templates/cm.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
data:
  test: 1.2345678e+07
  test: 12345678
```

**Describe the solution you'd like**
It can be fixed by:

a) using quotes as in commented out example in [values.yaml](https://github.com/actions-runner-controller/actions-runner-controller/blob/a98729b08b86aac59613769a592cf71cf3523c90/charts/actions-runner-controller/values.yaml#L26) 🙄
b) updating chart's [template](https://github.com/actions-runner-controller/actions-runner-controller/blob/a98729b08b86aac59613769a592cf71cf3523c90/charts/actions-runner-controller/templates/manager_secrets.yaml#L15)
```yaml
-  github_app_installation_id: {{ .Values.authSecret.github_app_installation_id | toString | b64enc }}
+  github_app_installation_id: {{ .Values.authSecret.github_app_installation_id | int | toString | b64enc }}
```
c) or maybe just add a comment like "quotes matters!" in values.yaml for a specially observant persons like me 🤷🏻‍♂️

Hope that at least this issue can save someone's time..

**Additional context**
chart: actions-runner-controller-0.12.8 
helm: v3.6.3

PR #869, opened 2021-10-04
chore: support app ids as int or strings

toast-gear:
Fixes https://github.com/actions-runner-controller/actions-runner-controller/issues/809


mumoshu (approved):
I don't even know those IDs are specified to be ints by GitHub, but yeah this may be worth it for our users...

PR #876, opened 2021-10-07
allow providing webhook secret token via flag instead of environment variable

Lazyshot:
Currently the webhook server only allows for providing the github token via environment variable. Our infrastructure only allows for us to provide secrets via either files or by extension argument flags.

### DIFF
--- a/cmd/githubwebhookserver/main.go
+++ b/cmd/githubwebhookserver/main.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"flag"
+	"fmt"
 	"net/http"
 	"os"
 	"sync"
@@ -48,6 +49,8 @@ const (
 	logLevelInfo  = "info"
 	logLevelWarn  = "warn"
 	logLevelError = "error"
+
+	webhookSecretTokenEnvName = "GITHUB_WEBHOOK_SECRET_TOKEN"
 )
 
 func init() {
@@ -65,7 +68,8 @@ func main() {
 		metricsAddr string
 
 		// The secret token of the GitHub Webhook. See https://docs.github.com/en/developers/webhooks-and-events/securing-your-webhooks
-		webhookSecretToken string
+		webhookSecretToken    string
+		webhookSecretTokenEnv string
 
 		watchNamespace string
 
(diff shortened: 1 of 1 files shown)


mumoshu:
> Our infrastructure only allows for us to provide secrets via either files or by extension argument flags

You can set a container envvar from a K8s secret. Wouldn't that be a solution?
I can hardly think about why and how your infrastructure can force you to do it in less secure way.

Lazyshot:
Secrets at our org is currently provided by an init container which pulls them from vault and passes them to the container on a volume (as opposed to a K8s SealedSecret or similar solution).

This behavior is also consistent with the controller.

mumoshu:
@Lazyshot Thanks for clarifying! How do you run actions-runner-controller? Perhaps you have some script to trigger `exec actions-runner-controller $ARGS`?

Then how about doing `SOME_TOKEN=$(cat path/to/the/secret/file) githubwebhookserver`?

> This proposed behavior is also consistent with the controller.

Yes, that's good point. I'm just confused. If you're going to add this for consistency only, that'd have made sense to me. But it seemed to me that you had to do this for another reason- that's where I was confused.

Lazyshot:
That's a fair suggestion. 

And nope, it's a relatively simple consistency thing between the two binaries.

cmd/githubwebhookserver/main.go
```diff
@@ -83,6 +83,7 @@ func main() {
 		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
 	flag.DurationVar(&syncPeriod, "sync-period", 10*time.Minute, "Determines the minimum frequency at which K8s resources managed by this controller are reconciled. When you use autoscaling, set to a lower value like 10 minute, because this corresponds to the minimum time to react on demand change")
 	flag.StringVar(&logLevel, "log-level", logLevelDebug, `The verbosity of the logging. Valid values are "debug", "info", "warn", "error". Defaults to "debug".`)
+	flag.StringVar(&webhookSecretToken, "github-webhook-secret-token", webhookSecretToken, "The personal access token of GitHub.")
```
mumoshu, line 86: I think this change can be exploited to let the webhook server dump the value of `GITHUB_WEBHOOK_SECRET_TOKEN` when run with `-h`.

I have no concrete attack scenario that I can come up with it, but it won't be that bad to make it secure by default!

Let me add a commit to tweak this, so that we can use either env or flag to set the token, without the possibility to leak the token to stdout.

mumoshu: Addressed in 7abcd48

mumoshu (approved):
Merging. Thanks a lot for your contribution @Lazyshot ☺️

Issue #882, opened 2021-10-12
PR-869 breaks secrets management techniques 


apr-1985:
**Describe the bug**
Since https://github.com/actions-runner-controller/actions-runner-controller/pull/869 App and Installation IDs can no longer be provided as strings.
This breaks our secrets management solution as it requires a string which is converted to the secret at run time.

It also goes against the commit message of the PR to `support app ids as int or strings` as it now does implicit conversion to `int` then implicit conversion to `string`.  So not really sure of the intention behind the change....

**Checks**

- [x] My actions-runner-controller version (v0.x.y) does support the feature
- [x] I'm using an unreleased version of the controller I built from HEAD of the default branch

**To Reproduce**
1. Add string as the value for .Values.authSecret.github_app_id
2. Install the Helm Chart
3. Check the secret has `MA==` as a value

**Expected behavior**
Allow App IDs and Installation IDs to be provided as strings for use with Secrets management solutions.

**Screenshots**

**Environment (please complete the following information):**
 - Controller Version [e.g. 0.18.2] HEAD
 - Deployment Method [e.g. Helm and Kustomize ] Helm v3.7.0
 - Helm Chart Version [e.g. 0.11.0, if applicable] HEAD

**Additional context**

PR #883, opened 2021-10-12
Revert "chore: support app ids as int or strings (#869)"

apr-1985:
This reverts commit 0a3d2b686e72ec78d578671bf5305d1183701c7b.
Fixes https://github.com/actions-runner-controller/actions-runner-controller/issues/882


toast-gear:
@aliusmiles we're going to revert the change to support providing the IDs as either strings or ints. Whilst that is a nicer user experience, it breaks some secret managment integrations. Given there isn't a problem with them being kept as strings beyond it being a bit superficially unintuitive it's better we roll back.

aliusmiles:
@toast-gear Well, I'm already using quotes there anyway. But I think with a new comment it should be more visible that ids must be strings, and if someone still makes the same mistake as I did -- now there are even couple of issues to search for :)

@apr-1985 I still believe that I'm right in my conclusion that github_app_installation_id won't work if passed as int, but only if it's more then 6 symbol long:

 I've tested chart v0.13.2 with following:

```yaml
authSecret:
  github_app_id: 111111 #literally '111111'
  github_app_installation_id: 111111 #same here
  github_app_private_key: "ACTUAL-KEY-CONTENT"
```
and with
```yaml
authSecret:
  github_app_id: 111111 #literally '111111'
  github_app_installation_id: 11111111 # 8 symbols instead of 6
  github_app_private_key: "ACTUAL-KEY-CONTENT"
```

First one (with 6 symbols app_installation_id) resulted in: `"error": "failed to list runners: Get \"https://api.github.com/orgs/REDUCTED/actions/runners?per_page=100\": could not refresh installation id 111111's token:`

Second one still produces: `Error: Client creation failed. authentication failed: using private key of size 0 (...): could not parse private key: Invalid Key: Key must be PEM encoded PKCS1 or PKCS8 private key` and I believe that's because of controller-manager secret containing value `1.1111111e+07` as a result of helm's int to string conversion. 

Let me guess, your github_app_installation_id is 6 symbol long and that's why int works for you? :)

To be honest, that error is quite misleading as it's complaining on invalid key while the issue is with installation_id not a key itself and I've spend at least one hour trying to debug why key was incorrect :) so that's why I've created that initial issue - trying to prevent this from happening to someone else. Now we have a comment in values and couple of issues on that so mission accomplished I guess :) I don't think it's worth an effort to change error handling in that particular case.

P.S. @apr-1985 Would you mind sharing name of secrets management tool you are using, or it's a custom in-house solution? Haven't heard of such before - all others are using CRDs to create secrets I believe, and your's seem to update secrets in-place?

PR #884, opened 2021-10-12
fix: allow GH priv key from env in helm chart

apr-1985:
Allowing the ability to get the GH priv key from the env when using the helm chart.
Feature Code added in https://github.com/actions-runner-controller/actions-runner-controller/pull/383

I have tested switching between Env and file by toggling the new helm chart variable and this works without issue.

This feature increases the interop with secrets managers that inject to ENV at run time.


apr-1985:
Is there anything else I need to do get this reviewed please?

marchenkov:
I have the same issue - we are using vault and all secrets must be passed as ENV (not as file `/etc/actions-runner-controller/github_app_private_key`).
Is there any chance to review/merge this PR?
Thanks

mumoshu:
We're going to review and merge this before our next controller minor release! Thanks for your patience.

mumoshu:
@marchenkov Hey! How does this PR solve your issue?

> we are using vault and all secrets must be passed as ENV

If you're using [vault-k8s' injector](https://github.com/hashicorp/vault-k8s) you should get files instead of envvars.
How are you integrating vault with your setup?

apr-1985:
> We're going to review and merge this before our next controller minor release! Thanks for your patience.

Many thanks mainly commented to stop the stale bot 😀

marchenkov:
> @marchenkov Hey! How does this PR solve your issue?
> 
> > we are using vault and all secrets must be passed as ENV
> 
> If you're using [vault-k8s' injector](https://github.com/hashicorp/vault-k8s) you should get files instead of envvars. How are you integrating vault with actions-runner-controller in your setup?

I'm use vault-secrets-webhook and no way to inject to file.
https://banzaicloud.com/docs/bank-vaults/mutating-webhook/configuration/
So this PR can solve my problem, because I can read PEM from secret (now I receive error that file `/etc/actions-runner-controller/github_app_private_key` not contain PEM cert).

mumoshu:
@marchenkov I was also wondering if this hard-coded env name and secretKeyRef values might affect your use-case

```
key: github_app_private_key
name: {{ include "actions-runner-controller.secretName" . }}
```

But after reading how bank-vaults mutating webhook works, perhaps that's also no issue.

Thanks for confirming!

charts/actions-runner-controller/templates/deployment.yaml
```diff
@@ -84,7 +84,15 @@ spec:
...
         - name: GITHUB_APP_PRIVATE_KEY
+          {{- if .Values.authSecret.github_private_key_from_disk }}
           value: /etc/actions-runner-controller/github_app_private_key
+          {{- else }}
+          valueFrom:
+            secretKeyRef:
+              key: github_app_private_key
+              name: {{ include "actions-runner-controller.secretName" . }}
+              optional: true
+          {{- end }}
```
mumoshu, line 95: @apr-1985 @marchenkov @toast-gear Can we just do this instead, without the extra flag?

```suggestion
          valueFrom:
            secretKeyRef:
              key: github_app_private_key
              name: {{ include "actions-runner-controller.secretName" . }}
              optional: true
```

Regardless of we use the local file or the secretKeyRef, the content of the private key is loaded from `controller-manager` secret anyway. Then, this won't be a breaking change, right?

mumoshu:
I'm now wondering if `authSecret.enabled=false` added via #937 with the existing `env` helm value can be an alternative to your use-case @apr-1985 @marchenkov WDYT?

charts/actions-runner-controller/templates/deployment.yaml
marchenkov: @mumoshu, I can confirm that your variant works for me too. My main problem - that file `/etc/actions-runner-controller/github_app_private_key` not found. So both options are valid for me, but your proposal the better. Thanks

mumoshu: Thanks for your confirmation! Let's go ahead without the if block then.

charts/actions-runner-controller/values.yaml
```diff
@@ -23,10 +23,12 @@ enableLeaderElection: true
 
 # Only 1 authentication method can be deployed at a time
 # Uncomment the configuration you are applying and fill in the details
+# Setting github_private_key_from_disk to false will load the key from ENV
 authSecret:
   enabled: true
   create: false
   name: "controller-manager"
+  github_private_key_from_disk: true
```
mumoshu, line 31:
```suggestion
```

charts/actions-runner-controller/values.yaml
```diff
@@ -23,10 +23,12 @@ enableLeaderElection: true
 
 # Only 1 authentication method can be deployed at a time
 # Uncomment the configuration you are applying and fill in the details
+# Setting github_private_key_from_disk to false will load the key from ENV
```
mumoshu, line 26:
```suggestion
#
# If authSecret.enabled=true these values are inherited to actions-runner-controller's controller-manager container's env.
#
# Do set authSecret.enabled=false and set env if you want full control over
# the GitHub authn related envvars of the container.
# See https://github.com/actions-runner-controller/actions-runner-controller/pull/937 for more details.
```

mumoshu (approved):
Merging. Thanks for your contribution and support @apr-1985 @marchenkov ☺️

PR #887, opened 2021-10-12
fix: cast so int and string supported

toast-gear:
**Context** https://github.com/actions-runner-controller/actions-runner-controller/pull/883
**Fixes** https://github.com/actions-runner-controller/actions-runner-controller/issues/882

@apr-1985 how do you feel about this as a compromise? Having gone through the Helm docs and tested this PR it seems to work fine. I was able to deploy a controller onto a kind cluster with my auth config set as either of the below:

```yaml
github_app_id: "1111111"
github_app_installation_id: "1111111"
github_app_private_key: |
  -----BEGIN RSA PRIVATE KEY-----
  ......
```

```yaml
github_app_id: 1111111
github_app_installation_id: 1111111
github_app_private_key: |
  -----BEGIN RSA PRIVATE KEY-----
  ......
```

### DIFF
--- a/charts/actions-runner-controller/templates/manager_secrets.yaml
+++ b/charts/actions-runner-controller/templates/manager_secrets.yaml
@@ -8,11 +8,13 @@ metadata:
     {{- include "actions-runner-controller.labels" . | nindent 4 }}
 type: Opaque
 data:
+# cast attribute like this so it can be provided as an int or string in values.yaml
 {{- if .Values.authSecret.github_app_id }}
-  github_app_id: {{ .Values.authSecret.github_app_id | int | toString | b64enc }}
+  github_app_id: {{ int .Values.authSecret.github_app_id | toString | b64enc }}
 {{- end }}
+# cast attribute like this so it can be provided as an int or string in values.yaml
 {{- if .Values.authSecret.github_app_installation_id }}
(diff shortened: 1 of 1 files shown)


apr-1985:
For the secrets management solution that we are using, basically you put in a path to an AWS parameter (string) and this is then converted to the value when it is loaded into the environment.
Therefore implicit int casting wouldn't work in this situation as the parameter store path will try to be cast to an int resulting in an incorrect value.  `MA==`

This situation is documented in the Helm docs https://helm.sh/docs/chart_best_practices/values/#make-types-clear
`Large integers like foo: 12345678 will get converted to scientific notation in some cases`
And their advice is.
```
Often, to avoid the integer casting issues, it is advantageous to store your integers as strings as well, and use {{ int $value }} in the template to convert from a string back to an integer.
```
Which to me means make the vars strings (as the current values.yaml has them).  Especially as the `b64enc` requires a string anyway, it makes more sense to me to have the variable as the type that is being encoded rather than casting through multiple types in the encoding line.

However I can totally see why you would want them as numbers in the values.yml for clarity reasons.  So I will leave it to the maintainers of the module to decide which way to take the chart.

toast-gear:
Closing in favour of reverting to the previous setup of having the ID as as strings https://github.com/actions-runner-controller/actions-runner-controller/pull/883

PR #894, opened 2021-10-15
Reduce number of http superfluous messages

zotrix:
write to http.ResponseWriter create HTTP OK response, so set *ok* to true will disable calling error code in defered function


mumoshu (approved):
Ah, good catch! Thanks a lot for fixing it ☺️

Issue #829, opened 2021-09-17
v0.20 Problem with authentication on controller-manager


grzegorzkawkaosik:
**Describe the bug**
It seems that controller-manager deployment from yaml is not using environment variables correctly.

**Checks**

- [ ] My actions-runner-controller version (v0.x.y) does support the feature
- [ ] I'm using an unreleased version of the controller I built from HEAD of the default branch

**To Reproduce**
1. deploy actions-runner-controller v0.20.0 from [release page](https://github.com/actions-runner-controller/actions-runner-controller/releases/download/v0.20.0/actions-runner-controller.yaml)
1. deploy secret with github app key and ids
1. Try deploying any runner
1. Get error in controller-manager saying that key file has length 0

**Expected behavior**
Github runner should be registered

**Environment (please complete the following information):**
 - Controller Version: 0.20.0
 - Deployment Method: `kubectl apply`

**Additional context**
I have created a workaround by manually providing required arguments to /manager binary, see below:

```
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    control-plane: controller-manager
  name: controller-manager
  namespace: actions-runner-system
spec:
  replicas: 1
  selector:
    matchLabels:
      control-plane: controller-manager
  template:
    metadata:
      labels:
        control-plane: controller-manager
    spec:
      containers:
      - name: manager
        command:
        - "/manager" 
        args:
        - --metrics-addr=127.0.0.1:8080
        - --enable-leader-election
        - --sync-period=10m
        - "-github-app-private-key" 
        - $(GITHUB_APP_PRIVATE_KEY)
        - "-github-app-id"
        - $(GITHUB_APP_ID)
        - "-github-app-installation-id"
        - $(GITHUB_APP_INSTALLATION_ID)
        env:
        - name: GITHUB_TOKEN
          valueFrom:
            secretKeyRef:
              key: github_token
              name: controller-manager
              optional: true
        - name: GITHUB_APP_ID
          valueFrom:
            secretKeyRef:
              key: github_app_id
              name: controller-manager
              optional: true
        - name: GITHUB_APP_INSTALLATION_ID
          valueFrom:
            secretKeyRef:
              key: github_app_installation_id
              name: controller-manager
              optional: true
        - name: GITHUB_APP_PRIVATE_KEY
          value: /etc/actions-runner-controller/github_app_private_key
        image: summerwind/actions-runner-controller:v0.20.0
        ports:
[...]

PR #907, opened 2021-10-24
Fail with detailed message on envvar parse error

mumoshu:
Currently, it's hard for users to notice envvar parse error(s). This fixes that, by adding a more detailed message on such error and existing with status 1 instead of just continuing the application startup process with possibly broken configuration.

Ref #829


toast-gear:
Fixes: https://github.com/actions-runner-controller/actions-runner-controller/issues/829

PR #915, opened 2021-10-27
Add WAIT_FOR_DOCKER env var to entrypoint script

### LINKED ISSUE #897: Enhance runner entrypoint to wait for docker
**Feature**
Enhance the runner entrypoint to wait for docker sidecar or docker daemon (dind image) to be up and running before starting the runner service

This will reduce the need for tweaks to the `STARTUP_DELAY_IN_SECONDS` env var

**Context**
https://github.com/actions-runner-controller/actions-runner-controller/issues/787

artificial-aidan:
Closes #897 

Signed-off-by: Aidan Jensen <aidan@artificial.com>


README.md
```diff
@@ -6,32 +6,40 @@ This controller operates self-hosted runners for GitHub Actions on your Kubernet
...
+    - [Runner Groups](#runner-groups)
+    - [Using IRSA (IAM Roles for Service Accounts) in EKS](#using-irsa-iam-roles-for-service-accounts-in-eks)
+    - [Use with Istio](#use-with-istio)
+    - [Stateful Runners](#stateful-runners)
+    - [Ephemeral Runners](#ephemeral-runners)
+    - [Software Installed in the Runner Image](#software-installed-in-the-runner-image)
+    - [Common Errors](#common-errors)
+      - [invalid header field value](#invalid-header-field-value)
+      - [Runner coming up before network available](#runner-coming-up-before-network-available)
+      - [Runner coming up before dind container is ready](#runner-coming-up-before-dind-container-is-ready)
```
toast-gear, line 42: Remove this structure change please, headings after a certain point become too small and so we don't want a root heading they all sit under

runner/entrypoint.sh
```diff
@@ -24,6 +24,11 @@ if [ ! -z "${STARTUP_DELAY_IN_SECONDS}" ]; then
   sleep ${STARTUP_DELAY_IN_SECONDS}
 fi
 
+if [ -n "${WAIT_FOR_DOCKER}" ]; then
+  echo "Waiting for docker to be available"
+  until docker info; do sleep 1; done
+  echo "Docker is now available"
+fi
```
toast-gear, line 31: We would like this to be mroe automatic so the end user doesn't need to set this setting, it should be automatic if the runner has docker enabled. 

If you add to the below to [here](https://github.com/actions-runner-controller/actions-runner-controller/blob/master/controllers/runner_controller.go#L777) you can check if docker is enabled in the entrypoint. If it is then we should wait for docker to be avaliable, if it isn't then we shouldn't.

add this:

```go
		{
			Name:  "DOCKER_ENABLED",
			Value: fmt.Sprintf("%v", dockerEnabled),
		},
```

before this:

```go
		{
			Name:  "DOCKERD_IN_RUNNER",
			Value: fmt.Sprintf("%v", dockerdInRunner),
		},

```

README.md
artificial-aidan: Will do. I think my formatter did it by default.

runner/entrypoint.sh
artificial-aidan: Do we want an option to opt out? Or just an always on option.

Why does a new controller need to be built and tested?

artificial-aidan: I'll see if I have time for that change this week. I was able to get this quick change with some free time last week, but I have less of that.

toast-gear: > Do we want an option to opt out? Or just an always on option.

To be honest I'm not why you would want to opt out of preventing a race condition. @mumoshu if you have any thoughts.

mumoshu: I think the default `WAIT_FOR_DOCKER` value should be dependent on `DOCKER_ENABLED`(=dockerEnabled from the controller) and I do also think an ability to opt-out `WAIT_FOR_DOCKER` would be a nice-to-have.

We don't yet have a controller functionality to auto-detect the version of GHES, so opting-out WAIT_FOR_DOCKER would be the only way to keep actions-runner-controller working on old versions of GHES after this change.

artificial-aidan:
Still wip

runner/entrypoint.sh
genisd: As for the opt-out, we could use "0" as the opt-out, and any other int we could interpret as yes we should wait for docker with the provided value beeing the upper limit (timeout) in seconds before we terminate from the startup script.

We could really use this as well, currently running with a fixed startup delay to account for docker (but just this morning some jobs failed because docker wasn't up in time). This would be a great addition/improvment

mumoshu:
I think this can be merged into #962 - thoughts? @toast-gear 

// Regardless of this PR being merged as-is or not, the credit for starting this enhancement should go to the original contributor @artificial-aidan so I'll definitely write a big thanks note to you in the changelog!

artificial-aidan:
Sounds good to me. Glad somebody took care of the enhancements, I got swamped with work, and had to use the original solution as is.

Issue #896, opened 2021-10-15
Support default imagePullSecrets in runner pod for private docker registries


fgalind1:
**Is your feature request related to a problem?.**
When using a private docker registry it is required to provide the docker image as a `imagePullSecrets` in the runner pod spec. Runner pod spec supports imagePullSecrets, however it's quite useful to support this at the controller level with a default value similarly as it's done for runner image

**Describe the solution you'd like**
In the same way as a cmdline argument is supported to provide `--runner-image`, I'd like to have an option `--runner-image-pull-secret-name` or something similar that contains the secret name that holds the docker config and then use that to add the `imagePullSecrets` in the runner pod with something like:

```
	if defaultRunnerImagePullSecretName != "" {
		pod.Spec.ImagePullSecrets = []corev1.LocalObjectReference{
			{Name: defaultRunnerImagePullSecretName},
		}
	}
```

**Describe alternatives you've considered**
Define the imagePullSecrets for every single **RunnerDeployment** which could be repetitive and prune to errors

**Additional context**
N/A

fgalind1:
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions.

waiting for review of PR https://github.com/actions-runner-controller/actions-runner-controller/pull/921

PR #921, opened 2021-10-28
Add support for default image pull secret name

fgalind1:
Currently it's supported to define `ImagePullSecrets` in the runner spec e.g as a `RunnerDeployment`. However the controller can specify a default runner image.

If this image is on a private registry it requires the image pull secret. This PR adds this option also to the controller to be consistent with supporting a default runner image. This makes it much
easier when using runner images from private registries

Fixes https://github.com/actions-runner-controller/actions-runner-controller/issues/896


fgalind1:
still needed

controllers/runner_controller.go
```diff
@@ -872,6 +873,12 @@ func newRunnerPod(template corev1.Pod, runnerSpec v1alpha1.RunnerConfig, default
 		}...)
 	}
 
+	if defaultRunnerImagePullSecretName != "" {
+		pod.Spec.ImagePullSecrets = []corev1.LocalObjectReference{
```
mumoshu, line 877: I think this should be effective only when `pod.Spec.ImagePullSecrets` is empty(Not sure about RunnerDeployment but at least RunnerSet allows you to specify it via runnerset.spec.template.imagePulSecrets` so I'd like to not break it :)

mumoshu:
@fgalind1 Sorry for the delay and thanks for your contribution! I'm going to review/merge this towards our next feature release(0.21.0)

controllers/runner_controller.go
fgalind1: sounds reasonable 👍  updated to be set only if it's not already specified via spec - thx

charts/actions-runner-controller/templates/deployment.yaml
```diff
@@ -46,6 +46,9 @@ spec:
         - "--sync-period={{ .Values.syncPeriod }}"
         - "--docker-image={{ .Values.image.dindSidecarRepositoryAndTag }}"
         - "--runner-image={{ .Values.image.actionsRunnerRepositoryAndTag }}"
+        {{- if .Values.image.actionsRunnerPullImageSecretName }}
+        - "--runner-image-pull-secret-name={{ .Values.image.actionsRunnerPullImageSecretName }}"
```
mumoshu, line 50: I'm a bit confused but you didn't actually define this command-line in the actions-runner-controller command yet?

charts/actions-runner-controller/values.yaml
```diff
@@ -39,6 +39,19 @@ image:
   actionsRunnerRepositoryAndTag: "summerwind/actions-runner:latest"
   dindSidecarRepositoryAndTag: "docker:dind"
   pullPolicy: IfNotPresent
+  # Defines secret to use as ImagePullSecrets in the runner pod spec
+  # actionsRunnerPullImageSecretName: someSecretName
+
+
+imagePullSecrets:
+- name: github-actions-docker-config
```
mumoshu, line 47: Apparently, this is used from nowhere?

```suggestion
```

charts/actions-runner-controller/templates/deployment.yaml
mumoshu: Addressed in 1f6facb775899f0b878b1c3a2e7f34fffe508242

charts/actions-runner-controller/values.yaml
mumoshu: Well, it turns out this already exists in L57 and is used to set image pull secrets for controller-manager, not runners 😄

charts/actions-runner-controller/values.yaml
```diff
@@ -39,6 +39,20 @@ image:
...
+
+# Optionally create a secret with the dockerconfigjson used for the runner pod spec
+# imagePullSecretsDockerConfig:
+#   secretName: someSecretName
+#   registry: somePrivateRegistry
+#   username: someUsername
+#   password: somePassword
```
mumoshu, line 55: Can we make this out of the scope of this PR or this chart, or can we make it more generalized like `serviceAccount`?

I mean, there are might be many image pull secrets you'd want to manage as part of your actions-runner-controller instance. Then you'd want ability to  define one or more image pull secrets like this:

```suggestion
# Optionally create a secret with the dockerconfigjson used for the runner pod spec
imagePullSecrets:
- name: blah
  type: kubernetes.io/dockerconfigjson
  stringDataTemplate: |
    .dockerconfigjson: |-
      {
      "auths": {
        "{{ .Values.imagePullSecretsDockerConfig.registry }}": {
            "username": "{{ .Values.imagePullSecretsDockerConfig.username }}",
            "password": "{{ .Values.imagePullSecretsDockerConfig.password }}"
          }
      }
    }
```

In the manager secret template we'd do this with the `tpl` function so that you can freely hard-code all the JSON values within the values.yaml or you can render the dockerconfigjson with values loaded via e.g. helm-secrets.

```
{{ range $_, $c := .Values.imagePullSecrets }}
---
apiVersion: v1
kind: Secret
type: {{ $c.type }}
metadata:
  name: {{ $c.name }}
stringData:
  {{ tpl $c.stringDataTemplate $ }}
{{ end }}
```

fgalind1: Thanks for the feedback and review @mumoshu. I've improved it on 2d62544e1a043834affc9bac91ff621924a2f013 and made it a bit more generic [here](https://github.com/actions-runner-controller/actions-runner-controller/pull/921/commits/2d62544e1a043834affc9bac91ff621924a2f013#diff-3dfd28f4387ada0d16b7e7cdc2efd38b3724c82f6821b4f0fa5b3045ff9788c4R36-R53). Let me know what your thoughts are 😃

charts/actions-runner-controller/values.yaml
```diff
@@ -33,12 +33,33 @@ authSecret:
   ### GitHub PAT Configuration
   #github_token: ""
 
+# Optionally create additional secrets required for your deployments
+extraSecrets: []
```
mumoshu, line 37: Thanks! Sorry if I wasn't clear enough but I'd like the chart to contain features that are directly related to actions-runner-controller only.

If this key named this way, I believe it can be out of the scope of actions-runner-controller chart, as no values/templates in it is intended to allow arbitrary resources that doesn't directly related to actions-runner-controller itself.

For example, resources like ingress, service, etc created by this chart are tightly-coupled to actions-runner-controller. You can't create arbitrary ingress or service resources that doesn't connect to the actions-runner-controller deployment with this chart. Similary you can't create artibrary service account that isn't used by actions-runner-controller. But this `extraSecrets` allows creating secrets that aren't used by actions-runner-controller at all.

I guess- the only way to make it into this chart might be to name this so that it works like the `serviceAccount` key.

I mean, name it:

```suggestion
imagePullSecrets: []
```

and let the chart map all the names to `--runner-image-secret-name` like

```
        {{- range .Values.imagePullSecrets }}
        - "--runner-image-pull-secret-name={{ .name }}"
        {{- end }}
```

so that it might become clear that this key is only for defining image pull secrets for actions-runner-controller and nothing else.

// To do this, you also need to tweak the golang-side of `--runner-image-pull-secret-name` so that it can accept multiple secret names by repeating the flag

Thanks anyway for your efforts! I hope my comment makes sense.

fgalind1: I've pushed one change with the feeedback 👍  [`fd881f7` (#921)](https://github.com/actions-runner-controller/actions-runner-controller/pull/921/commits/fd881f79771aa857dc198e391514e04e18db1184)

- I've removed for now the possibility to define the "content" of the secret, and I'll think it this could be added on a separate PR or maybe on a different chart.
- I've added support to specify +1 secret name via go cmdline
- I didn't use "imagePullSecrets" because that's already used to define the image pull secrets for the deployment and it could be confusing or we may want to have different image pull secrets for deployment vs runner. I'm using a string array `image.actionsRunnerImagePullSecrets` which can define all the default secret names that are going to be used in the go cmdline

Let me know what your thoughts are 👍

fgalind1: I see that one integration test failed - do you know if that test is flaky by any chance? I ran them locally but all seemed to passed for me using this PR

image

controllers/runner_controller.go
```diff
@@ -872,6 +873,15 @@ func newRunnerPod(template corev1.Pod, runnerSpec v1alpha1.RunnerConfig, default
 		}...)
 	}
 
+	if len(pod.Spec.ImagePullSecrets) == 0 && len(defaultRunnerImagePullSecrets) > 0 {
+		// runner spec didn't provide custom values and default image pull secrets are provided
+		for _, imagePullSecret := range defaultRunnerImagePullSecrets {
+			pod.Spec.ImagePullSecrets = []corev1.LocalObjectReference{
+				{Name: imagePullSecret},
+			}
```
mumoshu, line 881:
```suggestion
			pod.Spec.ImagePullSecrets = append(pod.Spec.ImagePullSecrets, []corev1.LocalObjectReference{
				{Name: imagePullSecret},
			})
```

otherwise pod.spec.ImagePullSecrets always contain one element which correlate to the last element in defualtRUnnerImagePullSecrets, which won't be what you want :)

charts/actions-runner-controller/values.yaml
mumoshu: This does seem like a test flake. Let me rerun it and see if it passes!

controllers/runner_controller.go
fgalind1: Thanks good catch - fixed in ffa0d582a28a3eeac825fadd4efede5189d6141c

mumoshu (approved):
LGTM. Thanks a lot for your efforts and contribution @fgalind1! ☺️

Issue #892, opened 2021-10-14
enterprise runner horizontal scale error with webhook 


roshvin:
**Describe the bug**
enterprise runner Horizontal auto scale error . and its not scaling 
but repo and org level horizontal scaling is working 

controller log 

2021-10-14T12:20:09.659Z        INFO    controllers.Runner      Scale target not found. If this is unexpected, ensure that there is exactly one repository-wide or organizational runner deployment that matches this webhook event  {"event": "workflow_job", "hookID": "45", "delivery": "12c4eb20-2ce9-11ec-9840-f5605bef85ef", "workflowJob.status": "in_progress", "workflowJob.labels": [], "repository.name": "repo", "repository.owner.login": "ROK-ACTION-TEST", "repository.owner.type": "Organization", "action": "started"}
2021-10-14T12:20:36.734Z        INFO    controllers.Runner      finding repository-wide runner  {"event": "workflow_job", "hookID": "45", "delivery": "22e8b180-2ce9-11ec-8058-205f3fdaa6e0", "workflowJob.status": "completed", "workflowJob.labels": [], "repository.name": "repo", "repository.owner.login": "ROK-ACTION-TEST", "repository.owner.type": "Organization", "action": "completed", "repository": "ROK-ACTION-TEST/repo"}
2021-10-14T12:20:36.734Z        ERROR   controllers.Runner      handling check_run event        {"event": "workflow_job", "hookID": "45", "delivery": "22e8b180-2ce9-11ec-8058-205f3fdaa6e0", "workflowJob.status": "completed", "workflowJob.labels": [], "repository.name": "repo", "repository.owner.login": "ROK-ACTION-TEST", "repository.owner.type": "Organization", "action": "completed", "error": "cache contained <nil>, which is not an Object"}
net/http.HandlerFunc.ServeHTTP
        /usr/local/go/src/net/http/server.go:2046
net/http.(*ServeMux).ServeHTTP
        /usr/local/go/src/net/http/server.go:2424
net/http.serverHandler.ServeHTTP
        /usr/local/go/src/net/http/server.go:2878
net/http.(*conn).serve
        /usr/local/go/src/net/http/server.go:1929

deployment yaml 
```

apiVersion: actions.summerwind.dev/v1alpha1
kind: RunnerDeployment
metadata:
  name: dot-runner
spec:
  replicas: 1
  template:
    spec:
      enterprise: test
      labels:
      - dot-enterprise-runner
      env: []
[...]

PR #922, opened 2021-10-28
Add WorkflowJob to GitHubEventScaleUpTriggerSpec types

pje:
Currently, `HorizontalRunnerAutoscalerSpec` allows you to scale up in response to GitHub events, but only specificies `CheckRun`, `PullRequest`, and `Push` in the types—**not** `WorkflowJob`, which is surprising because it's mentioned in the README as ["the preferred way of autoscaling"](https://github.com/actions-runner-controller/actions-runner-controller#:~:text=since%20the%20release%20of%20github's%20workflow_job%20webhook%2C%20webhook%20driven%20scaling%20is%20the%20preferred%20way%20of%20autoscaling).†

I hope this helps and that I've added the type correctly! (This is my first code contribution to the repo.)

†: (In practice, people may be using `githubEvent: {}` and (presumably?) just configuring their github repo settings to only send `workflow_job` events. [Example](https://github.com/actions-runner-controller/actions-runner-controller/issues/892))


controllers/integration_test.go
```diff
@@ -538,6 +538,109 @@ var _ = Context("INTEGRATION: Inside of a new namespace", func() {
...
+							Name: name,
+						},
+						MinReplicas:                       intPtr(1),
+						MaxReplicas:                       intPtr(5),
+						ScaleDownDelaySecondsAfterScaleUp: intPtr(1),
+						ScaleUpTriggers: []actionsv1alpha1.ScaleUpTrigger{
+							{
+								GitHubEvent: &actionsv1alpha1.GitHubEventScaleUpTriggerSpec{
+									WorkflowJob: &actionsv1alpha1.WorkflowJobSpec{
+										Status:       "queued",
```
mumoshu, line 603: Hey! Thanks for the PR. This looks good overall, but could you remove every fields under `WorkflowJobSpec`?

The whole point of `workflow_job` for me is that it frees us from manually mapping `status` and other fields to a subset of runners to be scaled.

For example, we currently map every `workflow_job` of `action=queued` to an internal `scale by 1` event, while mapping one with `action=completed` to an internal `scale down by 1` event. So, manually specifying this defeats our original motivation on using `workflow_job`, probably.

I thought it was the same for other fields like `repositories`, `names`, and `labels`. But I'd need to recheck 🤔

pje: > The whole point of workflow_job for me is that it frees us from manually mapping status and other fields to a subset of runners to be scaled.

I can imagine situations where users might want more control over scaling behavior in response to `WorkflowJob` webhooks. For example: scaling up only when the workflow name matches a pattern, which would require `WorkflowJobSpec.Names`. 

Is your intention that users could accomplish that use-case by different means, or is it not a use-case you want to support?

(Thanks for the review, Yusuke!)

brunnes:
Hi

Was going to raise a similar PR. One thing that would be handy to add is to use the workflow_job.runner_group_name github event.

Our scenario at the moment is we have 1:1 app/aws account/eks cluster so if we use unique labels (app1-dev-build) with our runner groups this is well targeted, but there is a chance that a dev by mistake could start a workflow and choose runs-on and a label outside of their application, and then we would have pods scaling on accounts/clusters not associated with the application. In this scenario, the workflow would fail as the repo would more than likely not be part of the repo list associated with the runner group but this would still trigger the workflow autoscaling trigger (havent tested).

Another benefit is we could have the label be pretty generic such as build/deploy and have the added event to match the runner_group_name so that the apps are scaling pods up into the correct accounts/clusters. 

E.g.
```
metadata:
  name: build-autoscaler
spec:
  scaleTargetRef:
    name: build
  minReplicas: 0
  maxReplicas: 12
  scaleUpTriggers:
  - githubEvent:
      workflowJob:
        runner_group_name: app1
    duration: "5m"

**Comment by mumoshu** [comment]
@brunnes Hey! Thanks for chiming in. I think I almost understand what you say, but still have one question- How does runner groups resolve that situation?

Do runner groups isolate runners in a way so that duplicate runner labels "across" runner groups are fine? (Just asking, because I had been blindly thinking that you need unique runner labels across runner groups. But I may have been completely missed the truth

**Comment by brunnes** [comment]
Hey @mumoshu  thanks for getting back to me so quick.

So we are specifying repos and assigning to runner groups.

```
Runner Groups:
  Name: Team1
  Repos:
    - team1-app1
    - team1-app2
  Name: Team2
  Repos:
    - team2-app1
    - team2-app2
```
At the moment if you wanted to run a runner in the Team1 Account/cluster then you would in the workflow
```
runs-on: ['self-hosted', 'team1-build']
```
and we have the runner deployments with
```
group: Team1
      labels:
        - self-hosted
        - team1-build
```
And that all works fine, but we were thinking a scenario might happen that someone might accidentally from a Team2 repo run a workflow and have runs-on: team1-build. The expected behavior is that the workflow would fail as the repo/runner group doesn't match, but I believe that it would still fire the webhook workflow_job event and expose the label, and then scale the team1-build pod.

To add to that, we were thinking (but havent tested) if we can make the autoscaler github event match to a runner group name, then a user could run-on: 'build' and use that same label for all repos and it would know which account/cluster to run a runner.

**Comment by toast-gear** [comment]
> And that all works fine, but we were thinking a scenario might happen that someone might accidentally from a Team2 repo run a workflow and have runs-on: team1-build. The expected behavior is that the workflow would fail as the repo/runner group doesn't match, but I believe that it would still fire the webhook workflow_job event and expose the label, and then scale the team1-build pod.

I believe the runner group information isn't included in the `workflow_job` event at the point of it being in a queued status, the event will look something like this:

```json
{
  "action": "queued",
  "workflow_job": {
...
    "status": "queued",
    "conclusion": null,
    "started_at": "2021-11-15T13:14:32Z",
    "completed_at": null,
    "name": "my-job-name",
    "steps": [

    ],
...
    "labels": [
      "ubuntu-latest"
    ],
    "runner_id": null,
    "runner_name": null,
    "runner_group_id": null,
    "runner_group_name": null
  },
...
```

The `runner_group_id` and `runner_group_name` will only be populated once the job has been allocated to a runner. How do you envision resolving the runner group?

brunnes:
@toast-gear Ah I see. Well what I was thinking won't work then. Will just have to have unique/targeted labels, which is fine.

toast-gear:
@brunnes As a workaround, although I really wish GitHub had included the repository runner groups in the event whilst queued so we could actually scale stuff correctly without a load of faff 😞, you could:

1. install an instance of `actions-runner-controller` per runner group 
2. expose each install with a unique endpoint
3. create a unique GitHub App per runner group and point the webhook in the app to the corresponding endpoint

Until GitHub provide proper support for runner groups in `workflow_jobs` events this would probably work.

brunnes:
@toast-gear Yeah agree it would be good if runner_group event was there in the queue status.

The workaround looks worth a try though, thanks for that.

pje:
> the runner group information isn't included in the `workflow_job` event at the point of it being in a `queued` status (it would be great if it was 😞)

@toast-gear The reason GitHub can't include runner/group data for `queued` events is by definition: `queued` means "not yet assigned to a runner". We literally don't know yet which runner will pick up the job.

pje:
> I really wish GitHub had included the repository runner groups in the event whilst queued so we could actually scale stuff correctly without a load of faff 😞

@toast-gear  Ah, I just re-read this earlier comment. So you're saying that, within each `workflow_job.queued` event, you'd like to have GitHub include a list of the runner groups that _could potentially_ run the job? Am I understanding that correctly? Could you describe that a little bit more?

toast-gear:
@pje 

> The reason GitHub doesn't include runner/group data for queued events is by definition: queued means "not yet assigned to a runner". We literally don't know yet which runner will pick up the job, nor which group that runner will be part of.

Correct, those attributes in the workflow_kjob event relate to the actual runner that takes on the job, not the repository.

> So you're saying that, within each workflow_job.queued event, you'd like to have GitHub include a list of the runner groups that could potentially run the job? Am I understanding that correctly? Could you describe that a little bit more?

This basically. Without this information you are relegated to doing something hacky:

1. Unique labels for each runner group. This is a bit rubbish as the workflow author now needs to know about the implementation details of the runner group setup, it's very leaky.
2. A unique GitHub App per runner group with the same set of repositories associated with the app as with the runner group. You would then subscribe to the `workflow_job` in the GitHub App. This is not exactly clean and lots of engineers who use GitHub don't know this level of detail about the platform.
3. With this PR merged the `repositories` attribute would work but this is more effort than it should be and makes changes more complicated as I now have to manage multiple things (the runner group and my HRAs)
4. Don't use runner groups and use labels to drive the entire thing. Labels are client-side, they can be changed in a PR and take effect in that PR. Runner groups are sever-side so I probably don't want to swap a server-side control for a client-side control.

As you can see there are ways around this limitation but it's all very messy.

In an ideal world I should be able to have a single install instance of actions-runner-controller, set up a single GitHub App and a single webhook and be able to scale runners accurately using any of GitHub Actions features, of which Runner Groups are one. I don't want to have to configure stuff in GitHub server-side and then have to make configuration changes to actions-runner-controller, I'd like all repository configuration information that could relate to how / what runners are sclaed in the payload at the queued state so keep my actions-runner-controller setup as simple as possible.

toast-gear:
@pje I think myself and many others have totally misunderstood the design of runner groups and how they are intended to be used, with how they are designed to be used it makes sense not including the runner group in the webhook https://github.com/actions-runner-controller/actions-runner-controller/issues/1055.

controllers/integration_test.go
toast-gear: long time no speak 😅 , could you remove the attributes so we can merge this? We want to have the event to be explicit as implemented via this PR with the plan to remove the default event entirely however we unsure about the `WorkflowJob` attributes so would like to get this merged without them first and go from there.

pje: Will do! Thanks for the reminder, @toast-gear

pje: @toast-gear: removed all `WorkflowJobSpec` fields in 1002e8f7e10b52dc32eff5ab6b5885fddcc1fb28. Now it's just an empty struct type.

toast-gear (approved):

controllers/integration_test.go
```diff
@@ -1150,7 +1248,7 @@ var _ = Context("INTEGRATION: Inside of a new namespace", func() {
 
 			// Scale-up to 2 replicas on first workflow_job webhook event
 			{
-				env.SendWorkflowJobEvent("test", "valid", "pending", "created", []string{"self-hosted"})
+				env.SendWorkflowJobEvent("test", "valid", "pending")
```
mumoshu, line 1251: We need to simulate the default and custom labels to make these tests passing. I'll follow-up with a commit to fix that.

mumoshu (approved):
Thanks for your patience and help @pje!

mumoshu:
Also see https://github.com/actions-runner-controller/actions-runner-controller/pull/1361 for the corresponding documentation update!

genisd:
Well done guys!

I looked at the readme in master, and i quickly navigated here somehow: https://github.com/actions-runner-controller/actions-runner-controller/blob/master/README.md?plain=1#L678

While `Example 1` covers this new way of doing things, I think this has to be the most preferred way to do the webhook scaling (at least from my experience). I think we should have it as the first example configuration snipped shown.

Are there any scenarios wherein anyone would want
```
  - githubEvent:
      checkRun:
        types: ["created"]
        status: "queued"
```
over the new method
```
  scaleUpTriggers:
  - githubEvent:
      workflowJob: {}
```
I cannot come up with a reason why anyone would want or need the older way.


And first impressions are great, it looks *very* promising!
We used to have starvation issues (we have a construction also where we skip steps).

Have to give it a bit more time, but the way it looks right now, for webhook scaling, I would almost suggest to release current master. It's a big improvement

toast-gear:
Yeh, raise a PR to swap them around, it's just an artefact of history really as to why we don;t have workflow job as the first example yaml.

> Have to give it a bit more time, but the way it looks right now, for webhook scaling, I would almost suggest to release current master. It's a big improvement

we're testing master atm to hopefully release soon

PR #926, opened 2021-10-29
refactor: split Reconciler from Reconcile in a few methods

fgalind1:
Reduce a bit the size of `Reconcile()` method from ~400 to ~300 to be easier to understand

Functionality kept the same, just some logic from `Reconcile` was abstracted into 3 additional methods

### DIFF
--- a/controllers/runner_controller.go
+++ b/controllers/runner_controller.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


mumoshu:
@fgalind1 Thanks for your contribution and sorry for the delay! We're planning to review and merge this BEFORE our next minor release(=a feature release). 

The rationale is that, although I believe you've done it perfectly, I'm not sure if this doesn't actually break anything, as our unit test and integration test suite are far from 100% coverage. So I'd like to make it as a part of a feature release rather than a patch release, so that we can better communicate via the semver that it may break something(unexpectedly) due to new features) than nothing.

mumoshu:
Maybe you introduced some compile errors when you kindly tried to rebase your work on our latest main branch! But I'll try to fix it myself. Wish I'll be able to manage it!

controllers/runner_controller.go
```diff
@@ -183,237 +144,186 @@ func (r *RunnerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
-			} else if offline {
+		} else if offline {
+			if registrationOnly {
+				log.Info(
+					"Observed that registration-only runner for scaling-from-zero has successfully been registered.",
+					"podCreationTimestamp", pod.CreationTimestamp,
+					"currentTime", currentTime,
+					"configuredRegistrationTimeout", registrationTimeout,
+				)
+			} else if registrationDidTimeout {
```
mumoshu, line 326: Seems this is duplicate of L327-L334

mumoshu: Fixed in 21a2ba9

mumoshu (approved):
Hey! I was able to fix the build error and I manually took diffs around these changes:

- Extraction of `processRunnerDeletion`
- Extraction of `processRunnerCreation`
- Return on `r.processRunnerCreation` so that you can decrease the indent-level of all the lines bellow that return (formerly it was intended within a `else` block)

Everything looks good now.

If there's next chance you kindly contribute to this project, if I could ask, it would be super helpful if you could split it into 3 commits like the above so that it is immediately clear for maintainers where to look for review!

Thank you for your contribution. This LGTM now. Merging.

fgalind1:
> If there's next chance you kindly contribute to this project, if I could ask, it would be super helpful if you could split it into 3 commits like the above so that it is immediately clear for maintainers where to look for review!


Will do, thanks on this.

> Thank you for your contribution. This LGTM now. Merging.

:)

PR #932, opened 2021-11-03
doc: Add GitHub App Permission for Workflow Job Webhook

mumoshu:
I think we've missed documenting about the app permission for receiving workflow_job events via the app webhook


toast-gear:
Good spot, that area probably needs slightly adjusting further as you can either create the webhook centrally in the app, or externally and so it probably needs some clarification above it

toast-gear:
@mumoshu how about that?

mumoshu:
@toast-gear Looks great! Thanks for the enhancement ☺️

PR #937, opened 2021-11-09
Add `authSecret.enabled` parameter to chart values

### LINKED ISSUE #936: Support configuring authentication without requiring secret to exist [Helm Chart]
**Is your feature request related to a problem? Please describe.**
We are using Vault to manage secrets and use an admission-control workflow that will inject secrets into container environment variables on pod startup. This means that we do not read any values from `authSecret`.  

As it stands we are configuring `env.GITHUB_TOKEN` with a value that is replaced by our admission controller and this works well. However, the current chart implementation requires that an `authSecret` is present in the controller namespace and therefore we must create a dummy secret in order to satisfy the secret volume mount in the controller deployment.

**Describe the solution you'd like**
Add a `authSecret.require` value to the Helm Chart which is `true` by default. When the value is `false`, references to the authentication secret are omitted in the deployment manifest. This would retain the existing behavior for normal use-cases whilst also allowing for an unopinionated secret workflow such as outlined above

**Describe alternatives you've considered**
We can create a dummy secret as a workaround.

phoban01:
Adds a new parameter to chart values: `authSecret.enabled`.

By default `authSecret.enabled` is `true` and there is no change to existing behavior.

When false the chart deployment template will not add GITHUB_*
environment variables to the manager container. In addition, the `volume`
and `volumeMount` for the secret will also be omitted from the
deployment manifest.

Fixes #936 

Signed-off-by: Piaras Hoban <phoban01@gmail.com>

### DIFF
--- a/charts/actions-runner-controller/templates/deployment.yaml
+++ b/charts/actions-runner-controller/templates/deployment.yaml
@@ -65,6 +65,7 @@ spec:
         - name: GITHUB_ENTERPRISE_URL
           value: {{ .Values.githubEnterpriseServerURL }}
         {{- end }}
+        {{- if .Values.authSecret.enabled }}
         - name: GITHUB_TOKEN
           valueFrom:
             secretKeyRef:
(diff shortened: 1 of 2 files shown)


mumoshu (approved):
LGTM 👍 I also wanted to note that this feature has potential to cover the use-case described in another PR #884.

Thank you for your contribution @phoban01 ☺️

PR #831, opened 2021-09-20
Ephemeral runner

rofafor:
The `--ephemeral` switch is now released and `--once` was (accidentally?) removed in https://github.com/actions/runner/pull/660/files:
- https://github.com/actions/runner/blob/7ffd9af644647114f924a8b6cede3aaf5d79b71f/src/Runner.Listener/CommandSettings.cs#L71-L72
- https://github.com/actions/runner/releases/tag/v2.282.0

Actually, the `--ephemeral` switch is already in the images published via CI, but missing one small fix. Here's my log from the current `latest` tagged image:
```
Passing --once to runsvc.sh to enable the legacy ephemeral runner.
.path=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
Starting Runner listener with startup type: service
Started listener process
Started running service
Unrecognized command-line input arguments: 'once'. For usage refer to: .\config.cmd --help or ./config.sh --help

√ Connected to GitHub
```

I had to leave `RUNNER_FEATURE_FLAG_EPHEMERAL` logic due to the fact that GitHub Enterprise Server 3.0 or 3.1 doesn't support this new ephemeral feature yet:
```
√ Connected to GitHub

# Runner Registration

An Internal Error Occurred. Activity Id: XXXXXXXX-e335-4995-9f33-659731da78eb.
Configuration failed. Retrying
```

PR #938, opened 2021-11-09
Removed --once flag in favor of --ephemeral

reiniertimmer:
Edit: probably a duplicate of #831 

Since the `--ephemeral` flag is now generally available, it is best to remove the `--once` flag altogether - it is undocumented and could be removed at any time. So, it would be best to switch over.

According to [the official docs for autoscaling](https://docs.github.com/en/actions/hosting-your-own-runners/autoscaling-with-self-hosted-runners#using-ephemeral-runners-for-autoscaling) this flag is the preferred way to implementing webhook-based autoscaling.

# Testing

I tested this on the latest runner image `v2.284.0` and it works without issues, no customizations required.
Since this update to the `entrypoint.sh` will only be included in newer images, there is no real need to keep
the old flag supported.

I only tested this with a single RunnerDeployment inside a local cluster. The pod logs the following output:

```
√ Connected to GitHub

Current runner version: '2.284.0'
2021-11-09 20:38:33Z: Listening for Jobs
2021-11-09 20:44:56Z: Running job: test
2021-11-09 20:45:02Z: Job test completed with result: Succeeded
√ Removed .credentials
√ Removed .runner
Runner listener exited with error code 0
Runner listener exit with 0 return code, stop the service, no retry needed.

```

As expected, it exits right after the job was finished, after which a new pod is started.

If you feel this requires additional testing then please let me know.

# Documentation

I basically cut out the entire part from the README.md documenting the experimental nature of the `--ephemeral`
flag. It now contains only a section decribing that the flag exists and not much more

### DIFF
--- a/README.md
+++ b/README.md
@@ -893,7 +893,7 @@ spec:
           tolerationSeconds: 10
       # true (default) = The runner restarts after running jobs, to ensure a clean and reproducible build environment
       # false = The runner is persistent across jobs and doesn't automatically restart
-      # This directly controls the behaviour of `--once` flag provided to the github runner
+      # This directly controls the behaviour of `--ephemeral` flag provided to the github runner
       ephemeral: false
       # true (default) = A privileged docker sidecar container is included in the runner pod.
       # false = A docker sidecar container is not included in the runner pod and you can't use docker.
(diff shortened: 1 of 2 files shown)


reiniertimmer:
Unit tests are failing because of the changed entrypoint.sh (nice!). Will look into it

reiniertimmer:
Seeing the discussion on #831 - this cannot be merged because of issues in GHES

Somehow I missed this other PR when looking into this.

Issue #951, opened 2021-11-16
Not able to scale Organization RunnerDeployment with "workflow_job"


sigurdfalk:
**Describe the bug**

Organization RunnerDeployment not able to scale with `workflow_job` event. Manifests looks like below:

```yaml
apiVersion: actions.summerwind.dev/v1alpha1
kind: RunnerDeployment
metadata:
  name: iac-runners
  namespace: actions-runner-system
spec:
  template:
    metadata:
      labels:
        app: iac-runners
        owner: platform
        environment: test
    spec:
      organization: gjensidige
      group: IaC
      labels:
        - iac
      ephemeral: true

```

```yaml
apiVersion: actions.summerwind.dev/v1alpha1
kind: HorizontalRunnerAutoscaler
metadata:
  name: iac-runners
  namespace: actions-runner-system
spec:
  scaleTargetRef:
    name: iac-runners
  scaleUpTriggers:
    - githubEvent: {}
      amount: 1
      duration: "5m"
  minReplicas: 2
  maxReplicas: 5
```

Seing the following in logs:

```
2021-11-16T20:04:01.274Z        DEBUG   controllers.Runner      Found 0 HRAs by key     {"key": "gjensidige/terraform-aks"}
2021-11-16T20:04:01.274Z        DEBUG   controllers.Runner      Found 1 HRAs by key     {"key": "gjensidige"}
2021-11-16T20:04:01.274Z        DEBUG   controllers.Runner      Found 0 HRAs by key     {"key": "enterprises/gjensidige"}
2021-11-16T20:04:01.274Z        DEBUG   controllers.Runner      no repository/organizational/enterprise runner found    {"event": "workflow_job", "hookID": "328719738", "delivery": "56d85c90-4718-11ec-8b7f-a4369b1789a2", "workflowJob.status": "queued", "workflowJob.labels": ["self-hosted", "iac"], "repository.name": "terraform-aks", "repository.owner.login": "gjensidige", "repository.owner.type": "Organization", "enterprise.slug": "gjensidige", "action": "queued", "repository": "gjensidige/terraform-aks", "organization": "gjensidige", "enterprises": "gjensidige"}
[...]

PR #953, opened 2021-11-17
fix(#951): add exception for self-hosted label in webhook search

clement-loiselet-talend:
Fix issue #951 

The webhook "workflowJob" pass the labels the job needs to the controller, who in turns search for them in its RunnerDeployment / RunnerSet. The current implementation ignore the search for `self-hosted` if this is the only label, however if multiple labels are found the `self-hosted` label must be declared explicitely or the RD / RS will not be selected for the autoscaling.

This PR fixes the behavior by ignoring this label, and add documentation on this webhook for the other labels that will still require an explicit declaration (OS and architecture). 

The exception should be temporary, ideally the labels implicitely created (self-hosted, OS, architecture) should be searchable alongside the explicitly declared labels.

code tested, work with `["self-hosted"]` and `["self-hosted","anotherLabel"]`

### DIFF
--- a/README.md
+++ b/README.md
@@ -610,6 +610,8 @@ spec:
     duration: "30m"
 ```
 
+This webhook requires you to explicitely set the labels in the RunnerDeployment / RunnerSet if you are using them in your workflow to match the agents (field `runs-on`). Only `self-hosted` will be considered as included by default.
+
 You can configure your GitHub webhook settings to only include `Workflows Job` events, so that it sends us three kinds of `workflow_job` events per a job run.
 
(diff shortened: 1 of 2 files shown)


mumoshu (approved):
It took a bit more time until I figure out how this aligns with all the other aspects of the system. After rereading your comments and lastly @fgalind1's [suggestion and summary](https://github.com/actions-runner-controller/actions-runner-controller/pull/1012#discussion_r771845538), I'm finally confident with this change.

LGTM. Thank you so much for your patience and contribution!

PR #952, opened 2021-11-17
add `additionalVolumes` and `additionalVolumeMounts` to helm chart

brunocous:
Fixes issue #950.

PR #966, opened 2021-11-26
fix: update volumes and volumes mounts indent

gscho:
I was getting `YAML parse error on actions-runner-controller/templates/deployment.yaml` and when I investigated it seems the indent by values were off by 2. 

These are the values I was testing with:

```
additionalVolumeMounts:
- mountPath: /etc/ssl/certs
  name: ghe-cert
  readOnly: true
additionalVolumes:
- name: ghe-cert
  secret:
    secretName: ghecert
```

Signed-off-by: Gregory Schofield <gscho@github.com>


mumoshu (approved):
Apparently, this has been introduced via #952. Thanks for fixing it @gscho!

PR #987, opened 2021-12-07
update readme instruction for webhook scaling

KannappanSomu:
Added note about mandatory parameters **minReplicas** & **maxReplicas** for webhook driven scaling without which **HorizontalRunnerAutoscaler** object will not deploy successfully


mumoshu (approved):
Good catch! Thanks a lot for your contribution @KannappanSomu ☺️

PR #988, opened 2021-12-07
fix(webhook): ingress for k8s v1.22.x support

samip5:
Also dropped the deprecated .Capabilities.KubeVersion.Gitversion usage in ingress template.

### DIFF
--- a/charts/actions-runner-controller/templates/githubwebhook.ingress.yaml
+++ b/charts/actions-runner-controller/templates/githubwebhook.ingress.yaml
@@ -41,7 +41,7 @@ spec:
                name: {{ $fullName }}
                port:
                  number: {{ $svcPort }}
-              {{- else -}}
+              {{- else }}
               serviceName: {{ $fullName }}
               servicePort: {{ $svcPort }}
               {{- end }}
--- a/charts/actions-runner-controller/templates/githubwebhook.ingress.yaml
+++ b/charts/actions-runner-controller/templates/githubwebhook.ingress.yaml
@@ -1,10 +1,12 @@
 {{- if .Values.githubWebhookServer.ingress.enabled -}}
 {{- $fullName := include "actions-runner-controller-github-webhook-server.fullname" . -}}
 {{- $svcPort := (index .Values.githubWebhookServer.service.ports 0).port -}}
-{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }}
 apiVersion: networking.k8s.io/v1beta1
-{{- else -}}
+{{- else if .Capabilities.APIVersions.Has "extensions/v1beta1" }}
 apiVersion: extensions/v1beta1
+{{- else -}}
+apiVersion: networking.k8s.io/v1
 {{- end }}
 kind: Ingress
 metadata:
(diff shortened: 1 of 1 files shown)


charts/actions-runner-controller/templates/githubwebhook.ingress.yaml
```diff
@@ -34,8 +36,15 @@ spec:
           {{- range .paths }}
           - path: {{ .path }}
             backend:
+              {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
```
mumoshu, line 39: You need to scope `Capabilityes` with `$`. Otherwise it is evaluated in the context of `range .paths`, which tries to evaluate something like `$path..Capabilities.APIVersions.Has` which results in `<.Capabilities.APIVersions.Has>: nil pointer evaluating interface {}.APIVersions` error

```suggestion
              {{- if $.Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
```

charts/actions-runner-controller/templates/githubwebhook.ingress.yaml
```diff
@@ -34,8 +36,15 @@ spec:
...
+              service:
+               name: {{ $fullName }}
+               port:
+                 number: {{ $svcPort }}
+              {{- else -}}
```
mumoshu, line 44: This needs to be:

```suggestion
              {{- else }}
```

Otherwise it breaks for any non-`networking.k8s.io/v1`-case like

```
spec:
  rules:
    - host: "example"
      http:
        paths:
          - path: mypath
            backend:serviceName: foo-actions-runner-controller-github-webhook-server
              servicePort: 80
```

mumoshu (approved):
LGTM. Thank you for your contribution @samip5!

charts/actions-runner-controller/templates/githubwebhook.ingress.yaml
```diff
@@ -1,10 +1,12 @@
 {{- if .Values.githubWebhookServer.ingress.enabled -}}
 {{- $fullName := include "actions-runner-controller-github-webhook-server.fullname" . -}}
 {{- $svcPort := (index .Values.githubWebhookServer.service.ports 0).port -}}
-{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }}
 apiVersion: networking.k8s.io/v1beta1
-{{- else -}}
+{{- else if .Capabilities.APIVersions.Has "extensions/v1beta1" }}
```
jsotelo, line 6: For kubernetes 1.19, both `extensions/v1beta1` and `networking.k8s.io/v1` coexist so this approach in fact breaks this helm chart since we end up with `apiVersion: extensions/v1beta1` here but also hit the `{{- if $.Capabilities.APIVersions.Has "networking.k8s.io/v1" }}` conditional below producing the following error:

```
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(Ingress.spec.rules[0].http.paths[0].backend): unknown field "service" in io.k8s.api.networking.v1beta1.IngressBackend
helm.go:88: [debug] error validating "": error validating data: ValidationError(Ingress.spec.rules[0].http.paths[0].backend): unknown field "service" in io.k8s.api.networking.v1beta1.IngressBackend
```

samip5: How do you suppose this should be fixed?

jsotelo: Maybe test for `"networking.k8s.io/v1"` explicitly here since we test explicitly below?

toast-gear: Could someone raise an issue for this please with the details so this can be fixed

jsotelo: Done: https://github.com/actions-runner-controller/actions-runner-controller/issues/1037

Issue #990, opened 2021-12-08
TotalNumberOfQueuedAndInProgressWorkflowRuns cant scale beyond 60 runners due to GH API pagination


larhauga:
**Describe the bug**
When scaling runners based on `TotalNumberOfQueuedAndInProgressWorkflowRuns` it seems that the maximum number of possible runners are currently 60.

Having testet with a workflow with 170 jobs (created by matrix), we only reached 50 runners.

We think the issue is related to missing pagination handling of the ListWorkflowJobs api here: https://github.com/actions-runner-controller/actions-runner-controller/blob/48923fec562126e2ffe575c9d7694d18c0a56f3d/controllers/autoscaling.go#L70
The default pagination of that API is 30, and that correlates with what we see in the controller manager logs.

**Checks**

- [ ] My actions-runner-controller version (v0.x.y) does support the feature
- [ ] I'm using an unreleased version of the controller I built from HEAD of the default branch

**To Reproduce**
1. Create a `HorizontalRunnerAutoscaler` that uses `TotalNumberOfQueuedAndInProgressWorkflowRuns`, with a limit of more than 60.
2. Trigger a workflow that has more than 60 parallel jobs.
3. Expect the number of runners to level out below what should be expected.

**Expected behavior**
Expected the number of runners to be equal to the number of jobs created in the workflow.

**Environment (please complete the following information):**
 - Deployment Method Helm
 - Helm Chart Version: 0.13.2

jbkc85:
FYI - I am not sure this fixes the pagination issue.  I have been running 0.20.4 for a few days now and ran into an issue where I can only have - at most - 24 runners up at a single time.  I can't necessarily tell where in the code this is happening (whether it is a runner call or the job call), but given this is the only code change that looks like it impacts the area around scaling to the workflow jobs I think this is causing it currently.

Do we have any metrics that can be exposed or even debug logging to indicate the amount of jobs returned by these calls?

PR #925, opened 2021-10-29
Update go-github from v37 -> v39

pje:
Like #800, but _also_ update `import`s to use the new version :)

I ran: 

```
go get -u github.com/google/go-github/v39
rg --files-with-matches 'github.com/google/go-github/v37/github' **/*.go | xargs sed -i '' "s|github.com/google/go-github/v37/github|github.com/google/go-github/v39/github|g"
go mod tidy
```

In particular this pulls in the changes @mumoshu made in https://github.com/google/go-github/pull/2103 that have since been merged into their main, so we no longer need the `replace` directive in go.mod.

PR #992, opened 2021-12-08
fix: pagination for ListWorkflowJobs in autoscaler (#990)

larhauga:
Adding handling of paginated results when calling `ListWorkflowJobs`. By default the `per_page` is 30, which potentially would return 30 queued and 30 in_progress jobs.

This change should enable the autoscaler to scale workflows with more than 60 jobs to the exact number of runners needed.

Problem: I did not find any support for pagination in the Github fake client, and have not been able to test this (as I have not been able to push an image to an environment where I can verify this).
If anyone is able to help out verifying this PR, i would really appreciate it.

Ref #990

### DIFF
--- a/controllers/autoscaling.go
+++ b/controllers/autoscaling.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/actions-runner-controller/actions-runner-controller/api/v1alpha1"
+	"github.com/google/go-github/v39/github"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


mumoshu:
@larhauga Thanks for your contribution! Rebased and I'm now trying to test this locally. If you could help double-checking, it would be great if you could follow [our contribution guide](https://github.com/actions-runner-controller/actions-runner-controller/blob/master/CONTRIBUTING.md#testing-controller-built-from-a-pull-request) for local testing.

controllers/autoscaling.go
```diff
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/actions-runner-controller/actions-runner-controller/api/v1alpha1"
+	"github.com/google/go-github/v37/github"
```
mumoshu, line 13:
```suggestion
	"github.com/google/go-github/v39/github"
```

This seems outdated since #925. I'll update your commit with the change.

mumoshu:
Sry it took much longer time than I had thought to have time to test this manually! Code LGTM so let merge this anyway and make follow-ups only after we find any issues layer.
Thanks for your contribution @larhauga ☺️

Issue #911, opened 2021-10-26
Random `Operation Cancelled` Runner Decommissions 


jbkc85:
**Describe the bug**

In the current deployment of actions-runner-controller, whenever a GitHub runner has an action cancelled (via GitHub UI) or has an action that fails (like a unit test), it seems to cause a cascading effect that triggers other random cancellations of actions running on other runners in the system.  Furthermore, there is a high chance that at the end of a GitHub workflow that one of the last remaining runners (IE: a runner that survives until a scale down is triggered via reconciling) is cancelled while 'busy' with a Workflow Job/Step.

**Checks**

- we are not using ephemeral runners at this time - though ephemeral runners appear to have the same issue
- we are on v`summerwind/actions-runner-controller:v0.20.1` and use the the provided GitHub-runner with our own custom OpenJDK installation
- controller is installed via Helm:
```yaml
syncPeriod: 5m
githubAPICacheDuration: 5m
nodeSelector:
  environment: ops
  purpose: infrastructure
```
- AWS EKS w/ Cluster Autoscaling and Spot Instances (confirmed spot instances aren't causing a shutdown issue)

**To Reproduce**
Steps to reproduce the behavior (though is more random than not):
1. Setup GitHub-action-controller with the above Helm deployment and the following spec:
```yaml
apiVersion: actions.summerwind.dev/v1alpha1
kind: RunnerDeployment
metadata:
  namespace: default
  name: org-runners
spec:
  template:
    metadata:
      annotations:
        cluster-autoscaler.kubernetes.io/safe-to-evict: "true"
    spec:
      dockerdWithinRunnerContainer: true
      organization: <Github Org>
      nodeSelector:
        capacity-type: SPOT
        purpose: github-runners
        environment: ops
      labels:
        - java
        - linux
        - eks
        - self-hosted
      image: <GitHub-runner image here>
      resources:
        requests:
          cpu: "1.0"
          memory: "10Gi"
---
apiVersion: actions.summerwind.dev/v1alpha1
kind: HorizontalRunnerAutoscaler
metadata:
  namespace: default
  name: runners-autoscaler
spec:
  scaleTargetRef:
    name: org-runners
  scaleDownDelaySecondsAfterScaleOut: 3600
  minReplicas: 1
  maxReplicas: 150
  metrics:
  - type: TotalNumberOfQueuedAndInProgressWorkflowRuns
    repositoryNames:
    - <your repository here>
```
[...]

PR #993, opened 2021-12-09
Workaround for offline+busy runners

mumoshu:
Ref #911

We're going to cut a patch release after merging this.


gmambro:
Many thanks for that! I planned to write the patch and try it on my cluster next week and then eventually submit the PR... I will then test this!

gmambro:
@mumoshu is this already in controller v0.24.1? I tried it and still got old behavior, now double-checking the cluster and also trying the "canary" controller.

Issue #920, opened 2021-10-28
condense `r.GitHubClient.ListRunners(ctx, enterprise, org, repo)` calls into a single call/cache


jbkc85:
**Is your feature request related to a problem? Please describe.**
The feature is related to a problem as it comes to GitHub API calls.  There are multiple instances in which the GitHubClient.ListRunners is called and returns potentially 100 runner statuses via the API, but as I can tell none of this is shared and rather than using the same response for checking various runners, the call is made recursively through the looping of each runner.  This means for each runner, the API is hit N*N times depending on various stages of the runner life cycle.

**Describe the solution you'd like**
The solution would be along the lines of storing the runners statuses every minute (perhaps?) or at a various time span that aligns with registration checks, unregistration attempts, etc.  Unless I am misunderstanding the behavior, I do not believe there has to be a specific reason why its called more than once per cycle.

**Describe alternatives you've considered**
N/A

**Additional context**
When running 100+ runners to sustain a workload, this call eventually eats through the API limit and kills all the runners.  I believe it is one of the main culprits.

mumoshu:
@sledigabel Thanks for the compliment!

Yeah, an in-memory time-based cache would work. I believe the most called APIs various flavors of "List Runners" API so the potential implementation of it could be added to:

https://github.com/actions-runner-controller/actions-runner-controller/blob/f7e14e06e87961a67f46e22cc33ba2c707b6e682/github/github.go#L214-L222

jbkc85:
https://github.com/actions-runner-controller/actions-runner-controller/pull/997

I do apologize for this PR likely not being in the contribution feed back...I wanted to share the progress and see what you all thought.

jbkc85:
also a side note here: Why isn't the GitHub API status pulled from GitHub and stored w/in the Runner's CRD?  I can imagine that we only need 1 call to the GitHub API per minute to list all runners, in which will update the `Status:` of an individual runner pod....that will eliminate a *ton* of GitHub Runner API calls.

mumoshu:
> Why isn't the GitHub API status pulled from GitHub and stored w/in the Runner's CRD?

@jbkc85 I believe it's k8s custom controller convention to not use resource status for cache. It's for reflecting the source of truth, not for storing the source data or cache.

mumoshu:
Linking #1101 from @Djiit

Djiit:
[...]

PR #997, opened 2021-12-09
cache GitHub runners request

jbkc85:
https://github.com/actions-runner-controller/actions-runner-controller/issues/920

Attempt to create a generic endpoint/call to avoid calling the runners endpoint N times a minute - where N is the number of runners available to be reconciled at any given time.

I was unable to figure out the best place to put this - and had to essentially replicate the code between the RunnerReconciler and the RunnerPodReconciler - but perhaps its better placed in the GitHubClient code?  That way its one call vs making multiple functions in the Reconciler code?  Is there a place to put this that can be used by both structs w/o having multiple footprints?

Regardless - this code allows for a maximum of 120 calls to the GitHub API for Runner status - 60 by the RunnerPodReconciler and 60 by the RunnerReconciler - which removes a potential ~8880 calls to the API (given I have scaled up to 150 runner pods before)


controllers/utils.go
```diff
@@ -1,5 +1,23 @@
...
+)
+
+// RunnerCache holds cached returns from github runner API calls
+// - default time for cache is 60 * time.Seconds()
+type RunnerCache struct {
+	ExpirationTime metav1.Time
+	Runners        []*gogithub.Runner
+}
+
+func (i *interface{}) test() {
```
sledigabel, line 17: This is probably not used?

controllers/utils.go
```diff
@@ -1,5 +1,23 @@
...
+import (
+	"fmt"
+
+	gogithub "github.com/google/go-github/v37/github"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// RunnerCache holds cached returns from github runner API calls
+// - default time for cache is 60 * time.Seconds()
+type RunnerCache struct {
```
sledigabel, line 12: The RunnerCache structure is missing the dimensions where you have multiple runners, per repo, orgs or enterprise. In the current implementation, it looks like it would only work with one type of runners in a single controller.

Instead, you could have a `RunnerCacheMap` structure that would point to a `RunnerCache` for each runner group.

controllers/runner_pod_controller.go
```diff
@@ -429,3 +431,20 @@ func (r *RunnerPodReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	now := time.Now().Add(-60 * time.Second)
+	if !r.RunnerCache.ExpirationTime.Time.IsZero() && now.Before(r.RunnerCache.ExpirationTime.Time) {
+		return nil
+	}
+	// if context of time is not passed and the cached_runners is not null, return cached runners
+	// else, call the runners and 'cache' it to return
+	runners, err := r.GitHubClient.ListRunners(ctx, enterprise, org, repo)
+	// always cache runners for one minute regardless of other caching in GitHub
+	r.RunnerCache = RunnerCache{
+		ExpirationTime: metav1.Time{Time: time.Now().Add(60 * time.Second)},
```
sledigabel, line 445: I would make this a constant or a config item instead of hardcoding it to 1 minute

controllers/runner_pod_controller.go
```diff
@@ -429,3 +431,20 @@ func (r *RunnerPodReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Named(name).
 		Complete(r)
 }
+
+func (r *RunnerPodReconciler) updateCachedRunners(ctx context.Context, enterprise, org, repo string) error {
+	now := time.Now().Add(-60 * time.Second)
```
sledigabel, line 436: I'm not sure what this means?

sledigabel:
Thank you for looking into this @jbkc85 !! FWIW, I think the general approach is good, but as I pointed out in the review there's a dimension missing in the current implementation

I'm not sure I was super clear :-) but if not lemme know I'll try to explain better.

controllers/utils.go
jbkc85: The `Runners` is used in the structure to host the actual GitHub Runners from the GitHub client.  I put it in Utils because I was using it in two files - and wasn't sure where else to put it unless I were to make another file specific to this cache inside of the controllers folder

controllers/utils.go
jbkc85: Gotcha!  I will take a look at that.

controllers/runner_pod_controller.go
jbkc85: Agreed. The goal was to at least reduce the GitHub API calls down to (at the most) 60 per minute per reconciler.  I figured it was a good approach to get it set to 1m in the MVP, then work on the configurable asset after - but we can definitely make it a `CONST` until then

jbkc85:
> Thank you for looking into this @jbkc85 !! FWIW, I think the general approach is good, but as I pointed out in the review there's a dimension missing in the current implementation
> 
> I'm not sure I was super clear :-) but if not lemme know I'll try to explain better.

not everything is super clear - but thats because I haven't had my head in the code long enough... I get what you are asking for and will look into the available code to see what I can come up with!  Thanks for the review 👍

controllers/utils.go
cbui: I think they were referring to this:

```
func (i *interface{}) test() {
```

jbkc85: > I think they were referring to this:
> 
> ```
> func (i *interface{}) test() {
> ```

oh - wow.  Didnt even know that was there.

jbkc85:
continued in https://github.com/INNOSOLProject/actions-runner-controller/pull/1 - I took a second go around and think came up with a better solution.  I Will be submitting a new PR in soon from this as its actually in production for us at this time.

PR #849, opened 2021-09-28
Push packages to GHCR

sollie:
A while back, in issue https://github.com/actions-runner-controller/actions-runner-controller/issues/566 I suggested adding packages to GHCR as an alternative to Docker Hub.

This is my stab at accomplishing this.

The repository needs a secret with a PAT that can write packages.
I've called my secret `GHCR_PAT`.

I'm also using `secrets.DOCKER_USER` instead of `env.DOCKERHUB_USERNAME` for the Dockerhub tags. I figured this works smoother if people forking the repo also want to build. My initial commit retains the `DOCKERHUB_USERNAME` env. I guess this could also be removed.

After an initial run, packages have to be made visible and preferably linked to the repo.

Issue #566, opened 2021-05-23
Move our DockerHub account to something more consistent with the GitHub organization name


mumoshu:
Extracted from https://github.com/actions-runner-controller/actions-runner-controller/issues/562#issuecomment-846336222

DockerHub doesn't allow hyphens in the organization name so it's impossible to just use `actions-runner-controller`.

I've taken [actionsrunnercontroller](https://hub.docker.com/orgs/actionsrunnercontroller) anyway but not yet sure if this is good enough.

sollie:
Have you considered hosting packages on GitHub/ghcr.io instead or in addition to DockerHub?

mumoshu:
Nope. Does it help? Anyone's using it as a primary container image source?

mabushey:
Docker the company is really imploding. They have been placing so many bandwidth restraints on DockerHub that it's getting unusable. I recommend ditching docker/dockerhub. There are so many container engines that docker itself is irrelevant. Even Kubernetes is deprecating Docker as a container runtime after v1.20.: https://kubernetes.io/blog/2020/12/02/dont-panic-kubernetes-and-docker/

sledigabel:
Dockerhub is sure losing some momentum.
Github container registry will integrate better with github actions too, the permissions are attached to the repository and it will integrate better with github actions as you don't need to add a username/password

https://github.com/marketplace/actions/build-and-publish-docker-image-to-github-packages-registry

mumoshu:
Thanks everyone. I hear you all. But this is about using another DockerHub account and all you're talking about how ghcr.io is great! I appreciate the information, but I don't see what you're asking.

Perhaps stop publishing to DockerHub while adding ghcr.io as the single registry that we publish images? But then I can easily imagine that another person doesn't like ghcr.io will start asking me to restart publishing to DockerHub. So it's very unlikely we stop using DockerHub.

To be fair, K8s deprecating Docker as a container runtime has nothing to do with DockerHub. I usually use dockerd or contained as my k8s container runtimes, and gcr.io, qyay.io, and docekrhub in my work. I do prefer using contained these days, but still using DockerHub. I don't dislike quay.io even though my experience says it's still more unreliable than DockerHub :)

Anyway, I'm just explaining my thoughts. If you want us to add ghcr.io as the another container registry to publish images, just create another issue, or even better contribute a pull request to enhance our github actions workflow to support it.

PR #1004, opened 2021-12-12
Push packages to GHCR (rebased. the credit goes to @sollie)

mumoshu:
I was going to handle #849 by awesome @sollie in a usual merge process but failed to do so due to some permission issues in GH and also perhaps some kind of (unusual?) conflicts among our git histories.

This pull request is made of the rebased and reworked version of the commit included in @sollie's, so that it should be good to be merged as usual. Thank you for your contribution @sollie!

Ref #849
Ref #566

### DIFF
--- a/.github/actions/setup-docker-environment/action.yaml
+++ b/.github/actions/setup-docker-environment/action.yaml
@@ -7,6 +7,12 @@ inputs:
   password:
     description: "Password"
     required: true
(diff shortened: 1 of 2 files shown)


sollie:
This is excellent. I try to avoid using hub.docker.com as much as possible after the rate limits were introduced.

Issue #335, opened 2021-02-19
Runner not installing due to webhook error


leovanhaaren:
Hey,

I'm facing a issue which I've seen before;

```
Error from server (InternalError): error when creating "runner.yaml": Internal error occurred: failed calling webhook "mutate.runnerdeployment.actions.summerwind.dev": Post "https://actions-runner-controller-webhook.actions-runner-system.svc:443/mutate-actions-summerwind-dev-v1alpha1-runnerdeployment?timeout=30s": service "actions-runner-controller-webhook" not found
```
controller-manager logs:
```
2021-02-19T10:56:29.669Z ERROR controller-runtime.controller Reconciler error {"controller": "runner", "request": "actions-runner-system/jellow-runner-dqgq8-qb2lw", "error": "Internal error occurred: failed calling webhook \"mutate.runner.actions.summerwind.dev\": Post \"https://actions-runner-controller-webhook.actions-runner-system.svc:443/mutate-actions-summerwind-dev-v1alpha1-runner?timeout=30s\": service \"actions-runner-controller-webhook\" not found"}
github.com/go-logr/zapr.(*zapLogger).Error
 /go/pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:128
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
 /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.4.0/pkg/internal/controller/controller.go:258
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
 /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.4.0/pkg/internal/controller/controller.go:232
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker
 /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.4.0/pkg/internal/controller/controller.go:211
k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1
 /go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190913080033-27d36303b655/pkg/util/wait/wait.go:152
k8s.io/apimachinery/pkg/util/wait.JitterUntil
 /go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190913080033-27d36303b655/pkg/util/wait/wait.go:153
k8s.io/apimachinery/pkg/util/wait.Until
[...]

Issue #293, opened 2021-02-05
connection reset by peer


rovin-ms:
After creating the runner.yaml the following error occurs when applying:

Error from server (InternalError): error when creating "runner.yaml": Internal error occurred: failed calling webhook "mutate.runner.actions.summerwind.dev": Post "https://webhook-service.actions-runner-system.svc:443/mutate-actions-summerwind-dev-v1alpha1-runner?timeout=30s": read unix @->/socket/konnectivity: read: connection reset by peer

mumoshu:
@rovin-ms I'm not really sure if this is really an error due to the controller.

Generally speaking, the connection-reset-by-peer error can happen when kubectl established a tcp session via iptables configured by kube-proxy via the K8s service, but the peer(maybe iptables or the controller pod behind the K8s service or the node) rejected it as unexpected, right?

Are your `controller-manager` pods ready and running? Do you have any firewall rule in between?(like seen in https://github.com/summerwind/actions-runner-controller/issues/68)?

rovin-ms:
No @mumoshu , the pods aren't running because of this issue: 

Similar issue with a new cluster:

kubectl apply -f org_runner.yaml

Error from server (InternalError): error when creating "org_runner.yaml": Internal error occurred: failedcalling webhook "mutate.runner.actions.summerwind.dev": Post "https://webhook-service.actions-runner-system.svc:443/mutate-actions-summerwind-dev-v1alpha1-runner?timeout=30s": context deadline exceeded

mumoshu:
@rovin-ms Sorry if I'm missing something, but could you confirm that you do see the difference between the runner pods and controller-manager pods?

You usually install the controller with https://github.com/summerwind/actions-runner-controller#installation and runners and/or runner deployments https://github.com/summerwind/actions-runner-controller#usage, and what you reported seems to indicate that the latter is failing due to the former. Investigating `controller-manager` logs may help finding any fatal error in the former.

Issue #68, opened 2020-07-06
Failed calling webhook...context deadline exceeded


raj-saxena:
Hi, 

I have been trying to get this working for some time now referring the guide in the README but it always fails at the last step. 
I have tried both the approaches of authentication - Github App as well as PAT.

I tried installing [Runners](https://github.com/summerwind/actions-runner-controller#repository-runners) as well as [RunnerDeployments](https://github.com/summerwind/actions-runner-controller#runnerdeployments) but the step fails with similar errors

### Here are the steps that I followed - 

```bash
# Install cert-manager
kubectl create namespace cert-manager
helm repo add jetstack https://charts.jetstack.io
helm repo update
helm install \
  cert-manager jetstack/cert-manager \
  --namespace actions-runner-system \
  --version v0.15.1 \
  --set installCRDs=true

# Create necessary resources in actions-runner-system namespace
kubectl apply -f https://github.com/summerwind/actions-runner-controller/releases/latest/download/actions-runner-controller.yaml

# Create secret with github PAT
kubectl create secret generic controller-manager \
    -n actions-runner-system \
    --from-literal=github_token=<GH_PAT>

# Copied the Runner.yml file from the example and changed the repo name
kubectl apply -f Runner.yaml

# Error for Runner
Error from server (InternalError): error when creating "Runner.yaml": Internal error occurred: failed calling webhook "mutate.runner.actions.summerwind.dev": Post https://webhook-service.actions-runner-system.svc:443/mutate-actions-summerwind-dev-v1alpha1-runner?timeout=30s: context deadline exceeded

# Error for RunnerDeployment
Error from server (InternalError): error when creating "RunnerDeployment.yaml": Internal error occurred: failed calling webhook "mutate.runnerdeployment.actions.summerwind.dev": Post https://webhook-service.actions-runner-system.svc:443/mutate-actions-summerwind-dev-v1alpha1-runnerdeployment?timeout=30s: context deadline exceeded
```

If it is relevant, I am on Google Kubernetes Engine and we run `nginx-ingress` for ingress management instead of default LoadBalancer. Do I need to add some firewall rule or ingress rule?

mumoshu:
The communication between the k8s api and the addmission webhook should be in-cluster as we can see the in-cluster domain is used in `https://webhook-service.actions-runner-system.svc:443`, which is also our default setup.

[...]

PR #1010, opened 2021-12-14
Add GKE firewall issues to common errors

Jacse:
A lot of people have issues with private GKE clusters and it seems they are all solved by setting up a firewall policy. I think it would be relevant to include this in a troubleshooting-section since so many people are searching around issues for it. I myself just spent most of my day trying to figure it out.

Issues where this is the solution:
* https://github.com/actions-runner-controller/actions-runner-controller/issues/293
* https://github.com/actions-runner-controller/actions-runner-controller/issues/335
* https://github.com/actions-runner-controller/actions-runner-controller/issues/68

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ ToC:
   - [Software Installed in the Runner Image](#software-installed-in-the-runner-image)
   - [Using without cert-manager](#using-without-cert-manager)
   - [Common Errors](#common-errors)
+- [Troubleshooting](#troubleshooting)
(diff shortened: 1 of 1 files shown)


mumoshu (approved):
LGTM. Thank you for your contribution!

We'd like to move those common errors into a dedicated trouble-shooting guide like one at https://github.com/shopsys/shopsys/blob/master/docs/docker/docker-troubleshooting.md so let me add a follow-up commit/pr to do that later.

PR #1012, opened 2021-12-16
Support runner groups with selected visibility in webhooks autoscaler

fgalind1:
The current implementation doesn't support yet runner groups with custom visibility (e.g selected repositories only). If there are multiple runner groups with selected visibility - not all runner groups may be a potential target to be scaled up. Thus this PR introduces support to allow having runner groups with selected visibility. This requires to query GitHub API to find what are the potential runner groups that are linked to a specific repository (whether using visibility all or selected).

This also improves resolving the `scaleTargetKey` that are used to match an HRA based on the inputs of the `RunnerSet`/`RunnerDeployment` spec to better support for runner groups.

This requires to configure github auth in the webhook server, to keep backwards compatibility if github auth is not provided to the webhook server, this will assume all runner groups have no selected visibility and it will target any available runner group as before

### DIFF
--- a/cmd/githubwebhookserver/main.go
+++ b/cmd/githubwebhookserver/main.go
@@ -28,6 +28,8 @@ import (
 
 	actionsv1alpha1 "github.com/actions-runner-controller/actions-runner-controller/api/v1alpha1"
 	"github.com/actions-runner-controller/actions-runner-controller/controllers"
+	"github.com/actions-runner-controller/actions-runner-controller/github"
+	"github.com/kelseyhightower/envconfig"
 	zaplib "go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
@@ -76,7 +78,16 @@ func main() {
 		enableLeaderElection bool
 		syncPeriod           time.Duration
 		logLevel             string
+
+		ghClient *github.Client
 	)
+
+	var c github.Config
+	err = envconfig.Process("github", &c)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error: processing environment variables: %v\n", err)
+		os.Exit(1)
+	}
 
 	webhookSecretTokenEnv = os.Getenv(webhookSecretTokenEnvName)
 
@@ -88,6 +99,11 @@ func main() {
 	flag.DurationVar(&syncPeriod, "sync-period", 10*time.Minute, "Determines the minimum frequency at which K8s resources managed by this controller are reconciled. When you use autoscaling, set to a lower value like 10 minute, because this corresponds to the minimum time to react on demand change")
 	flag.StringVar(&logLevel, "log-level", logLevelDebug, `The verbosity of the logging. Valid values are "debug", "info", "warn", "error". Defaults to "debug".`)
 	flag.StringVar(&webhookSecretToken, "github-webhook-secret-token", "", "The personal access token of GitHub.")
+	flag.StringVar(&c.Token, "github-token", c.Token, "The personal access token of GitHub.")
+	flag.Int64Var(&c.AppID, "github-app-id", c.AppID, "The application ID of GitHub App.")
+	flag.Int64Var(&c.AppInstallationID, "github-app-installation-id", c.AppInstallationID, "The installation ID of GitHub App.")
+	flag.StringVar(&c.AppPrivateKey, "github-app-private-key", c.AppPrivateKey, "The path of a private key file to authenticate as a GitHub App")
+
 	flag.Parse()
 
 	if webhookSecretToken == "" && webhookSecretTokenEnv != "" {
@@ -121,6 +137,15 @@ func main() {
 		}
 	})
 
+	if len(c.Token) > 0 || (c.AppID > 0 && c.AppInstallationID > 0 && c.AppPrivateKey != "") {
+		ghClient, err = c.NewClient()
+		if err != nil {
+			fmt.Fprintln(os.Stderr, "Error: Client creation failed.", err)
+			setupLog.Error(err, "unable to create controller", "controller", "Runner")
+			os.Exit(1)
+		}
+	}
+
 	ctrl.SetLogger(logger)
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
@@ -143,6 +168,7 @@ func main() {
 		Scheme:         mgr.GetScheme(),
 		SecretKeyBytes: []byte(webhookSecretToken),
 		Namespace:      watchNamespace,
+		GitHubClient:   ghClient,
 	}
 
 	if err = hraGitHubWebhook.SetupWithManager(mgr); err != nil {
--- a/controllers/horizontal_runner_autoscaler_webhook.go
+++ b/controllers/horizontal_runner_autoscaler_webhook.go
@@ -44,6 +44,7 @@ const (
 	scaleTargetKey = "scaleTarget"
 
 	keyPrefixEnterprise = "enterprises/"
+	keyRunnerGroup      = "/group/"
 )
 
 // HorizontalRunnerAutoscalerGitHubWebhook autoscales a HorizontalRunnerAutoscaler and the RunnerDeployment on each
@@ -534,7 +535,7 @@ func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) getScaleUpTargetWithF
 	}
 
 	for _, group := range organizationGroups {
-		if target, err := scaleTarget(owner + "/group/" + group); err != nil {
+		if target, err := scaleTarget(organizationalRunnerGroupKey(owner, group)); err != nil {
 			log.Error(err, "finding organizational runner group", "organization", owner)
 			return nil, err
 		} else if target != nil {
@@ -544,7 +545,7 @@ func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) getScaleUpTargetWithF
 	}
 
 	for _, group := range enterpriseGroups {
-		if target, err := scaleTarget(enterpriseKey(enterprise) + "/group/" + group); err != nil {
+		if target, err := scaleTarget(enterpriseRunnerGroupKey(enterprise, group)); err != nil {
 			log.Error(err, "finding enterprise runner group", "enterprise", owner)
 			return nil, err
 		} else if target != nil {
@@ -805,14 +806,14 @@ func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) SetupWithManager(mgr 
 			}
 			if rd.Spec.Template.Spec.Organization != "" {
 				if group := rd.Spec.Template.Spec.Group; group != "" {
-					keys = append(keys, rd.Spec.Template.Spec.Organization+"/group/"+rd.Spec.Template.Spec.Group) // Organization runner groups
+					keys = append(keys, organizationalRunnerGroupKey(rd.Spec.Template.Spec.Organization, rd.Spec.Template.Spec.Group)) // Organization runner groups
 				} else {
 					keys = append(keys, rd.Spec.Template.Spec.Organization) // Organization runners
 				}
 			}
 			if enterprise := rd.Spec.Template.Spec.Enterprise; enterprise != "" {
 				if group := rd.Spec.Template.Spec.Group; group != "" {
-					keys = append(keys, enterpriseKey(enterprise)+"/group/"+rd.Spec.Template.Spec.Group) // Enterprise runner groups
+					keys = append(keys, enterpriseRunnerGroupKey(enterprise, rd.Spec.Template.Spec.Group)) // Enterprise runner groups
 				} else {
 					keys = append(keys, enterpriseKey(enterprise)) // Enterprise runners
 				}
@@ -833,13 +834,13 @@ func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) SetupWithManager(mgr 
 			if rs.Spec.Organization != "" {
 				keys = append(keys, rs.Spec.Organization) // Organization runners
 				if group := rs.Spec.Group; group != "" {
-					keys = append(keys, rs.Spec.Organization+"/group/"+rs.Spec.Group) // Organization runner groups
+					keys = append(keys, organizationalRunnerGroupKey(rs.Spec.Organization, rs.Spec.Group)) // Organization runner groups
 				}
 			}
 			if enterprise := rs.Spec.Enterprise; enterprise != "" {
 				keys = append(keys, enterpriseKey(enterprise)) // Enterprise runners
 				if group := rs.Spec.Group; group != "" {
-					keys = append(keys, enterpriseKey(enterprise)+"/group/"+rs.Spec.Group) // Enterprise runner groups
+					keys = append(keys, enterpriseRunnerGroupKey(enterprise, rs.Spec.Group)) // Enterprise runner groups
 				}
 			}
 			autoscaler.Log.V(1).Info(fmt.Sprintf("HRA keys indexed for HRA %s: %v", hra.Name, keys))
@@ -859,4 +860,12 @@ func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) SetupWithManager(mgr 
 
 func enterpriseKey(name string) string {
 	return keyPrefixEnterprise + name
+}
+
+func organizationalRunnerGroupKey(owner, group string) string {
+	return owner + keyRunnerGroup + group
+}
+
+func enterpriseRunnerGroupKey(enterprise, group string) string {
+	return keyPrefixEnterprise + enterprise + keyRunnerGroup + group
 }
--- a/controllers/horizontal_runner_autoscaler_webhook.go
+++ b/controllers/horizontal_runner_autoscaler_webhook.go
@@ -37,6 +37,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/actions-runner-controller/actions-runner-controller/api/v1alpha1"
+	"github.com/actions-runner-controller/actions-runner-controller/github"
 )
 
 const (
@@ -56,6 +57,9 @@ type HorizontalRunnerAutoscalerGitHubWebhook struct {
 	// SecretKeyBytes is the byte representation of the Webhook secret token
 	// the administrator is generated and specified in GitHub Web UI.
 	SecretKeyBytes []byte
+
+	// GitHub Client to discover runner groups assigned to a repository
+	GitHubClient *github.Client
 
 	// Namespace is the namespace to watch for HorizontalRunnerAutoscaler's to be
 	// scaled on Webhook.
@@ -489,10 +493,12 @@ func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) getScaleUpTarget(ctx 
 func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) getJobScaleUpTargetForRepoOrOrg(
 	ctx context.Context, log logr.Logger, repo, owner, ownerType, enterprise string, labels []string,
 ) (*ScaleTarget, error) {
+
 	repositoryRunnerKey := owner + "/" + repo
 
+	// Search for repository HRAs
 	if target, err := autoscaler.getJobScaleTarget(ctx, repositoryRunnerKey, labels); err != nil {
-		log.Info("finding repository-wide runner", "repository", repositoryRunnerKey)
+		log.Error(err, "finding repository-wide runner", "repository", repositoryRunnerKey)
 		return nil, err
 	} else if target != nil {
 		log.Info("job scale up target is repository-wide runners", "repository", repo)
@@ -500,34 +506,41 @@ func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) getJobScaleUpTargetFo
 	}
 
 	if ownerType == "User" {
-		log.V(1).Info("no repository runner found", "organization", owner)
-
+		log.V(1).Info("user repositories not supported", "owner", owner)
 		return nil, nil
 	}
 
+	// Search for organization runner HRAs in default runner group
 	if target, err := autoscaler.getJobScaleTarget(ctx, owner, labels); err != nil {
-		log.Info("finding organizational runner", "organization", owner)
+		log.Error(err, "finding organizational runner", "organization", owner)
 		return nil, err
 	} else if target != nil {
 		log.Info("job scale up target is organizational runners", "organization", owner)
 		return target, nil
 	}
 
-	if enterprise == "" {
-		log.V(1).Info("no repository runner or organizational runner found",
-			"repository", repositoryRunnerKey,
-			"organization", owner,
-		)
-		return nil, nil
+	if enterprise != "" {
+		// Search for enterprise runner HRAs in default runner group
+		if target, err := autoscaler.getJobScaleTarget(ctx, enterpriseKey(enterprise), labels); err != nil {
+			log.Error(err, "finding enterprise runner", "enterprise", enterprise)
+			return nil, err
+		} else if target != nil {
+			log.Info("scale up target is default enterprise runners", "enterprise", enterprise)
+			return target, nil
+		}
 	}
 
-	if target, err := autoscaler.getJobScaleTarget(ctx, enterpriseKey(enterprise), labels); err != nil {
-		log.Error(err, "finding enterprise runner", "enterprise", enterprise)
+	// At this point there were no default organization/enterprise runners available to use, try now
+	// searching in runner groups
+
+	// We need to get the potential runner groups first to avoid spending API queries needless. Once/if GitHub improves an
+	// API to find related/linked runner groups from a specific repository this logic could be removed
+	availableEnterpriseGroups, availableOrganizationGroups, err := autoscaler.getPotentialGroupsFromHRAs(ctx, enterprise, owner)
+	if err != nil {
+		log.Error(err, "finding potential organization runner groups from HRAs", "organization", owner)
 		return nil, err
-	} else if target != nil {
-		log.Info("scale up target is enterprise runners", "enterprise", enterprise)
-		return target, nil
-	} else {
+	}
+	if len(availableEnterpriseGroups) == 0 && len(availableOrganizationGroups) == 0 {
 		log.V(1).Info("no repository/organizational/enterprise runner found",
 			"repository", repositoryRunnerKey,
 			"organization", owner,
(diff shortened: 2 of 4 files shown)


controllers/horizontal_runner_autoscaler_webhook.go
```diff
@@ -489,55 +493,158 @@ func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) getScaleUpTarget(ctx
...
+	} else {
+		// For backwards compatibility if GitHub authentication is not configured, we assume all runner groups have
+		// visibility=all to honor the previous implementation, therefore any available enterprise/organization runner
+		// is a potential target for scaling
+		enterpriseGroups = availableEnterpriseGroups
+		organizationGroups = availableOrganizationGroups
+	}
+
+	for _, group := range organizationGroups {
+		if target, err := autoscaler.getJobScaleTarget(ctx, owner+"/"+group, labels); err != nil {
```
mumoshu, line 572: I think this can break when there were both (1) a repository named `group` in the org and (2) runner group named `group` in the organization. If that's the case we might better use a dedicated key prefix for this case?

fgalind1: yep noticed that same I'm preparing a new commit where I'm using for runner groups `owner+"/group/"+group`

fgalind1: Added in 200518c45b9ce48872075536d2d0f5463c3c9d0e. I've also try to remove some duplicate code between `getScaleUpTarget` and `getJobScaleUpTargetForRepoOrOrg`

mumoshu: Ideally, logic change and refactoring should be a separate commit for easy reviews but let me try!
Thanks anyway for your prompt response. I'll do a full review in a day or two!

fgalind1: Awesome - thanks @mumoshu

controllers/horizontal_runner_autoscaler_webhook.go
```diff
@@ -716,31 +794,55 @@ func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) SetupWithManager(mgr
...
 
-			keys := []string{rd.Spec.Template.Spec.Repository, rd.Spec.Template.Spec.Organization}
-
+			keys := []string{}
+			if rd.Spec.Template.Spec.Repository != "" {
+				keys = append(keys, rd.Spec.Template.Spec.Repository) // Repository runners
+			}
+			if rd.Spec.Template.Spec.Organization != "" {
+				if group := rd.Spec.Template.Spec.Group; group != "" {
+					keys = append(keys, rd.Spec.Template.Spec.Organization+"/group/"+rd.Spec.Template.Spec.Group) // Organization runner groups
```
mumoshu, line 808:
```suggestion
					keys = append(keys, organizationalRunnerGroupKey(organizatrd.Spec.Template.Spec.Organization, rd.Spec.Template.Spec.Group))
```

Would you mind extracting this logic to a `enterpriseKey`-like function so that it can be used from within `controllers/horizontal_runner_autoscaler_webhook.go` too and that makes future refactoring easier?

controllers/horizontal_runner_autoscaler_webhook.go
```diff
@@ -716,31 +794,55 @@ func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) SetupWithManager(mgr
...
+				} else {
+					keys = append(keys, rd.Spec.Template.Spec.Organization) // Organization runners
+				}
+			}
 			if enterprise := rd.Spec.Template.Spec.Enterprise; enterprise != "" {
-				keys = append(keys, enterpriseKey(enterprise))
+				if group := rd.Spec.Template.Spec.Group; group != "" {
+					keys = append(keys, enterpriseKey(enterprise)+"/group/"+rd.Spec.Template.Spec.Group) // Enterprise runner groups
```
mumoshu, line 815:
```suggestion
					keys = append(keys, enterpriseRunnerGroupKey(enterprise, rd.Spec.Template.Spec.Group)) // Enterprise runner groups
```

Same as above

controllers/horizontal_runner_autoscaler_webhook.go
```diff
@@ -716,31 +794,55 @@ func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) SetupWithManager(mgr
...
+				keys = append(keys, rs.Spec.Organization) // Organization runners
+				if group := rs.Spec.Group; group != "" {
+					keys = append(keys, rs.Spec.Organization+"/group/"+rs.Spec.Group) // Organization runner groups
+				}
+			}
 			if enterprise := rs.Spec.Enterprise; enterprise != "" {
-				keys = append(keys, enterpriseKey(enterprise))
+				keys = append(keys, enterpriseKey(enterprise)) // Enterprise runners
+				if group := rs.Spec.Group; group != "" {
+					keys = append(keys, enterpriseKey(enterprise)+"/group/"+rs.Spec.Group) // Enterprise runner groups
```
mumoshu, line 842: This can be even more nicer if you could use https://github.com/actions-runner-controller/actions-runner-controller/pull/1012/files#r771778182

controllers/horizontal_runner_autoscaler_webhook.go
```diff
@@ -716,31 +794,55 @@ func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) SetupWithManager(mgr
...
-			keys := []string{rs.Spec.Repository, rs.Spec.Organization}
-
+			keys := []string{}
+			if rs.Spec.Repository != "" {
+				keys = append(keys, rs.Spec.Repository) // Repository runners
+			}
+			if rs.Spec.Organization != "" {
+				keys = append(keys, rs.Spec.Organization) // Organization runners
+				if group := rs.Spec.Group; group != "" {
+					keys = append(keys, rs.Spec.Organization+"/group/"+rs.Spec.Group) // Organization runner groups
```
mumoshu, line 836: It would be nicer if you could also use https://github.com/actions-runner-controller/actions-runner-controller/pull/1012/files#r771778109 here!

github/github.go
```diff
@@ -176,6 +176,93 @@ func (c *Client) ListRunners(ctx context.Context, enterprise, org, repo string)
...
+	var enterpriseRunnerGroups []string
+	var orgRunnerGroups []string
+
+	if org != "" {
+		runnerGroups, err := c.getOrganizationRunnerGroups(ctx, org, repo)
+		if err != nil {
+			return enterpriseRunnerGroups, orgRunnerGroups, err
+		}
+		for _, runnerGroup := range runnerGroups {
+			if runnerGroup.GetInherited() { // enterprise runner groups
```
mumoshu, line 190: Ah, so this is how you could see if an organizational runner group is coming from an enterprise or not, right? Good to know!

controllers/horizontal_runner_autoscaler_webhook.go
```diff
@@ -586,6 +658,9 @@ HRA:
 
 			// Ensure that the RunnerSet-managed runners have all the labels requested by the workflow_job.
 			for _, l := range labels {
+				if l == "self-hosted" {
```
mumoshu, line 661: I recently learned that this isn't quite true. `self-hosted` is included in a `workflow_job` event because you had included it in `runs-on` of your workflow definition yaml. So if we were to make it clear, you'd better do 

```suggestion
				if len(rs.Spec.Labels) == 0 && l == "self-hosted" {
```

So that it's clearer that actions-runner-controller just assumes the default RunnerSet.Spec.Labels and RunnerDeployment.Spec.Labels as `["self-hosted"]` if empty.

cmd/githubwebhookserver/main.go
```diff
@@ -121,6 +137,15 @@ func main() {
 		}
 	})
 
+	if len(c.Token) > 0 || (c.AppID > 0 && c.AppInstallationID > 0 && c.AppPrivateKey != "") {
+		ghClient, err = c.NewClient()
+		if err != nil {
+			fmt.Fprintln(os.Stderr, "Error: Client creation failed.", err)
```
mumoshu, line 143: Probably we can make this even more nicer by adding more info on what were there and not, like `Error: Client creation failed. AppIP is provided but Token, ApplicationID, AppPrivateKey were missing. You need to either set Token, or all of AppID, AppInstallationID and AppPrivateKey`. But let's do that in another pr :)

mumoshu: Nevermind! I just misunderstood when this error occurs.

github/github.go
fgalind1: yes - that matches with the right information based on some testing in our enterprise 👍

controllers/horizontal_runner_autoscaler_webhook.go
fgalind1: good one - added in a7014dc9d80f2f874316e81ac886c2731d839e95

controllers/horizontal_runner_autoscaler_webhook.go
fgalind1: added in a7014dc

controllers/horizontal_runner_autoscaler_webhook.go
fgalind1: yes - that's right `self-hosted` is included in `workflow_job`. However the proposed approach doesn't cover other examples, e.g if I have a workflow with `runs-on: ["self-hosted", "group-a"]` and I have `RunnerDeployment.Spec.Labels = ["group-a"]`, then job won't be scheduled. The only way will be to define `RunnerDeployment.Spec.Labels = ["self-hosted", "group-a"]` although that seems weird for two reasons:
1. all runners in this controller are self-hosted - if you register a runner without self-hosted in Labels it will be useless as it will never match a self-hosted runner.
2. `self-hosted` is a default label added by actions/runner which is not specified during registration - I believe in `Spec.Labels` we should set the custom labels that are going to be passed via config.sh which is not the case for self-hosted.

Any thoughts?

mumoshu: @fgalind1 Thanks for helping me shape this!

I believe this discussion relates to #953 which stems from the same thought as yours.

> self-hosted is a default label added by actions/runner which is not specified during registration

So the biggest thing I missed was this one!

I was blindly assuming that `self-hosted` is just a standard label that everyone uses just by convention. If this is the hard-coded default used by the actions/runner agents as well, actions-runner-controller should treat it as such too.

I'm going to merge #953 first as it's submitted earlier and after that, I'm fine with this change being included in this PR!

mumoshu (approved):
LGTM. Thank you for your contribution! I'm merging this.

We still lack a comprehensive E2E test suite so an awesome but relatively large change like this one may break things in an unexpected way.
So I'd appreciate it a lot if you could keep testing this for a while by perhaps building and deploying actions-runner-controller built from the latest main branch after merging!
Thanks in advance for our contribution and support ☺️

toast-gear:
@fgalind1 could I ask you what your driver was for introducing this change?

mumoshu:
@fgalind1 I've disucssed with @toast-gear and we now consider this is not the right way to do it.

AFAIK, a runner group works like an additional visibility toggle in addition to labels. So GitHub filters target runners by both labels and groups. But in this PR, you scaled "all" the runner groups associated to a repo/org/enterprisea regardless of labels, which doesn't aligh with the github actions behavior.

The end result seems to be that after this change ARC would scale up unnecessary runner groups, like runner groups that are visible to the repo/org/enterprise but don't match workflow job labels.

A better support for runner group visibilities would be to make ARC skip a RunnerDeployment/RunnerSet that found via `scaleTarget` func IF it is associated to a runner group and the group is invisible to the target repo/org/enterprise.

mumoshu:
Linking #1055

controllers/horizontal_runner_autoscaler_webhook.go
```diff
@@ -436,106 +441,174 @@ func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) getScaleTarget(ctx co
...
 		return nil, nil
 	}
 
-	if target, err := autoscaler.getScaleTarget(ctx, owner, f); err != nil {
-		log.Info("finding organizational runner", "organization", owner)
+	// Search for organization runner HRAs in default runner group
+	if target, err := scaleTarget(owner); err != nil {
+		log.Error(err, "finding organizational runner", "organization", owner)
 		return nil, err
 	} else if target != nil {
```
mumoshu, line 483: I believe what we had to do instead was to skip this `target` if it was associated to a runner group and that runner group was indivisible to this repository.

controllers/horizontal_runner_autoscaler_webhook.go
```diff
@@ -436,106 +441,174 @@ func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) getScaleTarget(ctx co
...
-		)
-
-		return nil, nil
+	if enterprise != "" {
+		// Search for enterprise runner HRAs in default runner group
+		if target, err := scaleTarget(enterpriseKey(enterprise)); err != nil {
+			log.Error(err, "finding enterprise runner", "enterprise", enterprise)
+			return nil, err
+		} else if target != nil {
+			log.Info("scale up target is default enterprise runners", "enterprise", enterprise)
```
mumoshu, line 494: As similar to https://github.com/actions-runner-controller/actions-runner-controller/pull/1012/files#r785624849, I believe what we had to head instead would be to skip this `target`(=a runnerdeployment that is associated to an enterprise) if it was associated to a runner group defined at enterprise-level and the group was configured to be invisible to this organization.

github/github.go
```diff
@@ -176,6 +176,93 @@ func (c *Client) ListRunners(ctx context.Context, enterprise, org, repo string)
...
+				}
+				if runnerGroup.GetVisibility() == "all" {
+					enterpriseRunnerGroups = append(enterpriseRunnerGroups, runnerGroup.GetName())
+				} else {
+					hasAccess, err := c.hasRepoAccessToOrganizationRunnerGroup(ctx, org, runnerGroup.GetID(), repo)
+					if err != nil {
+						return enterpriseRunnerGroups, orgRunnerGroups, err
+					}
+					if hasAccess {
+						enterpriseRunnerGroups = append(enterpriseRunnerGroups, runnerGroup.GetName())
```
mumoshu, line 202: `hasRepoAccessToOrganizationRunnerGroup` seems to check if a organizational runner group is available to the repository so this should be used to determine if you need to add this runner group to `orgRunnerGroups`, rather than `enterpriseRunnerGroups`.

controllers/horizontal_runner_autoscaler_webhook.go
```diff
@@ -716,31 +795,55 @@ func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) SetupWithManager(mgr
...
+				if group := rd.Spec.Template.Spec.Group; group != "" {
+					keys = append(keys, organizationalRunnerGroupKey(rd.Spec.Template.Spec.Organization, rd.Spec.Template.Spec.Group)) // Organization runner groups
+				} else {
+					keys = append(keys, rd.Spec.Template.Spec.Organization) // Organization runners
+				}
+			}
 			if enterprise := rd.Spec.Template.Spec.Enterprise; enterprise != "" {
-				keys = append(keys, enterpriseKey(enterprise))
+				if group := rd.Spec.Template.Spec.Group; group != "" {
+					keys = append(keys, enterpriseRunnerGroupKey(enterprise, rd.Spec.Template.Spec.Group)) // Enterprise runner groups
```
mumoshu, line 816: It turns out we don't need to find HRA by runner groups as GitHub finds the runner to scheduled a job by labels anyway.
What we need instead would be to list all the HRAs whose RunnerDeployment matches the workflow_job labels and filter out those that are associated to runner groups (both organization and enterprise groups) that are unavailable to the org / the repo.

fgalind1:
> @fgalind1 could I ask you what your driver was for introducing this change?

In our company we use enterprise and organization runner groups with selected visibility. You can think of a runner group is mapped to a certain team who owns a few repos in my company and only that team/repos should use that runner group. Prior this change if a webhook was coming from repo X, it could potentially scale up a runner group with selected visibility that doesn't include that repo

controllers/horizontal_runner_autoscaler_webhook.go
fgalind1: @mumoshu but you can use groups without labels, you coukd set runner groups with custom visibility and no custom labels right? (That's a common pattern in our company) if we dont take group into consideration, we could scale up the wrong runner group

toast-gear: That's not really the design philosophy behind runner groups though. If Github had intended on them as a means for defining exclusive access they would have:

1. included the repository runner group membership information in the workflow_jobs event whilst queued
2. enabled you to remove repositories from being members of the `Default` runner group.

Runner groups are a means for _extending_ what runners are avaliable to a repository via provided access via labels. They aren't designed to be a way of defining placement of runners beyond opening up access different classes of runners via labels which in turn implicitly defines placement.

fgalind1: > 1. included the repository runner group membership information in the workflow_jobs event whilst queued

This is one of the points that we're dicussing with GitHub (still pending) to potentially include that information in the webhook payload and have the full support as they do support custom visibility. That could simplify some of the logic here to avoid querying GitHub API and find the matching runner group

> 2. enabled you to remove repositories from being members of the `Default` runner group.

You can not add runners to the default runner group (this is what we're doing) that way qe preserve the intent of having only repository specific runner groups

> Runner groups are a means for _extending_ what runners are avaliable to a repository via labels. They aren't designed to be a way of defining placement of runners beyond opening up access different classes of runners via labels which in turn implicitly defines placement.

Uhm, that's why GitHub added the custom visibility isn't it? So that you could define some sort of protected runners yo a specific set of repositories. Labels won't do anything w.r.t protection or scoping some runners to some repositories only.

mumoshu: > @mumoshu but you can use groups without labels, you could set runner groups with custom visibility and no custom labels right? (That's a common pattern in our company) if we dont take group into consideration, we could scale up the wrong runner group

@fgalind1 Correct. But this seems like considering it in a wrong way. What's implemented in this PR only works when you only use a lot of runner groups with the same set of labels.

If you configure a different set of labels per runner groups, what's done in this PR scales wrong runner runnerdeployments.

mumoshu: Instead you'd need to do what I've explained in https://github.com/actions-runner-controller/actions-runner-controller/pull/1012#discussion_r785624849

fgalind1: > I believe what we had to do instead was to skip this target(=a runnerdepoyment that is associated to an organization) if it was associated to a runner group and that runner group was indivisible to this repository.

@mumoshu just to clarify with your previous comment, right now we find the potential HRA with the following precedence:

1. matching runner in repo from webhook
2. matching default runner in an org from webhook
3. matching default runner in enterprise from webhook
4. matching runner group in an org that matches repo from webhook and custom visibility of runner group
5. matching runner group in an org that matches repo from webhook and custom visibility of runner group

with your comment - are you suggesting to remove/skip 2 and 3? 

> it was associated to a runner group 

This is the part that wasn't clear to me what object where you referring to be associated to the runner group

mumoshu:
@fgalind1 Hey. I believe I understand your use case but I also see two problems:

1. This change seems to work as intended only when you don't differentiate labels per runner group. We'd need to make it work as intended regardless of you use custom labels with runner groups or not, as... GitHub allows it/I can imagine users who use runner groups with custom labels.

https://github.com/actions-runner-controller/actions-runner-controller/pull/1012#discussion_r787247036

2. This turns out to amplify the number of API calls per webhook

 2 can be addressed later with a big caution/enough documentation, but 1 needs to be addressed before the next release of ARC.

controllers/horizontal_runner_autoscaler_webhook.go
mumoshu: I'd say it should be like the below one instead:

1. matching repository-wide runner from the repository, the default org runner group, and the non-default org runner group that are visible to the repository, from webhook
2. matching organizational runner from the organization, the default enterprise runner group, and the non-default enterprise runner groups that are visible to the organization, from webhook
3. matching default runner in enterprise from webhook

You may already know but there're two variants of runner groups, organizational runner groups and enterprise runner groups. The visibility for the former can be controlled per repository, not organization. Similarly, the visibility for the latter can be controlled per organization under the enterprise.

fgalind1:
> This change seems to work as intended only when you don't differentiate labels per runner group. We'd need to make it work as intended regardless of you use custom labels with runner groups or not, as... GitHub allows it/I can imagine users who use runner groups with custom labels.

I'll take a look to see if this could be a simple fix/update to support that. So just to double-check, the example you're describing is that I could have 2 runner groups with different labels, and with the current implementation it may be possible that we could scale up (target) the runner group that doesn't match the webhook labels?

mumoshu:
> with the current implementation it may be possible that we could scale up (target) the runner group that doesn't match the webhook labels?

@fgalind1 Yes. That's exactly what I had in my mind!

mumoshu:
@fgalind1 Here's my note that may help fixing it: #1058

It isn't a working implementation of the fix but contains several TODO comments in places where I thought we'd add some code.

fgalind1:
@mumoshu I did some real-live testing with the current controller as-is and I'm not able to reproduce the incorrect behavior yet. Could you share the steps you have on your mind to reproduce the issue?

e.g I have:
- `HRA` and `RunnerDeployment` "foo" with enterprise=XXX, group: k8s-dev, labels: [fgalind1]
- Enterprise Runner group in GitHub with visibility to "`fgalind1-test`" repo

if I trigger a `workflow_dispatch` for `fgalind1-test` repo with a job containing only `self-hosted` label I get:
`Scale target not found. If this is unexpected, ensure that there is exactly one repository-wide or organizational`

if I trigger a `workflow_dispatch` for `fgalind1-test` repo with a job containing only `self-hosted, fgalind1` labels I properly scale the runner-group:
`scaled foo by 1`

Right now if we have N groups with custom visibility and custom labels, first we find all the potential candidates `autoscaler.getPotentialGroupsFromHRAs` and based on that we filter out the HRAs where its runner group doesn't match the labels

mumoshu:
Let's say you have:

- HRA and RunnerDeployment "foo" with enterprise=XXX, group: k8s-dev, labels: [fgalind1]
- HRA and RunnerDeployment "bar" with enterprise=XXX, labels: [fgalind1]
- Setup two repos, fgalind-foo-test and fgalind-bar-test
- Enterprise Runner group `k8s-dev` in GitHub with visibility to "fgalind-foo-test" repo only

What I had in my mind was that with the above setup, triggering a workflow run on `fgalind-bar-test` with `on: [fgalind1]` may result in scaling deployment `foo`, which doesn't seem correct.

fgalind1:
@mumoshu I tested the above example using the same input (just used org group instead of enterprise runner group) and got the following:

**initial setup**

foo runners (1 org runner + 1 runner group runner)
image

bar runners (1 org runner)
image

**the HRAs keys for the webhooks controller**
```
github-webhook-server 1.642610484236478e+09    DEBUG    controllers.Runner    HRA keys indexed for HRA bar: [XXX]
github-webhook-server 1.6426104842365727e+09    DEBUG    controllers.Runner    HRA keys indexed for HRA foo: [XXX/group/foo-group]   
```

**when I trigger a job in foo**
```
github-webhook-server 1.6426105445647357e+09    DEBUG    controllers.Runner    Found 0 HRAs by key    {"key": "XXX/personal.fgalind1.foo-test"}
github-webhook-server 1.6426105445647907e+09    DEBUG    controllers.Runner    Found 1 HRAs by key    {"key": "XXX"}
github-webhook-server 1.6426105445648148e+09    INFO    controllers.Runner    job scale up target is organizational runners    {"event": "workflow_job", "hookID": "", "delivery": "c71f9490-79
github-webhook-server 1.642610544564828e+09    INFO    controllers.Runner    Patching hra for capacityReservations update    {"before": [{"expirationTime":"2022-01-19T17:08:58Z","replicas":1}
github-webhook-server 1.6426105445775921e+09    INFO    controllers.Runner    scaled bar by 1         
```                

foo should scale up either of the deployments - it scaled up "bar"

**when I trigger a job in bar**
```
github-webhook-server 1.6426106025543697e+09    DEBUG    controllers.Runner    Found 0 HRAs by key    {"key": "XXX/personal.fgalind1.bar-test"}
github-webhook-server 1.6426106025544243e+09    DEBUG    controllers.Runner    Found 1 HRAs by key    {"key": "XXX"}
github-webhook-server 1.6426106025544486e+09    INFO    controllers.Runner    job scale up target is organizational runners    {"event": "workflow_job", "hookID": "", "delivery": "e9aeee70-79
github-webhook-server 1.6426106025554268e+09    INFO    controllers.Runner    Patching hra for capacityReservations update    {"before": [{"expirationTime":"2022-01-19T17:12:24Z","replicas":1
github-webhook-server 1.6426106025682225e+09    INFO    controllers.Runner    scaled bar by 1     
```

bar should scale up RunnerDeployment "bar" with org=XXX, labels: [fgalind1] and it scaled up "bar"

AFAIK logic is working fine as expected. as we're honoring labels just fine

mumoshu:
> bar should scale up RunnerDeployment "bar" with org=XXX, labels: [fgalind1] and it scaled up "bar"

Isn't that just a coincidence? ARC isn't really looking into the runner group when it's matching based on labels so it MAY still scale up `foo`(which is incorrect) from time to time.

mumoshu:
See this line:

https://github.com/actions-runner-controller/actions-runner-controller/blob/5e86881c30614aef7e5edbfdeb7854ada28101db/controllers/horizontal_runner_autoscaler_webhook.go#L440

ARC returns the first target when there were two or more scale targets found. For deployment "bar" ARC would find both `foo` and `bar` as scale targets, and if K8s API returned the RunnerDeployment `foo` earlier it would return it. Otherwise `bar` is returned. It's not deterministic.

What we'd need to do there would be to return the scale target with matching labels AND from a visible runner group, instead of blindly returning the first scale target with matching labels(which may or may not be coming from a visible runner group).

fgalind1:
okay I think I see the case of the discussion - this issue will happen if the "default runner group" uses also "selected repositories only". In the example I was testing bar was using "all repositories" but if I switch it to "selected repositories = bar) then yes the issue will occur. Working on the fix now

mumoshu:
I appreciate your efforts!

> this issue will happen if the "default runner group" uses also "selected repositories only"

Yeah if it switched to the group `bar` to "selected repositories = bar"` the issue will reproduce consistently (like 100%).
My point was though, it MAY still reproduce time to time even with "all repositories"`.

ARC should guard against both cases, and I guess, implementation-wise, somehow enhancing the `getScaleTarget` function could address both cases altogether.

fgalind1:
> I appreciate your efforts!
> 
> > this issue will happen if the "default runner group" uses also "selected repositories only"
> 
> Yeah if it switched to the group `bar` to "selected repositories = bar" the issue will reproduce consistently (like 100%). My point was though, it MAY still reproduce time to time even with "all repositories"`.
> 
> ARC should guard against both cases, and I guess, implementation-wise, somehow enhancing the `getScaleTarget` function could address both cases altogether.

@mumoshu let me know what do you think 😃 https://github.com/actions-runner-controller/actions-runner-controller/pull/1062

fgalind1:
> > I appreciate your efforts!
> > > this issue will happen if the "default runner group" uses also "selected repositories only"
> > 
> > 
> > Yeah if it switched to the group `bar` to "selected repositories = bar" the issue will reproduce consistently (like 100%). My point was though, it MAY still reproduce time to time even with "all repositories"`.
> > ARC should guard against both cases, and I guess, implementation-wise, somehow enhancing the `getScaleTarget` function could address both cases altogether.
> 
> @mumoshu let me know what do you think 😃 #1062. tested with the above flow using bar (selected repos and all repos) and everything looks good

@mumoshu just wonder if you had a chance to look at the improvements from #1062

mumoshu:
@fgalind1 I'll try my best to fully review your change this weekend. Thank you for your patience!

Issue #1007, opened 2021-12-14
Allow volumeStorageMedium for the 'work' mount


edspencer:
**Is your feature request related to a problem? Please describe.**
Yes - our runners grind to a halt, apparently because of the large amount of I/O performed by Yarn. We have a fairly standard Node JS + React application - on each push on any branch we run a series of jobs to build the Node JS API server artifact, the UI artifacts, run tests, etc.

Each job starts by running yarn install. As we're relatively package-heavy, this is extremely I/O intensive, making hundreds of thousands of reads and especially writes. This is partly due to Yarn not being particularly efficient, but that's another story, and we'd want this ability regardless.

We use the `volumeStorageMedium: 'Memory'` option to configure the `runner` mount to use the system's memory. This works great. I believe that if there was an equivalent config for the `work` mount, our jobs would speed up tremendously.

**Describe the solution you'd like**
I would like to be able to set the `Medium` of the `work` mount to `Memory`, just as I can for the `runner` mount using `volumeStorageMedium: 'Memory'`

**Describe alternatives you've considered**
Currently I'm only able to use the actions-runner-controller for 1 out of our dozen or so jobs, because running more than one at a time appears to saturate the underlying k8s node's drive with I/O requests. So our pipelines run pretty slowly because we have to rely on the github cloud-hosted runners for most 

**Additional context**

![Screen Shot 2021-12-13 at 11 28 39 PM](https://user-images.githubusercontent.com/6929/145933307-55e03c5e-207c-4d68-b97b-bf412064be19.png)

mumoshu:
@edspencer Hey! Thanks for reporting.

I think you can do that today, by setting `runner.spec.workDir` to somewhere else than the default `/runner/_work` and adding a corresponding pair of volume and volume mount definition to `runner.spec.volumes` and `runner.spec.volumeMounts` respectively.

If you do need to implement this, probably adding `workerVoumeStorageMedium` and use that from `runner_controller.go` would do the job.

That said, I'm not entirely sure if that's the right approach! AFAIK `medium: memory` consumes the node memory so you won't be able to serve a lot of runner pods on one node when your git repo and project dependencies(like node modules) can take GBs of a disk, which defeats the purpose of your original goal of making it possible for a single node to serve two or more runner pods.

Which cloud platform are you running actions-runner-controller on? I was going to say, if you're using e.g. AWS, you'd better use bigger EC2 instance types with more EBS throughput or with a NVMe SSD.

[...]

PR #1015, opened 2021-12-20
Allow providing the "work" volume(mount) as part of the RunnerDeployment

genisd:
This allows providing a different `work` Volume.

This should be a cloud agnostic way of allowing the operator to use (for example) NVME backed storage.

This is a working example where the workDir will use the provided volume, additionally here docker is placed on the same NVME.
```
kind: RunnerDeployment
metadata:
  name: runner-2
spec:
  template:
    spec:
      dockerdContainerResources: {}
      env:
      - name: POD_NAME
        valueFrom:
          fieldRef:
            fieldPath: metadata.name
      # this is to mount the docker in docker onto NVME disk
      dockerVolumeMounts:
      - mountPath: /var/lib/docker
        name: scratch
        subPathExpr: $(POD_NAME)-docker
      - mountPath: /runner/_work
        name: work
        subPathExpr: $(POD_NAME)-work
      volumeMounts:
      - mountPath: /runner/_work
        name: work
        subPathExpr: $(POD_NAME)-work
      dockerEnv:
      - name: POD_NAME
        valueFrom:
          fieldRef:
            fieldPath: metadata.name
      volumes:
      - hostPath:
          path: /mnt/disks/ssd0
        name: scratch
      - hostPath:
          path: /mnt/disks/ssd0
        name: work
      nodeSelector:
       cloud.google.com/gke-nodepool: runner-16-with-nvme
      ephemeral: false
      image: ""
      imagePullPolicy: Always
      labels:
        - runner-2
        - self-hosted
      organization: yourorganization

```


controllers/runner_controller.go
```diff
@@ -1123,3 +1148,21 @@ func removeFinalizer(finalizers []string, finalizerName string) ([]string, bool)
 
 	return result, removed
 }
+
+func workVolumePresent(items []corev1.Volume) (bool, int) {
```
genisd, line 1152: I could make this fn type agnostic, that it accepts `items []corev1.Volume` and `items []corev1.VolumeMount` like the fn below

controllers/runner_controller.go
```diff
@@ -671,10 +671,29 @@ func (r *RunnerReconciler) newPod(runner v1alpha1.Runner) (corev1.Pod, error) {
 	runnerSpec := runner.Spec
 
 	if len(runnerSpec.VolumeMounts) != 0 {
+		// if operater provides a work volume mount, use that
```
genisd, line 674: An alternative to stripping/removing the work VolumeMount here is simply adding it here if it's not provided by the user.
From reading the code, the functions where it's added should be like a "default" template, so I opted for this.

I tried skipping it in that default template, but the needed context is not always present.
The user provided volumes is not available, but easily refactored. Only issue I couldn't overcome was the "registration-only" codepath was somehow different.

genisd:
@mumoshu could you take a look at this please?
It's running on our own environment.

I think this solves https://github.com/actions-runner-controller/actions-runner-controller/issues/1007 as well, since now a full volume can be specified.

Like:
```
kind: RunnerDeployment
metadata:
  name: runner-2
spec:
  template:
    spec:
      volumes:
      - name: work
        emptyDir:
          medium: Memory
```

should work.

mumoshu:
@genisd Thanks for your patience and contribution! I'll review this in a day or two.

controllers/runner_controller.go
genisd: Once https://github.com/actions-runner-controller/actions-runner-controller/pull/1028 is merged I can simplify this pr and make it cleaner/neater.

mumoshu (approved):
I didn't really mean to call two weeks "a day or two" 😆 
After careful reading, this change looks straightforward enough and safe to be included in 0.21.0.
Sorry for overriding the milestone @toast-gear but I believe this is worth a lot!

mumoshu:
@genisd Thanks for your contribution! It would be great if you could add some documentation on this feature to our README, so that people suffering from issues like #1007 would easily find your awesome enhancement 😄

genisd:
Thank you, I'll definitely create a PR for documentation, it's indeed something I know i had not done yet.

I'm going to especially check things like if data gets left behind and such things, put a proper disclaimer next to the examples.
It's not an issue for us, but probably worth mentioning.

edspencer:
Thanks @genisd (and @mumoshu) - looking forward to abandoning my janky custom solution from https://github.com/actions-runner-controller/actions-runner-controller/issues/1007 in favor of your new magic when 0.21 comes out 🙌🏻

PR #1016, opened 2021-12-20
Remove duplicate self-hosted condition

fgalind1:
Duplicate condition caused after merge of #953 and #1012


mumoshu (approved):
LGTM. Thanks for your contribution as always @fgalind1 ☺️

PR #1017, opened 2021-12-21
feat: support routing GitHub API calls to custom proxy API

fgalind1:
GitHub currently has some limitations w.r.t permissions management on runner groups as they all require org admin, however at our company we're using runner groups to serve different internal teams (with different permissions), thus we needed to deploy a custom proxy API with our internal authentication to provide who has access to certain APIs depending on the repository/runner group on a given org/enterprise

This PR just allows to optionally send the GitHub API calls to an alternate custom proxy URL instead of cloud github (github.com) or an enterprise URL with basic authentication


github/github.go
```diff
@@ -78,6 +95,14 @@ func (c *Config) NewClient() (*Client, error) {
 	} else {
 		client = github.NewClient(httpClient)
 		githubBaseURL = "https://github.com/"
+		if len(c.ProxyUrl) > 0 {
+			proxyUrl, err := url.Parse(c.ProxyUrl)
+			if err != nil {
+				return nil, fmt.Errorf("proxy client creation failed: %v", err)
+			}
+			client.BaseURL = proxyUrl
+			client.UploadURL = proxyUrl
```
mumoshu, line 104: I believe the upload URL is hightly dependent on the environment.

The default base URL is `https://api.github.com/` and the upload URL is `https://uploads.github.com/` for GitHub Cloud. In GHES it's like `$GITHUB_ENDPOINT/api/uploads` or `api.$GITHUB_ENDPOINT` depending on the situation.

That said, would you mind adding another flag like `github-proxy-upload-url` for a more fine-grained configuration? The default value for that would be `github-proxy-url` + `/api/uploads/`.

mumoshu: Also, `githubBaserURL` should be assigned the same value as  `proxyURL` in this case, so that it is correctly propagated down to runners via `GITHUB_URL` env.

fgalind1: > That said, would you mind adding another flag like github-proxy-upload-url for a more fine-grained configuration?

good call - I'll add that - thanks

> Also, githubBaserURL should be assigned the same value as proxyURL in this case, so that it is correctly propagated down to runners via GITHUB_URL env.

This one I "intentionally" kept it as github.com as (in our example/infra) we still want runners to be registered with github.com (or we didn't had the need to have a custom proxy API for runners registration nor custom internal API calls triggered by the runner). What about having another flag `github-proxy-runner-url`  to let people chose if they want to use their custom proxy URL, github.com or whatever URL they would like to, they can configure it as needed, and probably default to proxy url if not defined/empty

fgalind1: let me know what do you think based on 7a976cc17b661a4d7c49e24a8c37549c752c5cb5

mumoshu: Thanks. Then, I'd name it `runner-github-url` so that it sounds more like it's basically `GITHUB_URL` only for runners :)

mumoshu: Also, I started to think it might communicate better if we omitted the `proxy` concept out of those names.

- `github-proxy-url`- can be just `github-url`. This is slightly more consistent with the existing `github-enterprise-url` imho
- `github-proxy-upload-url`- can also be just `github-upload-url`
- `github-proxy-runner-url` would better be `runner-github-url`
- `github-proxy-username` would better be `github-basicauth-username` so that it directly matches that of our HTTP client parameter
- `github-proxy-password` would better be `github-basicauth-password`. same as above

After all the new code path will look a bit different, like

```
client = github.NewClient(httpClient)
githubBaseURL = "https://github.com/"

if len(c.GitHubURL)  > 0 {
	baseUrl, err := url.Parse(c.GitHubURL)
	if err != nil {
		return nil, fmt.Errorf("proxy client creation failed: %v", err)
	}
	if !strings.HasSuffix(baseUrl.Path, "/") {
		baseUrl.Path += "/"
	}
	client.BaseURL = baseUrl
}
```

In the flag usage description one would introduce `github-basicauth-username` and `github-basicauth-password` are used only when you want the client to use basic auth instead of PAT or GitHub App authn in case you're running your own github API proxy.

I imagine there would be a github API proxy without even basic-auth. So an option to not enforce basic-auth when you're using proxy would also be nice.

So the below code needs a bit change from my perspective:

```
if len(c.ProxyUrl) > 0 {
		transport = ProxyTransport{BaseUrl: c.ProxyUrl, Username: c.ProxyUsername, Password: c.ProxyPassword}
	} else if len(c.Token) > 0 {
```

Perhaps it would better align with my idea if it was like:

```
if c.ProxyUsername != "" && c.ProxyPassword {
		transport = ProxyTransport{BaseUrl: c.ProxyUrl, Username: c.ProxyUsername, Password: c.ProxyPassword}
	} else if len(c.Token) > 0 {
```

Sorry for the loooong reply but I believe with some tweaks your change will be perfect! I appreciate your cooperation.

fgalind1: thanks for the feedback - the proposal looks good to me 👍 - I've done the renaming on 891f909dc5137b5dd5a54d10051ecfacc893b0d3

mumoshu:
@fgalind1 Thank you so much for your support! I've added cd99975 for minor tweaks but everything looks great to me now! I'll merge this once the build passes.

fgalind1:
> @fgalind1 Thank you so much for your support! I've added [cd99975](https://github.com/actions-runner-controller/actions-runner-controller/commit/cd99975e6a1fbb5fa6254222ebce6c058cb070cd) for minor tweaks but everything looks great to me now! I'll merge this once the build passes.

thanks @mumoshu  - just curious - I noticed that auth secrets values are using snake_case vs camelCase which is the standard for helm values.. was there any reason of choosing snake_case for those specific config values

mumoshu:
> I noticed that auth secrets values are using snake_case vs camelCase which (camelCase) is the standard for helm values.. was there any reason of choosing snake_case for those specific config values

@fgalind1 You found it! I kept it for backward-compatibility and consistency. Perhaps those originally slipped in when we first introduced the authSecret feature.

Honestly speaking, I don't like those being snake_case. So I'd be happy if there are some momentum to change it to lowerCamelCase in the future. We need to do it in a separate change, in a backward-compatible way though.

fgalind1:
> @fgalind1 You found it! I kept it for backward-compatibility and consistency. Perhaps those originally slipped in when we first introduced the authSecret feature.

Makes sense - I imagined it was some sort of carry over.. 

>  Honestly speaking, I don't like those being snake_case.

Totally +1 on that 👍 

> So I'd be happy if there are some momentum to change it to lowerCamelCase in the future. We need to do it in a separate change, in a backward-compatible way though.

Not speaking about this change but in general wonder what are your thoughts If the idea is to change these to camelCase at some point - I wonder if it's easier to drop the line for any new value introduced to be camelCase, regardless if others are values are already snake_case.. just to make it easier to transition at some point and have backward compatible changes just for less values

mumoshu:
> Not speaking about this change but in general wonder what are your thoughts If the idea is to change these to camelCase at some point - I wonder if it's easier to drop the line for any new value introduced to be camelCase, regardless if others are values are already snake_case.. just to make it easier to transition at some point and have backward compatible changes just for less values

I'm afraid I'm not reading the whole sentence correctly, but I think we're sharing the idea. I want any change to be backward-compatible. The only way to rename values fields then would be add lowerCamelCase versions of the existing fields, and make any missing lowerCamelCase field to fallback to it's snake_case variant(that we have today). Would you mind confirming?

mumoshu (approved):
LGTM. Thanks for your contribution, as always!

fgalind1:
> I'm afraid I'm not reading the whole sentence correctly, but I think we're sharing the idea. I want any change to be backward-compatible. The only way to rename values fields then would be add lowerCamelCase versions of the existing fields, and make any missing lowerCamelCase field to fallback to it's snake_case variant(that we have today). Would you mind confirming?

Yes I'm also in line with backwards compatible always 👍 - I guess my point/question was - would it make sense for any new configuration value introduced to be camelCase (like the new values introduced in this PR). That way the "tech-debt" to normalize/fix legacy values could be less and we don't keep adding more "tech-debt". Yes - it may be ugly having authSecret.github_token and authSecret.githubUrl - but at least in the future there are less fields to fix? Not a big deal just wondering 😄  - thanks for the review!

mumoshu:
@fgalind1 I basically agree with you! The only reason I opted to keep it snake_case in this PR was for consistency. If we were to add lowerCamelCase fields under `authSecret` where we already have legacy snake_case fields, it can be an even bigger tech-debt in case anyone didn't actually contribute to rename fields later ☺️

PR #1018, opened 2021-12-21
Refactor: Dockerfile [`UPX`]

pratikbin:
### Dockefile
- Move `ARG TARGETARCH TARGETOS` above build step. That will help buildx to run above
steps in local arch and after `ARG TARGETARCH TARGETOS` steps in target arch more <https://www.youtube.com/watch?v=0iN4dJB3OLk>
- Add upx to compress binaries, which make both `manager` and `github-webhook-server` binary size `~49MB` to `~19MB` without any performance penalty. Not working correctly in arm right now, but I'm working on that. More <https://github.com/upx/upx/>
  
    image

- Use GOOS="$TARGETOS" GOARCH="$TARGETARCH"

### CI: 
- Enable gha cache
- Use tonistiigi/binfmt for better emulation
- Use stdout based tar extracting with curl


.github/workflows/release.yml
```diff
@@ -39,6 +37,9 @@ jobs:
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v1
+        with:
+          image: tonistiigi/binfmt:latest
```
mumoshu, line 41: > Use tonistiigi/binfmt for better emulation

I'm still learning- How is it better than the default? 🤔

pratikbalar: Those are qemu static binaries, I don't know exact performance benefit but seen @crazy-max to use it everywhere

Dockerfile
```diff
@@ -18,13 +16,18 @@ RUN go mod download
...
-  export GOARCH=$(echo ${TARGETPLATFORM} | cut -d / -f2) && \
-  GOARM=$(echo ${TARGETPLATFORM} | cut -d / -f3 | cut -c2-) && \
+ENV CGO_ENABLED=0 GOOS="$TARGETOS" GOARCH="$TARGETARCH"
+RUN export GOARM=$(echo ${TARGETPLATFORM} | cut -d / -f3 | cut -c2-) && \
   go build -a -o manager main.go && \
   go build -a -o github-webhook-server ./cmd/githubwebhookserver
 
+## Compress binary with upx https://github.com/upx/upx/
+RUN upx -9 /workspace/manager || true && \
+  upx -9 /workspace/github-webhook-server || true
```
mumoshu, line 29: Thansks for the suggestion! But at this time I"m not yet convinced to use upx.

You said:

> Add upx to compress binaries which make final container image (extracted) 41MB from 97MB without any performance penalty. Not working correctly in arm right now but I'm working on that. more https://github.com/upx/upx/

I'm mainly concerned about the downloaded container image size(which is a set of image layers that are gzipped at the end as I understand) and I don't think compressing the executable binary results in a huge difference in the compressed, downloaded container image size.

pratikbalar: Well we are using `distroless/static` so the only thing we can compress is binaries right. :arrow_down: you can see UPX compression ratio for our binaries is `~39.38%` which is great right? That will directly downgrade the container's size at least `~39.38%`.

image

---
I have seen people also maintaining UPX'ed binaries and container separately, so we can do that also if you want. I'm In for UPX.

mumoshu: Thanks- Great. The compressed docker image size for actions-runner-controller is like 38MB these days. So upx is compressing it better than docker's native image compression!

I don't know all the side-effects of doing it though. I heard upx makes the resulting executable to decompress itself in-memory before running the original code. So would there be a bit CPU overhead and a tiny surge in disk read due to all the data in the container fs needs to be read on startup? Perhaps it can also affect some anti-virus or anti-malware detection to misbehave?

That said, it would be great if there were both upx-enabled version of the actions-runer-controller image and the standard version of it at the same time.

pratikbalar: > I don't know all the side-effects of doing it though. I heard upx makes the resulting executable to decompress itself in-memory before running the original code. So would there be a bit CPU overhead and a tiny surge in disk read due to all the data in the container fs needs to be read on startup? 

As long as someone doesn't run this in a very tight environment (i.e., edge)!!, we are good I guess.

> Perhaps it can also affect some anti-virus or anti-malware detection to misbehave?

Have to check in someone (I'm daily-ing Linux) else's Windows system with defender and other antivirus software,

> Thanks- Great. The compressed docker image size for actions-runner-controller is like 38MB these days. So upx is compressing it better than docker's native image compression!

Still, I haven't added go build flags `-trimpath -ldflags "-s -w"`, Testin it now

> That said, it would be great if there were both upx-enabled version of the actions-runer-controller image and the standard version of it at the same time.

I'm interested in contributing

pratikbalar: :scream:  with GO Build flags `-trimpath -ldflags "-s -w"` (`~35MB`) + `UPX`ed  = `~8.5MB`

image

that makes container image(extracted) only `19.6MB` :scream: 

image

> FYI, i don't know much about ``-trimpath -ldflags "-s -w"`` also :sweat_smile:  but i've seen people using it all the time

fgalind1: >> Perhaps it can also affect some anti-virus or anti-malware detection to misbehave?

> Have to check in someone (I'm daily-ing Linux) else's Windows system with defender and other antivirus software,

for what it's worth sometime ago we used upx in a windows project - but our company antivirus detected from time to time the binary as virus - and unfortunately we dropped using upx for windows

pratikbalar:
`UPX`ing in arm binaries somehow failing 

image

https://github.com/pratikbalar/actions-runner-controller/runs/4615575433?check_suite_focus=true#step:5:854

Dockerfile
pratikbalar: @fgalind1 What if we sign the `UPX`ed binary ? Does that help in Windows systems?

pratikbalar:
@mumoshu or @summerwind  why we are force rebuilding up-to-date packages when building new image with `-a` flag
https://github.com/actions-runner-controller/actions-runner-controller/blob/c5950d75fa5ded14e1a3cdcd476accdb141b6816/Dockerfile#L25-L26

pratikbalar:
- Add `docker-bake.hcl` 
- Used [goxx](https://github.com/crazy-max/goxx) which help `-a` to rebuild packages using cross-compilers
- Used [goreleaser-xx](https://github.com/crazy-max/goreleaser-xx) for generating multi arch build and handling UPX

Remain
- [ ] Docs for buildx bake targets and tags 
- [ ] Testing slim image in CI
- [ ] Adding slim support in makefile

mumoshu:
Thanks for your efforts but sorry, I'm not going to use goxx and goreleaser for this project as I'd like to keep the build dependencies minimal as possible so that everyone can clone the repo and get started easily when contributing.

Also, I'm a fan of cross-building with docker-buildx and AFAICS goxx and goreleaser doesn't work well with it.

mumoshu:
It would be great if you could submit a dedicated issue for formalizing which part of our build pipeline should be enhanced. Honestly speaking, I have a feeling that this is getting beyond of the scope of a "refactoring" PR.

.github/workflows/release.yml
```diff
@@ -39,27 +34,36 @@ jobs:
...
+          platforms: all
```
crazy-max, line 39: Looking at this project QEMU is not needed. You can do native cross compilation using the global scope ARG `BUILDPLATFORM` as shown in https://github.com/crazy-max/goreleaser-xx#minimal

Dockerfile
```diff
@@ -1,39 +1,118 @@
-# Build the manager binary
-FROM golang:1.17 as builder
+FROM --platform=$BUILDPLATFORM crazymax/goreleaser-xx:1.2.5 AS goreleaser-xx
```
crazy-max, line 1: Using goreleaser-xx or goreleaser is overkill here as the goal of this project seems to just release an image and not artifacts.

Dockerfile
```diff
@@ -1,39 +1,118 @@
...
+FROM --platform=$BUILDPLATFORM pratikimprowise/upx:3.96 AS upx
+FROM --platform=$BUILDPLATFORM crazymax/goxx:1.17 AS base
```
crazy-max, line 3: goxx is not needed here as CGO doesn't seem required in this project. `golang:1.17` is enough.

crazy-max (commented):
Dockerfile could be improved to automatically translate `TARGETPLATFORM` into into Go compiler's target.

So instead of 

```dockerfile
# Build
RUN export GOOS=$(echo ${TARGETPLATFORM} | cut -d / -f1) && \
  export GOARCH=$(echo ${TARGETPLATFORM} | cut -d / -f2) && \
  GOARM=$(echo ${TARGETPLATFORM} | cut -d / -f3 | cut -c2-) && \
  go build -a -o manager main.go && \
  go build -a -o github-webhook-server ./cmd/githubwebhookserver
```

You would have:

```dockerfile
# Build
RUN xx-go build -a -o manager main.go && \
  xx-go build -a -o github-webhook-server ./cmd/githubwebhookserver
```

And as I said in another comment, use native cross compilation with the local build platform to avoid pulling the base image for X platforms and using emulation:

```dockerfile
FROM --platform=$BUILDPLATFORM golang:1.17 as builder
```

.github/workflows/release.yml
crazy-max: > I'm still learning- How is it better than the default? 🤔

This image is already the default one: https://github.com/docker/setup-qemu-action#inputs

pratikbalar:
Created non `upx`ed, only dockerfile changes branch https://github.com/actions-runner-controller/actions-runner-controller/pull/1041

PR #912, opened 2021-10-26
Add dockerEnv variable for RunnerDeployment

pavelsmolensky:
https://github.com/actions-runner-controller/actions-runner-controller/issues/878

PR #1021, opened 2021-12-25
Bump chart version to fix outdated CRDs

gtaylor:
Changes happened to the CRDs since the last time that the Helm chart was bumped, so there are Helm users out there (like myself) who are running recent controllers with old CRDs. I am for example missing the new `dockerEnv` feature in the runnerdeployment resource.

I have bumped the chart version to pull the latest CRDs into a new point release. I've also nudged the appVersion up to match the latest release.

It'll be important to bump the chart version any time a change is made that results in different YAML rendering behavior to avoid future recurrence of the breakage.

### DIFF
--- a/charts/actions-runner-controller/Chart.yaml
+++ b/charts/actions-runner-controller/Chart.yaml
@@ -15,10 +15,10 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.15.1
+version: 0.15.2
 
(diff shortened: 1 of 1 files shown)


mumoshu:
@gtaylor Hey! Thank you for your contribution.

Yes, #912 added dockerEnv to the runner spec but we have no controller release (v0.21.0) for that feature yet.

My general recommendation is- if you're trying an unreleased version of the controller, use the chart from our main branch, not from the chart repo. Does that make sense?

Or should I just take this as a warm request to release the new controller 0.21.0 along with the new chart version earlier? 😃

gtaylor:
Whoops, that makes sense. Shall I close this PR?

toast-gear:
@gtaylor if you could, cheers

PR #1028, opened 2022-01-05
Stop creating registration-only runners on scale-to-zero

### LINKED ISSUE #859: Remove registration-only runners
**Description**
Since GitHub changed their [routing logic](https://docs.github.com/en/actions/hosting-your-own-runners/using-self-hosted-runners-in-a-workflow#routing-precedence-for-self-hosted-runners) to not fail pipelines that target a `runs-on` label that doesn't have a runner registered these are no longer needed

This work will also close some bugs linked with these pods:
- https://github.com/actions-runner-controller/actions-runner-controller/issues/595

mumoshu:
Resolves #859

### DIFF
--- a/controllers/runnerreplicaset_controller.go
+++ b/controllers/runnerreplicaset_controller.go
@@ -117,6 +117,15 @@ func (r *RunnerReplicaSetReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		desired = 1
 	}
 
+	// We had to have a registration-only runner to support scale-from-zero before.
+	// But since Sep 2021 Actions update on GitHub Cloud and GHES 3.3, it is unneceesary.
+	// See the below issues for more contexts:
+	// https://github.com/actions-runner-controller/actions-runner-controller/issues/516
+	// https://github.com/actions-runner-controller/actions-runner-controller/issues/859
+	//
+	// In the below block, we have a logic to remove existing registration-only runners as unnecessary.
+	// This logic is introduced since actions-runner-controller 0.21.0 and probably last one or two minor releases
+	// so that actions-runner-controller instance in everyone's cluster won't leave dangling registration-only runners.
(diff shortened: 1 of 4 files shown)


controllers/runnerreplicaset_controller.go
```diff
@@ -133,52 +142,11 @@ func (r *RunnerReplicaSetReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
 
-	// On scale to zero, we must have fully registered registration-only runner before we start deleting other runners, hence `desired == 0`
-	// On scale from zero, we must retain the registratoin-only runner until one or more other runners get registered, hence `registrationOnlyRunnerExists && available == 0`.
-	// On RunnerReplicaSet creation, it have always 0 replics and no registration-only runner.
-	// In this case We don't need to bother creating a registration-only runner which gets deleted soon after we have 1 or more available repolicas,
-	// hence it's not `available == 0`, but `registrationOnlyRunnerExists && available == 0`.
-	// See https://github.com/actions-runner-controller/actions-runner-controller/issues/516
-	registrationOnlyRunnerNeeded := desired == 0 || (registrationOnlyRunnerExists && current == 0)
+	if registrationOnlyRunnerExists {
+		if err := r.Client.Delete(ctx, &registrationOnlyRunner); err != nil {
```
mumoshu, line 146: I believe we'd better maintain this reg-only-runner clean up logic for a while(perhaps until 0.22 or 0.23) so that we can prevent remaining and dangling reg-only runners in everyone's cluster.

toast-gear: Makes sense, probably worth having a comment, e.g.
```
// TODO: remove registration runner check later (v0.23.0 or v0.24.0)
```

mumoshu: Thanks! Added 807e630 for the additional comment.

PR #1035, opened 2022-01-06
Add optional hostNetwork for controller

jsotelo:
Although not ideal, in some situations allowing this chart to set `hostNetwork: true` for the controller is the least invasive solution. This PR sets this field only if specified, otherwise it's a no-op.


KairaMiu:
Thank you very much, works perfectly.  
Would be nice to have it in a release.

toast-gear:
LGTM, cheers

toast-gear (approved):

PR #1039, opened 2022-01-07
refactor: reduce size of `Reconcile` method in runner controller

fgalind1:
Make size of `Reconcile` method smaller and abstract partial logic to analyze runner status into a separate method.  Reduces main method from ~330 lines down to ~190 lines. No logic changes just method abstraction


fgalind1:
still valid 👍

PR #1045, opened 2022-01-11
Add custom volume mount documentation

genisd:
one example for in-memory
and one example for NVME backed storage, also pointing out all the flaws/risks

in theory other volume mount types should be possible. Like a glusterfs mount and whatnot :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ ToC:
     - [Scheduled Overrides](#scheduled-overrides)
   - [Runner with DinD](#runner-with-dind)
   - [Additional Tweaks](#additional-tweaks)
+  - [Custom Volume mounts](#custom-volume-mounts)
   - [Runner Labels](#runner-labels)
   - [Runner Groups](#runner-groups)
   - [Runner Entrypoint Features](#runner-entrypoint-features)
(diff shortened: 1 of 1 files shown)


genisd:
Haven't found the time to properly investigate the issues regarding leaving data behind.
And when stuff is left behind. So that's why I marked it as a draw, because I haven't done the research yet (not easily observed while folks are running builds constantly).

We could simply reword it a little to bring it to the users attention that these are aspects of concern and leave it at that.

genisd:
@mumoshu I've checked the current state of things.

nvme setup:
- backend volume fills up over time due to no cleanup.
- with emphemeral: false i believe it leaves files behind on `/tmp`. I can see the on the backend volume and I have no reason to suspect that files in `/tmp` are not visible between buillds. they probably are, even with epmhemeral: true.

ramdisk setup:
here i believe nothing changes. i believe it's safe with emphemeral: true and flawed with emphemeral: false which is probably the case for any setup. emphemeral: true is probably better anyway to reclaim memory/disk space. We are at the moment running it with emphemeral: false :innocent:, though i will probably change it.

k8s will come with size limited ramdisks, so then we will be able to even have multiple workers on a very large node without overcommitting memory.
that sounds so awesome, can't wait. I think the current statis is alpha or beta.


but I think with these docs it's safe to show the new capabilities without neglecting due dilligence

genisd:
i just tested and verified.
ramdisk with emphemeral: true is safe.

but i fear that data from the workdir is left behind with emphemeral: fasle and persistent storage like nvme.
so I'm investigating that right now

genisd:
data in workdir is even left behind with emphemeral: true and nvme backend volume.
that is evil and we are running it like that right now :see_no_evil: 

so perhaps we should only document ramdisk setup.

genisd:
okay after some testing. with `ephemeral: true` data is cleared in the nvme setup on the work dir between job runs.
otherwise data **even leaks inside the workdir** :facepalm:

genisd:
I thinks it's good to go, for real now :innocent:

genisd:
@mumoshu I think this can be merged.
If you have suggestions for improvements, tomorrow I'm working and then I got a nice sweet holiday :palm_tree: :innocent:

Word on the street is there will be a release soon and this documentation should perhaps be included

mumoshu (approved):
Sorry for the radio silence! This does seem useful for i/o intensive builds. Thank you for your contribution ☺️

PR #1062, opened 2022-01-20
Improve support for runner groups with selected visibility

fgalind1:
This PR fixes additional cases from https://github.com/actions-runner-controller/actions-runner-controller/pull/1012

if GitHub auth is provided in the webhooks controller then runner groups with custom visibility is supported, otherwise all runner groups will be assumed to be visible to all repositories

`getScaleUpTargetWithFunction()` will check if there is an HRA available with the following flow:
1. search for **repository** HRAs - if so it ends here
2. get available HRAs in k8s
3. compute visible runner groups
  a. if github auth is provided - get all the runner groups that are visible to the repository of the incoming webhook using GitHub API calls.  
  b. if github auth is not provided - assume all runner groups are visible to all repositories
4. search for **default organization** runners
5. search for **default enterprise** runners
6. search for **organization runner groups**
7. search for **enterprise runner groups**


toast-gear:
This looks OK, it needs some changes however:
Helm template changes
Helm Chart docs
Main docs. There isn't any caching going on so they should warn that this is probably only usable in GHES as you will blow through your rate limit budget otherwise

fgalind1:
> This looks OK, it needs some changes however:
> 
> 1. Helm template changes to support the feature in Helm
> 2. Helm Chart docs
> 3. Main docs. There isn't any caching going on so they should warn that this is probably only usable in GHES as you will blow through your rate limit budget otherwise.

updated 👍

allistera:
Hi @toast-gear  - Would be awesome if this can be looked into being reviewed + merged. It is exactly what we need to allow flexible runners dedicated to runner groups. Thanks! :)

README.md
```diff
@@ -1045,6 +1045,20 @@ spec:
...
+Supporting custom visibility requires to do a few GitHub API calls to find out what are the potential runner groups that are visible to
+the webhook's repository, this may incur in increased API rate limiting when using github.com
+
+This option will be enabled when proper GitHub authentication options (token, app or basic auth is provided) in the GitHub webhook server and `enableCustomVisibilityRunnerGroups` is set to true, e.g.
+
+```yaml
+githubWebhookServer:
+  enabled: false
+  replicaCount: 1
+  enableCustomVisibilityRunnerGroups: true
```
mumoshu, line 1059: I believe this should eventually be the "default" behavior rather than "custom" in the future, because what's implemented in this PR is required to let ARC respect how GitHub designed runner groups.
That said, let me change this to a more neutral and simple form, like `useRunnerGroupsVisibility` 😃 

cc/ @toast-gear

github/github.go
```diff
@@ -224,51 +225,78 @@ func (c *Client) ListRunners(ctx context.Context, enterprise, org, repo string)
 	return runners, nil
 }
 
-func (c *Client) GetRunnerGroupsFromRepository(ctx context.Context, org, repo string, potentialEnterpriseGroups []string, potentialOrgGroups []string) ([]string, []string, error) {
+func (c *Client) GetRunnerGroupsFromRepository(ctx context.Context, org, repo string, potentialGroups utils.RunnerGroups) (utils.RunnerGroups, error) {
```
mumoshu, line 228:
```suggestion
func (c *Client) GetRunnerGroupsVisibleToRepository(ctx context.Context, org, repo string, potentialGroups utils.RunnerGroups) (utils.RunnerGroups, error) {
```

After carefully reading what's done in this function, I think this might be a better name that reflects the purpose 🤔

github/github.go
```diff
@@ -224,51 +225,78 @@ func (c *Client) ListRunners(ctx context.Context, enterprise, org, repo string)
...
 
-	var enterpriseRunnerGroups []string
-	var orgRunnerGroups []string
+	var visibleGroups utils.RunnerGroups
 
 	if org != "" {
 		runnerGroups, err := c.getOrganizationRunnerGroups(ctx, org, repo)
```
mumoshu, line 233: `repo` seems redundant (apparently unused within the function).

Also, (not your fault at all but)the name of this function(which is seemingly derived from the `ListOrganizationRunnerGroups` function provided by go-github) doesn't capture the fact that it also returns enterprise runners that are inherited to the organization.

That said, changing this to something like the below would improve the readability of code:

```suggestion
		runnerGroups, err := c.getDefinedAndInheritedRunnerGroups(ctx, org)
```

controllers/horizontal_runner_autoscaler_webhook.go
```diff
@@ -476,95 +477,99 @@ func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) getScaleUpTargetWithF
...
-	}
-
-	// At this point there were no default organization/enterprise runners available to use, try now
-	// searching in runner groups
-
-	// We need to get the potential runner groups first to avoid spending API queries needless. Once/if GitHub improves an
+	// Find the potential runner groups first to avoid spending API queries needless. Once/if GitHub improves an
 	// API to find related/linked runner groups from a specific repository this logic could be removed
-	availableEnterpriseGroups, availableOrganizationGroups, err := autoscaler.getPotentialGroupsFromHRAs(ctx, enterprise, owner)
+	availableGroups, err := autoscaler.getPotentialGroupsFromHRAs(ctx, enterprise, owner)
```
mumoshu, line 482: We have discrepancy on the name of available groups vs potential groups. (It's called `availableGroups` in the left hand and `PotentialGroups` in the right hand. `GetRunnerGroupsFromRepository` receives it as `potentialGroups`

Assuming what we're doing here is "Try to scale only one of the runner groups that are managed by ARC and are visible to the repository", how about consistently calling it `managed runner groups`?

cmd/githubwebhookserver/main.go
```diff
@@ -142,13 +142,19 @@ func main() {
 		}
 	})
 
+	// In order to support runner groups with custom visibility (selected repositories), we need to perform some GitHub API calls.
+	// Let the user define if they want to opt-in supporting this option by providing the proper GitHub authentication parameters
+	// Otherwise runner groups with custom visibility won't be supported and it will be assume all runner groups to be usable
+	// in all repositories to save API calls
```
mumoshu, line 148: Thanks! This greatly helped me understand how it works. Here's my suggestion to improve it even further.

```suggestion
	// Without an opt-in, runner groups with custom visibility won't be supported to save API calls.
	// That is, all runner groups managed by ARC are assumed to be visible to any repositories,
	// which is wrong when you have one or more non-default runner groups in your organization or enterprise.
```

fgalind1:
thanks @mumoshu for the feedback, addressed changed in 805a685ff54c7d01e0c8d8a17b14894c23553a59

controllers/horizontal_runner_autoscaler_webhook.go
```diff
@@ -476,95 +477,99 @@ func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) getScaleUpTargetWithF
...
-	}
-
-	// At this point there were no default organization/enterprise runners available to use, try now
-	// searching in runner groups
-
-	// We need to get the potential runner groups first to avoid spending API queries needless. Once/if GitHub improves an
+	// Find the potential runner groups first to avoid spending API queries needless. Once/if GitHub improves an
 	// API to find related/linked runner groups from a specific repository this logic could be removed
-	availableEnterpriseGroups, availableOrganizationGroups, err := autoscaler.getPotentialGroupsFromHRAs(ctx, enterprise, owner)
+	managedRunnerGroups, err := autoscaler.getManagedRunnerGroupsFromHRAs(ctx, enterprise, owner)
```
mumoshu, line 482: I was carefully inspecting the implementation of `getManagedRunnerGroupsFromHRAs` and noticed that this function conflates runner groups associated with different enterprises and organizations, which means this works well only when you have only one organization and enterprise, but not when you have many of those.

The problem seems to stem from the fact that the returned data structure, `utils. RunnerGroups` struct, is unable to differentiate runner groups across enterprises and organizations.

This `managedRunnerGroups` is passed to the `GetRunnerGroupsVisibleToRepository` function, so it results in `GetRunnerGroupsVisibleToRepository` thinking an unmanaged runner group as if it is managed, which would result in ARC adding a runner (and the runner pod that hosts it) that is never used by GitHub Actions.

Probably we need to introduce namespacing to the data structure of `managedRunnerGroups`, either at the organization level or the enterprise level, so that ARC won't add a runner that is never used by GitHub.

I'll try to experiment a bit more so that I can hopefully come up with a better approach/implementation, but if you've any insights to share on it I'd appreciate it! Thanks.

mumoshu: I found that, in getManagedRunnerGroupsFromHRAs, you did filter out any managed RunnerDeployments and RunnerSets that don't match the enterprise and the organization names read from the webhook event.
So, NVM! Probably it's not an issue.

mumoshu:
@fgalind1 [8612bc0](https://github.com/actions-runner-controller/actions-runner-controller/pull/1062/commits/8612bc04b4c25c224752a08fe10a5da559f2cd16) is my attempt to refactor your work.

My goal was to increase testability and cohesion and to reduce the repetition around:

- The sort order of groups based on their scopes (Enterprise vs Organization) and kinds (Default and Custom)
- The logic to match groups from GitHub API vs groups derived from HRAs

We now map groups from github and groups from hras into the same space represented by a `simulator.VisibleRunnerGroups`.  I named the package `simulator` because it simulates the GitHub's behavior on runner group visibility. `VisibleRunnerGroups` in combination with its element type `RunnerGroup` abstracts away the details of runner group equality and the sort order, so that the webhook autoscaler code becomes simpler. I hope you like it.

The next step is testing. Fortunately I got a GitHub Enterprise account for testing so I'll shortly deploy ARC built from this branch to the enterprise and see if how this works.

fgalind1:
> @fgalind1 [8612bc0](https://github.com/actions-runner-controller/actions-runner-controller/pull/1062/commits/8612bc04b4c25c224752a08fe10a5da559f2cd16) is my attempt to refactor your work.
> 
> My goal was to increase testability and cohesion, and to reduce the repetition around:
> 
> * The sort order of groups based on their scopes (Enterprise vs Organization) and kinds (Default and Custom)
> * The logic to match groups from GitHub API vs groups derived from HRAs
> 
> so that I can better understand and maintain your awesome work.
> 
> We now map groups from github and groups from hras into the same space represented by a `simulator.VisibleRunnerGroups`. I named the package `simulator` because it simulates the GitHub's behavior on runner group visibility. `VisibleRunnerGroups` in combination with its element type `RunnerGroup` abstracts away the details of runner group equality and the sort order, so that the webhook autoscaler code becomes simpler. I hope you like it.
> 
> The next step is testing. Fortunately I got a GitHub Enterprise account for testing so I'll shortly deploy ARC built from this branch to the enterprise and see if how this works.

Thanks @mumoshu I took a look at the commit and looks a really nice abstraction

mumoshu:
@fgalind1 Just finished testing this. I couldn't make it work reliably because it quickly consumes my PAT's API rate limit (5000 calls/h). I'd say this is unusable for GitHub Enterprise Cloud. However, there seems no regression when the feature flag is disabled, so probably this is good to go 😃 

Merging. I'd appreciate it if you could deploy and test this in your GHES environment.

Thank you again for your contribution!

mumoshu:
@fgalind1 FYI, I've tweaked the chart a bit so that you can use separate github creds for githubwebhookserver #1122

fgalind1:
> @fgalind1 FYI, I've tweaked the chart a bit so that you can use separate github creds for githubwebhookserver #1122

Awesome - thanks @mumoshu for merging this PR and the heads-up and merging the PR - I'll test this updated version on top of #1122 in our environment and share any results back 👍 

> I couldn't make it work reliably because it quickly consumes my PAT's API rate limit (5000 calls/h)

I guess you already may know if but if you create an Oauth app and a token from it - you'll get 15K calls instead of the 5K from a regular PAT. That's what we use to have more API calls :)

mumoshu:
> I guess you already may know if but if you create an Oauth app and a token from it - you'll get 15K calls instead of the 5K from a regular PAT. That's what we use to have more API calls :)

Yeah! I knew that but I tend to use PAT so that I can any issues related to calling too many APIs earlier than everyone.

It would be preferable to use a github app key in a production environment. Thanks for the reminder ☺️

mumoshu:
@fgalind1 Awesome!! Thank you for reporting and your contribution again ☺️

PR #1087, opened 2022-02-02
fix: Do recreate runner pod earlier on registration token update

mumoshu:
Apparently, we've been missed taking an updated registration token into account when generating the pod template hash which is used to detect if the runner pod needs to be recreated.

This shouldn't have been the end of the world since the runner pod is recreated on the next reconciliation loop anyway, but this change will make the pod recreation happen one reconciliation loop earlier so that you're less likely to get runner pods with outdated refresh tokens.

Ref https://github.com/actions-runner-controller/actions-runner-controller/pull/1085#issuecomment-1027433365

### DIFF
--- a/controllers/runner_controller.go
+++ b/controllers/runner_controller.go
@@ -626,6 +626,11 @@ func (r *RunnerReconciler) newPod(runner v1alpha1.Runner) (corev1.Pod, error) {
 		runner.ObjectMeta.Annotations,
 		runner.Spec,
 		r.GitHubClient.GithubBaseURL,
+		// Token change should trigger replacement.
+		// We need to include this explicitly here because
+		// runner.Spec does not contain the possibly updated token stored in the
+		// runner status yet.
+		runner.Status.Registration.Token,
 	)
 
 	objectMeta := metav1.ObjectMeta{


mumoshu:
> is it going to trigger a restart in the middle of a pod running? If so, we just have another potential active runner deletion.

@jbkc85 Yes, you're absolutely right!

To be clear, as ARC has no way to atomically delete the runner if and only if the runner isn't busy, `actions/runner` needs to be able to gracefully stop at any time. (I.e. polling "list runners" API to detect if the runner is busy or not and deleting the runner only after that, isn't a solution. it's still subject to race condition)

This fixes the issue that runner takes too much time to restart even after the token is already expired.
This doesn't fix the issue that actions/runner is unable to stop gracefully at any time.

jbkc85:
> This fixes the issue that runner takes too much time to restart even after the token is already expired.

Got it!  That makes perfect sense.

mumoshu:
@jbkc85 Good catch! Yeah we'd definitely better to take `IsPhemeral` into account to prevent potentially unnecessary restarts.

mumoshu:
@jbkc85 Ah sorry I'm pretty confused.

From my observation, the "runner" container for an ephemeral runner does exit with code 0 when it finishes running a job. 

We should definitely recreate the pod in that case. Otherwise, the pod gets stuck forever and you'll end up with zero ephemeral runner pods actually running the runner containers, right?

mumoshu:
> ARC has no way to atomically delete the runner if and only if the runner isn't busy

To be clear, this turned out to be my mistake. `RemoveRunner` seems like the API to atomically delete the runner. See https://github.com/actions-runner-controller/actions-runner-controller/pull/1125#issuecomment-1046003049

Issue #1086, opened 2022-02-01
Add support for new disable update flag


toast-gear:
https://github.blog/changelog/2022-02-01-github-actions-self-hosted-runners-can-now-disable-automatic-updates/

It just got shipped, we should add a new env var into the entrypoint to configure this new setting. The default should be as it is (keep self-updating enabled) as the actions runenr team are currently working on making the update process not cause a container to commit seppuku https://github.com/actions/runner/pull/1494 https://github.com/actions/runner/pull/1635. Once this work is complete self-updating would be preferred.

Issue #888, opened 2021-10-12
Enable Fetching and Installing Latest Runner Software On Container Start


toast-gear:
**Current Behaviour**
We provide a few runner images of which only the `latest` tag is continously updated to have the latest runner software in it automatically whilst retaining the same tag. There is no ability to pull in an alternative runner software version in any image.

**Suggestion**
Produce a solution to allow an end user to download and install the latest version of the runner software on container startup and use the baked in version as a fallback should this process fail.

**Rationale**
Containers don't always play with the self-update process, additionally, it's not ideal having to reference a public container (`latest`) from a community maintained open source project to always stay on the latest runner software. By providing an alternative bootstrap process enabled via a feature flag people can mitigate the self-update issues without the need to either constantly sync `latest` to their registry or have to rely on our image.

Furthermore, if you run a custom runner image you currently have to provide some solution for automatically pushing out a new version of your image when our image gets bumped.

Issues to Resolve:
1. If we call the GitHub release API every container restart we'll very quickly consume the 5000 authenticated API calls / hour rate-limit on even semi busy environments. Calls will need to be cached (to a configurable period) probably by sending them through a Go HTTP proxy sidecar implementation instead of going directly to Github.
2. Perhaps in addition we could allow a "pre-install" script to be ran in the container entrypoint a bit like philips-labs does for their user-data https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/modules/runners/templates/user-data.sh#L4. This would enable people to then implement their own solutions to the problem such as syncing the binaries to an S3 bucket and having their containers get the binary from an S3 bucket instead avoiding the rate-limit issue to begin with.

toast-gear:
https://github.com/actions/runner/issues/1396#issuecomment-983783264 this is probably not going to be needed in the new year 🎉

mumoshu:
Great news!

PR #1088, opened 2022-02-03
feat: disable automatic runner updates

cbui:
Closes https://github.com/actions-runner-controller/actions-runner-controller/issues/1086

I tested by setting the runner version in the images to 2.287.0, toggling the flag, and seeing if it would update when given a job.

> Once this work is complete self-updating would be preferred. 

@toast-gear by this do you mean disabling autoupdate by default?


runner/Dockerfile.ubuntu.1804
```diff
@@ -1,7 +1,7 @@
 FROM ubuntu:18.04
 
 ARG TARGETPLATFORM
-ARG RUNNER_VERSION=2.280.3
```
cbui, line 4: This version was a few behind, was there a reason for that?

reiniertimmer: It's just the default argument value I think. The actual value that is used for deploying is controlled by the GitHub workflow in https://github.com/actions-runner-controller/actions-runner-controller/blob/master/.github/workflows/runners.yml#L18

runner/entrypoint.sh
```diff
@@ -91,6 +91,10 @@ if [ "${RUNNER_FEATURE_FLAG_EPHEMERAL:-}" == "true" -a "${RUNNER_EPHEMERAL}" !=
   config_args+=(--ephemeral)
   echo "Passing --ephemeral to config.sh to enable the ephemeral runner."
 fi
+if [ "${RUNNER_DISABLE_UPDATE:-}" == "true" ]; then
```
reiniertimmer, line 94: How is this value passed to the entrypoint.sh? Is that something that should be configured in the runner spec?

In this case, a small bit of documentation would be helpful I guess

reiniertimmer: I guess all the entrypoint environments can be added here? https://github.com/actions-runner-controller/actions-runner-controller#runner-entrypoint-features

toast-gear:
> @toast-gear by this do you mean disabling autoupdate by default?

As in keep the current behavior, runner's should self-update.

runner/Dockerfile.ubuntu.1804
toast-gear: It's as @reiniertimmer says, the workflow passes in the runner version to be used. Renovate bumps this every new actions/runner release. Feel free to update the versions in the Dockerfiles as we might as well but they aren't the version that gets released.

`runner/Dockerfile.ubuntu.1804` itself is a image we no longer build, it's just hasn't been removed so it's there as a point of reference.

runner/entrypoint.sh
toast-gear: > Is that something that should be configured in the runner spec?

No, it's excessively complicated to set it in the CRD, we also already configure various runner software flags via environment variables so that is the convention.

toast-gear, line 94: I prefer `DISABLE_RUNNER_UPDATE`

toast-gear:
LGTM cheers pal

toast-gear:
Avoids the need for https://github.com/actions-runner-controller/actions-runner-controller/issues/888

PR #1098, opened 2022-02-08
chart: Fix ingress to make pathType and ingressClassName configurable for networking.k8s.io/v1

cw-sakamoto:
Signed-off-by: cw-sakamoto <sakamoto@chatwork.com>

The ingress resource was errored, so it has been fixed.


mumoshu (approved):
LGTM. Thank you for the fix @cw-sakamoto ☺️

PR #1099, opened 2022-02-08
Update the helm chart app version

FearlessHyena:
The Helm chart isn't using the latest app version `0.20.4` which was released a while back
Bumping it up


mumoshu (approved):
Good catch! Thank you for your contribution ☺️

Issue #1104, opened 2022-02-10
Add more linux archs to the image


cpanato:
**Is your feature request related to a problem? Please describe.**
I would like to run this controller in my raspi home cluster, but for that I need the image to be available in the arm arch as well

I can work on the addition, will be a pleasure 

**Describe the solution you'd like**
Add other Linux arch to the binaries/images

**Describe alternatives you've considered**
maybe fork this and build the project myself, but then i will need to keep the fork and other activities, which can be expensive and I prefer to maintain this project instead of my own fork, also maybe others would like to use the new archs as well

**Additional context**
n/a

mumoshu:
@cpanato Hey! I thought we have been already building and pushing images for arm. Isn't it `linux/arm64`? https://github.com/actions-runner-controller/actions-runner-controller/blob/b509eb4388cab249f2d88cad9c21220b86387385/.github/workflows/release.yml#L59

cpanato:
@mumoshu thanks for the reply, in my case we need the arm/v7 arch, but will be good to add arm/v8 as well 

```
pi@ccfde:~ $ uname -m
armv7l
```

mumoshu:
@cpanato Thanks! Then maybe all we need is to add `linux/arm/v7` to the platforms list?, right?

cpanato:
yep, i would say to add v8 as well as does not hurt :)

mumoshu:
@cpanato I'd definitely appreciate it if you could do so!

PR #1120, opened 2022-02-16
images: add linux arm/v7/v8 support

cpanato:
- images: add linux arm/v7/v8 support

Fixes: https://github.com/actions-runner-controller/actions-runner-controller/issues/1104

cc @mumoshu

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,10 +56,9 @@ jobs:
         uses: docker/build-push-action@v2
         with:
           file: Dockerfile
-          platforms: linux/amd64,linux/arm64
+          platforms: linux/amd64,linux/arm64,linux/arm/v7,linux/arm/v8
           push: true
           tags: |
             ${{ env.DOCKERHUB_USERNAME }}/actions-runner-controller:latest
             ${{ env.DOCKERHUB_USERNAME }}/actions-runner-controller:${{ env.VERSION }}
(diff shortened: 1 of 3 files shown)


cpanato:
looks like this dependency does not have the arm available https://github.com/Yelp/dumb-init

cpanato:
i will try to build locally, found other places that I was not aware when open this pr

cpanato:
@mumoshu if the [dumb-init](https://github.com/Yelp/dumb-init) is required, I think we will need to close this pr because that does not have the arm support and the underlining dependencies for that also does not have (for context https://github.com/pypa/manylinux/issues/84) so will take some time :/

mumoshu:
@cpanato Ah, good catch! I believe we don't have a strict requirement to stick with dumb-init. I thought there was a similar open source project that is written in Golang. IMHO a Golang-written alternative would have a better story on cross-platform.

mumoshu:
@toast-gear has suggested using `tini` instead, and I think it's great!
The only gotcha would be that we need to pass `-g` to `tini` to make it compatible with our usage(and the default behavior of dumb-init)

https://github.com/Yelp/dumb-init#session-behavior
https://github.com/krallin/tini#process-group-killing

cpanato:
sounds good! this is something you want to change and you want this in the project?

i will close this pr for now, thanks for all the support! 🎉

PR #1126, opened 2022-02-17
Fix RunerSet managed runner pods to terminate more gracefully

mumoshu:
Make RunnerSet-managed runners as reliable as RunnerDeployment-managed runners.

Ref https://github.com/actions-runner-controller/actions-runner-controller/issues/911#issuecomment-1042404460

### DIFF
--- a/controllers/unregister.go
+++ b/controllers/unregister.go
@@ -1,0 +1,49 @@
+package controllers
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/actions-runner-controller/actions-runner-controller/github"
+)
+
+// unregisterRunner unregisters the runner from GitHub Actions by name.
+//
+// This function returns:
+// - (true, nil) when it has successfully unregistered the runner.
+// - (false, nil) when the runner has been already unregistered.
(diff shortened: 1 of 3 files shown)


controllers/unregister.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+// unregisterRunner unregisters the runner from GitHub Actions by name.
+//
+// This function returns:
+// - (true, nil) when it has successfully unregistered the runner.
+// - (false, nil) when the runner has been already unregistered.
+// - (false, err) when it postponed unregistration due to the runner being busy, or it tried to unregister the runner but failed due to
+//   an error returned by GitHub API.
+func unregisterRunner(ctx context.Context, client *github.Client, enterprise, org, repo, name string) (bool, error) {
```
mumoshu, line 17: This function is derived from RunnerReconciler's unregisterRunner function which was problem-free.

We now use this from the runner controller (a counter-part of RunnerDeployment controller) and the runner pod controller (which is a counter-part of RunnerSet controller) to make runnsetset-managed runners as reliable as runnerdeployment's.

PR #1127, opened 2022-02-17
Add GitHub API cache to avoid rate limit

mumoshu:
Enhances ARC(both the controller-manager and github-webhook-server) to cache any GitHub API responses with HTTP GET and an appropriate Cache-Control header.

Ref #920

## Cache Implementation

`gregjones/httpcache` has been chosen as a library to implement this feature, as it is as recommended in `go-github`'s documentation:

https://github.com/google/go-github#conditional-requests

`gregjones/httpcache` supports a number of cache backends like `diskcache`, `s3cache`, and so on:

https://github.com/gregjones/httpcache#cache-backends

We stick to the built-in in-memory cache as a starter. Probably this will never becomes an issue as long as various HTTP responses for all the GitHub API calls that ARC makes, list-runners, list-workflow-jobs, list-runner-groups, etc., doesn't overflow the in-memory cache.

`httpcache` has an known unfixed issue that it doesn't update cache on chunked responses. But we assume that the APIs that we call doesn't use chunked responses. See #1503 for more information on that.

## Ephemeral runner pods are no longer recreated

The addition of the cache layer resulted in a slow down of a scale-down process and a trade-off between making the runner pod termination process fragile to various race conditions(shorter grace period before runner deletion) or delaying runner pod deletion depending on how long the grace period is(longer grace period). A grace period needs to be at least longer than 60s (which is the same as cache duration of ListRunners API) to not prematurely delete a runner pod that was just created.

But once I disabled automatic recreation of ephemeral runner pod, it turned out to be no more of an issue when it's being scaled via workflow_job webhook.

Ephemeral runner resources are still automatically added on demand by RunnerDeployment via RunnerReplicaSet(I've added `EffectiveTime` fields to our CRDs but that's an implementation detail so let's omit). A good side-effect of disabling ephemeral runner pod recreations is that ARC will no longer create redundant ephemeral runners when used with webhook-based autoscaler.

Basically, autoscaling still works as everyone might expect. It's just better than before overall.


github/github.go
```diff
@@ -82,8 +83,10 @@ func (c *Config) NewClient() (*Client, error) {
 		transport = tr
 	}
 
-	transport = metrics.Transport{Transport: transport}
-	httpClient := &http.Client{Transport: transport}
+	cached := httpcache.NewTransport(httpcache.NewMemoryCache())
+	cached.Transport = transport
+	metricsTransport := metrics.Transport{Transport: cached}
```
mumoshu, line 88: Perhaps this ordering results in the metrics not being distinguished between cached vs not-cached. But that's another story.

mumoshu:
ListRunners API calls are cached for 60 seconds now and that makes our IsRunnerBusy function and a pre-pod-deletion check almost useless, as it takes up to 60 seconds until you can see your runner's busy status.

Therefore, I've added [41e2a70](https://github.com/actions-runner-controller/actions-runner-controller/pull/1127/commits/41e2a70673ac97b918a453db2e6c6f8ac68c2b0b) as a rescue. We no longer check for busy status of runner before deletion. Instead, it tries its best to call the RemoveRunner API before deleting the pod.

It implements some grace period so that it's very unlikely to have the race issue discussed in https://github.com/actions-runner-controller/actions-runner-controller/pull/1085.

mumoshu:
> It also implements some grace period so that it's very unlikely to have the race issue discussed in https://github.com/actions-runner-controller/actions-runner-controller/pull/1085.

This either made the runner pod termination process fragile to various race conditions or delays runner pod deletion depending on how long the grace period is.

But those issues turned out to disappear for ephemeral runners once we stop recreating ephemeral runner pods (without recreating the whole runner resources on k8s). So I'm going ahead that way.

Previously, The desired number of runner pods was maintained by Runner recreating the completed runner pod which was subject to the related race condition by itself.
It's now maintained by RunnerDeployment adding replacement Runner resources for Runner resources deleted due to ephemeral runner completions.

It works very well so far and it also looks a very natural behavior to me 😃

acceptance/deploy.sh
```diff
@@ -37,7 +37,10 @@ fi
 
 tool=${ACCEPTANCE_TEST_DEPLOYMENT_TOOL}
 
+TEST_ID=${TEST_ID:-default}
+
 if [ "${tool}" == "helm" ]; then
+  set -v
```
Izzette, line 43: Should this be kept?

cmd/githubwebhookserver/main.go
```diff
@@ -47,11 +46,6 @@ var (
 )
 
 const (
-	logLevelDebug = "debug"
-	logLevelInfo  = "info"
-	logLevelWarn  = "warn"
-	logLevelError = "error"
-
```
Izzette, line 54: nice cleanup :)

github/github.go
```diff
@@ -82,8 +87,11 @@ func (c *Config) NewClient() (*Client, error) {
 		transport = tr
 	}
 
-	transport = metrics.Transport{Transport: transport}
-	httpClient := &http.Client{Transport: transport}
+	cached := httpcache.NewTransport(httpcache.NewMemoryCache())
+	cached.Transport = transport
+	loggingTransport := logging.Transport{Transport: cached, Log: c.Log}
+	metricsTransport := metrics.Transport{Transport: loggingTransport}
+	httpClient := &http.Client{Transport: metricsTransport}
```
sledigabel, line 94: @mumoshu Am I assuming right that we're reading [the TTL for each query](https://github.com/gregjones/httpcache/blob/901d90724c7919163f472a9812253fb26761123d/httpcache.go#L481) and honouring the `cache-control` header?

github/github.go
```diff
@@ -82,8 +87,11 @@ func (c *Config) NewClient() (*Client, error) {
 		transport = tr
 	}
 
-	transport = metrics.Transport{Transport: transport}
-	httpClient := &http.Client{Transport: transport}
+	cached := httpcache.NewTransport(httpcache.NewMemoryCache())
```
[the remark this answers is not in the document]
mumoshu: @TingluoHuang Not exactly. But I'd say 90% of requests are cached :)

I've added a new log level `-3` so see if each request to GitHub API is cached or not. To enable the log level of `-3` you would write values.yaml like...

```
logLevel: "-3"
githubWebhookServer:
  logLevel: "-3"
```

Tail the logs from ARC controllers...

```
kubectl -n actions-runner-system logs -l app.kubernetes.io/name=actions-runner-controller -f | tee arc.log
```

And extract logged HTTP requests:

```
tail -f -n +1 arc.log | jq -rRc 'fromjson? | select(.from_cache != null) | "\(.ratelimit_remaining) \(.method) \(.url)"'
```

You'll see:

```
null GET https://api.github.com/orgs/$ORG/$REPO/actions/runners?per_page=100
null GET https://api.github.com/orgs/$ORG/$REPO/actions/runners?per_page=100
null GET https://api.github.com/orgs/$ORG/$REPO/actions/runners?per_page=100
3548 DELETE https://api.github.com/orgs/$ORG/$REPO/actions/runners/591
null GET https://api.github.com/orgs/$ORG/$REPO/actions/runners?per_page=100
null GET https://api.github.com/orgs/$ORG/$REPO/actions/runners?per_page=100
null GET https://api.github.com/orgs/$ORG/$REPO/actions/runners?per_page=100
null GET https://api.github.com/orgs/$ORG/$REPO/actions/runners?per_page=100
null GET https://api.github.com/orgs/$ORG/$REPO/actions/runners?per_page=100
3546 DELETE https://api.github.com/orgs/$ORG/$REPO/actions/runners/595
null GET https://api.github.com/orgs/$ORG/$REPO/actions/runners?per_page=100
null GET https://api.github.com/orgs/$ORG/$REPO/actions/runners?per_page=100
null GET https://api.github.com/orgs/$ORG/$REPO/actions/runners?per_page=100
3544 DELETE https://api.github.com/orgs/$ORG/$REPO/actions/runners/589
```

The first column is the `x-ratelimit-remaining` header value returned by GitHub API that is logged only when it was NOT served from cache.

You can see that the list repository runners API are always cached(the cache-controller header for that response says `max-age=60s` so httpcache caches it for 60 seconds so all the runners for a repository shares the same cache), and  it occasionally calls the Delete Runner API.

github/github.go
mumoshu: @sledigabel Yes!

cmd/githubwebhookserver/main.go
mumoshu: Thank you 😄

acceptance/deploy.sh
mumoshu: Yep! I wanted to log the helm-upgrade command so that I can be sure that I have deployed it correctly, and that helps to isolate the cause of an e2e failure

github/github.go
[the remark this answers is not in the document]
TingluoHuang: 🆒 
will 90% still hold when the large customer using ARC to scale hundreds of ephemeral runners since the response from the ListRunner endpoint should change everytime on a busy repo/org/enterprise?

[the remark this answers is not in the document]
mumoshu: I guess it would scale well as this results in only one ListRunners call per 60s per RunnerDeployment.
So if a company has 1 enterprise runnerdeployment and 2 organizational runnerdeployments and 3 repository runnerdeployments, it might result in 6 calls per 60s.

Note that you only see a change in runner status (offline vs online, busy or not, runner existence, etc) only after a ListRunners API call after cache invalidation i.e. it will take up to 60s or so until you see changes in runner statuses.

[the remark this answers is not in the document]
sledigabel: FWIW I think this is acceptable. The default cache control for github is 60s as you said, and it sounds acceptable to have a bit of leeway of 1min to detect an idle runner. The `duration` parameter protects the runners from being decommissioned too quickly assuming it's greater than the cache control so I can't see any downside in this approach afaict.

sledigabel:
@mumoshu awesome work, this is really exciting 😎

github/github.go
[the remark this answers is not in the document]
mumoshu: > The duration parameter protects the runners from being decommissioned too quickly assuming it's greater than the cache control so I can't see any downside in this approach afaict.

Definitely! You should have `HRA.spec.scaleTrigggers[].duration` that is long enough to cover your longest-running workflow jobs anyway to let ARC not scale down the added runner too early in case of ARC failed to receive a webhook event of workflow_job completion so 60s of cache won't have real downside.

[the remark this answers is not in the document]
mumoshu: @tumoa and I have been discussion about a potential enhancement on how scale trigger duration works at:

https://github.com/actions-runner-controller/actions-runner-controller/issues/911#issuecomment-1049715625

mumoshu:
I'll shortly submit a pull request to add similar reliability enhancements to RunnerSet as well. Stay tuned!

PR #1128, opened 2022-02-18
fix(log): configure logger before trying to log

fgalind1:
Log about GitHub client not being initialized is not seen as logger is configured after adding the log


mumoshu (approved):
LGTM. Thanks a lot for your contribution as always ☺️

PR #1136, opened 2022-02-20
feat: copy dotfiles from asset to service dir

Fleshgrinder:
I am going to deploy this tomorrow in our enterprise and verify if the environment variables exported to the `.env` file are loaded as expected. Once I can confirm I will get back here and turn the draft into a proper pull request.

I tried to resists as much as possible from cleaning up the script with its POSIXish syntax to Bash. This is not the right place for a refactor. 😉


Fleshgrinder:
We tested this in our enterprise setup and it works as expected.

toast-gear (approved):
LGTM cheers pal, will release whenever GitHub stops exploding

PR #1140, opened 2022-02-22
Set UserAgent to `actions-runner-controller`  for all Http Client.

TingluoHuang:
I can't find any requests made by user agent `actions-runner-controller` in GitHub.com's telemetry in the past 7 days.

Turns out we only set user agent `actions-runner-controller` if we are configured to use BasicAuth which is not the case for most customers I think.

I update the code a little bit to make sure it always set `actions-runner-controller` as UserAgent for the GitHub HttpClient in ARC.

A further step would be somehow baking the ARC release version into the UserAgent as well. 😄


mumoshu (approved):
LGTM. Thank you for your contribution @TingluoHuang ☺️

PR #1145, opened 2022-02-22
Allow to use host network for the controller pod

samm-git:
This PR adds ability to use host networking for the controller pod. This is required to work on some environments like EKS+Calico. 

Some details why it is needed from [official documentation](https://projectcalico.docs.tigera.io/getting-started/kubernetes/managed-public-cloud/eks): 

> Note: Calico networking cannot currently be installed on the EKS control plane nodes. As a result the control plane nodes will not be able to initiate network connections to Calico pods. (This is a general limitation of EKS’s custom networking support, not specific to Calico.) As a workaround, trusted pods that require control plane nodes to connect to them, such as those implementing admission controller webhooks, can include hostNetwork:true in their [pod spec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#podspec-v1-core). See the Kuberentes API pod spec definition for more information on this setting.


abusaidm (approved):

toast-gear:
Thanks for the contribution, I've merged https://github.com/actions-runner-controller/actions-runner-controller/pull/1035 because their PR was older.

PR #1149, opened 2022-02-23
fix(chart): allow to use basic auth when authSecret.create is false

fgalind1:
When secret is created outside of the ARC chart using authSecret.create=false and basicAuth, the controller fails as we're not including the basic password as environment variable as the password value won't be inside the helm values.

This PR includes both environment variables for consistent regardless if those are set or not similar as the rest of the other auth options (e.g app_id, private  key, etc)

### DIFF
--- a/charts/actions-runner-controller/templates/githubwebhook.deployment.yaml
+++ b/charts/actions-runner-controller/templates/githubwebhook.deployment.yaml
@@ -94,17 +94,14 @@ spec:
               key: github_app_private_key
               name: {{ include "actions-runner-controller.githubWebhookServerSecretName" . }}
               optional: true
-        {{- if .Values.authSecret.github_basicauth_username  }}
         - name: GITHUB_BASICAUTH_USERNAME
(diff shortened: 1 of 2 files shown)


charts/actions-runner-controller/templates/githubwebhook.deployment.yaml
```diff
@@ -94,17 +94,14 @@ spec:
               key: github_app_private_key
               name: {{ include "actions-runner-controller.githubWebhookServerSecretName" . }}
               optional: true
-        {{- if .Values.authSecret.github_basicauth_username  }}
         - name: GITHUB_BASICAUTH_USERNAME
           value: {{ .Values.authSecret.github_basicauth_username }}
-        {{- end }}
```
mumoshu, line 100: I think this conditional block for `.Values.authSecret.github_basicauth_username` still has some value for consistency with others like `githubUploadURL`, `githubURL `, etc.
Let me revert that part only and merge other changes as-is 😄

mumoshu (approved):
Thanks a lot for your contribution as always!

PR #1156, opened 2022-02-24
Fix picking up wrong scale target when webhook labels are empty (such as for cloud running jobs)

tuomoa:
We are seeing scale down events due to some other jobs which are running in the github cloud. These events do not have labels so the first scale set is picked up. (I observed this while debugging other issue https://github.com/actions-runner-controller/actions-runner-controller/issues/911#issuecomment-1047523629)

What do you think could this work? I tested this in our own environment and it seems to work as expected, although I'm not sure if there is some unwanted side effects for some other scenarios?

### DIFF
--- a/controllers/horizontal_runner_autoscaler_webhook.go
+++ b/controllers/horizontal_runner_autoscaler_webhook.go
@@ -650,6 +650,11 @@ func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) getManagedRunnerGroup
 }
 
 func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) getJobScaleTarget(ctx context.Context, name string, labels []string) (*ScaleTarget, error) {
+	if len(labels) == 0 {
+		// No labels, can't be self-hosted runner
+		autoscaler.Log.V(1).Info(fmt.Sprintf("Skipping finding scale target for %s, because no labels", name))
+		return nil, nil
+	}
 	hras, err := autoscaler.findHRAsByKey(ctx, name)
 	if err != nil {
 		return nil, err


controllers/horizontal_runner_autoscaler_webhook.go
```diff
@@ -650,6 +650,11 @@ func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) getManagedRunnerGroup
 }
 
 func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) getJobScaleTarget(ctx context.Context, name string, labels []string) (*ScaleTarget, error) {
+	if len(labels) == 0 {
```
mumoshu, line 653: Ah, so we can safely assume that any workflow_job from a self-hosed runner has at least `self-hosted`, and OS-specific labels like `Linux`, and so on, right? I.e. this can be empty if and only if it's coming from a managed runner.

tuomoa: According to https://docs.github.com/en/actions/hosting-your-own-runners/using-self-hosted-runners-in-a-workflow#using-self-hosted-runners-in-a-workflow all self-hosted runners should have at least the `self-hosted` label yeah.

tuomoa: Hmm.. but if job labels were empty, can it be picked up by self hosted runner as well. I suppose it could? Does that make this fix invalid then 🤔

mumoshu: > I suppose it could?

Maybe..? But I don't think that happens reliably in practice. I thought there was no way to disable managed runners on a repository. Assuming that's true, why would one want (perhaps) a random self-hosted runner that is intended to run only workflow_jobs has corresponding labels to also run a job that had no labels(so it's likely that it was intended to run on a managed runner?

> when completed event comes is there really no way to know was it cloud runner or self-hosted which ran it?

Are you saying that a `workflow_job` of `completed` is triggered with empty labels even for one that is assigned to a self-hosted runner?

tuomoa: Yes, I think we are seeing workflow_job completed (and workflow_job in_progress) for every job, regardless which runner is running them.

tuomoa: I took a closer look at some of the payloads. Actually we might receive completed without labels also for a job which is skipped (and has labels):
```
{
  "action": "completed",
  "workflow_job": {
    ...
    "status": "completed",
    "conclusion": "skipped",
    ...
    "labels": [

    ],
```

so maybe this fix is not complete after all.

tuomoa: I think I misinterpreted ARC logs. Debugging it from the GitHub side it looks like the actual issue is that completed event without labels but with conclusion skipped is causing the scale-down for the first possible scale target 🤔

mumoshu: > But looked like we had some and kind of wanted to let you know what I think happened here. 😄

Yeah! I appreciate your cooperation like that. Thank you ☺️ 

> Actually we might receive completed without labels also for a job which is skipped (and has labels):

Interesting. Perhaps the right fix here would be to temporarily add a workaround in ARC to skip scaling down on a job with empty labels and conclusion=skipped?

We'd better also ask GitHub folks to improve it so that a workflow_job event with conclusion=skipped will have correct labels fulfilled?
I think that's crucial to enable ARC to clean up a runner pod that is added by an earlier workflow_job with status=queued, and then became unnecessary due to a job with status=completed + conclusion=skipped, in a timely manner.

tuomoa: > Interesting. Perhaps the right fix here would be to temporarily add a workaround in ARC to skip scaling down on a job with empty labels and conclusion=skipped?

I think this sounds like a way to go. But I think it's more like permanent fix? Because even so that there would be labels, it would still find a scale target to scale down, even though there was no scale up triggered by the skipped job 🤔 
 
> We'd better also ask GitHub folks to improve it so that a workflow_job event with conclusion=skipped will have correct labels fulfilled? I think that's crucial to enable ARC to clean up a runner pod that is added by an earlier workflow_job with status=queued, and then became unnecessary due to a job with status=completed + conclusion=skipped, in a timely manner.

It would make sense that also the skipped jobs would have correct labels. But I'm really not yet sure if there actually can be queued event for a job that ends up being `status=completed+conclusion=skipped` 🤔  Because if it's not known if it would be skipped, job is not yet queued? Then later on after other jobs are finishing and it is determinted to be queued or skipped, I think `status=queued` or `status=completed+conclusion=skipped` happens, but probably not both. Not sure though.

mumoshu: > actually can be queued event for a job that ends up being status=completed+conclusion=skipped 🤔

You're awesome! That's really a good point. Yeah, it can be a permanent fix then.

tuomoa: I think GitHub documentation is missing the `skipped` part, but my guess is that for skipped job there is no queued event.. not 100% sure though.

PR #1167, opened 2022-03-01
RunnerSet Reliability Improvements

mumoshu:
Based on my experience fixing RunnerDeployment reliability (as a part of https://github.com/actions-runner-controller/actions-runner-controller/pull/1127), I'm going to add several enhancements to the RunnerSet and RunnerPod controllers to make RunnerSets even more reliable.

- RunnerSet now creates one StatefulSet per Runner Pod, so that we have more control on when and how to scale down runner pods safely, as similar as we've done for RunnerDeployment in #1127
- RunnerSet spec got a new `EffectiveTime` field that is the same as RunnerDeployment's, which is used to prevent unnecessarily recreating completed runner pods when it's being scaled down quickly by webhook `workflow_job` completion events.
- A runner pod will never be marked for deletion before unregistration. This makes it possible for a runner pod that is running a long and time-consuming workflow job to not get terminated prematurely due to pod's `terminationGracePeriodSeconds`.

Ref #920

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -11,3 +11,4 @@ charts
 *.md
 *.txt
 *.sh
+test/e2e/.docker-build
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,22 +8,26 @@ WORKDIR /workspace
 ENV GO111MODULE=on \
   CGO_ENABLED=0
 
-# Copy the Go Modules manifests
-COPY go.mod go.sum ./
+# # Copy the Go Modules manifests
+# COPY go.mod go.sum ./
 
-# cache deps before building and copying source so that we don't need to re-download as much
-# and so that source changes don't invalidate our downloaded layer
-RUN go mod download
+# # cache deps before building and copying source so that we don't need to re-download as much
+# # and so that source changes don't invalidate our downloaded layer
+# RUN --mount=type=cache,target=/go/pkg/mod go mod download
 
 # Copy the go source
-COPY . .
+# COPY . .
+
+ARG TARGETOS
+ARG TARGETARCH
 
 # Build
-RUN export GOOS=$(echo ${TARGETPLATFORM} | cut -d / -f1) && \
-  export GOARCH=$(echo ${TARGETPLATFORM} | cut -d / -f2) && \
-  GOARM=$(echo ${TARGETPLATFORM} | cut -d / -f3 | cut -c2-) && \
-  go build -a -o manager main.go && \
-  go build -a -o github-webhook-server ./cmd/githubwebhookserver
+RUN --mount=target=. \
+  --mount=type=cache,mode=0777,target=/root/.cache/go-build \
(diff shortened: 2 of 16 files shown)


mumoshu:
Merging anyway. But I'll keep testing both RunnerDeployment and RunnerSet and perhaps submit a PR or two to tweak details and fix anything that can be considered as a regression!

msessa:
Hi @mumoshu,

I'd be happy to test these improvements on our live cluster (which has been suffering from the issues you're trying to solve here)

Do you provide nightly builds docker images of the controller by any chance ?

mumoshu:
@msessa Hey! Thanks for your help. I thought we'd been building controller images tagged with `canary` on each master build, which should be used for testing.
Would you mind checking it?

msessa:
Hi @mumoshu,

I've tried to run the canary images on my cluster ( arm64 architecture ) but unfortunately I stumbled on [this issue](https://github.com/actions-runner-controller/actions-runner-controller/issues/1184) when launching the webhook server.

msessa:
@mumoshu 

Just checking in on this one, after running the canary build for a few days in our live environment I can confirm this change has has a tremendous positive impact on the reliability of our runners especially with long-running jobs.

Even when running ARC with fairly aggressive scaling settings (`scaleDownDelaySecondsAfterScaleOut` set to 15 minutes) I have not encountered any premature pod termination or client disconnection so far.

The only thing I've encountered that is a bit off is that, when a runner pod is executing a long job and it ends up reaching the scale-down delay, the controller spams the log with several of these messages:

`{"level":"error","ts":1646927418.3731344,"logger":"actions-runner-controller.runnerpod","msg":"Failed to unregister runner before deleting the pod.","runnerpod":"runners/runner-amd64-nf79g-0","error":"failed to remove runner: DELETE https://api.github.com/orgs/service-victoria/actions/runners/2728: 422 Bad request - Runner \"runner-amd64-nf79g-0\" is still running a job\" []","stacktrace":"<omitted>"}`

I understand that is completely normal behavior at this point, so perhaps the error level should be lowered to `info`

mumoshu:
@msessa Hey! Thanks for the positive report. That really encourages me as I was still afraid if it can be just "works on my(@mumoshu's) environment" 😄 

Re log spam, that's really great feedback!
Let me try fixing it. Probably it doesn't need to retry that often in the first place, so I might make it only retry on, say, each 30 seconds.

msessa:
Looking forward to this new code making the next release :) meanwhile I will continue running the canary and monitor for issues. 

Thanks for your amazing work mate!

mumoshu:
@msessa Thanks for your support! FYI #1204 is my attempt to fix it.

mumoshu:
@msessa #1204 has been merged. I'd appreciate it if you could report back if it fixes it and the log will never get spammed 😄 (The intended behavior is that the 422 error is emitted at most once per a runner pod's whole lifecycle

PR #1168, opened 2022-03-01
docs: better runner group description

toast-gear:


README.md
```diff
@@ -1072,11 +1072,10 @@ spec:
...
-Supporting custom visibility requires to do a few GitHub API calls to find out what are the potential runner groups that are visible to
-the webhook's repository, this may incur in increased API rate limiting when using github.com
+authentication is included in the webhook server ARC will be assumed that all runner groups to be usable in all repositories.
+Currently, GitHub do not include the repository runner group membership information in the workflow_job event (or any webhook). To make the ARC "runner group aware" additional GitHub API calls are needed to find out what runner groups are visible to the webhook's repository. This behaviour will impact your rate-limit budget and so the option needs to be explicitly configured by the end user.
 
-This option will be enabled when proper GitHub authentication options (token, app or basic auth is provided) in the GitHub webhook server and `useRunnerGroupsVisibility` is set to true, e.g.
+This option will be enabled when proper GitHub authentication options (token, app or basic auth) is provided in the webhook server and `useRunnerGroupsVisibility` is set to true, e.g.
```
mumoshu, line 1078: In another pull request, I think it worth updating the values.yaml snippet below to also include `githubWebhookServer.secret` field like

```
githubWebhookServer:
  enabled: true
  replicaCount: 2
  useRunnerGroupsVisibility: true
  secret:
    enabled: true
    create: true
    ### GitHub Apps Configuration
    ## NOTE: IDs MUST be strings, use quotes
    #github_app_id: ""
    #github_app_installation_id: ""
    #github_app_private_key: |
    ### GitHub PAT Configuration
    #github_token: ""
```

Maybe the whole set of comments can be replaced with a single comment that links to the chart's default [values.yaml](https://github.com/actions-runner-controller/actions-runner-controller/blob/master/charts/actions-runner-controller/values.yaml) 🤔

README.md
```diff
@@ -1072,11 +1072,10 @@ spec:
 ```
 
 GitHub supports custom visilibity in a Runner Group to make it available to a specific set of repositories only. By default if no GitHub
-authentication is included in the GitHub webhook server it will be assumed that all runner groups to be usable in all repositories.
-Supporting custom visibility requires to do a few GitHub API calls to find out what are the potential runner groups that are visible to
-the webhook's repository, this may incur in increased API rate limiting when using github.com
+authentication is included in the webhook server ARC will be assumed that all runner groups to be usable in all repositories.
+Currently, GitHub do not include the repository runner group membership information in the workflow_job event (or any webhook). To make the ARC "runner group aware" additional GitHub API calls are needed to find out what runner groups are visible to the webhook's repository. This behaviour will impact your rate-limit budget and so the option needs to be explicitly configured by the end user.
```
mumoshu, line 1076: > This behaviour will impact your rate-limit budget and so the option needs to be explicitly configured by the end user.

👍 

Also note that it should be possible to use another GitHub App / another GitHub user's PAT to have a separate rate-limit budget for the webhook server only.

mumoshu (approved):
LGTM. Thanks a lot for your effort @toast-gear ☺️

PR #1175, opened 2022-03-03
Fix minor log in runner controller

fgalind1:
Log is mentioning registration only but this is about the standard runner pod


mumoshu (approved):
Good catch! Thanks for the fix ☺️

PR #1188, opened 2022-03-07
fix readme doc for The helm command for webhook installation

yourmoonlight:
The helm command for webhook installation is not working should be fixed

The cmd from readme doc should be working fine by directly copy and paste

### DIFF
--- a/README.md
+++ b/README.md
@@ -581,8 +581,9 @@ To enable this feature, you firstly need to install the webhook server, currentl
 _[see the values documentation for all configuration options](https://github.com/actions-runner-controller/actions-runner-controller/blob/master/charts/actions-runner-controller/README.md)_
 
 ```console
-$ helm upgrade actions-runner-controller actions-runner-controller/actions-runner-controller \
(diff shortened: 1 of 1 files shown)


yourmoonlight:
>

yourmoonlight:
> Hi @yourmoonlight! :)
> 
> Just happened to bump into the same one earlier today - cool!
> 
> Some comments:
> 
> 1. I guess a single "set" could be enough?
> 2. I noticed (as a MacOS and zsh user) - that I had to also add quotes around the set, like this:
> 
> ```
> 
> helm upgrade --install --namespace actions-runner-system --create-namespace \
>              --wait actions-runner-controller actions-runner-controller/actions-runner-controller \
>              --set "githubWebhookServer.enabled=true,githubWebhookServer.ports[0].nodePort=33080"
> ```
> 
> as otherwise zsh was unhappy...

Cool, thanks for your reply, I will update it.

toast-gear (approved):
LGTM, cheers pal

PR #1198, opened 2022-03-09
feat(kustomize): add github-webhook-server overlay

mgoodness:
This represents the minimum configuration required to run the GitHub webhook server component. All customization can be done via kustomize overlays.

I also removed the `--sync-period=10m` arg from config/default/manager_auth_proxy_patch.yaml, as 10 minutes is the coded default and overriding it should be done in a higher-level overlay.

### DIFF
--- a/README.md
+++ b/README.md
@@ -666,7 +666,7 @@ The primary benefit of autoscaling on Webhook compared to the pull driven scalin
 
 > You can learn the implementation details in [#282](https://github.com/actions-runner-controller/actions-runner-controller/pull/282)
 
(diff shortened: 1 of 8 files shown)


toast-gear (approved):
LGTM

Debelopercgx:
Hey anyone "out there" or "in here" CG. Bad, bad day, i think it's done!

Debelopercgx (approved):
Wow! you truly are special Boss. Sorry I've got life circumstances that basically prevent life right now but maybe down the road big guy. We can do a coding date or something. Thanks CG. Happy New Year!

Issue #1046, opened 2022-01-11
containers create by Github workflow have wrong dockerMTU


erichorwath:
**Describe the bug**
No network possible from Docker containers created by Github workflows (dind - Docker in Docker)

**To Reproduce**
1. create workflow with docker image, e.g.

```
name: Setup Go

on:
  workflow_dispatch:

jobs:
  setupgo:
    runs-on: [kubernetes]
    container:
      image: 'ubuntu:latest'
    steps:
    - name: Set up Go
      uses: actions/setup-go@v2
      with:
        go-version: 1.17
    
    - name: sleep
      shell: bash
      run: sleep 300
```

2. exec into the dind sidecar container and confirm that mtu parameter is propagated properly from runner deployment spec:

```
apiVersion: actions.summerwind.dev/v1alpha1
kind: RunnerDeployment
metadata:
  name: my-runner
spec:
  template:
    spec:
      organization: abc
      ephemeral: true
      dockerMTU: 1440
```

```
$ kubectl exec -it my-runner-xyz -c docker -- /bin/sh 
$ ps auxww
    1 root      0:00 docker-init -- dockerd --host=unix:///var/run/docker.sock --host=tcp://0.0.0.0:2376 --tlsverify --tlscacert /certs/server/ca.pem --tlscert /certs/server/cert.pem --tlskey /certs/server/key.pem --mtu 1440
   53 root      0:03 dockerd --host=unix:///var/run/docker.sock --host=tcp://0.0.0.0:2376 --tlsverify --tlscacert /certs/server/ca.pem --tlscert /certs/server/cert.pem --tlskey /certs/server/key.pem --mtu 1440
   61 root      0:04 containerd --config /var/run/docker/containerd/containerd.toml --log-level info
   ...
$ ip a
  ...
  4: eth0@if700: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1440 qdisc noqueue state UP
      link/ether d2:ee:d5:2f:04:ae brd ff:ff:ff:ff:ff:ff
      inet 100.96.2.200/32 brd 100.96.2.200 scope global eth0
         valid_lft forever preferred_lft forever
      inet6 fe80::d0ee:d5ff:fe2f:4ae/64 scope link
         valid_lft forever preferred_lft forever
```

3. run the workflow and exec into the workflow docker container

```
$ kubectl exec -it my-runner-xyz -c docker -- /bin/sh
$ docker ps
[...]

Issue #1137, opened 2022-02-20
Supply Chain Security of Base Image Sources


Fleshgrinder:
Various additional software is installed in the runner image from different sources. However, the authenticity of the downloaded software is not verified. Everything assumes that the transport encryption is sufficient. There sadly is no single solution for each software that is being added, because they originate from different sources.

> 🔥 Obviously supplying the images built as part of this project is also insecure, but I am going to create a dedicated issue for that later.

> 💡 Note that I am happy to help resolving these issues. I am only reporting it first so that there is a place to discuss. ☺️

## git

https://github.com/actions-runner-controller/actions-runner-controller/blob/0b9bef2c086a54b0731fa7aa0a6bd7cb856b48ee/runner/Dockerfile#L14

This imports the GPG key of the PPA maintainer into the global `apt-secure` trust store and consequently ultimately trusts if for any package from any repository.[^apt-key] It is also using the HTTP URL of Launchpad, because the `software-properties-common` version that is part of the Ubuntu base we are using does not have [the HTTPS patch](https://bugs.launchpad.net/ubuntu/+source/software-properties/+bug/1959015).

The proper way would be to download the GPG key from a keyserver via the long fingerprint. Adding it to `/usr/share/keyrings` and adding a source that locks both the architecture and the singing key.

## dumb-init

https://github.com/actions-runner-controller/actions-runner-controller/blob/0b9bef2c086a54b0731fa7aa0a6bd7cb856b48ee/runner/Dockerfile#L51

There is nothing with which we could verify the authenticity of the downloaded executable. The vendor provides some SHA hashes, but they are provided via the same channel. Hence, if someone gained access to change the release assets that someone can change those hashes as well. Quick fix would be to include the expected hash just like the version. Proper fix would be to either ask the vendor to provide a secure supply chain, or to switch to https://github.com/krallin/tini who provide proper signatures along their binaries.

## docker

https://github.com/actions-runner-controller/actions-runner-controller/blob/0b9bef2c086a54b0731fa7aa0a6bd7cb856b48ee/runner/Dockerfile#L59

There is [a proper Debian repository](https://docs.docker.com/engine/install/ubuntu/) from where the binaries can be retrieved. I would say that this is the most secure way to get this into the image (same approach as with [git](#git), download key from keyserver, etc.).

## actions-runner

[...]

PR #1201, opened 2022-03-10
feat: Add container to propagate host network MTU

tiagoblackcode:
Some network environments use non-standard MTU values. In these situations, the `DockerMTU` setting might be used to specify the MTU setting for the `bridge` network created by Docker. However, when the Github Actions workflow creates networks, it doesn't propagate the `bridge` network MTU which can lead to `connection reset by peer` messages.

To overcome this, I've created a new docker image called `summerwind/actions-runner-mtu` that shims the docker binary in order to propagate the MTU setting to networks created by Github workflows.

This is a follow-up on the discussion in #1046 and uses a separate image since there might be some unintended side-effects with this approach.

### DIFF
--- a/runner/docker-shim.sh
+++ b/runner/docker-shim.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -Eeuo pipefail
+
+if [[ ${ARC_DOCKER_MTU_PROPAGATION:-false} == true ]] &&
+  (($# >= 2)) && [[ $1 == network && $2 == create ]] &&
+  mtu=$(/usr/bin/docker network inspect bridge --format '{{index .Options "com.docker.network.driver.mtu"}}' 2>/dev/null); then
+  shift 2
+  set -- network create --opt com.docker.network.driver.mtu="$mtu" "$@"
+fi
+
+exec /usr/bin/docker "$@"
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -256,8 +256,28 @@ spec:
       env: []
 ```
 
-There may be more places you need to tweak for MTU.
-Please consult issues like #651 for more information.
+If the issue still persists, you can set the `ARC_DOCKER_MTU_PROPAGATION` to propagate the host MTU to networks created
+by the GitHub Runner. For instance:
+
+```yaml
+apiVersion: actions.summerwind.dev/v1alpha1
+kind: RunnerDeployment
+metadata:
+  name: github-runner
+  namespace: github-system
+spec:
+  replicas: 6
+  template:
+    spec:
+      dockerMTU: 1400
+      repository: $username/$repo
+      env:
+        - name: ARC_DOCKER_MTU_PROPAGATION
+          value: "true"
+```
+
+You can read the discussion regarding this issue in
+(#1406)[https://github.com/actions-runner-controller/actions-runner-controller/issues/1046].
 
 ## Unable to scale to zero with TotalNumberOfQueuedAndInProgressWorkflowRuns
 
--- a/runner/Makefile
+++ b/runner/Makefile
@@ -1,4 +1,5 @@
 DOCKER_USER ?= summerwind
+DOCKER ?= docker
 NAME ?= ${DOCKER_USER}/actions-runner
 DIND_RUNNER_NAME ?= ${DOCKER_USER}/actions-runner-dind
 TAG ?= latest
@@ -26,14 +27,14 @@ else
 endif
 
 docker-build-ubuntu:
-	docker build \
+	${DOCKER} build \
 	  --build-arg TARGETPLATFORM=${TARGETPLATFORM} \
 	  --build-arg RUNNER_VERSION=${RUNNER_VERSION} \
 	  --build-arg RUNNER_CONTAINER_HOOKS_VERSION=${RUNNER_CONTAINER_HOOKS_VERSION} \
 	  --build-arg DOCKER_VERSION=${DOCKER_VERSION} \
 	  -f actions-runner.dockerfile \
 	  -t ${NAME}:${TAG} .
-	docker build \
+	${DOCKER} build \
 	  --build-arg TARGETPLATFORM=${TARGETPLATFORM} \
 	  --build-arg RUNNER_VERSION=${RUNNER_VERSION} \
 	  --build-arg DOCKER_VERSION=${DOCKER_VERSION} \
@@ -41,8 +42,8 @@ docker-build-ubuntu:
 	  -t ${DIND_RUNNER_NAME}:${TAG} .
 
 docker-push-ubuntu:
-	docker push ${NAME}:${TAG}
-	docker push ${DIND_RUNNER_NAME}:${TAG}
+	${DOCKER} push ${NAME}:${TAG}
+	${DOCKER} push ${DIND_RUNNER_NAME}:${TAG}
 
 docker-buildx-ubuntu:
 	export DOCKER_CLI_EXPERIMENTAL=enabled ;\
@@ -50,14 +51,14 @@ docker-buildx-ubuntu:
 	@if ! docker buildx ls | grep -q container-builder; then\
 	  docker buildx create --platform ${PLATFORMS} --name container-builder --use;\
 	fi
-	docker buildx build --platform ${PLATFORMS} \
+	${DOCKER} buildx build --platform ${PLATFORMS} \
 	  --build-arg RUNNER_VERSION=${RUNNER_VERSION} \
 	  --build-arg RUNNER_CONTAINER_HOOKS_VERSION=${RUNNER_CONTAINER_HOOKS_VERSION} \
 	  --build-arg DOCKER_VERSION=${DOCKER_VERSION} \
 	  -f actions-runner.dockerfile \
 	  -t "${NAME}:${TAG}" \
 	  . ${PUSH_ARG}
-	docker buildx build --platform ${PLATFORMS} \
+	${DOCKER} buildx build --platform ${PLATFORMS} \
 	  --build-arg RUNNER_VERSION=${RUNNER_VERSION} \
 	  --build-arg RUNNER_CONTAINER_HOOKS_VERSION=${RUNNER_CONTAINER_HOOKS_VERSION} \
 	  --build-arg DOCKER_VERSION=${DOCKER_VERSION} \
--- a/runner/actions-runner-dind.dockerfile
+++ b/runner/actions-runner-dind.dockerfile
@@ -68,7 +68,7 @@ RUN export ARCH=$(echo ${TARGETPLATFORM} | cut -d / -f2) \
 	tar --extract \
 		--file docker.tgz \
 		--strip-components 1 \
-		--directory /usr/local/bin/ \
+		--directory /usr/bin/ \
 	; \
 	rm docker.tgz; \
 	dockerd --version; \
--- a/runner/actions-runner.dockerfile
+++ b/runner/actions-runner.dockerfile
@@ -59,7 +59,7 @@ RUN set -vx; \
     && if [ "$ARCH" = "amd64" ] || [ "$ARCH" = "i386" ]; then export ARCH=x86_64 ; fi \
     && curl -f -L -o docker.tgz https://download.docker.com/linux/static/${DOCKER_CHANNEL}/${ARCH}/docker-${DOCKER_VERSION}.tgz \
     && tar zxvf docker.tgz \
-    && install -o root -g root -m 755 docker/docker /usr/local/bin/docker \
+    && install -o root -g root -m 755 docker/docker /usr/bin/docker \
     && rm -rf docker docker.tgz \
     && adduser --disabled-password --gecos "" --uid 1000 runner \
     && groupadd docker \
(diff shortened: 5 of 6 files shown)


arpitjindal97:
Can this be done on same image itself, because MTU is different on every public cloud. This value is important and user must be aware about it. 

Always set `MTU` value in image with default `1400` and user has flexibility to override it via spec

arpitjindal97:
I mean to say, default value can be `1400` and use can modify it anytime.
Reason of `1400` is because GCP provides `1460` where setting MTU will become mandatory to set it

Ref: https://cloud.google.com/network-connectivity/docs/vpn/concepts/mtu-considerations

tiagoblackcode:
I'm not sure I'm following, but I don't think we should have a customisation specifically for a cloud provider. There are situations where the MTU can stay at 1500 and we should not decrease it implicitly.

mumoshu:
@tiagoblackcode Sorry for the delayed repsonse! I just merged https://github.com/actions-runner-controller/actions-runner-controller/pull/1257 which adds some documentation on the MTU issue worth being revised to also cover this work.

Also, I think we can include the shim in our standard container image while enhancing entrypoint.sh so that it toggles the use of the shim behind a feature flag(a runner container envvar specified via RunnerDeployment/RunnerSet specs).
WDYT?

Fleshgrinder:
There really is no need for a separate image, this can all be done in the same image we already have. Wrapping the original binary in a shell script is also no problem, if we `exec` the original binary upon execution.

Docker should not be installed to `/usr/local/bin` but rather to `/usr/bin` since it is not a user binary, but one we provide as part of this project to users. (It should also not be installed manually in the first place, but that is a different story we are going to fix as part of https://github.com/actions-runner-controller/actions-runner-controller/issues/1137.) Once that is done we can override the executable in `/usr/local/bin`. The following script would result in a safe override:

```bash
#!/usr/bin/env bash
set -Eeuo pipefail

if (($# >= 2)) && [[ $1 == network && $2 == create ]] && mtu=$(docker network inspect bridge --format '{{index .Options "com.docker.network.driver.mtu"}}' 2>&-); then
  shift 2
  set -- network create --opt com.docker.network.driver.mtu="$mtu" "$@"
fi

exec /usr/bin/docker "$@"
```

tiagoblackcode:
I'm okay with using the official image and wrapping the docker binary with the shell script. Should we wait until the docker binary is moved to `/usr/bin` as suggested by @Fleshgrinder or shall we go ahead?

Once docker binary is moved:

- `/usr/bin/docker` is the docker binary
- `/usr/local/bin/docker` is the docker shim (the shell script which propagates the MTU)

Before the docker binary is moved:

- `/usr/local/bin/docker-unwrapped` becomes the official docker binary (perhaps the naming isn't the best?)
- `/usr/local/bin/docker` is the docker shim

In this case we don't need to add further directories to the `PATH`.

Fleshgrinder:
You can move as part of this PR, these two lines need to be adjusted:

- https://github.com/actions-runner-controller/actions-runner-controller/blob/8a8ec43364f8c7335dbd0dca65c215d6f1aa6fa5/runner/actions-runner.dockerfile#L61
- https://github.com/actions-runner-controller/actions-runner-controller/blob/8a8ec43364f8c7335dbd0dca65c215d6f1aa6fa5/runner/actions-runner-dind.dockerfile#L71

tiagoblackcode:
Updated the PR with the changes. Haven't had the chance to test it properly yet, but it's still missing the flag to enable the MTU propagation.

@Fleshgrinder in your shell script, where does the `ARC_DOCKER_MTU_PROPAGATION` come from? Is that something that should be configurable in the `Runner` CRD?

Also, AFAIK the dind runner doesn't have this problem so I just placed the shim in the standard runner. If you wish I can added to the dind runner as well.

Fleshgrinder:
The `ARC_DOCKER_MTU_PROPAGATION` in my example script is an environment variable that the user can set as part of their deployment.

Fleshgrinder:
These are also turned into environment variables so that they can be passed to the Bash scripts. I leave the API design decisions here to @mumoshu.

tiagoblackcode:
I understand, but in terms of API I'd rather configure the runner using spec than an environment variable. I'll wait for @mumoshu 's input before committing to a solution.

Thanks for your input.

tiagoblackcode:
That makes sense. I'll add some documentation about the flag and do some testing.
I'll ping back when it's ready.

tomsherrod:
> That makes sense. I'll add some documentation about the flag and do some testing. I'll ping back when it's ready.

I've ran into the mtu issue. Built image with this update. Containers run with no connection issues. Thank you.

tiagoblackcode:
@mumoshu @Fleshgrinder I just picked this up again, sorry for the delay.

Made some tests with the `ARC_DOCKER_MTU_PROPAGATION` and it seems to be working correctly. Updated the `TROUBLESHOOTING.md` doc to mention the flag wrt to the MTU issue.

I can setup a test similar to those in `test/entrypoint/should_*` for the docker shim if you wish. Otherwise I think this is ready to merge.

runner/docker-shim.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -Eeuo pipefail
+
+if [[ ${ARC_DOCKER_MTU_PROPAGATION:-false} == true ]] &&
+  (($# >= 2)) && [[ $1 == network && $2 == create ]] &&
+  mtu=$(/usr/bin/docker network inspect bridge --format '{{index .Options "com.docker.network.driver.mtu"}}' 2>&-); then
```
mumoshu, line 7: I'm afraid of using `2>&-` as we encountered #1560 due to our lack of understanding of how `FD>&-` works.
Can we perhaps use `2>/dev/null` as a safe alternative?

mumoshu:
@tiagoblackcode Thanks for the update! This looks good to me overall. I'll merge this after I give it a try on my own cluster(although I don't have a quick access to a cluster with a non standard MTU but at least verifying it to work on my cluster would make it more confident that we won't break anything for a standard env 😄 

> I can setup a test similar to those in test/entrypoint/should_* for the docker shim if you wish. Otherwise I think this is ready to merge.

I'd greatly appreciate it if you could do so, in this PR or in another PR depending on the situation!

runner/docker-shim.sh
Fleshgrinder: Actually we do understand what it means to close a file descriptor, and there's no need to be afraid of it. Vigilant, yes, testing properly, yes, but no need to be afraid. Goroutines suffer from races and deadlocks, we still use them, and are not switching to Rust because of it. 😉 That said, the difference is marginal in this case. Swallowing stderr is often the bigger issue, as it might hide issues. Hence, we could also just let it inherit stderr and print.

mumoshu: Thanks! But did we already know why `1>&-` caused the problem in some envs and not in other envs? That's my concern.

It does look like a "correct" way to write a bash script, but I thought we are yet to see any proof or data point about why `1>&-` can be considered reliable, right? 🤔

mumoshu: > we do understand what it means to close a file descriptor

To be extra clear, my point is if closing it to suppress some output is the right way to do it, as it can break like in #1560. `2>/dev/null` doesn't close it(right?) so it won't result in issues like #1560. Why use `1>&-` then..?

Fleshgrinder: Yes, I explained it in https://github.com/actions-runner-controller/actions-runner-controller/pull/1560#issuecomment-1170961218. Using `FD>&-` is the same as `File.Close` in Go. We are essentially closing the file descriptor so that all writes to it simply never happen, no bits are ever being transferred. Using `FD>/dev/null` means that all bits are still transferred to the device. In this case a device that does nothing with the bits, but they still need to be transferred.

The problem with `FD>&-` arises if the caller does never check if the file descriptor (or file handle) is actually writable, and simply writes. This is obviously going to fail, since the file is not writable, it is closed.

Many programs are expecting FD1 to always be writable, no matter what. That is what we saw in the other issue. However, the same is not true for any FD greater one. Here is a little more on this in the context of C specifically: https://stackoverflow.com/a/4973065/1251219

I am totally fine with using `>/dev/null` everywhere for safety reasons. It is just important to me that everyone understands the difference. This really has nothing to do with Bash, it is just normal programming in relation to how to deal with file handles, and the same in every programming language where we have full control over file handles.

Fleshgrinder: https://www.kernel.org/doc/html/latest/block/null_blk.html according to this `/dev/null` does not actually execute any read/write operations. Maybe there really is no difference anymore in modern Linux.

tiagoblackcode:
> @tiagoblackcode Thanks for the update! This looks good to me overall. I'll merge this after I give it a try on my own cluster(although I don't have a quick access to a cluster with a non standard MTU but at least verifying it to work on my cluster would make it more confident that we won't break anything for a standard env 😄

You can actually make some kind of test:

```
# lower the main interface MTU in the k8s host(s)
# WARNING: this may cause issues in other pods
for host in "<host>" "<host>"
do 
  ssh "root@$host" "ip link set <interface> mtu 1420"
done

# update the Runner to use a lower MTU for the docker container and keep the original image
kubectl patch -n <namespace> runners.actions.summerwind.dev <runner> --type='merge' \
  -p '{"spec": {"dockerMTU": 1420}}'

# force the runners to recreate
for p in $(kubectl get pod -n <namespace> -o custom-columns="NAME:.metadata.name,CONTROLLER:.metadata.ownerReferences[].kind" | grep Runner | awk '{ print $1 }')
do
  kubectl -n <namespace> delete pod $p
done

# perform the test inside one of the runners

runner=$(kubectl get pod -n <namespace> -o custom-columns="NAME:.metadata.name,CONTROLLER:.metadata.ownerReferences[].kind" | grep Runner | awk '{ print $1 }' | head -n 1)

kubectl exec -n <namespace> -it $runner -c runner -- bash

# this is similar to what GitHub Actions does when running a workflow
docker network create no-mtu

# this is similar to what the `docker-shim` does when proxying the network creation to docker
docker network create mtu --opt com.docker.network.driver.mtu=1420

# this should not end correctly with the container using the original image
docker run --rm --network no-mtu curlimages/curl --max-time 10 https://www.google.com

# this should work with both patched and unpatched images
docker run --rm --network mtu curlimages/curl --max-time 10 https://www.google.com

# To tear the test down: 
# - restore the host MTU
# - patch the runner to use/remove the `dockerMTU` configuration
```

The test above should use the original image to assert the _current_ image has issues on docker networks created with the default configuration when the underlying network infra-structure has a lower MTU.

To test the new image, patch the runner CRD to use an image created with this PR, and re-run the test above. The test using the `no-mtu` network should now work properly. This should assert the _new_ image propagates the MTU to docker networks created with the default configuration.

runner/docker-shim.sh
mumoshu: First of thanks a lot for sharing your insights and knowledge. I learned something!

> Yes, I explained it in https://github.com/actions-runner-controller/actions-runner-controller/pull/1560#issuecomment-1170961218

Ahh, thanks! Sorry, I somehow missed it.

> The problem with FD>&- arises if the caller does never check if the file descriptor (or file handle) is actually writable, and simply writes. This is obviously going to fail, since the file is not writable, it is closed.

This makes sense a lot! Thanks again for clarifying.

Does it basically mean that when using `$FD>&-`, we need to be sure if the target program checks the availability of the FD before writing to it? 

In this specific case, the use of `2>&-` is fine as long as `docker network inspect bridge` either checks for the availability of the fd 2(stderr) before writing, or it handles stderr write errors gracefully, right?

tiagoblackcode: I've updated the code to use `2>/dev/null` redirection instead. According to @Fleshgrinder redirecting to `/dev/null` does actually execute writes so there's no difference.

tiagoblackcode:
> > I can setup a test similar to those in test/entrypoint/should_* for the docker shim if you wish. Otherwise I think this is ready to merge.
> 
> I'd greatly appreciate it if you could do so, in this PR or in another PR depending on the situation!

Regarding this, I need to explore a bit how to mock calls to `/usr/bin/docker` without interfering with the host and/or the script. Doing a `chroot` into a mocked FS might be an option. So I'd say we merge without the test if it's okay with you.

runner/docker-shim.sh
Fleshgrinder: @tiagoblackcode way to go.

@mumoshu yes, that is what one need to know about the program before file handles can be closed. Normal programs can deal with stderr being closed, but not stdin and stdout.

mumoshu (approved):
Manually tested this and as far as I can see from what MTUs are set on docker0 and net interfaces for containers and docker networks, this is working flawlessly!
LGTM. Thanks a lot for your contribution @tiagoblackcode!

BTW, it turned out we have to add support for dind runners and rootless dind runners. I'm going to submit a follow-up PR for that.

mumoshu:
Oops, you already touched dind runner's Dockerfile! This doesn't work because you missed updating supervisor.conf which is used to run the dind dockerd within the runner container, which points to the inexistent `/usr/local/bin/dockerd` (Notice the trailing "d"- the dockerd binary moved to /usr/bin along with the docker command).
Let me add a commit to fix it. Otherwise the users of dind runners will encounter their runners start failing after this gets merged.

mumoshu:
I'll follow up with rootless dind runners later.

PR #1225, opened 2022-03-15
docs: remove runnerset limitation

toast-gear:
This works great from testing now, this is no longer a limitation due to ARC now creating a statefulset per runner


mumoshu (approved):
Thanks for your confirmation!

Issue #1162, opened 2022-02-28
Use the new container friendly `run.sh` file


toast-gear:
As part of the latest runner softawre updates proper cotainer support was added https://github.com/actions/runner/releases/tag/v2.288.0. We need to test our runner using the newly patched `run.sh` file as well as remove our patched `RunnerService.js`

PR #1244, opened 2022-03-18
fix: unit tests were very broken + use run.sh

toast-gear:
- Fixes https://github.com/actions-runner-controller/actions-runner-controller/issues/1162


toast-gear:
This is good enough for now. Now the unit tests are fixed and are more readable and better documented we need to actually test that the `run.sh` doesn't break anything and that we no longer need our patched runner files.

mumoshu (approved):
LGTM. Thanks a lot for your awesome work ☺️

PR #1246, opened 2022-03-19
Improved Bash Logger

Fleshgrinder:
This is a first step towards having robust Bash scripts in the runner images. The changes _could_ be considered breaking, depending on our backwards compatibility definition.


toast-gear:
Resolve the conflicts and we can look at reviewing this, I don't have an issue with it fundamentally

Fleshgrinder:
@toast-gear done

runner/logger.bash
```diff
@@ -0,0 +1,73 @@
...
+  level=${level^^} # UPPERCASE
+
+  if [[ ! -v "LOG_${level}_DISABLED" ]]; then
+    instant=$(date '+%F %T.%-3N' 2>/dev/null || :)
+
+    # https://no-color.org/
+    if [[ -v NO_COLOR ]]; then
+      printf -- '%s  %s --- %s\n' "$instant" "$level" "$*" 1>&2 || :
+    else
+      printf -- '\033[0;%dm%s  %s --- %s\033[0m\n' "$color" "$instant" "$level" "$*" 1>&2 || :
```
toast-gear, line 56: Can I ask about the delimit choice? Why delimit with with spaces except for the message which is delimited via a `---`? Why not delimit entirely with commas? I haven't got strong feelings on this, just want to ensure we get it right first time as any subsequent changes will be potentially impactful.

Fleshgrinder: > Why not delimit entirely with commas?

I assume you meant spaces here. The pattern is the default console log pattern from Spring Boot (although simplified by removing the parts we don‘t need).[^1] We could use just spaces to separate the level from the message too. I cannot think of any functionality that could possibly break.

```diff
- '\033[0;%dm%s  %s --- %s\033[0m\n'
+ '\033[0;%dm%s  %s   %s\033[0m\n'
```

Afterwards we are still able to get each individual part with simple “index of” operations, e.g.:

```bash
line='2022-03-19 10:01:23.172  NOTICE   example message'
echo "date:    ${line%% *}"
(time=${line%%  *}; echo "time:    ${time#* }")
(level=${line%%   *}; echo "level:   ${level#*  }")
echo "message: ${line#*   }"
```

```
date:    2022-03-19
time:    10:01:23.172
level:   NOTICE
message: example message
```

Using a single space is also possible and works nicely with `cut`, although it makes parsing of the log line a little harder as a human (I guess that is generally the argument for the `---` separator):

```bash
line='2022-03-19 10:01:23.172 NOTICE example message'
echo "date:    $(cut -d' ' -f1 <<<"$line")"
echo "time:    $(cut -d' ' -f2 <<<"$line")"
echo "level:   $(cut -d' ' -f3 <<<"$line")"
echo "message: $(cut -d' ' -f4- <<<"$line")"
```

```
date:    2022-03-19
time:    10:01:23.172
level:   NOTICE
message: example message
```

That said, all of this is possible with the original format as well:

```bash
line='2022-03-19 10:01:23.172  NOTICE --- example message'
echo "date:    ${line%% *}"
(time=${line%%  *}; echo "time:    ${time#* }")
(level=${line%% --- *}; echo "level:   ${level#*  }")
echo "message: ${line#* --- }"
echo
echo "date:    $(cut -d' ' -f1 <<<"$line")"
echo "time:    $(cut -d' ' -f2 <<<"$line")"
echo "level:   $(cut -d' ' -f4 <<<"$line")"
echo "message: $(cut -d' ' -f6- <<<"$line")"
```

```
date:    2022-03-19
time:    10:01:23.172
level:   NOTICE
message: example message

date:    2022-03-19
time:    10:01:23.172
level:   NOTICE
message: example message
```

I have no strong feelings here either, and am happy to change it to any format. ☺️

> […] any subsequent changes will be potentially impactful.

Yes

[^1]: https://github.com/spring-projects/spring-boot/blob/d6f111ce59298a2166523b71ffa5d2071334f4c9/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/DefaultLogbackConfiguration.java#L70-L73

PR #1247, opened 2022-03-19
Removed `modprobe` Script

Fleshgrinder:
I was able to find out that this script originates from https://github.com/docker-library/docker/blob/master/modprobe.sh but our image does not have `lsmod` nor `modprobe` installed. Hence, if it were in use, it would fail every time. 🤔


toast-gear:
@Fleshgrinder could you resolve the conflicts so we can merge this please

toast-gear:
@Fleshgrinder resolve the conflicts please

PR #1251, opened 2022-03-21
Use const envvars

rofafor:
No functional changes, but make sure the same envvar definitions are used everywhere inside the controller.


toast-gear (approved):
LGTM cheers pal

Issue #1258, opened 2022-03-22
Missing conversionReviewVersions field in config/crd/patches/webhook_in_runners.yaml


MarioUhrik:
**Describe the bug**
When `config/crd/patches/webhook_in_runners.yaml` is used and `kubectl apply/create` is attempted on the generated Runner CRD, it fails due to missing required field `conversionReviewVersions`.

```
error: error validating "STDIN": error validating data: ValidationError(CustomResourceDefinition.spec.conversion.webhook): missing required field "conversionReviewVersions" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.WebhookConversion; if you choose to ignore these errors, turn validation off with --validate=false
```

**Checks**

- [x] My actions-runner-controller version (v0.x.y) does support the feature
- [ ] I'm using an unreleased version of the controller I built from HEAD of the default branch

**To Reproduce**
1. Edit kustomization.yaml such that patch `config/crd/patches/webhook_in_runners.yaml` is used
2. `kustomize build | kubectl apply -f -`
3. See error

**Expected behavior**
`kubectl apply` should finish without errors

**Environment (please complete the following information):**
 - Controller Version - [v0.22.0](https://github.com/actions-runner-controller/actions-runner-controller/tree/v0.22.0)
 - Deployment Method - kustomize
 - GKE K8s 1.21

PR #1259, opened 2022-03-22
Runner CRD: Add required conversionReviewVersions field

MarioUhrik:
Without that field, GKE 1.21 refuses to create the CRD with an error message that conversionReviewVersions is mandatory.

```
conversionReviewVersions is a required field when creating apiextensions.k8s.io/v1 custom resource definitions.
Webhooks are required to support at least one ConversionReview version understood by the current and previous API server.
````

Source: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/_print/#webhook-request-and-response

Resolves https://github.com/actions-runner-controller/actions-runner-controller/issues/1258

### DIFF
--- a/config/crd/patches/webhook_in_runners.yaml
+++ b/config/crd/patches/webhook_in_runners.yaml
@@ -8,6 +8,7 @@ spec:
   conversion:
     strategy: Webhook
     webhook:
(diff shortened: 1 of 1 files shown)


mumoshu (approved):

PR #1262, opened 2022-03-22
Fix deleting a runner when pod was deleted

fgalind1:
With the current implementation if a pod is deleted, controller is failing to delete the runner as it's trying to annotate a pod that doesn't exist as we're passing a new pod object that is not an existing resource. We need to send `nil` if pod wasn't found to make sure we don't attempt to annotate a non-existing pod

### DIFF
--- a/controllers/runner_controller.go
+++ b/controllers/runner_controller.go
@@ -106,15 +106,16 @@ func (r *RunnerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 			return ctrl.Result{}, nil
 		}
 	} else {
+		// Request to remove a runner. DeletionTimestamp was set in the runner - we need to unregister runner
 		var pod corev1.Pod
 		if err := r.Get(ctx, req.NamespacedName, &pod); err != nil {
 			if !kerrors.IsNotFound(err) {
 				log.Info(fmt.Sprintf("Retrying soon as we failed to get runner pod: %v", err))
 				return ctrl.Result{Requeue: true}, nil
 			}
+			// Pod was not found
+			return r.processRunnerDeletion(runner, ctx, log, nil)
 		}
-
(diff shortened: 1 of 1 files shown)


fgalind1:
@mumoshu do you know if something odd is going on with CI? interestingly a few different PRs started failing on the same after rebasing e.g this PR and this https://github.com/actions-runner-controller/actions-runner-controller/pull/1268

mumoshu (approved):
Makes sense! Thanks for the fix.
I thought we still needed `processRunnerDeletion` in that case to try unregister the runner for the disappeared pod, but turned out it's unnecessary because we now defer the registration to the runner pod controller.

mumoshu:
@fgalind1 It seems like the setup-go action went wild (or we totally misunderstood how it's configured) that it's now installing go 1.18.x even though we specify go 1.17.x

CleanShot 2022-04-05 at 11 26 11@2x

The fact that `controller-gen` isn't found in the path even though the log says go built it might be due to that Go 1.18 changed `go get` to not install the binary.
We'd need to replace `go get` we use in Makefile to install controller-gen to `go install`.

fgalind1:
> @fgalind1 It seems like the setup-go action went wild (or we totally misunderstood how it's configured) that it's now installing go 1.18.x even though we specify go 1.17.x
> 
> CleanShot 2022-04-05 at 11 26 11@2x
> 
> The fact that `controller-gen` isn't found in the path even though the log says go built it might be due to that Go 1.18 changed `go get` to not install the binary. We'd need to replace `go get` we use in Makefile to install controller-gen to `go install`.

interesting - thanks. This https://github.com/actions-runner-controller/actions-runner-controller/pull/1303 should at least unblock master branch and make sure there are no surprises. It's off because ^ should work and stick the minor version indeed as per semver docs

mumoshu (approved):

PR #1264, opened 2022-03-23
Use container name to detect runner container in Pod

ekarlso:
This switches to instead of using index which is easily breakable if using the sidecar pattern to using the name of "runner" instead ot determine the container inside of the runner Pod.

EDIT by @mumoshu: Note that the issue reproduces only when you use a mutating webhook other than ARC's own to inject arbitrary container at the index 0 of a runner pod.

### DIFF
--- a/controllers/runner_pod_controller.go
+++ b/controllers/runner_pod_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


mumoshu (approved):
LGTM. Thanks a lot for the fix @ekarlso!

PR #1268, opened 2022-03-23
feat: allow to discover runner statuses

fgalind1:
This is a new optional feature that allows to set the `.status.phase` of a Runner to a more specific/accurate phase based on the status of the runner. This leverages the recent runner feature https://github.com/actions/runner/pull/1737 that allows to set hooks on job start and job end. The end goal is to make it easier for admins or however manages runners via ARC to easily see what each runner is doing `e.g Created, Registering, Idle, Running, Idle, Running...`

`entrypoint.sh` will update the runner status to Registering and Idle as it goes trough the runner registration. Runner will set the status to Running (along with the run number and repo using that runner at that moment) on job start via the runner hook and runner back to idle on job end

In order to do so - this requires the Runner pod to be able to update its own `.status` by itself; thus, this PR makes sure that whenever we're creating a new runner, we create a custom role, service account and role binding that has access only to update the runner/status of his own.

All these things to make this work (RBAC and the runner status) can be optionally enabled with the chart option `useCustomRunnersRBAC`. By default is disabled and this should have no impact in the current behavior. At some point this may be defaulted to true or just remove the old behavior but this was implemented this way to have it as an opt-in feature

Example of how it looks like:

```
NAME                     ENTERPRISE   ORGANIZATION   REPOSITORY                    LABELS         STATUS       MESSAGE                  AGE
test-runner-wdgpj-2s6k7                              XXXX/personal.fgalind1.test   ["fgalind1"]   Idle                                  11m
test-runner-wdgpj-9m6g8                              XXXX/personal.fgalind1.test   ["fgalind1"]   Registering                           11m
test-runner-wdgpj-gngp2                              XXXX/personal.fgalind1.test   ["fgalind1"]   Running      Run 1925397027 from XXXX 10m
```


controllers/runner_controller.go
```diff
@@ -134,13 +139,16 @@ func (r *RunnerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
-	if runner.Status.Phase != phase || runner.Status.Ready != ready {
+	if (runner.Status.Phase != phase || runner.Status.Ready != ready) && !r.ConfigureRunnersRBAC || runner.Status.Phase == "" && r.ConfigureRunnersRBAC {
 		if pod.Status.Phase == corev1.PodRunning {
 			// Seeing this message, you can expect the runner to become `Running` soon.
 			log.V(1).Info(
 				"Runner appears to have been registered and running.",
 				"podCreationTimestamp", pod.CreationTimestamp,
 			)
+			if r.ConfigureRunnersRBAC {
+				phase = "Created"
```
mumoshu, line 150: Why do we need this treatment?

controllers/runner_controller.go
```diff
@@ -259,6 +321,27 @@ func (r *RunnerReconciler) processRunnerCreation(ctx context.Context, runner v1a
 	return ctrl.Result{}, nil
 }
 
+func (r *RunnerReconciler) createObject(ctx context.Context, obj client.Object, meta metav1.ObjectMeta, runner v1alpha1.Runner, owner metav1.Object, log logr.Logger) *ctrl.Result {
```
mumoshu, line 324: I believe the `owner` can be merged into `runner` if we received it as `*v1alpha1.Runner`

```suggestion
func (r *RunnerReconciler) createObject(ctx context.Context, obj client.Object, meta metav1.ObjectMeta, runner *v1alpha1.Runner, log logr.Logger) *ctrl.Result {
```

controllers/runner_controller.go
```diff
@@ -552,6 +637,10 @@ func newRunnerPod(runnerName string, template corev1.Pod, runnerSpec v1alpha1.Ru
 			Name:  EnvVarEphemeral,
 			Value: fmt.Sprintf("%v", ephemeral),
 		},
+		{
+			Name:  "RUNNER_CUSTOM_RBAC",
```
mumoshu, line 641: I'd prefer naming this envvar more around for what it works, like:

```suggestion
			Name:  "RUNNER_STATUS_UPDATE_HOOK",
```

The use of a custom RBAC can conceptually be optional if one prefers using a shared rolebinding and a serviceaccount that have broader permissions

runner/entrypoint.sh
```diff
@@ -3,6 +3,12 @@
 RUNNER_ASSETS_DIR=${RUNNER_ASSETS_DIR:-/runnertmp}
 RUNNER_HOME=${RUNNER_HOME:-/runner}
 
+if [ "${RUNNER_CUSTOM_RBAC:-}" == "true" ]; then
+  # Let runner execute these hooks
+  export ACTIONS_RUNNER_HOOK_JOB_STARTED=/job_start.sh
+  export ACTIONS_RUNNER_HOOK_JOB_COMPLETED=/job_end.sh
+fi
```
mumoshu, line 10:
```suggestion
# Let runner execute these hooks
export ACTIONS_RUNNER_HOOK_JOB_STARTED=/job_start.sh
export ACTIONS_RUNNER_HOOK_JOB_COMPLETED=/job_end.sh
```

I think it's fine to enable these hooks by default because we skip the status update anyway when the envvar is not specified.

runner/update_status.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+if [ "${RUNNER_CUSTOM_RBAC:-}" == "true" ]; then
```
mumoshu, line 2:
```suggestion
if [ "${RUNNER_STATUS_UPDATE_HOOK:-}" == "true" ]; then
```

controllers/runner_controller.go
```diff
@@ -237,6 +245,60 @@ func (r *RunnerReconciler) processRunnerCreation(ctx context.Context, runner v1a
 		return ctrl.Result{}, err
 	}
 
+	if r.ConfigureRunnersRBAC {
+		serviceAccount := &corev1.ServiceAccount{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      runner.ObjectMeta.Name + "-sa",
```
mumoshu, line 251:
```suggestion
				Name:      runner.ObjectMeta.Name,
```

I prefer not having a `-sa` prefix here, due to a reason similar to why I don't like naming a pod like `myrunner-pod` :)

controllers/runner_controller.go
```diff
@@ -237,6 +245,60 @@ func (r *RunnerReconciler) processRunnerCreation(ctx context.Context, runner v1a
...
+				APIGroups:     []string{"actions.summerwind.dev"},
+				Resources:     []string{"runners/status"},
+				Verbs:         []string{"get", "update", "patch"},
+				ResourceNames: []string{runner.ObjectMeta.Name},
+			},
+		}
+
+		role := &rbacv1.Role{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      runner.ObjectMeta.Name + "-role",
```
mumoshu, line 270:
```suggestion
				Name:      runner.ObjectMeta.Name,
```

Same as above

controllers/runner_controller.go
```diff
@@ -237,6 +245,60 @@ func (r *RunnerReconciler) processRunnerCreation(ctx context.Context, runner v1a
...
+			},
+			Rules: rules,
+		}
+		if res := r.createObject(ctx, role, role.ObjectMeta, runner, &runner, log); res != nil {
+			return *res, nil
+		}
+
+		roleBinding := &rbacv1.RoleBinding{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      runner.ObjectMeta.Name + "-rb",
```
mumoshu, line 281:
```suggestion
				Name:      runner.ObjectMeta.Name,
```

Same as above

controllers/runner_controller.go
```diff
@@ -237,6 +245,60 @@ func (r *RunnerReconciler) processRunnerCreation(ctx context.Context, runner v1a
...
+			},
+			RoleRef: rbacv1.RoleRef{
+				APIGroup: "rbac.authorization.k8s.io",
+				Kind:     "Role",
+				Name:     runner.ObjectMeta.Name + "-role",
+			},
+			Subjects: []rbacv1.Subject{
+				{
+					Kind:      "ServiceAccount",
+					Name:      runner.ObjectMeta.Name + "-sa",
```
mumoshu, line 292:
```suggestion
					Name:      runner.ObjectMeta.Name,
```

controllers/runner_controller.go
```diff
@@ -237,6 +245,60 @@ func (r *RunnerReconciler) processRunnerCreation(ctx context.Context, runner v1a
...
+
+		roleBinding := &rbacv1.RoleBinding{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      runner.ObjectMeta.Name + "-rb",
+				Namespace: runner.ObjectMeta.Namespace,
+			},
+			RoleRef: rbacv1.RoleRef{
+				APIGroup: "rbac.authorization.k8s.io",
+				Kind:     "Role",
+				Name:     runner.ObjectMeta.Name + "-role",
```
mumoshu, line 287:
```suggestion
				Name:     runner.ObjectMeta.Name,
```

charts/actions-runner-controller/templates/deployment.yaml
```diff
@@ -65,6 +65,9 @@ spec:
         {{- if .Values.runnerGithubURL  }}
         - "--runner-github-url={{ .Values.runnerGithubURL }}"
         {{- end }}
+        {{- if .Values.useCustomRunnersRBAC  }}
```
mumoshu, line 68:
```suggestion
        {{- if .Values.useRunnerStatusUpdateHookEphemeralRole  }}
```

After reading all this, perhaps this might better reflect (1)how log the custom RBAC resources persist(=as long as the runner pod exists) and (2)the purpose of the custom RBAC resources(=to enable runner status update hook in runner pods)?

charts/actions-runner-controller/README.md
```diff
@@ -72,6 +72,7 @@ All additional docs are kept in the `docs/` folder, this README is solely for do
 | `scope.watchNamespace`                                   | Tells the controller and the github webhook server which namespace to watch if `scope.singleNamespace` is true             | `Release.Namespace` (the default namespace of the helm chart).       |
 | `scope.singleNamespace`                                  | Limit the controller to watch a single namespace                                                                           | false                                                                |
 | `certManagerEnabled`                                     | Enable cert-manager. If disabled you must set admissionWebHooks.caBundle and create TLS secrets manually                   | true                                                                 |
+| `useCustomRunnersRBAC`                                   | Use custom RBAC for runners (role, role binding and service account), this will enable reporting runner statuses           | false                                                                |
```
mumoshu, line 75:
```suggestion
| `useRunnerStatusUpdateHookEphemeralRole `                                   | Use custom RBAC for runners (role, role binding and service account), this will enable reporting runner statuses           | false                                                                |
```

mumoshu (changes requested):
Sorry for the delayed review!
LGTM overall, but it would be great if you could confirm my comments 🙏

runner/entrypoint.sh
mumoshu: AFAIK we can only specify only one script per hook, and that makes me wonder how existing users already using the hooks to leverage this new runner status update hooks.

Perhaps we'd better introduce another layer of indirection, so that existing users can put their own hooks in a predefined location, and ARC's default hooks finds and runs those custom hooks along with these `job_start.sh` and `job_end.sh`?

For example, how about making `job_start.sh` look like:

```
for hook in $(ls /hooks/job_started); do
  $hook "$1" "$2"
done
```

And we put e.g. `job_start.sh` into `/hooks/job_started/job_start.sh`?

runner/Dockerfile.dindrunner
```diff
@@ -1,7 +1,7 @@
 FROM ubuntu:20.04
 
 ARG TARGETPLATFORM
-ARG RUNNER_VERSION=2.287.1
+ARG RUNNER_VERSION=2.289.1
```
mumoshu, line 4: I think we can safely omit this change from Dockerfiles now as we already bumped to more recent version of actions/runner which supports runner hooks.

runner/Dockerfile
```diff
@@ -110,7 +110,8 @@ RUN mkdir /opt/hostedtoolcache \
     && chgrp docker /opt/hostedtoolcache \
     && chmod g+rwx /opt/hostedtoolcache
 
-COPY entrypoint.sh /
+COPY entrypoint.sh job_start.sh job_end.sh update_status.sh /
+RUN chmod 755 entrypoint.sh job_start.sh job_end.sh update_status.sh
```
mumoshu, line 114: Why did you need to chmod entrypoint.sh too? (I thought it's already executable by any user/group

mumoshu:
@fgalind1 I tried my best to rebase your awesome work onto our main branch. Would you mind testing it?

runner/update_status.sh
```diff
@@ -0,0 +1,15 @@
...
+    APISERVER=https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT_HTTPS}
+    SERVICEACCOUNT=/var/run/secrets/kubernetes.io/serviceaccount
+    NAMESPACE=$(cat ${SERVICEACCOUNT}/namespace)
+    TOKEN=$(cat ${SERVICEACCOUNT}/token)
+    CACERT=${SERVICEACCOUNT}/ca.crt
+
+    curl --silent -X PATCH --cacert ${CACERT} -H "Content-Type: application/merge-patch+json" -H "Authorization: Bearer ${TOKEN}" \
+        ${APISERVER}/apis/actions.summerwind.dev/v1alpha1/namespaces/${NAMESPACE}/runners/${HOSTNAME}/status \
+        -d "{\"status\":{\"phase\": \"$1\", \"message\": \"$2\"}}"
+fi
```
Fleshgrinder, line 15: All the files in this patch are missing the final line feed required by git (and many other tools). You can see an error icon being display here on GitHub in the file view. Please add a final line feed to every file.

runner/update_status.sh
Fleshgrinder, line 2:
```suggestion
#!/usr/bin/env bash
set -Eeuo pipefail

if [[ ${1:-} == '' ]]; then
  # shellcheck source=runner/logger.bash
  source logger.bash
  log.error "Missing required argument -- '<phase>'"
  exit 64
fi

if [[ ${RUNNER_CUSTOM_RBAC:-false} == true ]]; then
```

runner/update_status.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+if [ "${RUNNER_CUSTOM_RBAC:-}" == "true" ]; then
+    export HTTPS_PROXY=
+    export https_proxy=
```
Fleshgrinder, line 4: What is the purpose of these two lines? Do we want to unset them?

runner/update_status.sh
```diff
@@ -0,0 +1,15 @@
...
+
+    APISERVER=https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT_HTTPS}
+    SERVICEACCOUNT=/var/run/secrets/kubernetes.io/serviceaccount
+    NAMESPACE=$(cat ${SERVICEACCOUNT}/namespace)
+    TOKEN=$(cat ${SERVICEACCOUNT}/token)
+    CACERT=${SERVICEACCOUNT}/ca.crt
+
+    curl --silent -X PATCH --cacert ${CACERT} -H "Content-Type: application/merge-patch+json" -H "Authorization: Bearer ${TOKEN}" \
+        ${APISERVER}/apis/actions.summerwind.dev/v1alpha1/namespaces/${NAMESPACE}/runners/${HOSTNAME}/status \
+        -d "{\"status\":{\"phase\": \"$1\", \"message\": \"$2\"}}"
```
Fleshgrinder, line 14: We have to use `jq` to create the JSON that we are about to post, because JSON requires valid UTF-8 and certain bytes need to be escaped properly (just think of `"` in either `$1` or `$2`).

Local variables should always be written in lowercase. By convention the OS and shell use uppercase variables for environment and system variables, we do not want to interact with them.

Using long options and most of the time preferable over short options, as it makes the code more readable. Exceptions are very well known options (e.g. `rm -f`) or commands that do not provide long options (e.g. `jq -n`).

```suggestion
    serviceaccount=/var/run/secrets/kubernetes.io/serviceaccount
    namespace=$(cat $serviceaccount/namespace)
    token=$(cat $serviceaccount/token)
    jq -n --arg phase "$1" --arg message "${2:-}" '.status = $ARGS.named' | curl \
      --cacert $serviceaccount/ca.crt \
      --data @- \
      --header "Authorization: Bearer $token" \
      --header 'Content-Type: application/merge-patch+json' \
      --show-error \
      --silent \
      --request PATCH \
      "https://$KUBERNETES_SERVICE_HOST:$KUBERNETES_SERVICE_PORT_HTTPS/apis/actions.summerwind.dev/v1alpha1/namespaces/$namespace/runners/$HOSTNAME/status"
```

runner/Dockerfile
```diff
@@ -113,6 +113,10 @@ RUN mkdir /opt/hostedtoolcache \
 # We place the scripts in `/usr/bin` so that users who extend this image can
 # override them with scripts of the same name placed in `/usr/local/bin`.
 COPY entrypoint.sh logger.bash /usr/bin/
+RUN chmod 755 entrypoint.sh
+
+COPY job_start.sh job_end.sh update_status.sh /
+RUN chmod 755 job_start.sh job_end.sh update_status.sh
```
Fleshgrinder, line 119: Why would we not install the scripts into `/usr/bin` and thus allow resolution from `PATH` and users to override it with their own scripts in `/usr/local/bin`?

I see no reason for the `.sh` suffix for the filenames, any reason for it? Personally I dislike underscores (`_`) in filenames because it is an old Unix tradition to use hyphens (`-`), a tradition that came into existence before most of us where even born. That said, the scripts are so simple, the could be combined into a single one. Unless we want to give users the ability to override them, in which case placing them in `/usr/bin` is again the more logical choice. Or, what @mumoshu proposed, a dedicated directory where the should be stored too, in this case I would propose the following that follows Unix conventions once more:

```
/etc/arc/runner/job
+-- started
+-- started.d/
|   +-- user-provided-script
|   +-- another-user-provided-script
|   +-- ...
+-- stopped
+-- stopped.d/
    +-- user-provided-script
    +-- another-user-provided-script
    +-- ...
```

controllers/runner_controller.go
Fleshgrinder: I'm still generally afraid of using `RUNNER_` variables for ARC specific things. GitHub uses them and we are only asking for a future collision by continuing this practice. We really should switch to prefixing them with `ARC_`.

We also need a way to tell variables apart that are meant to be passed down to the runner, and those that should not be passed down. (You surely remember the automatic, prefix-based unsetting idea for the entrypoint script.) Not sure how we could do that in a sustainable way.

fgalind1: Updated to RUNNER_STATUS_UPDATE_HOOK for now as prefixing env variables may be of a larger scope for this PR as most likely we'll want consistency in all env vars

runner/entrypoint.sh
fgalind1: I'm merging this together with @Fleshgrinder 's idea from https://github.com/actions-runner-controller/actions-runner-controller/pull/1268#discussion_r874052393

controllers/runner_controller.go
mumoshu: > GitHub uses them

I thought they tend to use `ACTIONS_RUNNER_` prefix these days. Do you have any references on the use of `RUNNER_`-prefixed envvars by GitHub?(Probably I lost my memory? 🤔

mumoshu: Anyway, it would be nicer if we could migrate to a more unique prefix in the future, like `ARC_`.

controllers/runner_controller.go
fgalind1: not needed - I've removed in the next commits

runner/update_status.sh
fgalind1: I wanted to avoid using proxy when running the curl command in case people uses proxies in the pod

runner/Dockerfile
fgalind1: removed - wasn't needed 👍

fgalind1:
> Sorry for the delayed review!
LGTM overall, but it would be great if you could confirm my comments 🙏

> @fgalind1 I tried my best to rebase your awesome work onto our main branch. Would you mind testing it?

Thanks @mumoshu as always for the review - I've address the comments from you and @Fleshgrinder 

I've tested this flow with the feature on and off and it's working as expected 👍

mumoshu:
@fgalind1 Thanks! We're almost there- the last thing is the failing TestNewRunnerPodFromRunnerController. Seems like you need to add the new envvar to the testdata

runner/hooks/completed.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# shellcheck source=runner/logger.bash
+source logger.bash
+
+log.notice "Running ARC Completed Hooks"
```
mumoshu, line 6:
```suggestion
log.notice "Running ARC Job Completed Hooks"
```

We'd better say it's for jobs, or it can be read like "Runner Started Hook" which sounds like something else!

runner/hooks/completed.sh
```diff
@@ -0,0 +1,10 @@
...
+
+for hook in $(ls /etc/arc/hooks/completed.d); do
```
mumoshu, line 8:
```suggestion
for hook in $(ls /etc/arc/hooks/job_completed.d); do
```

I prefer having the `job_` prefix. Same for `completed.sh`

runner/hooks/started.d/job-start
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
mumoshu, line 1: Probably this file might better be named `job_started.d/update_status`, so that we won't over complete this script in the future(I'd like to keep one hook script do only one thing, which might be better enforced by naming this script `update_status` rather than a more general `job-start`.

fgalind1: I've renamed all hook files for consistency to use job prefix and dashes

```
runner@k8s:/$ find /etc/arc/
/etc/arc/
/etc/arc/hooks
/etc/arc/hooks/job-completed.d
/etc/arc/hooks/job-completed.d/update-status
/etc/arc/hooks/job-started.d
/etc/arc/hooks/job-started.d/update-status
/etc/arc/hooks/job-started.sh
/etc/arc/hooks/job-completed.sh

```

runner/update-status
```diff
@@ -0,0 +1,28 @@
...
+set -Eeo pipefail
+if [[ ${1:-} == '' ]]; then
+  # shellcheck source=runner/logger.bash
+  source logger.bash
+  log.error "Missing required argument -- '<phase>'"
+  exit 64
+fi
+
+if [ "${RUNNER_STATUS_UPDATE_HOOK:-false}" == "true" ]; then
+
```
Fleshgrinder, line 12: We should never use `test` (`[]`) because it works weirdly compared Bash’s `[[]]`.

```suggestion
if [[ ${RUNNER_STATUS_UPDATE_HOOK:-false} == true ]]; then
```

runner/update-status
```diff
@@ -0,0 +1,28 @@
...
+    apiserver=https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT_HTTPS}
+    serviceaccount=/var/run/secrets/kubernetes.io/serviceaccount
+    namespace=$(cat ${serviceaccount}/namespace)
+    token=$(cat ${serviceaccount}/token)
+
+    jq -n --arg phase "$1" --arg message "$2" '.status.phase = $phase | .status.message = $message' | HTTPS_PROXY= http_proxy= curl \
```
Fleshgrinder, line 18: The second argument is optional, we have to handle this.

```suggestion
    jq -n --arg phase "$1" --arg message "${2:-}" '.status.phase = $phase | .status.message = $message' | HTTPS_PROXY= http_proxy= curl \
```

We could also treat all arguments that follow as the message:

```suggestion
    phase=$1
    shift
    jq -n --arg phase "$phase" --arg message "${*:-}" '.status.phase = $phase | .status.message = $message' | HTTPS_PROXY= http_proxy= curl \
```

runner/update-status
```diff
@@ -0,0 +1,28 @@
...
+        --cacert ${serviceaccount}/ca.crt \
+        --data @- \
+        --header "Content-Type: application/merge-patch+json" \
+        --header "Authorization: Bearer ${token}" \
+        --show-error \
+        --silent \
+        --output /dev/null \
+        --request PATCH \
+        ${apiserver}/apis/actions.summerwind.dev/v1alpha1/namespaces/${namespace}/runners/${HOSTNAME}/status > /dev/null
```
Fleshgrinder, line 27: - If we want to disable proxies we should use `--noproxy '*'` that disables it for all hosts. However, I'm not sure if we actually want this, it should be a user choice.
- If we want curl to be truly silent then we can just close its stdout. (We should not ignore the error output, and we are not right now, curl failures will fail our script because we use `set -e` and this is how it should be.)
- We have to quote the URL to avoid issues with possible whitespace.

```suggestion
    jq -n --arg phase "$1" --arg message "$2" '.status.phase = $phase | .status.message = $message' | curl \
        --cacert ${serviceaccount}/ca.crt \
        --data @- \
        --header "Content-Type: application/merge-patch+json" \
        --header "Authorization: Bearer ${token}" \
        --noproxy '*' \
        --show-error \
        --silent \
        --request PATCH \
        "${apiserver}/apis/actions.summerwind.dev/v1alpha1/namespaces/${namespace}/runners/${HOSTNAME}/status"
        1>&-
```

runner/hooks/job-completed.sh
```diff
@@ -0,0 +1,11 @@
...
+
+# shellcheck source=runner/logger.bash
+source logger.bash
+
+log.notice "Running ARC Job Completed Hooks"
+
+for hook in $(ls /etc/arc/hooks/job-completed.d); do
+  log.notice "Running hook: $hook"
+  /etc/arc/hooks/job-completed.d/$hook "$1" "$2"
+done
```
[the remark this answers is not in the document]
fgalind1: I've address all the changes except the .sh - we can run any script regardless of the extension - the extension requirement is only for the script used in the ACTIONS_RUNNER_HOOK_JOB_* env. I expanded the comment on this where those env are set

fgalind1:
thanks for the feedback @Fleshgrinder. I've addressed the feedback on https://github.com/actions-runner-controller/actions-runner-controller/pull/1268/commits/068de406909b5278fd3b2366b682a83438270287

fgalind1:
@mumoshu in case you have a chance to take a look after the feedback :) thanks

mumoshu:
@fgalind1 Thanks for the reminder! We'll be merging this next week. We're currently in a code-freeze period and therefore this week's going to be used for fixes (if any) to the recently released 0.24.0

fgalind1:
> @fgalind1 Thanks for the reminder! We'll be merging this next week. We're currently in a code-freeze period and therefore this week's going to be used for fixes (if any) to the recently released 0.24.0

awesome - thanks for the note 👍

fgalind1:
@mumoshu - looks like runner workflow are missing some secrets? 
image

toast-gear:
@fgalind1  merged master into your branch and see if that fixes it

toast-gear:
I've update the workflow to just use a hardcoded dockerhub name, it's not a secret anyway

fgalind1:
> I've update the workflow to just use a hardcoded dockerhub name, it's not a secret anyway @fgalind1 pull and merge master again

awesome - thanks @toast-gear - after merge master it's working again 👍

fgalind1:
> @fgalind1 Thanks for the reminder! We'll be merging this next week. We're currently in a code-freeze period and therefore this week's going to be used for fixes (if any) to the recently released 0.24.0

@mumoshu could this change be processed? :) appreciate the help

mumoshu:
I did my best fixing conflicts in [96d7925](https://github.com/actions-runner-controller/actions-runner-controller/pull/1268/commits/96d7925f85d25e0627365aea0769007b353ae622). Wishing it works

config/rbac/role.yaml
```diff
@@ -1,10 +1,8 @@
-
----
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   creationTimestamp: null
-  name: manager-role
+  name: {{ include "actions-runner-controller.managerRoleName" . }}
```
mumoshu, line 5: This seems messed up. This is a resource loaded by ARC's kustomize config, which has no access to these Helm template expressions

charts/actions-runner-controller/values.yaml
```diff
@@ -66,6 +66,7 @@ image:
 imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
+useRunnerStatusUpdateHookEphemeralRole: false
```
mumoshu, line 69: I started to wonder if this can be improved after https://github.com/actions-runner-controller/actions-runner-controller/pull/1546.
One might want to choose if they want ARC to create a service account for the runner pod, and separately enable additional features like the kubernetes container mode and/or this status update hooks.
Considering all the possibles variations of configs, perhaps something like this would be better:

```
runner:
  serviceAccount:
    create: true # This will let ARC create a service account for each runner pod
    #name: <specify the name of existing SA. Makes sense only when `create: false`
  statusUpdateHook:
    enabled: true #=> This enables te runner status update hook and adds the role and role binding to either the existing or the dynamically created service account
  #Uncomment if you want the kubernetes container mode.
  # In case runner.serviceAccount.created=true,runner.statusUpdateHook.enabled=true,containerMode=kubernetes is specified,
  # ARC will create a runner service account for each runner pod and adds all the permissions required for both the container mode and the status update hooks
  #containerMode: kubernetes
```

fgalind1: I believe these are indeed 2 different features but the scope a bit different. containerMode is something you want to add to your runners, while status update hook I believe is a more horizonal feature (e.g this PR adds a new column in the runner's table to show the running job, which applies for all runners. In other words I believe it makes more sense to be an on/off feature but to the controller - rather than be a switch per runner. While for the containerMode I do believe that makes sense in the context of the runner. Thoughts?

config/rbac/role.yaml
fgalind1: good call - right. fixed in https://github.com/actions-runner-controller/actions-runner-controller/pull/1268/commits/9c0c9add18c8fa1f376162081c27f03da30d3e8f

fgalind1:
> @fgalind1 Hey! This has been always in my head. Sorry for the delay, I do realize this is now 3 months old and I'm very eager to get this merged!
> 
> I had reviewed this once again this morning and, unfortunately, it turned out that a part of this enhancement (conceptually) conflicts with the new container mode support #1546.
> 
> The new container mode expects the user to pre-create a service account used by the runner pod, whereas this enhancement dynamically creates a service account for each runner pod(which lacks permissions necessary for the new container mode). Please correct me if I'm saying something wrong.
> 
> I'm planning to modify this a bit so that (1)the dynamically created service account would also have any permissions required for the new container mode(so that this feature can be used along with the new container mode) and/or (2)this feature can optionally use a pre-created serviceaccount, so that the user can provided either of both sets of permissions required for the new container mode and this feature, to the pre-created serviceaccount.
> 
> I'm working hard to make it happen earlier, but probably it won't make it before 0.25.0, which I'm planning to cut this weekend if everything went well. So, this might get merged in next or the week after the next.
> 
> Hope it works. Thanks for your contribution and patience 🙏

I've added option #1 here https://github.com/actions-runner-controller/actions-runner-controller/pull/1268/commits/d3e1f96e0f0ae8eabc121aa7bd73d8cccb23415e - let me know what you think

That way is consistent - it can let user provide a custom service account (that may be used for kubernetes or this feature) or if it is not provided we create one with the right permissions for kubernetes and/or statuses

charts/actions-runner-controller/values.yaml
mumoshu: Thanks for confirming! You're absolutely correct.

In my previous comment, I made two mistakes.

First, I confused the runner service account vs workflow job pod's service account. This feature affects the former and the new container mode is related to the latter.

Second, my suggested values.yaml snippet did look like containing settings for runners, not for the controller. Configuring runners is out of the scope of the ARC's chart. It might make sense a bit if it was for default runner configs, as we already have some runner defaults related values like `image.actionsRunnerRepositoryAndTag`. However, this feature should indeed be a controller-level feature as you pointed out.

mumoshu: That said, perhaps it would look ok if we just omitted the containerMode key from my previous example?

```
runner:
  serviceAccount:
    create: true # This will let ARC create a service account for each runner pod
  statusUpdateHook:
    enabled: true #=> This enables the runner status update hook. If serviceaccount.create=true, this also creates the role and bind it to the dynamically created service account
```

To let ARC use an existing runner pod service account, you'd say:

```
runner:
  serviceAccount:
    name: mysa # The name of an existing SA. Valid only when `create: false`
    # In this case you don't need to give ARC the permission to create SA at all, which might be preferable to some folks for security reasons.
  statusUpdateHook:
    enabled: true # This doesn't automatically create a role and rolebinding associated to the existing SA, similarly to the top-level `serviceAccount.name` which his used to specify the existing SA for the controller-manager pod.
```

Perhaps it might be a good idea to imply `runner.serviceAccount.create=true` when you set `runner.statusUpdateHook.enabled=true`:

```
runner:
  statusUpdateHook:
    enabled: true #=> This enables the runner status update hook. In order to do so, ARC automatically creates a runner pod serviceaccount and a role with necessary permissions, and binds the role to the SA.
```

This way, it would look like a controller-wide on/off feature, while providing enough flexibility to let the user decide whether to bring their own SA for runner pods, or not.

WDYT?

mumoshu: > First, I confused the runner service account vs workflow job pod's service account. This feature affects the former and the new container mode is related to the latter.

Sorry, this was again a complete misunderstanding of mine! The new container mode works by giving an existing SA to every runner pod so that it can create workflow job pods from within runner pods. Never mind about this. But I think my suggestion in https://github.com/actions-runner-controller/actions-runner-controller/pull/1268#discussion_r912317146 is still valid.
It's just that we won't need to provide the container mode related settings in helm charts as it's a runner-level feature.

controllers/runner_controller.go
```diff
@@ -256,6 +261,91 @@ func (r *RunnerReconciler) processRunnerCreation(ctx context.Context, runner v1a
 		return ctrl.Result{}, err
 	}
 
+	needsServiceAccount := runner.Spec.ServiceAccountName != "" && (r.UseRunnerStatusUpdateHookEphemeralRole || runner.Spec.ContainerMode == "kubernetes")
```
mumoshu, line 264: I think `runner.Spec.ServiceAccountName != ""` is an indication that the user wants ARC to use the specified existing services ccount, whereas this condition results in the opposite. Perhaps we'd better invert it so that it looks like:

```suggestion
	needsServiceAccount := runner.Spec.ServiceAccountName == "" && (r.UseRunnerStatusUpdateHookEphemeralRole || runner.Spec.ContainerMode == "kubernetes")
```

controllers/runner_controller.go
```diff
@@ -256,6 +261,91 @@ func (r *RunnerReconciler) processRunnerCreation(ctx context.Context, runner v1a
...
+				{
+					APIGroups:     []string{"actions.summerwind.dev"},
+					Resources:     []string{"runners/status"},
+					Verbs:         []string{"get", "update", "patch"},
+					ResourceNames: []string{runner.ObjectMeta.Name},
+				},
+			}...)
+		}
+
+		if runner.Spec.ContainerMode == "kubernetes" {
```
mumoshu, line 289: Awesome! Thanks for the addition ☺️

charts/actions-runner-controller/values.yaml
mumoshu: I think my suggestion in this thread can be distilled into just one sentence considering the scope of this pull request. That's "Can we rename useRunnerStatusUpdateHookEphemeralRole to runner.statusUpdateHook.enabled, so that the values will be kept structured great even if we added an ability to let ARC use a pre-created SA as the default runner pod SA in the future?"

In other words, even I mentioned it in my previous examples, we don't need to add support for `runner.serviceAccount.create` or `runner.serviceAccount.name` in this PR. That might be out of the scope of this PR.

controllers/runner_controller.go
fgalind1: good catch - fixed in [`ec02bb7` (#1268)](https://github.com/actions-runner-controller/actions-runner-controller/pull/1268/commits/ec02bb77a613d173d79364958a19b58c1e706b0b)

charts/actions-runner-controller/values.yaml
fgalind1: sounds good to me :) I've done the rename to `runner.statusUpdateHook.enabled` in [`ec02bb7` (#1268)](https://github.com/actions-runner-controller/actions-runner-controller/pull/1268/commits/ec02bb77a613d173d79364958a19b58c1e706b0b)

mumoshu:
I'm currently testing this on top the current branch and hopefully merge this soon as long as I don't see any major issues. Thanks for your patience!

controllers/runner_controller.go
mumoshu: Enabling this resulted in ARC unable to create the role due to an error like:

```
{"level":"error","ts":"2022-07-10T01:31:04Z","logger":"actions-runner-controller.runner","msg":"Retrying as failed to create Role org-runnerdeploy-ngmsw-pwd5r resource","runner":"default/org-runnerdeploy-ngmsw-pwd5r","error":"roles.rbac.authorization.k8s.io \"org-runnerdeploy-ngmsw-pwd5r\" is forbidden: user \"system:serviceaccount:actions-runner-system:actions-runner-controller\" (groups=[\"system:serviceaccounts\" \"system:serviceaccounts:actions-runner-system\" \"system:authenticated\"]) is attempting to grant RBAC permissions not currently held:\n{APIGroups:[\"\"], Resources:[\"pods/exec\"], Verbs:[\"get\" \"create\"]}\n{APIGroups:[\"\"], Resources:[\"pods/log\"], Verbs:[\"get\" \"list\" \"watch\"]}\n{APIGroups:[\"\"], Resources:[\"secrets\"], Verbs:[\"create\" \"delete\"]}","stacktrace":"github.com/actions-runner-controller/actions-runner-controller/controllers.(*RunnerReconciler).processRunnerCreation\n\t/workspace/controllers/runner_controller.go:322\ngithub.com/actions-runner-controller/actions-runner-controller/controllers.(*RunnerReconciler).Reconcile\n\t/workspace/controllers/runner_controller.go:133\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.2/pkg/internal/controller/controller.go:121\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.2/pkg/internal/controller/controller.go:320\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.2/pkg/internal/controller/controller.go:273\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.2/pkg/internal/controller/controller.go:234"}
```

As the error indicates, this is probably due to ARC's controller-manager role lacks permissions for `pods/exec` and `pods/log`, and create/delete of `secrets`. Without the permissions, Kubernetes seems to think ARC is trying to do  a privilege escalation (from ARC controller-manager's lesser permission set, to the runner pod's greater permission set), which is apparently forbidden by Kubernetes.

mumoshu: Probably we'd need another chart value flag to make this privilege escalation possible, and it should be orthogonal to `statusUpdateHook` 🤔

controllers/runner_controller.go
```diff
@@ -256,6 +261,91 @@ func (r *RunnerReconciler) processRunnerCreation(ctx context.Context, runner v1a
...
+				{
+					APIGroups: []string{""},
+					Resources: []string{"pods/log"},
+					Verbs:     []string{"get", "list", "watch"},
+				},
+				{
+					APIGroups: []string{""},
+					Resources: []string{"secrets"},
+					Verbs:     []string{"get", "list", "create", "delete"},
+				},
```
mumoshu, line 311: While testing this manually, I realized that this lacks permissions for jobs, resulting in a workflow job to fail like this:

CleanShot 2022-07-10 at 11 53 35@2x

mumoshu (approved):
After fixing two minor issues I found (https://github.com/actions-runner-controller/actions-runner-controller/pull/1268/files#r912319021 and https://github.com/actions-runner-controller/actions-runner-controller/pull/1268/files#r917331632), it worked with the "kubernetes" container mode, too 🎉 I'll submit a follow-up PR or two for those issues.

One last thing to note- the hook updates the runner status to e.g. `Registering` and `Idle` before/after the registration. It's great. But I suspect it might have a race condition with the runner controller's logic that syncs the runner pod status to the runner resource:

https://github.com/actions-runner-controller/actions-runner-controller/blob/10b88bf0702ebbf8e0e2f45a5634bc1111d9b0ac/controllers/runner_controller.go#L138-L157

If the runner controller's resync period passed and hence it reconciled the runner resource immediately after the hook update the runner status to `Idle`, the runner status might be "corrected" to `Running`(=pod status phase). So it can look a bit flappy if you watched the runner status phase transitions continuously (by e.g. running `kubectl get runner -w` and triggering a bunch of workflow runs). It might not be a big deal, but might be worth a fix.

I think I'm confident this works almost perfectly, and we know what the remaining issues are and how to address those. That said, I consider this is now good to be merged!

Thanks again for your patience and contribution @fgalind1 😄

fgalind1:
> this works almost perfectly, and we know what the remaining issues a

Lovely! thanks also for your help @mumoshu

PR #1285, opened 2022-03-30
docs: add limitations to runnersets + reorder

toast-gear:
**Changes**
- Added PVC / PV build up limitation
- Moved persistent runners further up as it's a runner mode and so should be group with the kinds and it has major implications for RunnerSets


mumoshu (approved):
LGTM! Thanks a lot for your continued effort on improving our documentation :bow:

PR #1313, opened 2022-04-06
Add more usages of RUNNER_VERSION to Renovate config.

### LINKED ISSUE #1278: Get Renovate to bump all RUNNER_VERSION references when bumping the version
We declare the runner version in a few spots but only bump the pipeline reference currently. It would be great if our Renovate setup could patch all references when raising a runner version bump, current locations:

* https://github.com/actions-runner-controller/actions-runner-controller/blob/master/.github/workflows/runners.yml#L18
* https://github.com/actions-runner-controller/actions-runner-controller/blob/master/runner/Makefile#L7
* https://github.com/actions-runner-controller/actions-runner-controller/blob/master/runner/Dockerfile#L4
* https://github.com/actions-runner-controller/actions-runner-controller/blob/master/runner/Dockerfile.dindrunner#L4

danxmoran:
Closes #1278

I'm not sure of the best way to test this 🤔

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -14,8 +14,25 @@
       // use https://github.com/actions/runner/releases
       "fileMatch": [
         ".github/workflows/runners.yml"
-        ],
+      ],
       "matchStrings": ["RUNNER_VERSION: +(?<currentValue>.*?)\\n"],
+      "depNameTemplate": "actions/runner",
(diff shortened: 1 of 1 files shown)


toast-gear:
Cheers pal, I'll test it on 1 of our test repositories and see what happens! :D

PR #1277, opened 2022-03-27
refactor(runner/entrypoint): check for externalstmp

lovesegfault:
This fixes these annoying messages in the runner logs:
```
# Authentication
√ Connected to GitHub
# Runner Registration
√ Runner successfully added
√ Runner connection is good
# Runner settings
√ Settings Saved.
Runner successfully configured.
﻿{
  "agentId": 123,
  "agentName": "runner-deployment-123",
  "poolId": 1,
  "poolName": "Default",
  "ephemeral": true,
  "serverUrl": "https://pipelines.actions.githubusercontent.com/123",
  "gitHubUrl": "https://github.com/123",
  "workFolder": "/runner/_work"
mkdir: cannot create directory './externals': File exists
mv: cannot stat './externalstmp/*': No such file or directory
}
√ Connected to GitHub

Current runner version: '2.289.1'
2022-03-27 11:04:41Z: Listening for Jobs
```

Emphasis:

```
mkdir: cannot create directory './externals': File exists
mv: cannot stat './externalstmp/*': No such file or directory
```

PR #1315, opened 2022-04-06
refactor(runner/entrypoint): don't mv externalstmp if it's not there

lovesegfault:
This is a follow-up to #1277, which was edited before landing causing the issue
to not be fixed.

Essentially, we don't want to try moving `./externalstmp` unless it actually
exists, otherwise we get spurious error messages on the `mv` call.

This just adds that check.


mumoshu (approved):
LGTM. Thanks for fixing it 🙏

PR #1334, opened 2022-04-12
chore: migrate to actions stale bot

toast-gear:
Move to using the actions stale bot instead of probot's stale bot. Might be wrong but from reading the probot docs it isn't possible to only stale only issues and ignore PRs. The Probot stale project looks fairly dead too based on thea activity in the codebase https://github.com/probot/stale.


mumoshu (approved):
LGTM. Thanks!

PR #1340, opened 2022-04-13
Enhance RunnerSet to optionally retain PVs accross restarts

mumoshu:
This is my initial attempt to bring back the ability to retain PVs across runner pod restarts when using RunnerSet.
The implementation is composed of two new controllers, `runnerpersistentvolumeclaim-controller` and `runnerpersistentvolume-controller`.

It all starts from our existing `runnerset-controller`. The controller now tries to mark any PVCs created by StatefulSets created for the RunnerSet.

Once the controller terminated statefulsets, their corresponding PVCs are clean up by `runnerpersistentvolumeclaim-controller`. Shortly after that, PVs are unbound from their corresponding PVCs by `runnerpersistentvolume-controller` so that they can be reused by future PVCs created for future StatefulSets that share the same StorageClass.

Ref #1286

### DIFF
--- a/controllers/persistent_volume_claim_controller.go
+++ b/controllers/persistent_volume_claim_controller.go
@@ -1,0 +1,76 @@
+/*
+Copyright 2022 The actions-runner-controller authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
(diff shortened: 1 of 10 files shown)


erplsf:
looks very promising indeed, although I don't see how acceptance/testdata converts into an actual test of this implementation.

mumoshu:
@erplsf What do you mean by "actual test of this implementation"? It's used as a part of `e2e_test.go` which I use to test ARC.

mumoshu:
Merging this anyway so that you can test the `canary` version of ARC released for each commit to the master branch. I'd appreciate it if anyone could help testing it! Thanks.

controllers/persistent_volume_controller.go
```diff
@@ -0,0 +1,72 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// RunnerPersistentVolumeReconciler reconciles a PersistentVolume object
+type RunnerPersistentVolumeReconciler struct {
```
Gentoli, line 33: Is watching `PersistentVolume`s necessary? Should it be up to the PV controller to decide how to reclaim/recreate when PVs the PVC is deleted.

controllers/persistent_volume_claim_controller.go
```diff
@@ -0,0 +1,76 @@
...
+
+func (r *RunnerPersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := r.Log.WithValues("pvc", req.NamespacedName)
+
+	var pvc corev1.PersistentVolumeClaim
+	if err := r.Get(ctx, req.NamespacedName, &pvc); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	log.Info("Reconciling runner pvc")
```
Gentoli, line 53: This confusingly logs PVC skipped in `syncPVC`. 

image

Issue #1357, opened 2022-04-19
Fix make `docker-buildx` target on a Mac M1


toast-gear:
The controller no longer builds when using the make target docker-build on a Macbook (M1 at least, maybe Intel too)

mumoshu:
I believe you'd need `buildx` and use `make docker-buildx` instead, because we now rely on the cache mount https://github.com/actions-runner-controller/actions-runner-controller/blob/81951780b10577d645b3ab9f5cb27c6f80134676/Dockerfile#L37-L38 (and probably that's not Mac-specific

toast-gear:
We probably need to just remove the non-buildx target entirely from the Makefile

PR #1360, opened 2022-04-20
refactor: remove legacy controller Docker build

toast-gear:
Ref https://github.com/actions-runner-controller/actions-runner-controller/issues/1357

Changes: 
1. Removed the non-buildx target
2. I also enabled buildkit in the Makefile as I don't see a reason not to use it? It's off by default https://github.com/moby/moby/issues/40379
3. Added the RUNNER_VERSION the same as the runner Makefile for consistency and added it to Renovate

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -21,7 +21,8 @@
     },
     {
       "fileMatch": [
-        "runner/Makefile"
+        "runner/Makefile",
+        "Makefile"
       ],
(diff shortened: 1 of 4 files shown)


toast-gear:
When this is merged the commit message should be edited to include `[skip ci]` as we are building from the `runner/Makefile` having a change which we don't use in the CI build so there's no need to release a new runner version

PR #1361, opened 2022-04-20
doc: Removed the default `githubEvent: {}` requiring a event to be defined

### LINKED ISSUE #1358: Remove the default `githubEvent: {}` requiring a event to be defined
Ref https://github.com/actions-runner-controller/actions-runner-controller/pull/922

With the introduction of a explicit attribute for the workflow_job event (to be released in [v0.23.0](https://github.com/actions-runner-controller/actions-runner-controller/milestone/3)) we are removing the implicit default of workflow_job requiring people to explicitly define all webhook events they want to listen for, reasons:

1. Explicit > implicit
2. GitHub may introduce some other super awesome webhook event that is better than the workflow_job event meaning it becomes the new event would become the preferred default event. By avoiding having a default at all we avoid possible complications with implementing the newer event and cut down on end user configuration issues being raised.
3. The current syntax could be easily confused as meaning all events if someone didn't read the documentation correctly.

This
```yaml
  scaleUpTriggers:
  - githubEvent: {}
    duration: "30m"
````
Becomes:
```yaml
  scaleUpTriggers:
  - githubEvent:
      workflowJob: {}
    duration: "30m"
````

soham4abc:
Fixes #1358

### DIFF
--- a/README.md
+++ b/README.md
@@ -735,7 +735,8 @@ spec:
     # Uncomment the below in case the target is not RunnerDeployment but RunnerSet
     #kind: RunnerSet
   scaleUpTriggers:
-  - githubEvent: {}
+  - githubEvent:
+      workflowJob: {}
     duration: "30m"
 ```
 


mumoshu:
@soham4abc Thanks for the PR! Please also see #1379 for the code change required to make this documentation enhancement true.

mumoshu (approved):
LGTM. Thanks for your contribution!

Issue #1067, opened 2022-01-21
Installing the ARC on multiple orgs using GH App authentication


casanova-21:
Hi, team. 👋  I'm working on setting the ARC up for a high profile customer, and we're running into an issue that I'm not sure how to get past.

We have successfully configured the ARC for a single org using the GH App authentication method. We're using the webhook on the GH App to trigger the autoscaling of the runner pods. We're now trying to enable this for their other orgs, so we made the app public and the installed it on a second org. We then deployed a second runner-deployment.yaml that contains the second org, and I can see in the "describe runner" output that it is unable to update the registration token for this other org. I can also see the "403 Resource not accessible by integration" error in the controller pod logs. I believe this is due to the fact that GH App installation ID for the second org is unique and doesn't match the installation ID that is stored in the controller-manager secret in the actions-runner-system namespace.

What is the proper method of getting this installed across multiple orgs? My customer does not want to use PAT authentication to avoid having to rotate the service account every six months that would be associated with the PAT. And as far as I'm aware, you can't install the GH App at the enterprise level.

toast-gear:
Deploy a controller per organisation https://github.com/actions-runner-controller/actions-runner-controller#deploying-multiple-controllers, ARC only takes a single GitHub App configuration and as you've seen, the installtion ID is unique per install.

casanova-21:
@toast-gear Thanks for the response! I did read through that section on deploying multiple controllers. Our customers that want to adopt the ARC at scale are likely going to be unwilling to run a separate controller for each of their orgs. We've got customers in some cases with hundreds of orgs, and a controller per org will be too much overhead for them. Is the PAT authentication going to be their only option then? If that is the case, we'll need some kind of feature that allows the GH App method to scale better. My customer doesn't want to use PAT authentication as the PAT will be tied to a service account which will need to be rotated every six months. Their concerned about incurring downtime when that rotation occurs.

I looked at the Runner Deployment deployment file and it appears that "organization" is a string as opposed to an array which implies to me that you can only ever specify a single org per runner deployment. Is the best option here going to be a Runner Group?

cbui:
> Is the best option here going to be a Runner Group?

Sorry, could you clarify your question regarding Runner Groups? Are you asking if you could use Runner Groups to share runners across orgs?

casanova-21:
[...]

PR #1371, opened 2022-04-22
Enhance support for multi-tenancy

mumoshu:
This enhances every ARC controller and the various K8s custom resources so that the user can now configure custom GitHub API credentials, that are different from the default ones configured per the ARC instance.

Beware that the runner groups visibility support for the webhook-based autoscaler does not cover this feature, as there seemed like no clean way to implement it. We need to determine the scale-target to obtain the custom GitHub credentials per scale-target, but to determine the scale-target, we need to consult visible runner groups, which needs GitHub credentials!

Does this worth the effort? If so, how do you deal with the problem of the runner-group-visibility support?

Ref https://github.com/actions-runner-controller/actions-runner-controller/issues/1067#issuecomment-1043716646


mumoshu:
> Beware that the runner groups visibility support for the webhook-based autoscaler does not cover this feature, as there seemed like no clean way to implement it. We need to determine the scale-target to obtain the custom GitHub credentials per scale-target, but to determine the scale-target, we need to consult visible runner groups, which needs GitHub credentials!

Probably we'd better just enhance the HRA spec as well to support the new githubAPICredentialsFrom field.

README.md
```diff
@@ -1434,6 +1435,56 @@ $ helm --upgrade install actions-runner-controller/actions-runner-controller \
   admissionWebHooks.caBundle=${CA_BUNDLE}
 ```
 
+### Multitenancy
+
+Since [#1371](https://github.com/actions-runner-controller/actions-runner-controller/pull/1371) and the version 0.25.0, ARC supports multi-tenancy.
```
toast-gear, line 1440: can we make this in the style of all the other minimum controller version requirements

mumoshu: @toast-gear Sounds good! Fixed in [5ce2fab](https://github.com/actions-runner-controller/actions-runner-controller/pull/1371/commits/5ce2fabea21d2b388215ed94c9b6c7b154500a75)

mumoshu:
Let me merge this so that the `canary` image gets updated and everyone interested in this feature can give it a shot

nascit:
Hello, is this going to be available on next release?

toast-gear:
yes, it'll be in the next non patch level release which is the next planned release unless some critical bug is discovered that needs fixing and releasing ASAP

mumoshu:
@nascit Hey! Let me add that it's already available via ARC's `canary` image tag and the helm chart available in our main branch. It would be great if you could give it a shot and report back anything you found on the way- that would help us decide it's ready for release!

Issue #942, opened 2021-11-12
Helm chart for managing runners (but controller)


mumoshu:
Extracted from https://github.com/actions-runner-controller/actions-runner-controller/issues/368#issuecomment-966707909 via @AndrewFarley

Is anyone interested in contributing this, if we got to think this is in the scope of this project?

saravanan30erd:
@mumoshu  Created a PR for this one.

PR #1375, opened 2022-04-22
helm charts for actions runner

saravanan30erd:
Fixes: #942 

helm charts for actions runner, currently its having only RunnerDeployment and Autoscaler resources.

Looks like deployment order is important here, facing the below issue if Autoscaler deployed first and then autoscaling not working as expected.
```
2022-04-21T12:13:08Z    DEBUG    controllers.webhookbasedautoscaler    RunnerDeployment not found with scale target ref name test-actions-runner for hra test-actions-runner-autoscaler
```
Helm doesn't support [ordering](https://github.com/helm/helm/issues/8439) for custom resources. So using List to overcome this issue, didn't use helm chart hooks for ordering since its not [tracked](https://helm.sh/docs/topics/charts_hooks/#hook-resources-are-not-managed-with-corresponding-releases) after creation.


charts/actions-runner/values.yaml
```diff
@@ -0,0 +1,59 @@
+image:
+  repository: summerwind/actions-runner
+  tag: 2.290.1
```
yardbirdsax, line 3: Just a quick FYI that this doesn't appear to be a valid tag based on [the list of available ones](https://hub.docker.com/r/summerwind/actions-runner/tags?page=1&name=2.290.1).

yardbirdsax:
I was about to start working on this and then saw that you helpfully already have! 😄 One suggestion would be to add the ability to specify the Runner Group. I added this and created saravanan30erd/actions-runner-controller#1 to get that into your branch.

charts/actions-runner/values.yaml
saravanan30erd: yeah, thanks. It was a mistake initially I put it as placeholder then didn't change.

mumoshu:
@saravanan30erd @yardbirdsax Hey! First of all, thank you so much for your contribution and patience.

Sorry for the radio silence here. We've been discussing how to maintain this. The main problem is that maintaining the manifests, the kustomize config, and the chart for ARC itself is already too much to maintain for a team of only two, and we weren't sure if we could commit to maintain another right now.
I do believe this chart is super valuable to the community if maintained well, though.
That said, could you let me merge this under a directory like `contrib/examples` first, and start collecting feedbacks from early adopters/advanced users, and then graduate to the official chart under `charts` once we gain more confidence?

saravanan30erd:
>  That said, could you let me merge this under a directory like `contrib/examples` first, and start collecting feedbacks from early adopters/advanced users, and then graduate to the official chart under `charts` once we gain more confidence?

@mumoshu  Sure.

mumoshu (approved):
LGTM. Thanks for your contribution and understanding!

Issue #1196, opened 2022-03-09
Remove the ability to use the `--once` flag entirely only allowing `--ephemeral` to be set


toast-gear:
At the moment you can pick between where your ephemeral runners will be configured witwh the legacy `--once` flag or the supported new `--ephemeral` flag. We are not aware of any reason to use the `--once` flag, the only discussion we've seen on it is here https://github.com/actions/runner/issues/1339 and it is centred around rate limiting which we don't think will be an issue. 

We also state in the docs that ARC requires => 3.3 which supports the `--ephemeral` flag.

If you are running a GHES environment you **need** to upgrade your GHES instance to => version 3.3 if you wish to maintain an ARC upgrade path and not have down time. In order to remove the --once flag we will need to remove the logic from our runner container [[1]](https://github.com/actions-runner-controller/actions-runner-controller/blob/master/runner/entrypoint.sh#L166-L170) [[2]](https://github.com/actions-runner-controller/actions-runner-controller/blob/master/runner/entrypoint.sh#L90-L93), this will be done at the point of the linked milestone being released and so if you are using the latest tag or build a custom container from ours you will be impacted effectively at a random time. Migration notes:

1. Just upgrade to GHES => 3.3 and everything will continue to work (it was released on 07-12-2021)
2. If you won't / can't you will need to move to persistent runners via setting the attribute `ephemeral: false`, this will create a significant change in behaviour and so has impacts you may not desire.
3. Provde your own image from scratch not built from ARC's allowing you to set the `--once` flag as part of your images entrypoint (don't expect tonnes of help from ARC if you go this direction)

This issue is to be done in the subsequent release to https://github.com/actions-runner-controller/actions-runner-controller/issues/1189

mumoshu:
We ended by making `--ephemeral` the default and adding the new option to stick with `--once` instead.
We'll probably leave the new option for a while, but don't expect it to work forever. GitHub Actions and actions/runner may eventually remove the support for `--once`.
Before they remove it, upgrade to a newer version of GHES, ASAP!

mumoshu:
Let's keep this open until we completely remove the support for `--once`.

PR #1384, opened 2022-04-25
runner: Remove the ability to use the deprecated `--once` flag

mumoshu:
ARC would run with or without this patch. It's just that once this change has been merged and hence released as new runner images, `RUNNER_FEATURE_FLAG_EPHEMERAL` becomes noop and whether to pass `--ephemeral` to the config.sh is determined solely by the presence of `RUNNER_FEATURE_FLAG_EPHEMERAL` envvar.

Ref #1196


gjkim42:
This is a breaking change for the users who have used the RUNNER_FEATURE_FLAG_EPHEMERAL environment variable.
It might be better to notify there are a new env and the removed env to release notes.

mumoshu:
@gjkim42 Hey. Since this change the --ephemeral flag is enabled by default. How can it be a breaking change? 🤔

gjkim42:
Ah, I just found that the current actions-runner-controller does only support github enterprise >= 3.3.

I think our environment is broken because we have been using github enterprise 3.2 which does not support --ephemeral flag and we have been using summerwind/actions-runner:latest image.

gjkim42:
> Hey. Since this change the --ephemeral flag is enabled by default.

We've set RUNNER_FEATURE_FLAG_EPHEMERAL=false to disable --ephemeral flag but this PR makes the old environment variable useless.

mumoshu:
@gjkim42 Thanks for clarifying. That makes sense. It's unfortunate that you temporarily broke your environment(it should be "temporarily" because you can still opt-in to use --once via RUNNER_FEATURE_FLAG_ONCE since this change, right?)

We've been advocating folks to upgrade GHES to 3.3+ since probably almost a year ago but apparently, it wasn't enough to make it 100% alright.

gjkim42:
> it should be "temporarily" because you can still opt-in to use --once via RUNNER_FEATURE_FLAG_ONCE since this change, right?

Yes, we can.

---

One casual suggestion:

Is there any plan to split summerwind/actions-runner from this repository? or to make official docker image in [actions/runner](https://github.com/actions/runner)?

Maybe we need to split summerwind/actions-runner part(which builds the container image of the actions-runner) from this repository so that we can manage the release cycle of the actions-runner.

mumoshu:
> Is there any plan to split summerwind/actions-runner from this repository? or to make official docker image in [actions/runner](https://github.com/actions/runner)?

Unfortunately no, almost entirely because of the lack of our resources.

We're a team of only two maintainers and I even wondered if we waited too long (almost a year) before making this change, being afraid if we wasted our human resources to maintain support for those legacy GHES versions no one is using anymore(which turned wrong just now).

> Maybe we need to split summerwind/actions-runner part

Ideally yes, that would be a great idea!

gjkim42:
I understood the points.
Anyway, thank you for all your great work. This project is awesome.

toast-gear:
@gjkim42 to be clear this change was breaking on purpose (with the ability for end users to fix the problem using the new env var). We've gone to great efforts to highlight `--once` is being removed but assumed there are probably people out there using it still. We decided to cause a sort of "brown out" as the last step in making people aware `--once` support is being removed.

See https://github.com/actions-runner-controller/actions-runner-controller/issues/1196 for ways of working around its removal if you are unable to get onto => GHES 3.3. If you do have a reason why keeping `--once` would be great please do comment on https://github.com/actions-runner-controller/actions-runner-controller/issues/1196.

PR #1308, opened 2022-04-06
refactor: set sync period default to 1m

toast-gear:
Fixes: https://github.com/actions-runner-controller/actions-runner-controller/issues/1294

Not to be merged until linked milestone in linked issue is valid

Issue #1294, opened 2022-03-31
Change the --sync-period default from 10m to 1m


toast-gear:
Now that we are caching API calls https://github.com/actions-runner-controller/actions-runner-controller/pull/1210 we should change the sync period default from 10m to 1m as this is a better balance between performance and the "out of the box" risk of unintentional rate limited

PR #1388, opened 2022-04-25
fix: forgot to do the chart

toast-gear:


mumoshu:
Linking #1294 and #1308

mumoshu:
chart test is failing due to `flag provided but not defined: -default-scale-down-delay` which seems to come from the fact that we still use ARC 0.22.3 for chart testing. Probably we'd better figure out how to test it against both the latest release version of ARC and the canary version of ARC, or just test it against the canary version so that it won't fail when the chart depends on features that are available only in the canary version of ARC? 🤔

01SMD (commented):
👍

Issue #1056, opened 2022-01-17
Add label support to `TotalNumberOfQueuedAndInProgressWorkflowRuns` metric


toast-gear:
The `TotalNumberOfQueuedAndInProgressWorkflowRuns` metric relies on `ListWorkflowJobs` method https://github.com/google/go-github/blob/fa34afad23649db61eff69fecba7c48a3839e891/github/actions_workflow_jobs.go#L70 which is a wrapper to https://docs.github.com/en/free-pro-team@latest/rest/reference/actions/#list-jobs-for-a-workflow-run. You can see in the response that a list of all labels in the workflow is included. 

Labels in the response are just grouped together rather than mapped to their corresponding job, as a result there will be some significant limitations to labels with this metric.

We can't know how many self-hosted runners we need to scale exactly. We would have to assume that if there is a label match then we need a runner from that match for every job in the workflow.

There are 2 implementations of this that could be done (bronze and gold):

- Bronze - Check that the RunnerDeployment / RunnerSet has the matching labels before deciding if it needs to scale up. At the moment we scale all RunnerDeployments / RunnerSets to the depth of the queue (within the replica constraints). This would, in addition, to scaling only on workflows that actually have some jobs that target self-hosted runners, prevent overscaling when running multiple controllers as at the moment if both controllers have runners against the repository they will both scale up even if the second controllers runners don't have a relevant label.
- Gold - Between all of the RunnerDeployments / RunnerSets that match only scale enough to cover the depth of the queue. In the case of multiple RunnerDeployments / RunnerSets matching and the queue depth being less than the maxReplica count between all the matching sets, pick a runner group randomly (with `Default` being preferred if that can fulfil the label requirements) and opting to scale that one up over the other so we don't overscale.

mumoshu:
The Bronze variant of this has been implemented in #1373

PR #1390, opened 2022-04-26
Fix label support to TotalNumberOfQueuedAndInProgressWorkflowRuns metric

mumoshu:
In #1373 we checked if all the runner labels are included in the job labels and only after that it marked the target as eligible for scale.
It should definitely be the opposite!
Otherwise, for example, jobs with `self-hosted` + `custom` labels wont result in scaling runner with `self-hosted` + `custom` + `custom2`.

Ref #1056
Ref #1373


toast-gear (approved):
LGTM

PR #1392, opened 2022-04-26
Force ArgoCD to use replace on CRDs

### LINKED ISSUE #1391: Helm/CRDs cannot be applied during update process
Using ArgoCD one cannot easily update CRDs. Just because of: https://github.com/prometheus-community/helm-charts/issues/1500
Potential Fix:
* https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#replace-resource-instead-of-applying-changes

Vladyslav-Miletskyi:
Resolves #1391


toast-gear:
We don't want to bake in annotations that are tool specific into our CRDs, is it possible to set annotations in the CRDs via the values?

Vladyslav-Miletskyi:
Hey @toast-gear, it is not possible to add helm values to the CRDs :/ so, I will go with multi app approach - controller installation without CRDs and CRDs install with forced re-create.

Closing PR. Will add comment to the issue with possible ArgoCD solution.

Issue #1369, opened 2022-04-21
Pods stuck in terminating state


larhauga:
We experience that some runner pods are stuck in terminating state.
The pod is still registered as a runner in github. No change if the runner is deleted forcefully in gha api.
The actions-controller is continuously logging: `actions-runner-controller.runnerpod     Runner pod is annotated to wait for completion `.

This seams to happend if a node is being deleted with a runner.

We are running spot nodes and cluster-autoscaler, which seams to be making this issue a bit more apparent.

The last log event for the runner pod is: https://github.com/actions-runner-controller/actions-runner-controller/blob/e7200f274d592729b46848218c1d9c54214065c9/controllers/runner_graceful_stop.go#L115

As the pod has an error exit code (127) we think that a forceful delete of the runner is needed. https://github.com/actions-runner-controller/actions-runner-controller/blob/af8d8f7e1da4b32d837428f013b7b68510347343/controllers/runner_graceful_stop.go#L115 will only requeue the failed reconcile, and in this case never delete the pod.

unregisterRunner should probably be run if exist != 0 exists.

```
Name:                      runner
Namespace:                 monorepo
Start Time:                Wed, 20 Apr 2022 14:17:24 +0200
Labels:                    actions-runner-controller/inject-registration-token=true
                           pod-template-hash=7cb656579c
                           runner-deployment-name=runner
                           runner-template-hash=54d74c6f69
                           runnerset-name=runner-f87wk-wr4qk
Annotations:               actions-runner-controller/token-expires-at: 2022-04-20T12:43:57Z
                           actions-runner/id: 115340
                           actions-runner/runner-completion-wait-start-timestamp: 2022-04-20T12:17:56Z
                           actions-runner/unregistration-request-timestamp: 2022-04-20T12:14:13Z
                           actions-runner/unregistration-start-timestamp: 2022-04-20T12:17:55Z
                           sync-time: 2022-04-20T11:57:34Z
Status:                    Terminating (lasts 138m)
Termination Grace Period:  0s
Controlled By:  Runner/runner-f87wk-wr4qk
Containers:
  runner:
    State:          Waiting
      Reason:       ContainerCreating
    Last State:     Terminated
      Reason:       ContainerStatusUnknown
      Message:      The container could not be located when the pod was deleted.  The container used to be Running
      Exit Code:    137
      Started:      Mon, 01 Jan 0001 00:00:00 +0000
[...]

PR #1395, opened 2022-04-28
Force restartPolicy "Never" to prevent runner pods from stucking in Terminating when the container disappeared

mumoshu:
~~Note that there might be the same issue in runner pods managed by RunnerSet, but the fix for that isn't trivial.
That's why this PR contains the fix for RunnerDeployment only.~~ It turned out we are able to mutate pod specs in the mutating webhook so it's alright. This should work for both RunnerDeployment and RunnerSet.

Ref #1369

### DIFF
--- a/controllers/runner_controller.go
+++ b/controllers/runner_controller.go
@@ -629,9 +629,7 @@ func newRunnerPod(runnerName string, template corev1.Pod, runnerSpec v1alpha1.Ru
 
 	pod := template.DeepCopy()
 
-	if pod.Spec.RestartPolicy == "" {
-		pod.Spec.RestartPolicy = "OnFailure"
-	}
+	forceRunnerPodRestartPolicyNever(pod)
 
 	if mtu := runnerSpec.DockerMTU; mtu != nil && dockerdInRunner {
 		runnerContainer.Env = append(runnerContainer.Env, []corev1.EnvVar{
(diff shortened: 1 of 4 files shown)


toast-gear (approved):

controllers/runner_controller.go
```diff
@@ -630,7 +630,7 @@ func newRunnerPod(runnerName string, template corev1.Pod, runnerSpec v1alpha1.Ru
 	pod := template.DeepCopy()
 
 	if pod.Spec.RestartPolicy == "" {
-		pod.Spec.RestartPolicy = "OnFailure"
+		pod.Spec.RestartPolicy = corev1.RestartPolicyNever
 	}
```
toast-gear, line 634: add a comment for why we need to use `RestartPolicyNever` instead of another restart policy

mumoshu: Thanks for your suggestion! Addressed in c36e99a.

Also, I realized that we can make it work for RunnerSet too by forcing restartPolicy=Never in the mutating webhook. Updated code accordingly, so that the initial description of this pull request that stated this change works only for RunnerDeployment is no longer true.

Note that, as we can now force restartPolicy=Never on every runner pod, we may never see the code path added in https://github.com/actions-runner-controller/actions-runner-controller/pull/1420 is executed in real-world scenarios.

PR #1409, opened 2022-05-02
Runner Entrypoint: fix daemon.json

Vladyslav-Miletskyi:
Do not overwrite daemon.json if it already exists.
Usage: custom images, which are using a public image as source.

### DIFF
--- a/runner/startup.sh
+++ b/runner/startup.sh
@@ -20,7 +20,7 @@ function wait_for_process () {
 sudo /bin/bash <<SCRIPT
 mkdir -p /etc/docker
 
-echo "{}" > /etc/docker/daemon.json
+[ ! -f /etc/docker/daemon.json ] && echo "{}" > /etc/docker/daemon.json
 
 if [ -n "${MTU}" ]; then
 jq ".\"mtu\" = ${MTU}" /etc/docker/daemon.json > /tmp/.daemon.json && mv /tmp/.daemon.json /etc/docker/daemon.json
--- a/runner/startup.sh
+++ b/runner/startup.sh
@@ -20,7 +20,9 @@ function wait_for_process () {
 sudo /bin/bash <<SCRIPT
 mkdir -p /etc/docker
 
-[ ! -f /etc/docker/daemon.json ] && echo "{}" > /etc/docker/daemon.json
+if [ ! -f /etc/docker/daemon.json ]; then
+  echo "{}" > /etc/docker/daemon.json
+fi
 
 if [ -n "${MTU}" ]; then
 jq ".\"mtu\" = ${MTU}" /etc/docker/daemon.json > /tmp/.daemon.json && mv /tmp/.daemon.json /etc/docker/daemon.json


runner/startup.sh
```diff
@@ -20,7 +20,7 @@ function wait_for_process () {
 sudo /bin/bash <<SCRIPT
 mkdir -p /etc/docker
 
-echo "{}" > /etc/docker/daemon.json
+[ ! -f /etc/docker/daemon.json ] && echo "{}" > /etc/docker/daemon.json
```
toast-gear, line 23:
```suggestion
if [ ! -f /etc/docker/daemon.json ]; then
  echo "{}" > /etc/docker/daemon.json
fi
```

We've used the long form in the rest of the file so we should be consistent

UrosCvijan:
Having the same issue... Is there a way to provide daemon.json that is custom to regular summerwind/actions-runner-dind:latest
As our datacenter has the same network as docker default network, so we want to change the docker network using daemon.json
I wanted to create my own image using FROM the the above and to add my daemon.json but it gets overwritten. 
Is there some other way of doing it?

mumoshu:
@UrosCvijan How about modifying `startup.sh`?

UrosCvijan:
@mumoshu Aha, sorry for late reply. My idea was not to touch the "core" of everything that was used to build the docker image. As later something might be added to the startup.sh so I would not touch any of those but just add my daemon, basically do the seamless upgrades. But if there wont be a change like this, I guess this will have to be the way.

mumoshu:
@UrosCvijan Thanks for clarifying! I hear you- probably we should merge this and after that you'd be able to provide your own daemon.json in your custom runner image's Dockerfile. Does that sound good to you?

@UrosCvijan @Vladyslav-Miletskyi

The only reason I was a bit hesitant to improve startup.sh like this is that now you have another vector to break ARC so we'd need to improve our [already very lengthy bug report form](https://github.com/actions-runner-controller/actions-runner-controller/issues/new?assignees=&labels=bug&template=bug_report.yml&title=Bug) to also include a question like "Paste your custom docker daemon.json if any". If the only way you can customize it was by modifying start.sh, almost certainly the user is advanced enough so that they don't need basic user support so I won't include such question in the bug report form.

Perhaps you'd think it's better to merge this and enhance the bug report form, rather than not merging this at all? If so, I'd go that way anyway!

UrosCvijan:
@mumoshu Yeah, that is what i currently did with the base docker:dind image. As I don't want to "break" the future upgrades and flow, this way, I don't touch any of the mechanics of the startup.sh or entrypoint.sh, I can just use your image and add stuff I need, like packages new deamon.json etc.. 
So from all of this, as you said, i could change startup.sh but then i "interfere" with the things you are doing in the setup of the main docker image. 
The thing is that it worked for us when we just added it in original dind image, but as we started burning up a lot of runners, i was asked if i can use only one container, did the same steps but deamon.json was {}, didn't even think that it could be in startup.sh script :) 
I would not touch the Bug report form, as people that make changes and are adding their own deamon.json, know the reason behind it.

mumoshu:
@UrosCvijan Thanks for confirming! Got it. We'll do our best to maintain it :) But beware that we may or may not unexpectedly break this as it currently lacks a unit test. It would be great if you could submit a PR to add that. I remember we had a few tests for entrypoint.sh in https://github.com/actions-runner-controller/actions-runner-controller/tree/master/test/entrypoint. It would be great if you could write something similar for startup.sh.

mumoshu:
At least this seems to not break existing behavior. I've verified that by running a few tests with the updated image locally.

mumoshu (approved):
LGTM. Thanks for your contribution @Vladyslav-Miletskyi!

PR #1410, opened 2022-05-02
Allow customizing the controller webhook port

### LINKED ISSUE #1314: Allow customizing the controller webhook port
**Is your feature request related to a problem? Please describe.**
As noted in the [troubleshooting guide](https://github.com/actions-runner-controller/actions-runner-controller/blob/master/TROUBLESHOOTING.md#deployment-fails-on-gke-due-to-webhooks), deploying this to a private GKE cluster generally requires creating a firewall rule to allow traffic on port 9443. However, this recommendation is easier said then done in some complex or security-conscious environments.

**Describe the solution you'd like**
Ideally, the controller (and all associated deployment methods) would allow specifying the port on which it listens. This is the approach taken by a few other projects, [cert-manager being a notable example](https://github.com/cert-manager/cert-manager/blob/d212165c8da228437e71a74c2e117ae9d62c7f24/deploy/charts/cert-manager/values.yaml#L225-L232).

**Describe alternatives you've considered**
Other than creating a firewall rule (and my proposal above), my use case could be satisfied by creating a custom image that listens on the desired port, but that seems excessive considering that is the only change I would make.

**Additional context**
- While I am interested in this use case because of GKE private clusters, I would imagine this kind of port flexibility may be useful for others for security/consolidation purposes.
- This port flexibility is probably unnecessary for other components (such as the GitHub webhook service) as they will likely be going through some load balancer.
- Further, I recognize that this port is assumed at a few layers, including the Go application and the Helm chart. These would need to be parameterized, likely through command-line flags and `values.yaml` entries respectively.

Finally, I would be more than happy to submit a pull request for this functionality if interest exists.

nafarlee:
Closes #1314

I was initially having some issues performing local testing, as the [contribution guide](https://github.com/actions-runner-controller/actions-runner-controller/blob/master/CONTRIBUTING.md#developing-the-controller) and [`Makefile`](https://github.com/actions-runner-controller/actions-runner-controller/blob/800d6bd58667f384fe0b3fdd26225d0fe6830035/Makefile#L142) are (at least in-part) still structured around a `make` rule (`docker-build`) that was [removed recently](https://github.com/actions-runner-controller/actions-runner-controller/commit/059481b6103fe4ccefd8fce638a7565f9feaa12d). Considering the relatively small scope of this PR, I thought it would be easier to submit it, then go back and fix anything I had missed.


main.go
```diff
@@ -113,6 +114,7 @@ func main() {
 	flag.StringVar(&c.RunnerGitHubURL, "runner-github-url", c.RunnerGitHubURL, "GitHub URL to be used by runners during registration")
 	flag.DurationVar(&gitHubAPICacheDuration, "github-api-cache-duration", 0, "DEPRECATED: The duration until the GitHub API cache expires. Setting this to e.g. 10m results in the controller tries its best not to make the same API call within 10m to reduce the chance of being rate-limited. Defaults to mostly the same value as sync-period. If you're tweaking this in order to make autoscaling more responsive, you'll probably want to tweak sync-period, too")
 	flag.DurationVar(&defaultScaleDownDelay, "default-scale-down-delay", controllers.DefaultScaleDownDelay, "The approximate delay for a scale down followed by a scale up, used to prevent flapping (down->up->down->... loop)")
+	flag.IntVar(&port, "port", 9443, "The port to which the endpoint should bind")
```
mumoshu, line 117:
```suggestion
	flag.IntVar(&port, "port", 9443, "The port to which the admission webhook endpoint should bind")
```

We might better clarify it this way so that you can differentiate it from e.g. metrics port!

mumoshu (approved):
LGTM. Thanks for your contribution ☺️

PR #1417, opened 2022-05-06
fix(chart): add missing namespace to webhook.ingress

maigl:
**I think I found a small bug:**
When deploying the helm chart with:
```
helm template --namespace 'my-runner-namespace' ... | kubectl apply -f -
```

the ingress of the webhook server is deployed into namespace `default` instead
of `my-runner-namespace`. All other resources seem to be deployed into the correct namespace.  

**proposed fix:**
The ingress needs to be deployed in the very same namespace as the service it is forwarding to.

<sub>Michael Kuhnt <michael.kuhnt@mercedes-benz.com> Mercedes-Benz Tech Innovation GmbH ([ProviderInformation](https://github.com/mercedes-benz/daimler-foss/blob/master/PROVIDER_INFORMATION.md))
</sub>

### DIFF
--- a/charts/actions-runner-controller/templates/githubwebhook.ingress.yaml
+++ b/charts/actions-runner-controller/templates/githubwebhook.ingress.yaml
@@ -11,6 +11,7 @@ apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
   name: {{ $fullName }}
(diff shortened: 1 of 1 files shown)


mumoshu (approved):

Issue #1298, opened 2022-04-01
Feature: add scorecard to your project


laurentsimon:
I ran the [OpenSSF's scorecard](https://github.com/ossf/scorecard) on this project and found that branch protection and code reviews are not enforced.

I chatted with @josepalafox and he suggested I create this issue to start a discussion.

Given the popularity of this project and its sensitivity (access to build pipeline, etc), it would be beneficial to install scorecard as suggested in the [workflow hardening guideline](https://docs.github.com/en/enterprise-server@3.4/actions/security-guides/security-hardening-for-github-actions#using-openssf-scorecards-to-secure-workflows).

/cc @nebuk89

mumoshu:
@laurentsimon Thanks for bringing this up. I just ran `docker run -e GITHUB_AUTH_TOKEN=$OPENSSF_SCORECARD_GITHUB_TOKEN gcr.io/openssf/scorecard:stable --show-details --repo=https://github.com/actions-runner-controller/actions-runner-controller` to see what it finds. You can expect me to submit a few PRs to address each finding!

mumoshu:
We've managed to score 8.2 out of 10 now.

<details>

```
RESULTS
-------
Aggregate score: 8.2 / 10

Check scores:
|---------|------------------------|--------------------------------|-----------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------|
|  SCORE  |          NAME          |             REASON             |                                                  DETAILS                                                  |                                               DOCUMENTATION/REMEDIATION                                               |
|---------|------------------------|--------------------------------|-----------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------|
| 10 / 10 | Binary-Artifacts       | no binaries found in the repo  |                                                                                                           | https://github.com/ossf/scorecard/blob/5758364c82f7fc72b256f9a8cfc89dc550d7dd66/docs/checks.md#binary-artifacts       |
|---------|------------------------|--------------------------------|-----------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------|
| 8 / 10  | Branch-Protection      | branch protection is not       | Info: 'force pushes' disabled                                                                             | https://github.com/ossf/scorecard/blob/5758364c82f7fc72b256f9a8cfc89dc550d7dd66/docs/checks.md#branch-protection      |
[...]

PR #1421, opened 2022-05-09
Enable CodeQL Alerts following the OpenSSF Security Scorecards recommendation

mumoshu:
as documented in https://github.com/ossf/scorecard/blob/5758364c82f7fc72b256f9a8cfc89dc550d7dd66/docs/checks.md#sast

This isn't tested but should just work. If it doesn't, I'll shortly follow-up with additional PR(s).

Ref #1298

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,26 @@
+name: "Code Scanning"
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+  schedule:
+    - cron: '30 1 * * 0'
+
+jobs:
+  CodeQL-Build:
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,26 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v3.0.2
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v2.1.9
+        with:
+          languages: go
+      - name: Autobuild
+        uses: github/codeql-action/autobuild@v2.1.9
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@v2.1.9
```
toast-gear, line 26: These will need to reference the commit hash, ref https://github.com/actions-runner-controller/actions-runner-controller/pull/1422

mumoshu: Thanks! Fixed in 854fbeaf9d50b99cbfcb0b3d75e0e84581c06327

toast-gear (approved):

PR #1460, opened 2022-05-18
chore: move HOME to more logical place

toast-gear:


runner/actions-runner-dind.dockerfile
```diff
@@ -122,6 +120,7 @@ RUN echo "PATH=${PATH}" > /etc/environment \
 
 # No group definition, as that makes it harder to run docker.
 USER runner
+ENV HOME=/home/runner
```
mumoshu, line 123: Probably this breaks `ENV PATH="${PATH}:${HOME}/.local/bin"` in line 115?

mumoshu (approved):
Tested locally and verified it works without any problems! Thanks for your effort 😄

Issue #1362, opened 2022-04-20
Configuring A Docker Image Cache (And Possibly Others)


TheDarkula:
Coming from [this discussion](https://github.com/actions-runner-controller/actions-runner-controller/issues/1355#issuecomment-1103390564), I currently have this:
```
apiVersion: actions.summerwind.dev/v1alpha1
kind: RunnerSet
metadata:
  name: themelio
spec:
  replicas: 8
  image: summerwind/actions-runner-dind
  organization: themeliolabs
  selector:
    matchLabels:
      app: themelio
  serviceName: themelio
  template:
    metadata:
      labels:
        app: themelio
    spec:
      containers:
        - name: runner
          env: []
          resources:
            limits:
              cpu: "24000m"
              memory: "24Gi"
            requests:
              cpu: "2850m"
              memory: "3Gi"
        - name: docker
          volumeMounts:
            - mountPath: /var/lib/docker
              name: themelio-data
  volumeClaimTemplates:
    - metadata:
        name: themelio-data
      spec:
        accessModes:
          - ReadWriteOnce
        resources:
          requests:
            storage: 100Gi
```

The first time I ran the CI, [it succeeded](https://github.com/themeliolabs/mellis/actions/runs/2197196138).

The second time, it hit [no space left on device](https://github.com/themeliolabs/mellis/actions/runs/2197223699).

Is this a misconfiguration on my part?

mumoshu:
@TheDarkula Hey. I don't know the details of how docker pull works but anyway, try giving your host's root volume more capacity(or the one that backs your host's /tmp). This doesn't look like an issue in ARC.

toast-gear:
I would avoid symlinking as your limit will be the root volume, not your PV. I'd try physically moving the Docker data directory with the `-g` flag. Something like DOCKER_OPTS="-g /mnt/docker"

TheDarkula:
@toast-gear Brilliant!
That sorted out the free space issue. The curious thing is, it does not appear to be using a cache for the docker images.
If you look [here](https://github.com/themeliolabs/mellis/runs/6116909747?check_suite_focus=true), you will see that it is still pulling the latest image:
```
Status: Downloaded newer image for ghcr.io/themeliolabs/windows-32-bit:latest
```
I ran the CI twice so the runners would already have the image downloaded, but it is not functioning properly.
The link I pasted is the second run of the `RunnerSet`.

TheDarkula:
[...]

PR #1464, opened 2022-05-20
doc: Use RunnerSet to retain various cache by leveraging PV

mumoshu:
In relation to #1286 and as a follow-up for #1340

### DIFF
--- a/README.md
+++ b/README.md
@@ -1246,6 +1246,8 @@ spec:
 
 #### Docker image layers caching
 
+> Note: Ensure that the volume mount is added to the container that is running the Docker daemon!.
+
 `docker` stores pulled and built image layers in the [daemon's (note not client)](https://docs.docker.com/get-started/overview/#docker-architecture) [local storage area](https://docs.docker.com/storage/storagedriver/#sharing-promotes-smaller-images) which is usually at `/var/lib/docker`.
 
 By leveraging RunnerSet's dynamic PV provisioning feature and your CSI driver, you can let ARC maintain a pool of PVs that are
--- a/README.md
+++ b/README.md
@@ -1246,7 +1246,7 @@ spec:
 
 #### Docker image layers caching
 
-> Note: Ensure that the volume mount is added to the container that is running the Docker daemon!.
+> **Note**: Ensure that the volume mount is added to the container that is running the Docker daemon.
 
 `docker` stores pulled and built image layers in the [daemon's (note not client)](https://docs.docker.com/get-started/overview/#docker-architecture) [local storage area](https://docs.docker.com/storage/storagedriver/#sharing-promotes-smaller-images) which is usually at `/var/lib/docker`.
 
@@ -1282,34 +1282,7 @@ spec:
       storageClassName: var-lib-docker
 ```
 
-The examples below is based on the all in one runner image (requires `dockerdWithinRunnerContainer: true`) where the Docker daemon is running in the runner container itself with the client(`docker` commands invoked by your workflow job steps or `actions/runner`) rather than in a sidecar:
-
-```yaml
-kind: RunnerSet
-metadata:
-  name: example
-spec:
-  dockerdWithinRunnerContainer: true
-  runnerImage: ghcr.io/actions-runner-controller/actions-runner-controller/actions-runner-dind:latest
-  template:
-    spec:
-      containers:
-      # Notice that we added the volume mount for the `runner` container this time!
-      - name: runner
-        volumeMounts:
-        - name: var-lib-docker
-          mountPath: /var/lib/docker
-  volumeClaimtemplates:
-  - metadata:
-      name: var-lib-docker
-    spec:
-      accessModes:
-      - ReadWriteOnce
-      resources:
-        requests:
-          storage: 10Mi
-      storageClassName: var-lib-docker
-```
+With `dockerdWithinRunnerContainer: true`, you need to add the volume mount to the `runner` container.
 
 #### Go module and build caching
 
--- a/README.md
+++ b/README.md
@@ -1246,7 +1246,7 @@ spec:
 
 #### Docker image layers caching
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1242,6 +1244,124 @@ spec:
...
 ```
 
+#### Docker image layers caching
+
+`docker` stores pulled and built image layers in the [daemon's (note not client)](https://docs.docker.com/get-started/overview/#docker-architecture) [local storage area](https://docs.docker.com/storage/storagedriver/#sharing-promotes-smaller-images) which is usually at `/var/lib/docker`.
+
+By leveraging RunnerSet's dynamic PV provisioning feature and your CSI driver, you can let ARC maintain a pool of PVs that are
+reused across runner pods to retain `/var/lib/docker`.
+
+_The examples below are based on the all in one runner image (`dockerdWithinRunnerContainer: true`) where the Docker daemon is running in the runner container itself with the client rather than in a sidecar:_
```
mumoshu: The below example was almost intended for the case you didn't set `dockerdWithinRunnerContainer: true`. But it's confusing anyway!
How about d7fe702?

README.md
```diff
@@ -1242,6 +1244,124 @@ spec:
...
+    spec:
+      containers:
+      - name: runner
+        volumeMounts:
+        # Cache docker image layers, in case dockerdWithinRunnerContainer=true
+        - name: var-lib-docker
+          mountPath: /var/lib/docker
+      - name: docker
+        volumeMounts:
+        # Cache docker image layers, in case dockerdWithinRunnerContainer=true
```
mumoshu: This had to be

```suggestion
        # Cache docker image layers, in case dockerdWithinRunnerContainer=false
```

README.md
```diff
@@ -1242,6 +1244,124 @@ spec:
     emphemeral: true # VERY important. otherwise data inside the workdir and /tmp is not cleared between builds
 ```
 
+#### Docker image layers caching
+
+`docker` stores pulled and built image layers in the [daemon's (note not client)](https://docs.docker.com/get-started/overview/#docker-architecture) [local storage area](https://docs.docker.com/storage/storagedriver/#sharing-promotes-smaller-images) which is usually at `/var/lib/docker`.
```
toast-gear, line 1249:
```suggestion
#### Docker image layers caching

> Note: Ensure that the volume mount is added to the container that is running the Docker daemon!.

`docker` stores pulled and built image layers in the [daemon's (note not client)](https://docs.docker.com/get-started/overview/#docker-architecture) [local storage area](https://docs.docker.com/storage/storagedriver/#sharing-promotes-smaller-images) which is usually at `/var/lib/docker`.
```

toast-gear: We've generally put tips in `> Note: my awesome note` syntax, we should follow this format here

README.md
```diff
@@ -1242,6 +1244,125 @@ spec:
     emphemeral: true # VERY important. otherwise data inside the workdir and /tmp is not cleared between builds
 ```
 
+#### Docker image layers caching
+
+> **Note**: Ensure that the volume mount is added to the container that is running the Docker daemon.
```
mumoshu, line 1249: GitHub MD seems to add a little icon before `**Note**` only when it's surrounded with `**`. Perhaps it's a bit more appealing with that? 😄 

https://github.com/actions-runner-controller/actions-runner-controller/blob/pv-doc/README.md#docker-image-layers-caching

README.md
```diff
@@ -1242,6 +1244,125 @@ spec:
...
+    spec:
+      accessModes:
+      - ReadWriteOnce
+      resources:
+        requests:
+          storage: 10Mi
+      storageClassName: var-lib-docker
+```
+
+With `dockerdWithinRunnerContainer: true`, you need to add the volume mount to the `runner` container.
```
mumoshu, line 1285: @toast-gear Removed the secondary example entirely and put one sentence instead for this special case. WDYT?

toast-gear (approved):

README.md
```diff
@@ -1242,6 +1244,125 @@ spec:
...
+metadata:
+  name: example
+spec:
+  template:
+    spec:
+      containers:
+      - name: docker
+        volumeMounts:
+        - name: var-lib-docker
+          mountPath: /var/lib/docker
```
Gentoli, line 1272: Would this also cache state or container volume? If this is only for layers, would `/var/lib/docker/overlay2` be enough?

mumoshu: I'm not a docker expert, you might be correct but how can you confirm that's actually correct? To me, the roles of those subdirectories seems not explained in the official docker doc.

Also, I see some overlay2 related metadata reside in `/var/lib/docker/image/overlay2`. Can we safely omit that? 🤔

Gentoli: I'm not very sure on this either, but I think we wouldn't want anything other the the layers persists through different runs. So `/var/lib/docker` seems a bit vague.

I have `canary` setup but haven't able to get PV persist over runs.

Gentoli: I'm not very sure on this either, but I think we wouldn't want anything other the the layers persists through different runs. So `/var/lib/docker` seems a bit vague.

I have `canary` setup but haven't able to get PV persist over runs.

mumoshu: @Gentoli You need several job runs to let the next runner pod reuse the existing PV, because ARC doesn't block the runner pod creation while other PVCs and PVs are being unbound(otherwise every runner pod creation will be delayed by previous runner pod's PV ubounding)

mumoshu: > So /var/lib/docker seems a bit vague.

@Gentoli Yeah, but trying to rely on an undocumented thing doesn't sound like a good idea, especially if you're going to use it in production.

Gentoli: > You need several job runs
Oh, when it gets to have to PV, caching with `/var/lib/docker` did work. `/var/lib/docker/overlay2` didn't.

mumoshu:
@toast-gear I removed the explanation on legacy runnerset limitation. It should be good to go now!

toast-gear:
Ref https://github.com/actions-runner-controller/actions-runner-controller/issues/1362

PR #1465, opened 2022-05-20
chore: remove stale app config

toast-gear:
https://github.com/actions-runner-controller/actions-runner-controller/blob/master/.github/workflows/stale.yaml we use the action version as you can't get the app version to ignore pull requests and the [codebase](https://github.com/probot/stale) looks fairly dead


mumoshu (approved):
Good catch! Thanks

PR #1471, opened 2022-05-21
fix: Avoid duplicate volume and mount name error for generic ephemeral volume as "work"

mumoshu:
While manually testing configurations being documented in #1464, I discovered that the use of dynamic ephemeral volume for the "work" directory was not working correctly due to the valiadation error.

This fixes the runner pod generation logic to not add the default volume and volume mount for the "work" directory so that the error disappears.

Ref #1464


toast-gear (approved):
LGTM

Issue #1321, opened 2022-04-08
Race condition in upscale webhook


genisd:
There is a race condition in the web base upscale, where 2 events occur at the same time.
Essentially two events should scale +1, yielding in 2extra  runners. But due to concurrency only 1 extra runner is booted.

Here a log from the webhook server, 4events are handled. But only 3 runners got reserved/booted.
I've been suspecting this and I'm super happy I found this, because it means we can fix it.

Could explain many issues reported (and we ourselves are hitting this multiple times daily).
https://github.com/actions-runner-controller/actions-runner-controller/issues/1279
https://github.com/actions-runner-controller/actions-runner-controller/issues/1238
```
2022-04-08T13:55:31Z	DEBUG	controllers.webhookbasedautoscaler	Patching hra runner-16-ramdisk for capacityReservations update	{"before": 0, "expired": 0, "amount": 1, "after": 1}
2022-04-08T13:55:31Z	INFO	controllers.webhookbasedautoscaler	scaled runner-16-ramdisk by 1
2022-04-08T13:55:33Z	DEBUG	controllers.webhookbasedautoscaler	Found some runner groups are managed by ARC	{"event": "workflow_job", "hookID": "332460040", "delivery": "8f26cfa0-b743-11ec-93e7-d12e236a53a0", "workflowJob.status": "queued", "workflowJob.labels": ["runner-16-ramdisk", "self-hosted"], "repository.name": "companyname-frontend", "repository.owner.login": "companyname", "repository.owner.type": "Organization", "enterprise.slug": "", "action": "queued", "groups": "RunnerGroup{Scope:Organization, Kind:Default, Name:}, RunnerGroup{Scope:Organization, Kind:Default, Name:}"}
2022-04-08T13:55:33Z	DEBUG	controllers.webhookbasedautoscaler	groups	{"event": "workflow_job", "hookID": "332460040", "delivery": "8f26cfa0-b743-11ec-93e7-d12e236a53a0", "workflowJob.status": "queued", "workflowJob.labels": ["runner-16-ramdisk", "self-hosted"], "repository.name": "companyname-frontend", "repository.owner.login": "companyname", "repository.owner.type": "Organization", "enterprise.slug": "", "action": "queued", "groups": "RunnerGroup{Scope:Organization, Kind:Default, Name:}, RunnerGroup{Scope:Organization, Kind:Default, Name:}"}
[...]

PR #1477, opened 2022-05-24
Make webhook-based scale race-free

mumoshu:
This prevents race condition in the webhook-based autoscaler when it received a webhook event while processing another webhook event and both ended up scaling up the same horizontal runner autoscaler at the same time.
    
Ref #1321

### DIFF
--- a/cmd/githubwebhookserver/main.go
+++ b/cmd/githubwebhookserver/main.go
@@ -93,7 +93,7 @@ func main() {
 		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
 	flag.DurationVar(&syncPeriod, "sync-period", 10*time.Minute, "Determines the minimum frequency at which K8s resources managed by this controller are reconciled. When you use autoscaling, set to a lower value like 10 minute, because this corresponds to the minimum time to react on demand change")
 	flag.StringVar(&logLevel, "log-level", logging.LogLevelDebug, `The verbosity of the logging. Valid values are "debug", "info", "warn", "error". Defaults to "debug".`)
-	flag.IntVar(&queueLimit, "queue-limit", controllers.DefaultQueueLimit, `The maximum length of the scale operation queue. The scale opration is enqueued per every matching webhook event, and the server returns a 500 HTTP status when the queue was already full on enqueu attempt.`)
+	flag.IntVar(&queueLimit, "queue-limit", controllers.DefaultQueueLimit, `The maximum length of the scale operation queue. The scale opration is enqueued per every matching webhook event, and the server returns a 500 HTTP status when the queue was already full on enqueue attempt.`)
 	flag.StringVar(&webhookSecretToken, "github-webhook-secret-token", "", "The personal access token of GitHub.")
 	flag.StringVar(&c.Token, "github-token", c.Token, "The personal access token of GitHub.")
 	flag.Int64Var(&c.AppID, "github-app-id", c.AppID, "The application ID of GitHub App.")
--- a/cmd/githubwebhookserver/main.go
+++ b/cmd/githubwebhookserver/main.go
@@ -72,6 +72,7 @@ func main() {
 		enableLeaderElection bool
 		syncPeriod           time.Duration
 		logLevel             string
+		queueLimit           int
 
 		ghClient *github.Client
 	)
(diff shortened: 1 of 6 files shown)


cmd/githubwebhookserver/main.go
```diff
@@ -92,6 +93,7 @@ func main() {
 		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
 	flag.DurationVar(&syncPeriod, "sync-period", 10*time.Minute, "Determines the minimum frequency at which K8s resources managed by this controller are reconciled. When you use autoscaling, set to a lower value like 10 minute, because this corresponds to the minimum time to react on demand change")
 	flag.StringVar(&logLevel, "log-level", logging.LogLevelDebug, `The verbosity of the logging. Valid values are "debug", "info", "warn", "error". Defaults to "debug".`)
+	flag.IntVar(&queueLimit, "queue-limit", controllers.DefaultQueueLimit, `The maximum length of the scale operation queue. The scale opration is enqueued per every matching webhook event, and the server returns a 500 HTTP status when the queue was already full on enqueu attempt.`)
```
toast-gear, line 96:
```suggestion
	flag.IntVar(&queueLimit, "queue-limit", controllers.DefaultQueueLimit, `The maximum length of the scale operation queue. The scale opration is enqueued per every matching webhook event, and the server returns a 500 HTTP status when the queue was already full on enqueue attempt.`)
```

controllers/horizontal_runner_autoscaler_webhook_worker.go
```diff
@@ -0,0 +1,53 @@
...
+			case t := <-w.scaleTargetQueue:
+				work(t)
+			}
+		}
+	}()
+
+	return w
+}
+
+// Add adds the scale target to the bounded queue, returning the result as a bool value. It returns true on successful enqueue, and returns false otherwise.
```
toast-gear, line 35:
```suggestion
// Add the scale target to the bounded queue, returning the result as a bool value. It returns true on successful enqueue, and returns false otherwise.
```

toast-gear (approved):

cablespaghetti:
I've deployed this to our staging environment. So far it's looking good 👍

cablespaghetti:
We're having an intermittent issue where there is a large delay in scaling up, but it does seem to happen eventually. It seems like it's when there are some other jobs running and for some reason the controller is waiting for those jobs/runners to finish before actioning the scale up from the webhook?

toast-gear (approved):

mumoshu:
@cablespaghetti Thanks for your detailed report!

> the scale down delay blocking scale up

You seem to have discovered a new and unrelated bug while testing this! Thank you so much. I'll try to fix it asap and include it in the upcoming 0.25.0.

mumoshu:
I dug it deeper and still unsure what was going on. `Calculated desired replicas of 32` in the first log indicates that it already passed through the scaleDownDelayAfterScaleOut check, so theoretically it shouldn't be due to scaleDownDelayAfterScaleOut 🤔

mumoshu:
@cablespaghetti Can you confirm that your story was like this- you first maxed out the number of runners at `32` as you set `maxReplicas: 32` in your HRA spec and by receiving approx 30 webhook events to trigger scale-ups? Then you had to wait until the scaleDownDelayAfterScaleOut passed before ARC finally "recreate" the completed pods to match the number of active runners back to `32`?

ARC's design ensures that the completed runner pods are recreated only after RunnerDeployment and RunnerReplicaSet's `effectiveTime` field or `replicas` field is updated by HRA so that it won't end up flapping. I think I found in ARC's codebase that RunnerDeployment doesn't propagate EffectiveTime to RunnerReplicaSet when `replicas` isn't updated, which might explain your issue...

cablespaghetti:
> @cablespaghetti Can you confirm that your story was like this- you first maxed out the number of runners at `32` as you set `maxReplicas: 32` in your HRA spec and by receiving approx 30 webhook events to trigger scale-ups? Then you had to wait until the scaleDownDelayAfterScaleOut passed before ARC finally "recreate" the completed pods to match the number of active runners back to `32`?
> 
> ARC ensures that the completed runner pods are recreated only after RunnerDeployment and RunnerReplicaSet's `effectiveTime` field or `replicas` field is updated by HRA so that it won't end up flapping(=ARC recreates a pod but then it receives a webhook event of `workflow_job` with `status=completed` which decreases the desired replicas and therefore terminating the just recreated pod....). I think I found in ARC's codebase that RunnerDeployment doesn't propagate EffectiveTime to RunnerReplicaSet when `replicas` isn't updated, which might explain your issue...

Yes I think setting maxReplicas might be the root cause of our problem then. Thanks for investigating 👍

mumoshu:
@cablespaghetti Thanks a lot for reporting & confirming! I believe #1568 fixes the issue. I'd appreciate it if you could give it a try 🙏  Both this PR and #1568 have been merged so you might better build a canary version of ARC from our current main branch if you have any chance.

cablespaghetti:
@mumoshu Sorry just saw this. I'll run 0.25.0 and report back if we have a repeat of the problem. Thanks for being so responsive in fixing bugs!

PR #1484, opened 2022-05-25
ci: align pipeline files and setups

toast-gear:
- go back to using tags, commit hashes are too noisy with Renovate
- use the same composite action to setup docker everywhere
- general tidy up of syntax, format and file extensions of workflows so we are consistent across all pipelines
- added additional trigger ignores on some pipelines
- renamed files and workflows so they all (apart from runners) do this verb-noun thing for consistency

Note this should be merged with `[skip ci]` and the status check will need adjusting post merge, I guess _technically_ we should probably expect the result (`Code scanning results / CodeQL Successful in 3s — No new or fixed alerts`) rather than the pipeline run itself


toast-gear:
@mumoshu this is good to review, probably worth doing the v0.24.0 release first before this is merged (assuming no changes needed)

toast-gear:
https://github.com/moby/buildkit/issues/2885 Added a scope to the cache so we get layer caching for both runner images. This was based on https://github.com/moby/buildkit/issues/2885

.github/workflows/publish-arc.yaml
```diff
@@ -66,4 +58,6 @@ jobs:
             ${{ env.DOCKERHUB_USERNAME }}/actions-runner-controller:latest
             ${{ env.DOCKERHUB_USERNAME }}/actions-runner-controller:${{ env.VERSION }}
             ${{ env.DOCKERHUB_USERNAME }}/actions-runner-controller:${{ env.VERSION }}-${{ steps.vars.outputs.sha_short }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
```
crazy-max, line 62: You're using the same scope in this workflow and the canary one and therefore will overlap with each other. You should define a scope:

```yaml
cache-from: type=gha,scope=arc
cache-to: type=gha,mode=max,scope=arc
```

.github/workflows/validate-chart.yaml
```diff
@@ -36,7 +36,8 @@ jobs:
           chmod 755 kube-score
 
       - name: Kube-score generated manifests
-        run: helm template  --values charts/.ci/values-kube-score.yaml charts/* | ./kube-score score -
+        run: |
+          helm template  --values charts/.ci/values-kube-score.yaml charts/* | ./kube-score score -
```
mumoshu, line 40: This step was failing with:

```
Run helm template  --values charts/.ci/values-kube-score.yaml charts/* | ./kube-score score -
Error: open charts/.ci/values-kube-score.yaml: no such file or directory
/home/runner/work/_temp/fa25a86f-4d07-4fb2-a01a-4260a335971a.sh: line 2: --ignore-test: command not found
```

mumoshu: You'd need to either (1) add `\` at the end of each line in the command or (2) revert it to use `run:` instead of `run: |`. I personally prefer (2) as we might have been using `run: ` to avoid (1) in the first place 😃

.github/workflows/run-codeql.yaml
```diff
@@ -1,26 +1,32 @@
...
-    branches: [master]
+    branches:
+      - master
   schedule:
     - cron: '30 1 * * 0'
 
 jobs:
-  CodeQL-Build:
+  codeql-scan:
+    name: CodeQL Scan
```
mumoshu, line 15: If we got this way, we might better use just `Scan` because with `CodeQL Scan` our build status title ends up with `CodeQL Scan / CodeQL Scan` which isn't clear which is the workflow name or the job name, whereas with `Scan` it might be like `CodeQL / Scan`.

PR #1491, opened 2022-05-26
Allow customization of ServiceMonitor namespace

evan-hines-firebolt:
Adds a values.yaml field to customize the namespace that the ServiceMonitor is deployed into. Prometheus allows for only specific namespaces to be monitored for service monitors and this eases the constraint that it must include 'default'

It still defaults to 'default' for backwards compatibility.

If we prefer to change the serviceMonitor field's type to an object, something like the following may be preferable.

```yaml
serviceMonitor:
  enabled: true
  namespace: default
```


charts/actions-runner-controller/templates/githubwebhook.serviceMonitor.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
   name: {{ include "actions-runner-controller-github-webhook-server.serviceMonitorName" . }}
+  namespace: {{ .Values.serviceMonitorNamespace }}
```
mumoshu, line 11: It looks like you missed `.metrics` in-between?

```suggestion
  namespace: {{ .Values.metrics.serviceMonitorNamespace }}
```

evan-hines-firebolt: Fixed:
```
helm template . | grep -a100 namespace | grep -b1 test-service
17755-  name: release-name-actions-runner-controller-service-monitor
17818:  namespace: test-service-monitor
17852-spec:
--
18586-  name: release-name-actions-runner-controller-github-w-service-monitor
18658:  namespace: test-service-monitor
18692-spec:

File: charts/actions-runner-controller/templates/githubwebhook.serviceMonitor.yaml [hunk 2]
```diff
@@ -8,6 +8,7 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
   name: {{ include "actions-runner-controller-github-webhook-server.serviceMonitorName" . }}
+  namespace: {{ .Values.metrics.serviceMonitorNamespace }}
```
mumoshu, line 11: @evan-hines-firebolt Hey!! Sorry for the radio silence. I was about to finish all the works towards the next version of ARC and finally had some time to review this once again.

In the middle of doing so, I realized that this may not work as flexibly as I first thought.

We use `namespace: {{ .Release.Namespace }}` for the `controller-manager` and `github-webhook-server` deployments and pods like declared in https://github.com/actions-runner-controller/actions-runner-controller/blob/4446ba57e1a21077136eb371795b7fe0acbc72c7/charts/actions-runner-controller/templates/githubwebhook.deployment.yaml#L6.

This `{{ .Values.metrics.serviceMonitorNamespace }}` doesn't prevent us from setting it to another namespace that's totally different from `{{ .Values.metrics.serviceMonitorNamespace }}`. Is it supposed to work? Perhaps not. In L22-24 we have this:

```
  selector:
    matchLabels:
      {{- include "actions-runner-controller-github-webhook-server.selectorLabels" . | nindent 6 }}
```

This basically says that this service monitor finds the target pods by labels within the same namespace as the service monitor. So, if `{{ .Values.metrics.serviceMonitorNamespace }}` wasn't equal to `{{ .Release.Namespace }}`, it will never able to monitor anything, right? 🤔 

Note that I'm not an expert in the prometheus-operator so your opinion and confirmation is very helpful!
If I'm not missing anything, probably we should update it to:

```suggestion
  namespace: {{ .Release.Namespace }}
```

and remove `metrics.serviceMonitorNamespace` as now unnecessary? WDYT?

mumoshu: @evan-hines-firebolt Let's go ahead with my sugegstion! I'd appreciate it if you could submit a follow-up PR or a comment if this turns wrong.

charts/actions-runner-controller/templates/controller.metrics.serviceMonitor.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
   name: {{ include "actions-runner-controller.serviceMonitorName" . }}
+  namespace: {{ .Values.metrics.serviceMonitorNamespace }}
```
mumoshu, line 11:
```suggestion
  namespace: {{ .Release.Namespace }}
```

mumoshu:
To be clear, this issue appears only when you used `helm template`. `helm template` is known to not automatically set `metafata.name` so every chart author needs to add `name: {{ .Release.Namespace }}` to the metadata of every resource template.
Based on the discussion at https://github.com/actions-runner-controller/actions-runner-controller/pull/1491#discussion_r912321046 I'm going to update this to just include `name: {{ .Release.Namespace }}` in the metadata, so that it works for `helm template` too.

charts/actions-runner-controller/values.yaml
```diff
@@ -104,6 +104,7 @@ metrics:
   serviceAnnotations: {}
   serviceMonitor: false
   serviceMonitorLabels: {}
+  serviceMonitorNamespace: default
```
mumoshu, line 107:
```suggestion
```

mumoshu (approved):
Thanks a lot for your contribution @evan-hines-firebolt!

nicklev:
Hi there!

namespace: {{ .Release.Namespace }} is not very flexible.

For example the ARC is deployed in a dedicated namespace but kube-prometheus-stack usually is deployed under "monitoring" namespace. 
This creates the following problem. We want all ServiceMonitors to be added under "monitoring" namespace for easier Prometheus configuration but namespace: {{ .Release.Namespace }} deploys the ServiceMonitor under ARC namespace.

In my opinion https://github.com/actions-runner-controller/actions-runner-controller/pull/1491/commits/d50be98bef03d890b4757624999f23ebdfd0a186
is a more flexible approach for ServiceMonitor deployment.

PR #1500, opened 2022-06-01
added lower function for labels value on compare

Sajadorouji:


mumoshu:
@Sajadorouji Hey! Why do you need this? AFAIK this case-insensitive thing isn't a standard convention in the K8s eco-system at least.

toast-gear:
Does this fix a difference between how GitHub handles labels and how ARC handles labels? This change would only make sense if GitHub's label to runner allocation system is case-insensitive and ARC currently is case-sensitive.

PR #1512, opened 2022-06-08
ci: fix permisions

toast-gear:
The canary pipeline currently fails with `error: failed to solve: server message: insufficient_scope: authorization failed`. I think we need additional permissions. It's not clear which permission relates to caching but this looked like the most likely one.

EDIT Added the write permission for packages seen as we require a login to GitHub Packages as part of the composite action, we are planning on publish all images to GitHub Packages in addition to Dockerhub so it's needed for a subsequent PR anyway. We also weren't logging in due to conditionals on the composite action. This action is now used ot set-up Docker everywhere and so are now a problem.

### DIFF
--- a/controllers/utils_test.go
+++ b/controllers/utils_test.go
@@ -32,4 +32,3 @@ func Test_filterLabels(t *testing.T) {
 		})
 	}
 }
-
--- a/controllers/utils_test.go
+++ b/controllers/utils_test.go
@@ -32,3 +32,4 @@ func Test_filterLabels(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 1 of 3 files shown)


mumoshu (approved):
LGTM. Thanks for your effort 🙏

controllers/utils_test.go
```diff
@@ -32,3 +32,4 @@ func Test_filterLabels(t *testing.T) {
 		})
 	}
 }
+
```
mumoshu, line 35:
```suggestion
```

mumoshu (approved):

PR #1517, opened 2022-06-09
Feat/shellcheck

CASABECI:
Thanks for resource available


Fleshgrinder (changes requested):
I am working on rewriting the scripts to be proper Bash, and pass all shellcheck tests. Hence, I would prefer it if this PR is not merged. For instance, the `logger.sh` script already does not exist anymore and was replaced by `logger.bash` in https://github.com/actions-runner-controller/actions-runner-controller/pull/1246. Similarly, the `startup.sh` script saw a total rewrite in https://github.com/actions-runner-controller/actions-runner-controller/pull/1454 (which was reverted in https://github.com/actions-runner-controller/actions-runner-controller/pull/1561 but actually already has a proper fix in https://github.com/actions-runner-controller/actions-runner-controller/pull/1560 so that it can be merged again).

.github/workflows/runners.yml
```diff
@@ -40,11 +40,28 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Setup Docker Environment
+      - name: Lint shell
+        run: |
+           make -C runner lint-scripts
```
mumoshu, line 45: Hey @CASABECI! Thanks for a lot for your contribution.
I finally managed to revisit this PR and have some time to integrate your work into the project,
and realized that we now have an exclusion rulle for `runner/Makefile` so if we added this `lint-scripts` target to the this file, it won't get triggered when the Makefile has changed.
Probably we'd better roll a new workflow dedicated for running spellcheck... Let me try.

runner/startup.sh
```diff
@@ -1,4 +1,8 @@
 #!/bin/bash
+
+# SC1091: source /opt/bash-utils/logger.sh won't exist when shellcheck is ran
+# shellcheck disable=SC1091
+
```
mumoshu, line 5: FYI TIL that you can tell shellcheck the path to look from the scripts.
`shellcheck --source-path runner` made this mark unnecessary.

mumoshu:
＠CASABECI Hey! Sorry for the massive delay, but we finally managed to introduce shellcheck into our CI workflow and fixed many lint errors along the way. See linked pull requests for more information. Your contribution enabled me to get started with shellcheck very quickly. Thanks a lot for your work!

PR #1524, opened 2022-06-13
ci: publish controller canary to github packages

toast-gear:


mumoshu (approved):
Thanks! Let's see if it passes through the pipeline.

PR #1525, opened 2022-06-13
ci: add GitHub packages to arc release

toast-gear:


.github/workflows/publish-arc.yaml
```diff
@@ -58,6 +58,9 @@ jobs:
             ${{ env.DOCKERHUB_USERNAME }}/actions-runner-controller:latest
             ${{ env.DOCKERHUB_USERNAME }}/actions-runner-controller:${{ env.VERSION }}
             ${{ env.DOCKERHUB_USERNAME }}/actions-runner-controller:${{ env.VERSION }}-${{ steps.vars.outputs.sha_short }}
+            ghcr.io/actions-runner-controller/actions-runner-controller:latest
+            ghcr.io/actions-runner-controller/actions-runner-controller:${{ env.VERSION }}
+            ghcr.io/actions-runner-controller/actions-runner-controller:${{ env.VERSION }}-${{ steps.vars.outputs.sha_short }}
```
mumoshu, line 63: I guess we might also need the permission for package writing at least, as similar as we do for the canary builds (#1524):

https://github.com/actions-runner-controller/actions-runner-controller/blob/2733c36d0ed30a8761ae8b42636975efaee1ae2c/.github/workflows/publish-canary.yaml#L23-L25

toast-gear: we don't have any exlicit permissions set at the moment so the token will be write all. I'll do a little test to figure out which we need and add them to his PR.

.github/workflows/publish-arc.yaml
```diff
@@ -5,6 +5,11 @@ on:
     types:
       - published
 
+# https://docs.github.com/en/rest/overview/permissions-required-for-github-apps
+permissions:
+ contents: write
```
mumoshu, line 10:
```suggestion
 contents: read 
```

Perhaps `read` might be enough as our canary pipeline seems to be working fine with `read`? (I wonder if we need `write` for cache writing but apparently not?

mumoshu (approved):
One comment but LGTM. Thanks for your contribution as always!

Issue #1533, opened 2022-06-14
[Meta] Difficult/impossible to create issues using bug template


danxmoran:
### Controller Version

N/A

### Helm Chart Version

N/A

### CertManager Version

N/A

### Deployment Method

Other

### cert-manager installation

N/A

### Checks

- [X] This isn't a question or user support case (For Q&A and community support, go to [Discussions](https://github.com/actions-runner-controller/actions-runner-controller/discussions). It might also be a good idea to contract with any of contributors and maintainers if your business is so critical and therefore you need priority support
- [X] I've read [releasenotes](https://github.com/actions-runner-controller/actions-runner-controller/tree/master/docs/releasenotes) before submitting this issue and I'm sure it's not due to any recently-introduced backward-incompatible changes
- [X] My actions-runner-controller version (v0.x.y) does support the feature
- [X] I've already upgraded ARC (including the CRDs, see charts/actions-runner-controller/docs/UPGRADING.md for details) to the latest and it didn't fix the issue

### Resource Definitions

```yaml
N/A
```

### To Reproduce

```markdown
1. Create a bug using this template, filling out the required fields using the specified info
```

### Describe the bug

GitHub refuses to create the new issue, failing with:
```
There was an error creating your issue: body is too long, body is too long (maximum is 65536 characters). 
```

I assumed the controller log was to blame for this, so I tried deleting the contents of that box in the template. The UI still blocks issue creation (the "Submit new issue" button is deactivated), showing the same warning text.

### Describe the expected behavior

I expected the issue to be created 😄 

### Controller Logs

```shell
N/A
```

### Runner Pod Logs

```shell
N/A
```

mumoshu:
@danxmoran Thanks for reporting! I believe this has been addressed via #1539. Please let me close as resolved but feel free to submit another issue if necessary.

PR #1539, opened 2022-06-15
chore: suggest gist first

toast-gear:
Ref https://github.com/actions-runner-controller/actions-runner-controller/issues/1533

image

The default text area cuts off the existing instructions to provide the logs via a Gist. Moved this to the top so people see it making them more likely to provide them via a gist and not have issues when creating bugs by hitting the max issue length


mumoshu (approved):
Nice! Thanks for the enhancement ☺️

PR #1541, opened 2022-06-15
Fix: Match Ingress API Version correctly.

john-delivuk:
Using version 1.17 I run into rendering issues. 
```
Error: Failed to render chart: exit status 1: Error: unable to build kubernetes objects from release manifest: resource mapping not found for name: "actions-runner-controller-github-webhook-server" namespace: "github-runners" from "": no matches for kind "Ingress" in version "networking.k8s.io/v1"
  ensure CRDs are installed first
```

The conditional managing this only looks at API versions, which can have several kinds, so the condition would match incorrectly. 

Here is an example of how others have [solved this](https://github.com/runatlantis/helm-charts/blob/main/charts/atlantis/templates/ingress.yaml#L5-L11)


mumoshu (approved):
LGTM. Thanks a lot for your fix!

PR #1548, opened 2022-06-17
chore: move runner docker check

toast-gear:
Moved the docker check further down as in theory that will result in a slightly quicker runner boot time by giving Docker a chance to become up and and be available without blocking the entrypoint. At the moment we check quite early in the entrypoint so often Docker isn't up yet but is highly likely to come up fine, it makes sense to check this at the end as the Docker boot process is asynchronous to the runner entrypoint script


mumoshu (approved):
LGTM. Thanks for your continued effort ☺️

PR #1549, opened 2022-06-17
ci: runners trigger on push

toast-gear:
GitHub secrets are not available in a pipeline run when the run is triggered by a pull request has been triggered from someone outside the organisation. 

GitHub secrets are available from a push trigger. 

Updating our runner CI when releasing a new runner to trigger from a push rather than a closed pull request (i.e. a merged pull request). This will allow us to do releases of the runners without the need for someone in the organisation to also make a change.


mumoshu (approved):
LGTM. Thanks!

Issue #1557, opened 2022-06-23
Pod startup failing due to new startup.sh


rich-bain:
### Controller Version

0.24.1

### Helm Chart Version

0.19.1

### CertManager Version

1.8.1

### Deployment Method

Other

### cert-manager installation

- Installation process with Helm Via flux. Tested and working correctly with custom domain.

### Checks

- [X] This isn't a question or user support case (For Q&A and community support, go to [Discussions](https://github.com/actions-runner-controller/actions-runner-controller/discussions). It might also be a good idea to contract with any of contributors and maintainers if your business is so critical and therefore you need priority support
- [X] I've read [releasenotes](https://github.com/actions-runner-controller/actions-runner-controller/tree/master/docs/releasenotes) before submitting this issue and I'm sure it's not due to any recently-introduced backward-incompatible changes
- [X] My actions-runner-controller version (v0.x.y) does support the feature
- [X] I've already upgraded ARC (including the CRDs, see charts/actions-runner-controller/docs/UPGRADING.md for details) to the latest and it didn't fix the issue

### Resource Definitions

```yaml
apiVersion: actions.summerwind.dev/v1alpha1
kind: RunnerDeployment
metadata:
  name: default-org-runners
  namespace: actions-runner-system
spec:
  replicas: 1
  template:
    spec:
      image: summerwind/actions-runner-dind. # <-- was unpinned
      dockerdWithinRunnerContainer: true
      organization: <redacted>
      env: []
```

### To Reproduce

```markdown
Try to run on a GKE cluster with the latest image `latest@sha256:331d12b3c2bb35436f82dccfc661afd13bb093afd3461d19cc5fce74b64e896c`. Pod correctly schedules but `runner`
```

### Describe the bug

See `textPayload` key below.
```json
{
  "textPayload": "tee: 'standard output': Bad file descriptor",    
  "insertId": "vrxebg7zrlowahpw",
  "resource": {
    "type": "k8s_container",
    ...
    }
  },
  "timestamp": "2022-06-23T03:45:09.572895611Z",
  "severity": "ERROR",
  "labels": {
       ...
  },
   ...
}
```

This is a `k8s_container` resource which means the pod was correctly scheduled but failed during startup.

Symptoms are a restart loop of pods every ~1 second.

### Describe the expected behavior

[...]

PR #1560, opened 2022-06-23
Fix startup.sh of runner from exiting `Bad file descriptor` error

kyontan:
This may fix https://github.com/actions-runner-controller/actions-runner-controller/issues/1557.

The original line `sudo tee $dockerd_config_path 1>&- <<<"$dockerd_config" may intent to write string `"$dockerd_config"` to `$dockerd_config_path` without echoing to terminal (`1>&-`).
But it doesn't work in the runner pod's bash (`bash 5.0.17(1)`) because of `Bad file descriptor` error for the part of command: `1>&-`.

The `1>&-` is a special use of bash's redirection that closes 1 (stdout) and I thought it causes the error.
ref: https://www.gnu.org/software/bash/manual/html_node/Redirections.html

To fix the error, I rewrite it to `1>/dev/null`.

### DIFF
--- a/runner/startup.sh
+++ b/runner/startup.sh
@@ -27,7 +27,7 @@ if [[ $MTU ]]; then
   # session.
   if [[ $MTU =~ ^[1-9][0-9]*$ ]]; then
     # See https://docs.docker.com/engine/security/rootless/
-    sudo tee -a $dockerd_supervisor_config_path 1>&- <<<"environment=DOCKERD_ROOTLESS_ROOTLESSKIT_MTU=$MTU"
+    sudo tee -a $dockerd_supervisor_config_path 1>/dev/null <<<"environment=DOCKERD_ROOTLESS_ROOTLESSKIT_MTU=$MTU"
     dockerd_config=$(jq -S ".mtu = $MTU" <<<"$dockerd_config")
   else
     # shellcheck source=runner/logger.bash
@@ -42,7 +42,7 @@ if [[ ${DOCKER_REGISTRY_MIRROR:-} != '' ]]; then
 fi
 
 sudo mkdir -p ${dockerd_config_path%/*}
-sudo tee $dockerd_config_path 1>&- <<<"$dockerd_config"
+sudo tee $dockerd_config_path 1>/dev/null <<<"$dockerd_config"
 
 for config in $dockerd_config_path $dockerd_supervisor_config_path; do
   (


BeyondEvil:
Is it possible to add a test so that this doesn't happen again?

kyontan:
I think it's difficult now.  To prevent such a degradation, we need somewhat E2E testing.

BeyondEvil:
Wouldn't it be sufficient to just call the startup script in this case?

mumoshu:
@kyontan FYI, we've reverted the original change in #1561 but we'd certainly revisit this when we redo #1454. Thanks a lot for the detailed report and your fix! Good job 👍

Fleshgrinder (approved):

Fleshgrinder:
@mumoshu this is the proper fix for the problem encountered in https://github.com/actions-runner-controller/actions-runner-controller/issues/1557. The `1>&-` closes the file handle, but if a program treats the file handle as a file and it is closed it fails. Switching from `1>&-` to `1>/dev/null` ensures that there is a file handle that can be `stat`ed.

What I don't understand is why this was not a problem for anyone of us while testing. But, we all knew that our test coverage for this is not sufficient. We need proper integration tests, nothing else can help here in the future.

mumoshu:
@Fleshgrinder Hey! Thanks for sharing your insight. I did try reproducing it with a bash snippet and it did reproduce on my macOS with bash 3.x, but it didn't on my “bash” version 3/4/5 containers on linux(ubuntu 20.04, and I thought our runner images are also based on ubuntu 20.04) . So, I don't really know what kind of integration tests could have caught it. Before coming up with a perfect integration test that covers issues like that, perhaps we'd better have a guideline to write a portable bash script in the first place... or rewrite it in Go which should be more portable(and easy to test).

Fleshgrinder:
Btw. it is very common that programs expect stdout to be a proper fd, but are lenient in regards to stderr. So, I would refrain from actually stating that tee is broken, wrong, or anything. It's my fault, I should have known better, and it should have been caught while testing, even if just locally (as illustrated by the simple reproducer).

mumoshu:
Thanks for your response!

> I don't see how rewriting it in Go would help

The only reason is Go is seemingly more portable. At least, it doesn't fall down like this when we coded a Go program to not forward stdout from child processes.

I'd been sticking with bash-based entrypoint and startup scripts just because it might be easier to maintain, but if it started to look untrue.

Even if we had a comprehensive(it can't be comprehensive btw) integration test, it doesn't justify maintaining advanced bash scripts. We'd better architect it correctly and then integration-test it, rather than integration-testing it to make bad arch ok.

Fleshgrinder:
Oh, don't get me wrong, I'm not trying to convince you to not do it in Go. Quite the contrary, the only reason I'm working on the Bash scripts is because they are, well, sorry to be so frank, bad. Hence, if you (the maintainers) feel like you can do it better in Go, please go ahead.

My only concern is that it is always very hard with normal programming languages to do this kind of stuff, without shelling out all the time. We are interacting with a lot of programs here, switching between current and super user, etc. pp. This is simply what Bash was intended for.

Also, imho, there is no reason to panic just because of one bug. There is absolutely no guarantee to not have bugs in any other program without tests. And, well, we do not have tests for this, like, none at all. This particular issue would have been found right away if we would have had a test suite. I am much more concerned about all the MTU stuff we are doing all over the place, this seems more complicated to verify.

😊

mumoshu:
> This particular issue would have been found right away if we would have had a test suite

True! So the most needed contribution to this area would be to add tests, rather than to make it better... It's sad that I had been short on time to prioritize that. If you think we can keep entrypoint.sh and startup.sh maintainable and testable, that's good, if anyone could contribute tests for startup.sh, that would also be good (because then I don't need to rewrite it in Go)

It's just that I had observed no positive signals to convince me that those scripts are maintainable these days.

Fleshgrinder:
I don't want to convince you. It's a question for you, the core maintainers. Happy to think about ways of testing and contributing, also happy to continue supporting with the Bash stuff. But, no matter what, we have to keep it to the absolute minimum. Bash isn't simple.

The entire goal of the improvements is that we can introduce shellcheck and shfmt. These are going to help every contributor to write better scripts.

mumoshu:
@Fleshgrinder Thanks for the suggestion and your clarification! Let me say I do agree with the goal. I just wanted to make it crystal clear that I do take the reliability and maintainability of ARC seriously and that I'm willing to do whatever I can to improve it. Bash vs Go is not my goal but an implementation detail/idea.

Fleshgrinder:
Couldn't agree more. But, if Bash, good Bash. 😉

BeyondEvil:
+1 on shellcheck and shfmt.

toast-gear:
Thank you for raising a PR to fix the broken image, I opted to just do a revert of the PR as that was quicker and sure to "fix" the problem. Since then the scripting has been overhauled so this PR is no longer needed. Thanks again for your work though!.

PR #1562, opened 2022-06-23
docs: fix logic explanation for scale down delay

giuscri:
Signed-off: Giuseppe Crinò <giuscri@gmail.com>


mumoshu (approved):
Ah, good catch! Thanks for fixing it @giuscri ☺️

PR #1565, opened 2022-06-24
Adds way to allow additional environment variables from secretKeyRef

jonathanwiemers:
First-off: Thanks for the great work here! 

### Problem
We are using a CSI-provder to sync secrets between AWS and EKS. The way that secrets are bound into the controller deployment doesn't really fit our workflow, so I'd be happy to specify my own environment variables from secrets, kind of the same way as I can specify additional volumes and mounts. 

This PR would allow something like this: 
```yaml
env:
  - name: GITHUB_APP_ID
    valueFrom:
      secretKeyRef:
        key: PROD_GHA-CONTROLLER_APP-ID
        name: prod-gha-controller-app-id
        optional: true
  - name: GITHUB_APP_INSTALLATION_ID
    valueFrom:
      secretKeyRef:
        key: PROD_GHA-CONTROLLER_APP-INSTALLATION-ID
        name: prod-gha-controller-app-installation-id
        optional: true
  - name: GITHUB_APP_PRIVATE_KEY
    valueFrom:
      secretKeyRef:
        key: PROD_GHA-CONTROLLER_APP-PRIVATE-KEY
        name: prod-gha-controller-app-private-key
        optional: true
```

Since i didn't want to break the existing `env` prop I've added a new one. The name is not perfect, maybe there're other suggestions :)

### DIFF
--- a/charts/actions-runner-controller/templates/deployment.yaml
+++ b/charts/actions-runner-controller/templates/deployment.yaml
@@ -118,10 +118,6 @@ spec:
               name: {{ include "actions-runner-controller.secretName" . }}
               optional: true
         {{- end }}
-        {{- range $key, $val := .Values.env }}
-        - name: {{ $key }}
-          value: {{ $val | quote }}
-        {{- end }}
         {{- if kindIs "slice" .Values.env }}
         {{- toYaml .Values.env | nindent 8 }}
         {{- else }}
--- a/charts/actions-runner-controller/templates/deployment.yaml
+++ b/charts/actions-runner-controller/templates/deployment.yaml
@@ -122,6 +122,9 @@ spec:
         - name: {{ $key }}
           value: {{ $val | quote }}
         {{- end }}
+        { { - if .Values.additionalFullEnv } }
+          { { - toYaml .Values.additionalFullEnv | nindent 8 } }
+        { { - end } }
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default (cat "v" .Chart.AppVersion | replace " " "") }}"
         name: manager
         imagePullPolicy: {{ .Values.image.pullPolicy }}
--- a/charts/actions-runner-controller/templates/deployment.yaml
+++ b/charts/actions-runner-controller/templates/deployment.yaml
@@ -122,9 +122,14 @@ spec:
         - name: {{ $key }}
           value: {{ $val | quote }}
         {{- end }}
-        { { - if .Values.additionalFullEnv } }
-          { { - toYaml .Values.additionalFullEnv | nindent 8 } }
-        { { - end } }
+        {{- if kindIs "slice" .Values.env }}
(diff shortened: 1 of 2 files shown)


charts/actions-runner-controller/templates/deployment.yaml
```diff
@@ -122,6 +122,9 @@ spec:
         - name: {{ $key }}
           value: {{ $val | quote }}
         {{- end }}
+        { { - if .Values.additionalFullEnv } }
+          { { - toYaml .Values.additionalFullEnv | nindent 8 } }
+        { { - end } }
```
mumoshu, line 127: Thanks for your effort! One thing- Probably we'd better guide existing users to use this over the legacy style, without requiring the key change.

So, could you try something like:

```suggestion
        {{- if kindIs "slice" .Values.env }}
        {{- toYaml .Values.env | nindent 8 }}
        {{- else }}
        {{- range $key, $val := .Values.env }}
        - name: {{ $key }}
          value: {{ $val | quote }}
        {{- end }}
        {{- end }}
```

Also, I think adding a space between `{`s would never work(AFAIK).
Have you actually tested this? (Sorry if you did!
Try running `helm upgrade --install $RELEASE_NAME ./charts/actions-runner-controller` locally so that you can test the updated version of the chart before submitting a PR.

jonathanwiemers: I did, but ofc forgot to push the fix... Fridays :)

jonathanwiemers:
@mumoshu should we add an example somewhere, so people understand that i'ts possible without reading the templates?

jonathanwiemers:
> @jonathanwiemers Thanks a lot for the fix! Regarding the example, I think it might be nice if it's covered in the default values.yaml as a comment

@mumoshu i've added an example, thanks for the quick answering here!

charts/actions-runner-controller/templates/deployment.yaml
```diff
@@ -122,6 +122,14 @@ spec:
         - name: {{ $key }}
           value: {{ $val | quote }}
         {{- end }}
```
mumoshu, line 124: Now the above 4 lines are duplicates of L128-131 which results in invalid env when you've specified a slice as env:

```
$ helm template foo ./charts/actions-runner-controller --set env[0].name=env1,env[0].value=val1 | grep val1 -C 2
              optional: true
        - name: 0
          value: "map[name:env1 value:val1]"
        - name: env1
          value: val1
        image: "summerwind/actions-runner-controller:v0.24.1"
        name: manager
```

jonathanwiemers: fixed

jonathanwiemers:
@mumoshu pls review once more :)

mumoshu (approved):
LGTM. Thanks a lot for your contribution (and patience)!

Issue #1374, opened 2022-04-22
Broken PercentageRunnersBusy autoscaling when stalled runner


alanbover:
**Describe the bug**
We are currently using PercentageRunnersBusy metric for autoscale the runners. We have found out that sometimes (because an unknown reason) workers are infinitely kept running a workflow.

When this happens and that pod is marked for downscaling (we are using ephemeral runners) the pod is marked with flag AnnotationKeyRunnerCompletionWaitStartTimestamp (here https://github.com/actions-runner-controller/actions-runner-controller/blob/master/controllers/runner_graceful_stop.go#L178). I understand that this happens because Github replies a 422 error code, and hence the runner should be kept registered in Github. It makes sense also that `tickRunnerGracefulStop` kinds of "forgets" about that pod and leaves the reconciler to handle the pod termination event.

What we are seeing though (maybe it's just because some timeout on a heartbeat between the stalled runner and Github) is that the runner is not registered in Github.

This is actually a problem because that pod is not handled then correctly for autoscaling (here https://github.com/actions-runner-controller/actions-runner-controller/blob/master/controllers/autoscaling.go#L312).

As an example, we have 3 runners in the cluster (all bussy) but only 2 are registered in Github. 
```
fractionBusy := float64(2) / float64(3)       // fractionBusy = 0.66 and should be 1
```

I believe that if autoscaling would take into consideration the runners that are not registered in Github this issue would disappear, something like:
```
fractionBusy := (float64(numRunnersBusy) + float64(numRunnersBusyNotRegistered)) / float64(desiredReplicasBefore)         //  fractionBusy = (2 + 1) / (3) = 1
```

We are though currently trying to migrate to Webhook driven scaling, so I don't think we will be able to provide more information (hopefully what it's being reported is enough)

**Checks**

- [x] My actions-runner-controller version (v0.22.0) does support the feature
- [ ] I'm using an unreleased version of the controller I built from HEAD of the default branch

**To Reproduce**
Steps to reproduce the behaviour:
I'm not sure how to reproduce it, as I don't fully understand why the runner gets unregistered from Github if it's still running

**Expected behaviour**
Autoscaler is able to detect that a pod with AnnotationKeyRunnerCompletionWaitStartTimestamp annotation is not registered in Github and it counts it as an extra bussy runner

**Screenshots**

**Environment (please complete the following information):**
 - Controller Version [e.g. 0.22.0]
 - Deployment Method [Helm]
[...]

PR #1579, opened 2022-06-29
Fix PercentageRunnersBusy scaling delay

mumoshu:
This should fix the up to 60 seconds delay in PercentageRunnersBusy scaling due to our GitHub API cache.

This fills the gap between ARC and GitHub to make scaling RunnerDeployment earlier.
We’ve recently introduced GitHub API cache to cache ListRunners API responses according to the cache-control header (usually it tells the client to cache the response for 60s) in #1127. A side-effect of that was PercentageRunnersBusy needs more time to recognize the latest view of busy runners, hence delayed scaling.

This change enhances ARC to treat a runner to be busy when it already failed to unregister the runner due to 422 error. 422 error means the runner is busy, so we can safely say the runner is busy even though ListRunners API response doesn’t say it’s busy (yet).

Ref #1374

### DIFF
--- a/controllers/runner_graceful_stop.go
+++ b/controllers/runner_graceful_stop.go
@@ -67,6 +67,25 @@ func annotatePodOnce(ctx context.Context, c client.Client, log logr.Logger, pod 
 
 	return updated, nil
 }
+func labelPod(ctx context.Context, c client.Client, log logr.Logger, pod *corev1.Pod, k, v string) (*corev1.Pod, error) {
+	if pod == nil {
+		return nil, nil
+	}
+
+	updated := pod.DeepCopy()
+	if updated.Labels == nil {
+		updated.Labels = map[string]string{}
+	}
+	updated.Labels[k] = v
+	if err := c.Patch(ctx, updated, client.MergeFrom(pod)); err != nil {
+		log.Error(err, fmt.Sprintf("Failed to patch pod to have %s annotation", k))
+		return nil, err
+	}
+
+	log.V(2).Info("Labeled pod", "key", k, "value", v)
+
+	return updated, nil
+}
 
 // If the first return value is nil, it's safe to delete the runner pod.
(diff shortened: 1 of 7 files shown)


toast-gear (approved):

controllers/runner_graceful_stop.go
```diff
@@ -187,6 +210,11 @@ func ensureRunnerUnregistration(ctx context.Context, retryDelay time.Duration, l
 		}
 
 		if runnerBusy {
+			_, err := labelPod(ctx, c, log, pod, AnnotationKeyUnregistrationFailureMessage, runnerUnregistrationFailureMessage)
```
everpcpc, line 213: hi, @mumoshu 
this line introduces an error like:
```
Failed to patch pod to have actions-runner/unregistration-failure-message annotation","runnerpod":"actions-runner/databend-nqjfv-gvv7v","error":"Pod \"databend-nqjfv-gvv7v\" is invalid: [metadata.labels: Invalid value: \"Bad request - Runner \\\"databend-nqjfv-gvv7v\\\" is still running a job\\\"\": must be no more than 63 characters, metadata.labels: Invalid value: \"Bad request - Runner \\\"databend-nqjfv-gvv7v\\\" is still running a job\\\"\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')]",
```

mumoshu: @everpcpc Ahhh good catch! I also realized this while testing it before 0.25.0 release but I lost my memory since then.
At least this doesn't necessarily be a full error message, so truncating it to 63 characters should be better than the current state.
I'm also wondering if we need to strip some characters that aren't valid within an annotation value.
I'm going to be busy for a few days- Would you mind submitting a pull request to fix it, so that I can review it and cut a new release asap? Thanks!

everpcpc: ref: https://github.com/actions-runner-controller/actions-runner-controller/pull/1615

PR #1580, opened 2022-06-29
Drop support for --once

mumoshu:
Once merged, this removes the support for the legacy `--once` flag for `actions/runner`.

Upgrade your GHES to 3.3 or greater if you use our runner image tagged with `latest`, or fork our dockerfile and use a custom runner image if you absolutely need the `--once` flag! But note and beware that `--once` has never been officially supported even by `actions/runner`.

Ref #1196


mumoshu:
I"ve locally tested this. Merging.

PR #1582, opened 2022-06-29
Extracted validations to a single point for runner spec

nikola-jokic:
Follow up on https://github.com/actions-runner-controller/actions-runner-controller/pull/1546#discussion_r905638576


mumoshu (approved):
LGTM. Thanks a lot for your contribution @nikola-jokic!

PR #1583, opened 2022-06-29
ci: add paths to push trigger

toast-gear:


mumoshu (approved):

PR #1594, opened 2022-06-30
fix type README.md descriptrion.

SeongJuMoon:
This PR tries to fix a typos I found when reading the README files.
i think `template.medatada` mean to `template.metadata`.


mumoshu (approved):
medatada 😂 Good catch! Thanks for fixing it

PR #1596, opened 2022-07-02
Fix a regression in webhook-based autoscaler

mumoshu:
The regression resulted in the webhook-based autoscaler be unable to find visible runner groups and therefore unable to scale up and down the target RunnerDeployment/RunnerSet at all when the webhook-based autoscaler was provided GitHub API credentials to enable the runner groups support. This fixes that.

The regression was introduced via #1578 which is not released yet. Users of existing ARC releases are therefore not affected.


mumoshu:
Thansk for your review!

pnjihia:
Has this been released?

pnjihia:
Thanks! It was a mistake on my end with the scaling settings, before (worked in prior version):
```
scaleUpTriggers:
    - githubEvent: {}
```
Changed to:
```
  scaleUpTriggers:
    - githubEvent:
        workflowJob: {}
```

PR #1599, opened 2022-07-03
Make it easier to E2E test on various K8s versions

mumoshu:
According to the Kubernetes version skew policy, all the currently supported Kubernetes versions are 1.22, 1.23, and 1.24.
We ensure ARC works on all those three versions by running the E2E test against those three versions.

This change makes it easy to do so- Just set `ARC_E2E_KUBE_VERSION` to the desired Kubernetes major and minor version numbers before running the E2E test. For example, you can run it against K8s 1.24.x with `ARC_E2E_KUBE_VERSION=1.24 go test -run TestE2E ./...`.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -122,11 +122,12 @@ func TestE2E(t *testing.T) {
 		t.Skip("Skipped as -short is set")
 	}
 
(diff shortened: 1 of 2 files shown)


toast-gear (approved):

mumoshu:
Thanks for reviewing!

PR #1603, opened 2022-07-04
Fix charts

nobiit:


nobiit:
Correcting the mistake of secretName

mumoshu:
@nobiit Hey! The helper is actually defined at https://github.com/actions-runner-controller/actions-runner-controller/blob/5a9e8545aa194ce517227c1e714bb44e87fe0dba/charts/actions-runner-controller/templates/_helpers.tpl#L71-L73 and it's working for me. Could you clarify a bit more on your problem?

nobiit:
Hi @mumoshu . If `githubWebhookServer.useRunnerGroupsVisibility` is enabled. It will mount a GITHUB_TOKEN environment variable from secret. And it looks like it's taking the wrong secret

mumoshu:
@nobiit Why do you think it's referring to a wrong secret?

nobiit:
https://github.com/actions-runner-controller/actions-runner-controller/blob/master/charts/actions-runner-controller/templates/githubwebhook.secrets.yaml#L13
https://github.com/actions-runner-controller/actions-runner-controller/blob/master/charts/actions-runner-controller/templates/manager_secrets.yaml#L27
https://github.com/actions-runner-controller/actions-runner-controller/blob/master/charts/actions-runner-controller/templates/deployment.yaml#L90

mumoshu:
> If githubWebhookServer.useRunnerGroupsVisibility is enabled. It will mount a GITHUB_TOKEN environment variable from secret

To be clear, I think you described exactly how it's supposed to work. Only a small correction is that it can source either PAT (GITHUB_TOKEN) or GitHub App credentials.

mumoshu:
@nobiit Ooo, good catch! It turns out I have never tried it with `githubWebhookServer.useRunnerGroupsVisibility` AND `githubWebhookServer.secret.create=true`. I had been bringing my own secret by specifying `githubWebhookServer.secret.enabled` and `githubWebhookServer.secret.name` when using `githubWebhookServer.useRunnerGroupsVisibility` and probably that's why I didn't notice anything up until now.

That said, the secret name is correct. What's missing in the chart is rather an ability to create a correct secret from values.

mumoshu:
More concretely, what we might need instead would be to add a set of kv templates to the secret data similarly to https://github.com/actions-runner-controller/actions-runner-controller/blob/bfc5ea4727f3dfb1a8cc590306f04f3d32133d4b/charts/actions-runner-controller/templates/manager_secrets.yaml#L15-L28

mumoshu:
To be clear, the reason the github webhook server accepts a different secret for PAT and github app credentials is to let it consume a different API rate limit budget than that of controller-manager. GitHub gives a dedicated budget for each GitHub user/app installation so giving a distinct set of API creds per each ARC component allows us to double the budget.

nobiit:
Hi @mumoshu . Oh. I understand the problem, maybe I should readjust my PR

nobiit:
Hi @mumoshu , Please check again

mumoshu (approved):
LGTM. Thanks for fixing it!

mumoshu:
Waiting for CI to pass now

charts/actions-runner-controller/templates/githubwebhook.secrets.yaml
```diff
@@ -12,5 +12,20 @@ data:
...
+{{- end }}
+{{- if .Values.githubWebhookServer.secret.github_app_private_key }}
+  github_app_private_key: {{ .Values.githubWebhookServer.secret.github_app_private_key | toString | b64enc }}
+{{- end }}
+{{- if .Values.githubWebhookServer.secret.github_token }}
+  github_token: {{ .Values.githubWebhookServer.secret.github_token | toString | b64enc }}
+{{- end }}
+{{- if .Values.githubWebhookServer.secret.github_basicauth_password }}
+  github_basicauth_password: {{ .Values.githubWebhookServer.secret.github_basicauth_password | toString | b64enc }}
+{{- end }}
```
mumoshu, line 29: Although it's harmless, strictly speaking, we might not need this at the moment, as we seem to refer to the auth secret for this in the webhook server deployment spec anyway:

https://github.com/actions-runner-controller/actions-runner-controller/blob/bfc5ea4727f3dfb1a8cc590306f04f3d32133d4b/charts/actions-runner-controller/templates/githubwebhook.deployment.yaml#L106

nobiit: Oh. Maybe I should delete it to avoid confusion

mumoshu (approved):
LGTM. Thank you so much for your contribution!

PR #1606, opened 2022-07-05
Fixed Simple typo

ViktorLindgren95:
Simple Typo on the Kind of VolumeClaimTemplates. Small T will cause problems


mumoshu (approved):

PR #1608, opened 2022-07-06
Document how to add Windows self-hosted runners

ian-flores:
This PR documents the process we used @voltrondata to add Windows self-hosted runners to be managed by the `actions-runner-controller`. It specifies changes that need to be made to the `cert-manager` and the `actions-runner-controller`. It also specifies how to deploy `RunnerDeployment`'s configured for Windows.


README.md
```diff
@@ -1744,6 +1745,115 @@ $ helm --upgrade install actions-runner-controller/actions-runner-controller \
...
+
+RUN Invoke-WebRequest -Uri 'https://aka.ms/install-powershell.ps1' -OutFile install-powershell.ps1; ./install-powershell.ps1 -AddToPath
+
+RUN powershell Set-ExecutionPolicy Bypass -Scope Process -Force; [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072; iex ((New-Object System.Net.WebClient).DownloadString('https://community.chocolatey.org/install.ps1'))
+
+RUN powershell choco install git.install --params "'/GitAndUnixToolsOnPath'" -y
+
+RUN powershell choco feature enable -n allowGlobalConfirmation
+
+CMD [ "pwsh", "-c", "./config.cmd --name $env:RUNNER_NAME --url https://github.com/$env:RUNNER_REPO --token $env:RUNNER_TOKEN --labels $env:RUNNER_LABELS --unattended --replace --ephemeral; ./run.cmd"]
```
mumoshu, line 1821: @ian-flores Hey! Thanks for submitting this awesome pull request 🙏 

Although I'm not an expert in managing Windows machines as K8s nodes, I tried my best to review this and everything looked generally good.

My only concern is in this line- how does it handle termination signals sent by Kubernetes?

On Linux, the PID 0 of each container gets SIGTERM on pod termination and usually, the PID 0 should gracefully stop all its child processes. We implement this for ARC with dumb-init, as you can see at:
https://github.com/actions-runner-controller/actions-runner-controller/blob/8b619e7c6fa9b8b07ff184e7882cec4108d0a52e/runner/actions-runner.dockerfile#L131-L132

Is PowerShell supposed to handle it, or do you need your own way?

amaldonadomat: Hello @mumoshu 

We are also quite new with Windows, but I'll try to answer here. As far as I know there are not termination signals on Windows, it is handled with the CRTL_SHUTDOWN_EVENT.  On the[ kubernetes documentation](https://kubernetes.io/docs/concepts/windows/intro/) I found this:

terminationGracePeriodSeconds - this is not fully implemented in Docker on Windows, see the [GitHub issue](https://github.com/moby/moby/issues/25982). The behavior today is that the ENTRYPOINT process is sent CTRL_SHUTDOWN_EVENT, then Windows waits 5 seconds by default, and finally shuts down all processes using the normal Windows shutdown behavior. The 5 second default is actually in the Windows registry [inside the container](https://github.com/moby/moby/issues/25982#issuecomment-426441183), so it can be overridden when the container is built.

I've found also this interesting article explaining windows shutdown process:

https://technoresult.com/windows-shutdown-process-behind-the-scean/

Which also says that at some point all running processes are shutdown with a grace period of 5 seconds with the same signal, the timeout for this is configurable as well.  

As you know, on Linux you can handle everything in a much easier way, but Windows is different. We can play with shutdown timeouts and implement some logic for handling the CTRL_SHUTDOWN_EVENT signal with our entrypoint, but I don't think it's worth it because we would have to send more signals to processes, which the windows shutting down process already does.

mumoshu: @amaldonadomat Good to know and totally understood. Thank you so much for sharing your knowledge and thoughts!

README.md
```diff
@@ -1744,6 +1745,115 @@ $ helm --upgrade install actions-runner-controller/actions-runner-controller \
...
+The main two steps in enabling Windows self-hosted runners are:
+
+- Using `nodeSelector`'s property to filter the `cert-manger` and `actions-runner-controller` pods
+- Deploying a RunnerDeployment using a Windows-based image
+
+For the first step, you need to set the `nodeSelector.kubernetes.io/os` property in both the `cert-manager` and the `actions-runner-controller` deployments to `linux` so that the pods for these two deployments are only scheduled in Linux nodes. You can do this as follows:
+
+```yaml
+nodeSelector:
+      kubernetes.io/os: linux
```
mumoshu, line 1759:
```suggestion
  kubernetes.io/os: linux
```

README.md
```diff
@@ -1744,6 +1745,115 @@ $ helm --upgrade install actions-runner-controller/actions-runner-controller \
...
+        kubernetes.io/arch: amd64
+      repository: <owner>/<repo>
+      labels:
+        - windows
+        - X64
+        - devops-managed
+```
+
+#### Dockerfile
+
```
mumoshu, line 1799:
```suggestion

> Note that you'd need to patch the below Dockerfile if you need a graceful termination.
> See https://github.com/actions-runner-controller/actions-runner-controller/pull/1608/files#r917319574 for more information.

```

mumoshu (approved):

mumoshu:
@ian-flores Thank you so much for your efforts and contribution ❤️

PR #1613, opened 2022-07-08
Fix typos in README

KTamas:
"emphemeral" -> "ephemeral"


mumoshu (approved):
Ah, good catch! Thanks for fixing ☺️

PR #1615, opened 2022-07-08
fix: annotate pod instead of label on UnregistrationFailureMessage

everpcpc:
The error message should go to annotation instead of pod label, since we only check annotations on autoscale:

https://github.com/actions-runner-controller/actions-runner-controller/blob/master/controllers/autoscaling.go#L338-L342

Then there is no need to truncate or normalize the error message.


mumoshu (approved):
@everpcpc Good catch! You're absolutely correct. Thank you so much for pointing it out and fixing it!

PR #1618, opened 2022-07-09
docs: remove runner kind

toast-gear:
The `Runner` kind is no longer to be used by end the end user and should be considered an internal implementation detail now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,6 +283,7 @@ We go into details about the differences between the 2 later, initially lets loo
 To launch a single self-hosted runner, you need to create a manifest file that includes a `RunnerDeployment` resource as follows. This example launches a self-hosted runner with name *example-runnerdeploy* for the *actions-runner-controller/actions-runner-controller* repository.
 
 ```yaml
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -270,52 +271,50 @@ Alternatively, you can install each controller stack into a unique namespace (re
 - The organization level
 - The enterprise level
 
-There are two ways to use this controller:
+Runners can be deployed as 1 of 2 abstractions: 
 
-- Manage runners one by one with `Runner`.
-- Manage a set of runners with `RunnerDeployment`.
+- A `RunnerDeployment` (based on k8s's `Deployments`)
```
mumoshu, line 276: Nit but- `RunnerDeployment`'s API and implementation doesn't exactly follow/inherit Deployment's.
Perhaps something like the below would better summarize it?

```suggestion
- A `RunnerDeployment` (similar to k8s's `Deployments`, based on `Pods`)
```

mumoshu (approved):
LGTM. Thanks for taking your  time 🙏

PR #1627, opened 2022-07-11
docs: add default label considerations

toast-gear:
+ removed the commented out file names as we don't talk about applying the manifests like we do early in the docs meaning the hyperthetical file names are pointless


mumoshu (approved):
LGTM. Thank you so much for your efforts!

PR #1631, opened 2022-07-12
Remove github-api-cache-duration flag and code

mumoshu:
This removes the flag and code for the legacy GitHub API cache. We already migrated to fully use the new HTTP cache based API cache functionality which had been added via #1127 and available since ARC 0.22.0. Since then, the legacy one had been no-op and therefore removing it is safe.

Ref #1412


toast-gear (approved):

mumoshu:
Thanks for reviewing ☺️

PR #1632, opened 2022-07-12
chart: Remove support for extensions/v1beta1 and networking.k8s.io/v1beta1

mumoshu:
`networking.k8s.io/v1` has been available since v1.19.

As of today, AWS EKS supports v1.19+ and Oracle Cloud supports v1.20+. GKE and AKS supports v1.21+. The upstream Kubernetes project maintains v1.22+.
So it should be safe to remove it now.


toast-gear (changes requested):
I noticed we reference `networking.k8s.io/v1/Ingress` in a few other places in this file

mumoshu:
@toast-gear Good catch! Fixed in 9148457

toast-gear (approved):

mumoshu:
Thanks for reviewing 🙏

PR #1638, opened 2022-07-12
Minor improvements for runner groups

fgalind1:
- Add group in runners columns
- Add constant for runner group and labels


mumoshu (approved):
LGTM. Thanks for the improvements!

Issue #1654, opened 2022-07-15
Print the ARC version number on startup


mumoshu:
This would make it unnecessary for me to ask anyone how they upgraded ARC or which ARC version they are running when providing user support.

ViktorLindgren95:
Just a simple const that we set manually each update that print at the start of init/main function call or is there an internal version handling that should be called instead?

This is low hanging enough for me to do 🤣

mumoshu:
This has been resolved via #1659.
Thank you so much for your contribution @ViktorLindgren95!!

Issue #1161, opened 2022-02-28
Add the ARC release version into the UserAgent


toast-gear:
Pulled from https://github.com/actions-runner-controller/actions-runner-controller/pull/1140

It would be helpful to include the release version in the user agent for telemetry purposes.

mumoshu:
This has also been addressed by #1659.
Thanks for your contribution @ViktorLindgren95!

PR #1659, opened 2022-07-19
Print Version Number on startup

ViktorLindgren95:
Based on [Issue 1654](https://github.com/actions-runner-controller/actions-runner-controller/issues/1654)

Built a function that will print the current ARC version by fetching the enviroment variable ARC_VERSION with a fallback incase the enviroment variable isnt set yet.

ARC_VERSION is set inside the dockerfile which gets its value from the [Publish ARC workflow](https://github.com/actions-runner-controller/actions-runner-controller/blob/79a494b2aaf11fcc95008dbdb54afb0efa664cd6/.github/workflows/publish-arc.yaml#L39)

Works on my local windows machine, Havent confirmed on either my linux machine or K8s cluster. I will in future contributions setup a local test enviroment


ViktorLindgren95:
Im not totally sure how the docker build and push works as i personally use buildah but it should be the same logic, as you set the enviroment variable on the runner machine and the docker daemon is hosted on the same "container" so it should inherit the enviorment variable. If not then maybe we can pass it.

main.go
```diff
@@ -87,7 +92,7 @@ func main() {
 
 		commonRunnerLabels commaSeparatedStringSlice
 	)
-
+	fmt.Println("ARC VERSION:", getVersion("ARC_VERSION", "VERSION NOT SET"))
```
toast-gear, line 95: Concept is fine, I'd rather this was output as part of the existing initilisation log output here https://github.com/actions-runner-controller/actions-runner-controller/blob/98cefd105611e8eb81b975ba42d7052e9a50941d/main.go#L220 with the key `version`.

toast-gear:
It's probably worth doing https://github.com/actions-runner-controller/actions-runner-controller/issues/1161 as part of this PR or in a follow up PR

ViktorLindgren95:
> 



> It's probably worth doing #1161 as part of this PR or in a follow up PR

I have a quick question then. Shall i put the get version function inside github.go.  I dont feel like it belongs there really, as its not really related to github per say. But it makes sense inheritance wise, that way we can have one function that both github.go calls internally and the same function that main.go calls, so we dont have two diffrent functions / code blocks. Could you make a  "design" call on this kindly

Dockerfile
```diff
@@ -5,7 +5,8 @@ WORKDIR /workspace
 
 # Make it runnable on a distroless image/without libc
 ENV CGO_ENABLED=0
-
+# Enviroment Variable used inside main to display current ARC version
+ENV ARC_VERSION=$VERSION
```
mumoshu, line 9: Can we propagate this via go-build ldflags so that we can actually embed the value into the binary rather than reading it from a runtime environment variable? It's more error-prone as you can't accidentally alter the version number at run time, and slightly more maintainable as I think it's [a well-known practice](https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications) to embed the version number.

More concretely, we would be able to create a directory named `build/` in the repository root and create a source file `build/version.go`.

The source would look like:

```
package build

// This is overridden at build-time using go-build ldflags
var Version = "N/A"
```

In each Dockerfile, you add `-ldflags` to every `go build` invocation like:

```
go build -ldflags="-X 'github.com/actions-runner-controller/actions-runner-controller/build.Version=v0.26.0'"
```

Finally, `getVersion` functions would use `build.Version` instead of `os.Getenv`.

WDYT?

github/github.go
```diff
@@ -134,8 +134,8 @@ func (c *Config) NewClient() (*Client, error) {
 			}
 		}
 	}
-
-	client.UserAgent = "actions-runner-controller"
+	version := getVersion("ARC_VERSION", "dev")
+	client.UserAgent = "actions-runner-controller:" + version + ""
```
mumoshu, line 138: Can we instead use the `product/version` pattern as I believe it's relatively more [standard-ish](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent)?

```suggestion
	client.UserAgent = "actions-runner-controller/" + version
```

mumoshu:
@ViktorLindgren95 Thanks a lot for your effort! The direction looks very good. Implementation-wise, I have two comments- would you mind reviewing those comments and, if that makes sense to you, make corresponding changes, so that we can merge this before 0.26.0? Thanks in advance for your help ☺️

github/github.go
ViktorLindgren95: Yeah that is fine i just choose the image:tag was because it was the first to come to mind

Dockerfile
ViktorLindgren95: That looks better! I will look into it and try to get a PR before the weekend . When is 0.26 expected merge ?

ViktorLindgren95:
> @ViktorLindgren95 Thanks a lot for your effort! The direction looks very good. Implementation-wise, I have two comments- would you mind reviewing those comments and, if that makes sense to you, make corresponding changes, so that we can merge this before 0.26.0? Thanks in advance for your help ☺️

No problem im just happy to do some coding in my free time 🥰

toast-gear (changes requested):
The default didn't seem to take when I tested (built using the Makefile `make docker-buildx`). Additionally, the Makefile needs adjusting to support passing in a version if desired.  We can probably move the default value from the Go code to the Dockerfile via an ARG with a default value. The project is designed to run from the image produced from its Dockerfile so I don't see an issue with doing that.

ViktorLindgren95:
> The default didn't seem to take when I tested (built using the Makefile `make docker-buildx`). Additionally, the Makefile needs adjusting to support passing in a version if desired (I still expected to see the verison as `dev` but instead I got `version": "",`). We can probably move the default value being set from the Go code to being set via an ARG in the Dockerfile with a default value if no arg is passed in. The project is designed to run from the image produced from its Dockerfile so I don't see an issue with doing that.

Im on the changes, ignore my re-request

ViktorLindgren95 (commented):
~~ignore~~

ViktorLindgren95:
> 



> The default didn't seem to take when I tested (built using the Makefile `make docker-buildx`). Additionally, the Makefile needs adjusting to support passing in a version if desired (I still expected to see the verison as `dev` but instead I got `version": "",`). We can probably move the default value being set from the Go code to being set via an ARG in the Dockerfile with a default value if no arg is passed in. The project is designed to run from the image produced from its Dockerfile so I don't see an issue with doing that.

Ok i think i know what happens. If the $VERSION variable is empty, aka not set, the version becomes empty instead of dev. I just confirmed this in my local enviroment.

Because i got the version from the github actions variable its not fetched from any local instance. If you set version tag inside your own enviroment do you see a version being printed or is it still empty?

I unfortunately dont know much about makefiles but, Would the makefile create it with the dockerfile? then try setting the version env in your enviroment and confirm for me

toast-gear:
You're going to want to use `CONTROLLER_VERSION ?= dev` so Make sets the var only to dev only if it hasn't been set by the user. Then pass [here](https://github.com/actions-runner-controller/actions-runner-controller/blob/master/Makefile#L119) that var like this:

```
	docker buildx build --platform ${PLATFORMS} \
		--build-arg RUNNER_VERSION=${RUNNER_VERSION} \
		--build-arg DOCKER_VERSION=${DOCKER_VERSION} \
		--build-arg CONTROLLER_VERSION=${CONTROLLER_VERSION} \
		-t "${NAME}:${VERSION}" \
		-f Dockerfile \
		. ${PUSH_ARG}
```

then have CONTROLLER_VERSION as an ARG in the Dockerfile with a default so it's set for those not using the Makefile `ARG CONTROLLER_VERSION=dev`

toast-gear:
mmmm looks like some changes are needed:

1. https://github.com/actions-runner-controller/actions-runner-controller/blob/master/Makefile#L7 needs to become `dev` instead of latest
2. https://github.com/actions-runner-controller/actions-runner-controller/blob/master/Makefile#L121 a --build-arg added for passing in VERSION to the Dockerfile
3. https://github.com/actions-runner-controller/actions-runner-controller/blob/master/Dockerfile#L28 `ARG VERSION=dev`
4. The default removed from the Go code

ViktorLindgren95:
> mmmm looks like some changes are needed:
> 
>     1. https://github.com/actions-runner-controller/actions-runner-controller/blob/master/Makefile#L7 needs to become `dev` instead of latest
> 
>     2. https://github.com/actions-runner-controller/actions-runner-controller/blob/master/Makefile#L121 a --build-arg added for passing in VERSION to the Dockerfile
> 
>     3. https://github.com/actions-runner-controller/actions-runner-controller/blob/master/Dockerfile#L28 `VERSION=dev` added to the ARG list with a default for those not using the Makefile
> 
>     4. The default removed from the Go code (build/version.go)
> 
> 
> EDIT @ViktorLindgren95 really appreciate your work so far, we're keen to get this in the current milestone v0.26.0 so if you do find some time to make those changes sooner rather than later we'd very much appreciate it 🙏

I will try to get a push before tommorow evening. When is 0.26 deadline?

Also thanks for the patience on your end. This is all pretty new to me, the contributing to opensource and all that ^^

toast-gear:
> I will try to get a push before tommorow evening. When is 0.26 deadline?

We don't have specific deadlines for releases as the maintainers have full time jobs. There's only this + a small refactor to do and then we are ready to release so we are hoping in the next week or 2.

mumoshu (dismissed):
LGTM. Thank you so much for your efforts and contribution @ViktorLindgren95!!

Dockerfile
```diff
@@ -25,7 +24,7 @@ RUN go mod download
 # With the above commmand,
 # TARGETOS can be "linux", TARGETARCH can be "amd64", "arm64", and "arm", TARGETVARIANT can be "v7".
 
-ARG TARGETPLATFORM TARGETOS TARGETARCH TARGETVARIANT
+ARG TARGETPLATFORM TARGETOS TARGETARCH TARGETVARIANT VERSION=dev
```
TingluoHuang, line 27: Do we need to also update the workflow yaml file to set the `VERSION` arg?
https://github.com/actions-runner-controller/actions-runner-controller/blob/master/.github/workflows/publish-arc.yaml

TingluoHuang: nvm... `build.Version=${VERSION}` will read from the ENV.

mumoshu: @TingluoHuang Ahhh, that's actually a good catch!
Although our go-build snippet refers to the VERSION env, I don't think we now pass that env as a build-arg for the docker build.
Probably we might need to add something like `build-args: NPM_TOKEN=${{env.VERSION}}` to the docker-build-push action step.

mumoshu: Addressed in [da14999](https://github.com/actions-runner-controller/actions-runner-controller/pull/1659/commits/da149994f51d794159cbb91be865002aff035883)

mumoshu:
Updated the `N/A` to `NA` so that it looks a bit more natural in the default User-Agent string `actions-runner-controller/N/A` vs `actions-runner-controller/NA`. I don't think `NA` as an abbrev for `Not-Available` is not so common but I think it's a bit better than breaking UA string 😁 

Also fixed the failing test related to the UA string generation in [a4b3019](https://github.com/actions-runner-controller/actions-runner-controller/pull/1659/commits/a4b30197e71accd177333cd89b02b8ef1d966ebb)

mumoshu (approved):

mumoshu:
Merged. The next release of our canary image which will be available within 10 minutes or so would show `canary` as the version number.
It would be great if anyone could confirm it by looking into the controller logs or seeing outgoing HTTP requests or etc.

PR #1688, opened 2022-08-02
feat: Set runner UID and docker GID to match github actions runner

suhasgaddam-trueaccord:
https://dev.to/jj/designing-containers-for-github-actions-5h84

Similar to https://github.com/actions-runner-controller/actions-runner-controller/pull/778#issue-983548504

Similar to https://github.com/aws/jsii/pull/3018

### DIFF
--- a/runner/actions-runner-dind-rootless.dockerfile
+++ b/runner/actions-runner-dind-rootless.dockerfile
@@ -14,6 +14,10 @@ ARG DEBUG=false
 RUN test -n "$TARGETPLATFORM" || (echo "TARGETPLATFORM must be set" && false)
 
 ENV DEBIAN_FRONTEND=noninteractive
+
+# Use 1001 for compatibility with GitHub-hosted runners
+ARG RUNNER_UID=1000
+
 RUN apt update -y \
     && apt-get install -y software-properties-common \
     && add-apt-repository -y ppa:git-core/ppa \
@@ -56,7 +60,7 @@ RUN apt update -y \
     && rm -rf /var/lib/apt/lists/*
 
 # Runner user
-RUN adduser --disabled-password --gecos "" --uid 1000 runner
+RUN adduser --disabled-password --gecos "" --uid $RUNNER_UID runner
 
 ENV HOME=/home/runner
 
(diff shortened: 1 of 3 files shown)


mumoshu:
@suhasgaddam-trueaccord Hey! Thanks for submitting a pull request.
Would you mind clarifying what are the examples of exact issues & problems which can be fixed or addressed by this change?
Was there any specific error encountered in some action when uid/gid doesn't match with that of the managed actions runner? 🤔

mumoshu:
@suhasgaddam-trueaccord Hey! Sorry for the long silence. I finally managed to pull some time to revisit this today.

I read https://github.com/actions-runner-controller/actions-runner-controller/pull/1688#issuecomment-120405220,1 several times and I'm now convinced that this should have been the default. Thanks a lot for the idea and for sharing the background. However, I can also see the point about some folks possibly relying on current `uid:gid` settings of `1000:1000`...

Probably a feasible way to move this forward would be to have another set of recommended runner images(probably with a unique suffix, like `summerwind/actions-runner-recommended:v$(RUNNER_VER)`) with `1001:121`, and encourage everyone to migrate to one of the recommended images ASAP, and eventually deprecate and remove the older images with 1000:1000.

Would you agree with that?

mumoshu:
@suhasgaddam-trueaccord I'm gonna change the UID/GID part to refer to build args/env so that we can specify the UID/GID used in the image at build-time. After that, I would either try (1)publishing additional images with the new UID/GID along with existing images or (2) just add some instructions in our docs to explain how to build your custom images using alternative UID/GID. WDYT?

toast-gear (approved):

toast-gear:
@suhasgaddam-trueaccord if you can rebase this we'll keep it at `1000` and `1001` for the 20.04 image and update them in the new 22.04 images

mumoshu:
@suhasgaddam-trueaccord I've rebased this but I couldn't force-push the result to your branch. Let me open another pull request and continue the work there. Thanks again for your contribution!

mumoshu:
Continued to #2077

PR #1691, opened 2022-08-03
Create QuickStartGuide.md

Vijay-train:
Creating a new Quickstart guide that captures simple onboarding instructions. The intent is for first time users to be able to follow this guide and get their environment running and try out ARC. A link to this guide would be added to the repo readme once this PR get merged.

### DIFF
--- a/QuickStartGuide.md
+++ b/QuickStartGuide.md
@@ -107,7 +107,7 @@ spec:
 
 Apply this file to your K8s cluster
 ```shell
-$ kubectl apply -f runnerdeployment.yaml
+kubectl apply -f runnerdeployment.yaml
 ````
  
 
--- a/QuickStartGuide.md
+++ b/QuickStartGuide.md
@@ -44,16 +44,6 @@ kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/
 
 <details><summary>Helm deployment</summary>
 
- ##### Create a new file called `custom-values.yaml` containing...
-
- ```yaml
-authSecret:
-  github_token: REPLACE_YOUR_TOKEN_HERE
-  create: true
-
-```
-<sub> *note:- Replace REPLACE_YOUR_TOKEN_HERE with your PAT that was generated in Step 1 </sub>
-
 ##### Add repository
 ```shell
 helm repo add actions-runner-controller https://actions-runner-controller.github.io/actions-runner-controller
@@ -61,10 +51,12 @@ helm repo add actions-runner-controller https://actions-runner-controller.github
 
 ##### Install Helm chart
 ```shell
-helm install -f custom-values.yaml --wait --namespace actions-runner-system \
-  --create-namespace actions-runner-controller \
-  actions-runner-controller/actions-runner-controller
+helm upgrade --install --namespace actions-runner-system --create-namespace\
+  --set=authSecret.create=true\
+  --set=authSecret.github_token="REPLACE_YOUR_TOKEN_HERE"\
+  --wait actions-runner-controller actions-runner-controller/actions-runner-controller
 ```
+<sub> *note:- Replace REPLACE_YOUR_TOKEN_HERE with your PAT that was generated in Step 1 </sub>
 </details>
 
 <details><summary>Kubectl deployment</summary>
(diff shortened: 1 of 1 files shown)


QuickStartGuide.md
```diff
@@ -0,0 +1,142 @@
...
+```yaml
+apiVersion: actions.summerwind.dev/v1alpha1
+kind: RunnerDeployment
+metadata:
+  name: example-runnerdeploy
+spec:
+  replicas: 1
+  template:
+    spec:
+      repository: mumoshu/actions-runner-controller-ci
```
mumoshu, line 104:
```suggestion
      repository: mumoshu/actions-runner-controller-ci
```

We might better provide an example repo that contains a simple github actions workflow, so that folks can fork it and use it to test RunnerDeployment. WDYT?

QuickStartGuide.md
```diff
@@ -0,0 +1,142 @@
...
+1️⃣ Deploy  and configure ARC on your K8s cluster. You may use Helm or Kubectl.
+
+
+<details><summary>Helm deployment</summary>
+
+ ##### Create a new file called `custom-values.yaml` containing...
+
+ ```yaml
+authSecret:
+  github_token: REPLACE_YOUR_TOKEN_HERE
```
mumoshu, line 51: Generally speaking, placing a sensitive value like this one in values.yaml isn't recommended.

Perhaps we'd better use a secret created via `kubectl create secret` like explained in the `Configure Personal Access Token` section? The helm chart uses the secret named `controller-manager` by default as the container for github API credentials so I believe we can just omit this part and move the `Configure Personal Access Token` section before `Install Helm Chart`.

Vijay-train: Thanks for the feedback. I have now fixed this.

QuickStartGuide.md
Vijay-train: Thanks for the comment. Yes, this would help. As discussed, we will add this post merge of this PR incrementally

Vijay-train (commented):
I have responded to the comments. Thanks.
- Resolved one issue by removing the need to put PAT in values.yaml file
- Postponed fixing the other comment about needing an example repo, Will add that in future

QuickStartGuide.md
```diff
@@ -0,0 +1,134 @@
+## Introduction
+
+GitHub Actions can be run in GitHub-hosted cloud or self hosted environments. Self-hosted runners offer more control of hardware, operating system, and software tools than GitHub-hosted runners provide
+
+With just a few steps, you can set up your kubernetes (K8s) cluster to be a self-hosted environment.
+In this guide, we will setup pre-requistes ,deploy Actions Runner controller (ARC) and then target that cluster to run GitHub Action workflows.
```
nebuk89, line 6: Typo 
prerequisite, deploy

QuickStartGuide.md
```diff
@@ -0,0 +1,134 @@
...
+```shell
+$ kubectl get runners
+NAME                             REPOSITORY                             STATUS
+example-runnerdeploy2475h595fr   mumoshu/actions-runner-controller-ci   Running
+
+$ kubectl get pods
+NAME                           READY   STATUS    RESTARTS   AGE
+example-runnerdeploy2475ht2qbr 2/2     Running   0          1m
+````
+Also,this runner has been registered directly to the specified repository, you can see it in repository settings. For more information , see [settings](https://docs.github.com/en/actions/hosting-your-own-runners/monitoring-and-troubleshooting-self-hosted-runners#checking-the-status-of-a-self-hosted-runner)
```
nebuk89, line 122: Typo
Also, this

nebuk89 (dismissed):
Couple of minor grammar typos otherwise all tidy :D they are minor enough to just ship for now as this is awesome 🎉

QuickStartGuide.md
Vijay-train: Thanks for the comment, fixed in the latest commit.

QuickStartGuide.md
Vijay-train: Thanks for the comment, fixed in the latest commit.

mumoshu (approved):
LGTM 👍

Even though I think this kind of documentation is very hard to maintain by a few folks in general (I guess any getting started guide for a K8s controller can become a getting started guide for K8s which is complex and whose setup greatly varies across environments), I believe this has a lot of potentials to help both users and maintainers by ensuring everyone has the same foundational base-line knowledge on this solution.

Thanks a lot for your contribution!

// P.S. Shall we add a pointer to this doc in our bug report form, that looks like "If it's your first time to use K8s and ARC, have you already followed the quick-start guide"?

Issue #1607, opened 2022-07-05
Rationalise supported webhook events to just `workflow_job` removing `check_run`, `pull_request` and `push` events


toast-gear:
Currently ARC supports the following webhook events:

1. workflow_job
2. check_run
3. pull_request
4. push

Events 2 - 4 were introduced before the `workflow_job` event was created by GitHub and added to ARC. The ARC maintainers believe that there is no really need at a technical level any event type other than `workflow_job` for the ARC project nor is there any community demand for any other event types.

We are planning on **removing support** for all events other than `workflow_job`.

If you have a legitimate use case used in the world today for the other event types please do comment on this issue.

Currently having these unneeded events results in end user confusion and additional support overhead for the maintainers.

**Note** we are to keep the contract
```yaml
  scaleUpTriggers:
  - githubEvent:
      workflowJob: {}
    duration: "30m"
```

mumoshu:
@raunyoliveira-hotmart Do you really need to restrict a `workflow_job` event trigger scaling only on a specific repository?

To be clear, a `repository` runner is by definition tied to a specific repository and I thought ARC is able to scale the repository-specific runners on a `workflow_job` event just fine without explicit additional filters, as ARC already reads the repository contained within the `workflow_job` event and match that to an appropriate HorizontalRunnerAutoscaler+RunnerDeployment of the said repository.

raunyoliveira-hotmart:
hey @mumoshu, thanks for answering!

just to make sure I understood correct, instead of having the `HorizontalRunnerAutoscaler` being restrict by specific repositories:

```
apiVersion: actions.summerwind.dev/v1alpha1
kind: HorizontalRunnerAutoscaler
metadata:
  name: <name>
spec:
  minReplicas: <min>
  maxReplicas: <max>
  scaleTargetRef:
    name: <targetref>
  scaleUpTriggers:
  - githubEvent:
      checkRun:
        types: ["created"]
        status: "queued"
        repositories:
        - Org/repo1
        - Org/repo2
    amount: 1
```

I should configure an HRA and a RunnerDeployment in the following way:

```
kind: HorizontalRunnerAutoscaler
metadata:
  name: <name>
spec:
  minReplicas: <min>
  maxReplicas: <max>
  scaleTargetRef:
    name: <targetref>
  scaleUpTriggers:
  - githubEvent:
      workflowJob: {}
    duration: "30m"
```

and on the `RunnerDeployment` I filter by the repository:

```
[...]

PR #1709, opened 2022-08-10
docs: webhook scaling

toast-gear:
Ref https://github.com/actions-runner-controller/actions-runner-controller/issues/1607

We are rationalising down to just supporting the `workflow_job` event and so we are updating the docs to only talk about this event before removing the code for the other events.

Contains a few other minor doc clean up changes


README.md
```diff
@@ -690,44 +683,64 @@ spec:
...
----
+1. The new runner gets allocated the job and processes it
+2. Upon the job ending GitHub sends another `workflow_job` event to ARC but with `status=completed`
+3. The HRA removes a capacity reservation from its `capacityReservations` and picks a runner to terminate ensuring it isn't busy via the GitHub API beforehand
+
+If the job is cancelled before it is allocated to a runner or the runner is never used due to other runners matching needed runner group and required runner labels are allocated the job then the lifecycle looks like this:
 
-The primary benefit of autoscaling on Webhooks compared to the pull driven scaling is that it is far quicker as it allows you to immediately add runner resources rather than waiting for the next sync period.
+1. The new runner is retained until its `EffectiveTime` passes
+3. The HRA removes a capacity reservation from its `capacityReservations` and terminates the expired runner ensuring it isn't busy via the GitHub API beforehand
```
mumoshu, line 738:
```suggestion
1. The HRA removes a capacity reservation from its `capacityReservations` and terminates the expired runner ensuring it isn't busy via the GitHub API beforehand
```

`EffectiveTime` is used to prevent creating a redundant runner, in case a runner pod has successfully finished running a job and completed with exist status 0. So I'd say it isn't involved in this scenario.

`EffectiveTime` works as follows:

There is a race between the timing that the runner pod finishes running and the timing that GitHub sends a `status=completed` `workflow_job` event.

When the former happened before the latter(which is almost 100% cases), ARC used to end up with recreating a runner pod because RunnerDeployment specs > available runner pod replicas.

With EffectiveTime and saving the EffectiveTime that ARC saw when it ran the last reconciliation on the RunnerDeployment, ARC can prevent this pod recreation because it now knows that it already created all the desired replicas after seeing that EffectiveTime.

README.md
```diff
@@ -144,16 +144,17 @@ _Note: Links are provided further down to create an app for your logged in user
...
 
 **Subscribe to events**
 
-At this point you have a choice of configuring a webhook, a webhook is needed if you are going to use [webhook driven scaling](#webhook-driven-scaling). The webhook can be configured centrally in the GitHub app itself or separately. In either case the event details are:
+At this point you have a choice of configuring a webhook, a webhook is needed if you are going to use [webhook driven scaling](#webhook-driven-scaling). The webhook can be configured centrally in the GitHub app itself or separately. In either case you need to subscribe to the same events:
 
-* Check run (required for all webhook driven scaling events)
-* Workflow job (optionally) (required for [webhook driven scaling with workflow_job events](https://github.com/actions-runner-controller/actions-runner-controller#example-1-scale-on-each-workflow_job-event)
+* Check run
+* Workflow job
```
mumoshu, line 157:
```suggestion
* Workflow job
* Check run
```

I think we'd generally recommend `Workflow job` first so that users won't try `check run` and experience various gotchas of it.

Or, perhaps we'd better remove the reference to `check run` at all, assuming all the GHES in the world has already migrated to versions that supports `workflow_job` events? 😁

README.md
```diff
@@ -483,10 +482,6 @@ You can also read the design and usage documentation written in the original pul
 
 Under the hood, `RunnerSet` relies on Kubernetes's `StatefulSet` and Mutating Webhook. A `statefulset` is used to create a number of pods that has stable names and dynamically provisioned persistent volumes, so that each `statefulset-managed` pod gets the same persistent volume even after restarting. A mutating webhook is used to dynamically inject a runner's "registration token" which is used to call GitHub's "Create Runner" API.
 
-**Limitations**
-
-* For autoscaling the `RunnerSet` kind only supports pull driven scaling or the `workflow_job` event for webhook driven scaling.
-
```
mumoshu, line 489: Ah, good catch! Yes, this limitation is obsolete now.

README.md
```diff
@@ -690,44 +683,63 @@ spec:
...
+        workflowJob: {}
+      duration: "30m"
 ```
 
-With the above example, the webhook server scales `example-runners` by `1` replica for 5 minutes on each `check_run` event with the type of `created` and the status of `queued` received.
+The lifecycle of a runner provisioned from a webhook is different to a runner provisioned from the pull based scaling method:
 
-Of note is the `HRA.spec.scaleUpTriggers[].duration` attribute. This attribute is used to calculate if the replica number added via the trigger is expired or not. On each reconciliation loop, the controller sums up all the non-expiring replica numbers from previous scale-up triggers. It then compares the summed desired replica number against the current replica number. If the summed desired replica number > the current number then it means the replica count needs to scale up.
+1. GitHub sends a `workflow_job` event to ARC with `status=queued`
+2. ARC finds HRAs with a `workflow_job` webhook scale trigger that and back a RunnerDeployment / RunnerSet with matching runner labels
```
mumoshu, line 725:
```suggestion
2. ARC finds an HRA with a `workflow_job` webhook scale trigger that backs a RunnerDeployment / RunnerSet with matching runner labels
```

Nit but perhaps we should be more explicit about that a `workflow_job` will ever result in scaling only one RunnerDeployment / RunnerSet, by using a singular noun?

README.md
```diff
@@ -690,44 +683,63 @@ spec:
...
+3. The matched HRAs adds a unit to its `capacityReservations` list
```
mumoshu, line 726:
```suggestion
2. ARC finds an HRA with a `workflow_job` webhook scale trigger that backs a RunnerDeployment / RunnerSet with matching runner labels
3. The matched HRA adds a unit to its `capacityReservations` list
```

Nit but perhaps we should be more explicit about that a `workflow_job` will ever result in scaling only one RunnerDeployment / RunnerSet, by using a singular noun?

README.md
```diff
@@ -690,44 +683,63 @@ spec:
...
+4. ARC adds a replica and sets the EffectiveTime of that replica to current + `HRA.spec.scaleUpTriggers[].duration`
```
mumoshu, line 727:
```suggestion
4. ARC adds a replica and sets the EffectiveTime of the whole RunnerDeployment to the current time
```

EffectiveTime is used to ensure that the replicas are added only once per each the RunnerDeployment controller detected the changes in the desired `replicas`. Probably this is a bit more correct explanation

README.md
```diff
@@ -690,44 +683,63 @@ spec:
...
 
-As mentioned previously, the `scaleDownDelaySecondsAfterScaleOut` property has the final say still. If the latest scale-up time + the anti-flapping duration is later than the current time, it doesn’t immediately scale down and instead retries the calculation again later to see if it needs to scale yet.
+At this point there are a few things that can happen, either the job gets allocated to the runner or the runner is left dangling due to it not being used, if the runner gets assigned the job that triggered the scale up the lifecycle looks like this:
 
----
+1. The new runner gets allocated the job and processes it
+2. Upon the job ending GitHub sends another `workflow_job` event to ARC but with `status=completed`
+3. The HRA removes a capacity reservation from its `capacityReservations` and picks a runner to terminate ensuring it isn't busy via the GitHub API beforehand
```
mumoshu, line 733:
```suggestion
3. The HRA removes the oldest capacity reservation from its `capacityReservations` and picks a runner to terminate ensuring it isn't busy via the GitHub API beforehand
```

so that it won't end up in scaling in/removing unnecessarily more runners.

README.md
```diff
@@ -690,44 +683,63 @@ spec:
...
+
+If the job is cancelled before it is allocated to a runner or the runner is never used due to other runners matching needed runner group and required runner labels are allocated the job then the lifecycle looks like this:
```
mumoshu, line 735:
```suggestion
If the job is cancelled before it is allocated to a runner then the lifecycle looks like this:

1. Upon the job cancellation GitHub sends another `workflow_job` event to ARC but with `status=cancelled`
2. The HRA removes the oldest capacity reservation from its `capacityReservations` and picks a runner to terminate ensuring it isn't busy via the GitHub API beforehand

If runner is never used due to other runners matching needed runner group and required runner labels are allocated the job then the lifecycle looks like this:

1. The scale trigger duration specified via `HRA.spec.scaleUpTriggers[].duration` elapses
2. The HRA thinks the capacity reservation is expired, removes it from HRA's `capacityReservations` and terminates the expired runner ensuring it isn't busy via the GitHub API beforehand
```

README.md
```diff
@@ -690,44 +683,63 @@ spec:
...
 
-The primary benefit of autoscaling on Webhooks compared to the pull driven scaling is that it is far quicker as it allows you to immediately add runner resources rather than waiting for the next sync period.
+1. The HRA removes a capacity reservation from its `capacityReservations` and terminates the expired runner ensuring it isn't busy via the GitHub API beforehand
 
-> You can learn the implementation details in [#282](https://github.com/actions-runner-controller/actions-runner-controller/pull/282)
+Your `HRA.spec.scaleUpTriggers[].duration` value should be set long enough to account for 2 things:
+
+1. the potential amount of time it could take for a pod to become `Running` e.g. you need to scale horizontally because there isn't a node avaliable 
+2. the amount of time it takes for GitHub to allocate a job to that runner
```
mumoshu, line 742:
```suggestion
1. the potential amount of time it could take for a pod to become `Running` e.g. you need to scale horizontally because there isn't a node avaliable 
2. the amount of time it takes for GitHub to allocate a job to that runner
3. the amount of time it takes for the runner to notice the allocated job and starts running it
```

mumoshu: 3 is also important because, without it, ARC may remove the runner before the runner starts running the job and becomes busy.

README.md
mumoshu: Perhaps this might better reflect the implementation as a job cancellation can trigger an immediate scale down

mumoshu (approved):
LGTM. Thanks a lot for your effort!

PR #1712, opened 2022-08-11
feat: add queue size limit to github webhook server helm template

Sajadorouji:
We recently hit the queue size limit that is 100 by default and as i find out there is no way to pass the value through the helm chart

Here is a fix to enable users to add their queue size limit in case of need


toast-gear (changes requested):
LGTM beyond the docs need updating charts/actions-runner-controller/readme.md

Sajadorouji:
Hey thanks for the review
`Readme.md` file has been updated with the correct name and value

charts/actions-runner-controller/values.yaml
```diff
@@ -270,3 +270,5 @@ githubWebhookServer:
     enabled: false
     # minAvailable: 1
     # maxUnavailable: 3
+  ## Queue limit size override, by default its 100
+  # queueLimit: 1000
```
debugger24, line 274: Above `charts/actions-runner-controller/README.md` says default value set in `values.yaml` is `100` but we are not setting any value.

Lets either set the default value to 100 here or update `README.md` above to blank.

debugger24: Should we refactor to move this default value out of `horizontal_runner_autoscaler_webhook.go` to `values.yaml`? Moving this default config to `values.yaml` will give users more clarity. WDYT?

mumoshu: Good point!

What I had in my mind was that one can still deploy ARC by writing their own K8s manifest yaml or their own Helm chart . The hard-coded default value can be handy and user-friendly in that case.
Perhaps we can just make it in sync with the coded default like:

```suggestion
  # queueLimit: 100
```

and add some code comment in horizontal_runner_autoscaler_webhook.go to let the programmer know that they need to change the default value included in the default values.yaml along with the hard-coded default?

WDYT?

debugger24: > one can still deploy ARC by writing their own K8s manifest yaml or their own Helm chart

Got it, ya in that case lets keep that default in our code and update the README.md

Sajadorouji:
@debugger24 @mumoshu Thanks for the review and comments
I've commit your suggestion on the helm value, do we need to bump the chart version as well or not

debugger24 (approved):

toast-gear (changes requested):
Apologies I've not been on this, work is busy.

I am a believer of convention over configuration. We historically have not put the default in the values.yaml docs if it's set in the app itself. Perhaps that was mistake but I'd prefer for us to be consistent so could we remove the stated default from the docs as it's inconsistent with the rest of the doc as well as `## Queue limit size override, by default its 100` as this is the same thing. I'm fine with the commented out `queueLimit: 100`

toast-gear (approved):
LGTM

mumoshu:
Thanks a lot for your contribution! @Sajadorouji ☺️
Regarding the comment-out of the default value in values.yaml, I think that's probably the way to go for us, as it should result in less issues when e.g. we updated the default value in Go code but forgot to do so in values.yaml.

However, we do have some inconsistencies in other values within the same file, like, we probably want to comment out the default `syncPeriod` setting as well for the same reason and for consistency.

PR #1740, opened 2022-08-22
Change `type:` to `kind:`

apanzerj:
Per the CRD spec here https://github.com/actions-runner-controller/actions-runner-controller/blob/master/charts/actions-runner-controller/crds/actions.summerwind.dev_horizontalrunnerautoscalers.yaml#L115-L127

It's `kind:` not `type:`


mumoshu (approved):
Good catch!! Thanks a lot for the fixes ☺️

PR #1744, opened 2022-08-23
ci: include sha in canary version

toast-gear:
Given the canary tag is mutable it might be handy to include the SHA in the version so it's easier to be sure what canary code you're running.


mumoshu (approved):
LGTM. Thanks a lot for the improvement!

PR #1765, opened 2022-08-31
Use numeric USER for nonroot:nonroot in Dockerfile

funkypenguin:
Hey guys,

While tightening our security posture, I noticed the following error when deploying ARC:

```
  Warning  Failed     10s (x4 over 24s)  kubelet            Error: container has runAsNonRoot and image has non-numeric user
(nonroot), cannot verify user is non-root (pod: "actions-runner-controller-github-webhook-server-c68bcb8c4-
x7qqp_actions-runner-controller(545601de-f9bd-460e-b706-23ffd3f3c7da)", container: github-webhook-server)
```

This PR simply solves the above by updating the Dockerfile to use the numeric UID of the `nonroot` user supplied by the `distroless:nonroot` image :)

Cheers!
D

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,6 +48,6 @@ WORKDIR /
 COPY --from=builder /out/manager .
 COPY --from=builder /out/github-webhook-server .
 
-USER nonroot:nonroot
+USER 65532:65532
 
 ENTRYPOINT ["/manager"]


mumoshu:
@funkypenguin Hey! Thanks a lot for your contribution. I'm still trying to understand when and how this is useful. So, does this error occur only when you configured `runAsNonRoot: true` without `runAsUser: $YOUR_UID`? The benefit of this modification is we now omit `runAsUser` even when you specified `runAsRoot: true`, right?

funkypenguin:
Hey @mumoshu!

Huh. TIL. you're right, I didn't know that Kubernetes would run the pod if you explicitly set `runAsUser` / `runAsGroup`. So this works:

```
  securityContext:
    capabilities:
      drop:
      - ALL  
    seccompProfile:
      type: RuntimeDefault   
    runAsNonRoot: true 
    runAsUser: 65532
    runAsGroup: 65532
    allowPrivilegeEscalation: false 
```

But this doesn't:

```
  securityContext:
    capabilities:
      drop:
      - ALL  
    seccompProfile:
      type: RuntimeDefault   
    runAsNonRoot: true 
    allowPrivilegeEscalation: false 
```

So yes, it turns out that the only benefit is really that the security-conscious user doesn't have to explicitly set `runAsUser` :)

mumoshu (approved):
@funkypenguin Perfect! Thanks a lot for your confirmation and contribution. This LGTM 👍

PR #1725, opened 2022-08-16
Fix/multitenancy enterprise url

Jalmeida1994:
Fix #1714. Tested it in an Enterprise Server environment. Followed @mumoshu's suggestion on how to fix this issue.

Issue #1714, opened 2022-08-12
Multitenancy: New RunnerDeployment not using "githubEnterpriseServerURL" value


Jalmeida1994:
### Controller Version

canary

### Helm Chart Version

0.20.2 (main branch)

### CertManager Version

1.9

### Deployment Method

Helm

### cert-manager installation

Followed every procedure.

### Checks

- [X] This isn't a question or user support case (For Q&A and community support, go to [Discussions](https://github.com/actions-runner-controller/actions-runner-controller/discussions). It might also be a good idea to contract with any of contributors and maintainers if your business is so critical and therefore you need priority support
- [X] I've read [releasenotes](https://github.com/actions-runner-controller/actions-runner-controller/tree/master/docs/releasenotes) before submitting this issue and I'm sure it's not due to any recently-introduced backward-incompatible changes
- [X] My actions-runner-controller version (v0.x.y) does support the feature
- [X] I've already upgraded ARC (including the CRDs, see charts/actions-runner-controller/docs/UPGRADING.md for details) to the latest and it didn't fix the issue

### Resource Definitions

```yaml
---
# Source: secret.yaml
apiVersion: v1
kind: Secret
metadata:
  name: organization-github-app-secret  
  namespace: organization
data:
  github_app_id: ***
  github_app_installation_id: ***
  github_app_private_key: ***
---
# Source: hra.yaml
apiVersion: actions.summerwind.dev/v1alpha1
kind: HorizontalRunnerAutoscaler
metadata:
  name: organization-autoscaler
  namespace: organization
spec:
  githubAPICredentialsFrom:
    secretRef: 
      name: organization-github-app-secret
  scaleDownDelaySecondsAfterScaleOut: 300
  scaleTargetRef:
    name: organization-runners
  minReplicas: 0
  maxReplicas: 5
  scaleUpTriggers:
  - githubEvent: {}
    amount: 1
    duration: 30m
---
# Source: runner_deployment.yaml
apiVersion: actions.summerwind.dev/v1alpha1
kind: RunnerDeployment
metadata:
  name: organization-runners
  namespace: organization
spec:
  template:
    metadata:
    spec:
      githubAPICredentialsFrom:
        secretRef: 
          name: organization-github-app-secret 
      ephemeral: true
      labels:
        - organization-runners
      organization: organization
      containers:
        - name: runner
          imagePullPolicy: IfNotPresent
```

### To Reproduce

```markdown
[...]

PR #1781, opened 2022-09-06
Fix bug with enterpriseURL for multi-tenancy

### LINKED ISSUE #1764: Multi-tenancy doesn't work with Github Apps with no explicit Enterprise URL defined.
### Controller Version

623c84f

### Helm Chart Version

0.20.2

### CertManager Version

1.2.0

### Deployment Method

Helm

### cert-manager installation

We have installed cert-manager from an official helm chart. Anyhow, certificates aren't really the issue here. 

### Checks

- [X] This isn't a question or user support case (For Q&A and community support, go to [Discussions](https://github.com/actions-runner-controller/actions-runner-controller/discussions). It might also be a good idea to contract with any of contributors and maintainers if your business is so critical and therefore you need priority support
- [X] I've read [releasenotes](https://github.com/actions-runner-controller/actions-runner-controller/tree/master/docs/releasenotes) before submitting this issue and I'm sure it's not due to any recently-introduced backward-incompatible changes

### Resource Definitions

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: sre-controller
  namespace: nolimits
data:
  github_app_id: ***
  github_app_installation_id: ***
  github_app_private_key: ***
---
kind: RunnerDeployment
metadata:
  name: sre-runner
  namespace: nolimits
spec:
  template:
    spec:
      image: summerwind/actions-runner-dind
      organization: XXX
      labels:
        - k8s
      githubAPICredentialsFrom:
        secretRef:
          name: sre-controller
```

### To Reproduce

```markdown
1. Create a k8s secret containing the Github App data. Ensure that this secret is different from the one being used as the `controller-manager` secret.
2. Create a runnerdeployment and reference the secret using the `GithubAPICredentialsFrom` field.
```

### Describe the bug

The controller is not able to spawn runner pods for `RunnerDeployment` Objects (as well as HRAs) with `GithubAPICredentialsFrom` defined. 

It fails due to the error defined here:  https://gist.github.com/bm1216/090d015d1fee081857e5f6add01ce936.

 This doesn't seem to be a problem with the provided app credentials, as copying and renaming the `controller-manager` secret and providing it via `GithubAPICredentialsFrom` also leads to the same error. I can see that its picking up the  secret in the `GitHubAPICredentialsFrom` field as the installation id in the logs matches the secret. 

Upon further investigation, the issue seems to be with the URL its using. its using a `https://github.com/api/v3` URL as opposed to using `https://api.github.com` for the token calls. This is odd, considering I haven't describe an enterprise URL at any point during the setup

The controller should be able to spawn runner pods based on the Custom Resource provided. 

### Controller Logs

```shell
https://gist.github.com/bm1216/090d015d1fee081857e5f6add01ce936
```

### Runner Pod Logs

```shell
The runner pod is not able to spawn.
```

### Additional Context

I understand that the Multi-tenancy feature isn't officially released yet. However, we really like the feature and are looking forward to using it ASAP. Following this comment - https://github.com/actions-runner-controller/actions-runner-controller/pull/1371#issuecomment-1210057401, I tried to use the latest image and ran into this issue. Any help would be appreciated. I'm thinking there might be a difference between the default client created by the controller and the client created using the GithubAPICredentialsRef secret.

Thanks for all the work you are doing - we really appreciate it.

bm1216:
Addresses and fixes #1764.

### Changes:
- Add `IsEnterprise` bool to the GitHub client struct. It is set when the client is an enterprise client.
- Check if the main client is an enterprise client before setting `conf.EnterpriseURL` for clients created using `githubAPICredentialsFrom`. 
- Add command line argument for setting `EnterpriseURL` in GitHub config.

### Reasoning
Currently, ARC creates a GitHub client (let's call it client A) when the controller is created - this client is passed to the `RunnerReconciler`. As part of the multi-tenancy feature, a new client (let's call it client B) is created when a `RunnerDeployment` is added with the field `githubAPICredentialsFrom`. 

The GitHub wrapper creates a client from a `config` struct. When creating client B, it sets the `enterpriseURL` in its config to the `GithubBaseURL` of client A<sup>1</sup>. This causes issues when client A is not an enterprise client, as client B now has its `enterpriseURL` defined. The GitHub wrapper behaves differently when  an `enterpriseURL` defined in the `config` struct, leading to the newly created client B calling the wrong API endpoint for tasks such as runner registration.

To fix this, client A needs a basic idea of whether it is using an `enterpriseURL`. We have decided to use a bool (as opposed to doing checks on the baseURL) as it's just cleaner and less likely to go wrong.

Also, added command line support for `enterpriseURL` as it was the only thing that was missing.

<sup>1</sup> This was introduced as part of #1725 and issue #1714. The concern was valid because if the config for client B has no `enterpriseURL` defined it should be using the enterpriseURL of the original client (A). Nevertheless, the above bug was introduced.


bm1216:
@Jalmeida1994 for visibility. I tested it out and it should still work for your use case.

bm1216:
@mumoshu @toast-gear thoughts?

mumoshu (approved):
Thank you very much for diagnosing and fixing it! I love your detailed explanation of the issue and the fix too-it helped me quickly understand and confirm this is the right fix!

Definitely LGTM 👍

Jalmeida1994:
> @Jalmeida1994 for visibility. I tested it out and it should still work for your use case.

Hey @bm1216 nice work! When I get the time I'll test it also on an Enterprise Environment but the code and the reasoning makes sense! Cheers, and sorry again for introducing this bug @mumoshu @toast-gear and @bm1216!

bm1216:
Thanks @mumoshu @Jalmeida1994 for taking a look. Appreciate all the comments

PR #1794, opened 2022-09-09
Add golangci-lilnt to CI

cory-miller:
This introduces a linter to PRs to help with code reviews and code hygiene. I've also gone ahead and fixed (or ignored) the existing lints.

I've only setup the default linters right now. There are many more options that are documented at https://golangci-lint.run/.

The GitHub Action should add appropriate annotations to the lint job for the PR. Contributors can also lint locally using `make lint`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -155,3 +155,7 @@ GINKGO_FOCUS='[It] should create a new Runner resource from the specified templa
 #### Helm Version Bumps
 
 In general we ask you not to bump the version in your PR, the maintainers in general manage the publishing of a new chart.
+
+#### Running linters locally
+
+To run the `golangci-lint` tool locally, we recommend you use `make lint` to run the tool using a Docker container matching the CI version.
--- a/controllers/horizontal_runner_autoscaler_batch_scale.go
+++ b/controllers/horizontal_runner_autoscaler_batch_scale.go
@@ -79,6 +79,7 @@ func (s *batchScaler) Add(st *ScaleTarget) {
 				for {
 					select {
 					case <-after:
+						// TODO: What was the intention here? Close the channel?
 						after = nil
 						break batch
 					case st := <-s.queue:
--- a/controllers/horizontal_runner_autoscaler_batch_scale.go
+++ b/controllers/horizontal_runner_autoscaler_batch_scale.go
@@ -79,8 +79,6 @@ func (s *batchScaler) Add(st *ScaleTarget) {
 				for {
 					select {
 					case <-after:
-						// TODO: What was the intention here? Close the channel?
-						after = nil
 						break batch
 					case st := <-s.queue:
 						nsName := types.NamespacedName{
--- a/controllers/runner_controller.go
+++ b/controllers/runner_controller.go
@@ -715,7 +715,7 @@ func runnerHookEnvs(pod *corev1.Pod) ([]corev1.EnvVar, error) {
 				},
 			},
 		},
-		corev1.EnvVar{
+		{
 			Name:  "ACTIONS_RUNNER_REQUIRE_SAME_NODE",
 			Value: strconv.FormatBool(isRequireSameNode),
 		},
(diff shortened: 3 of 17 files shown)


controllers/horizontal_runner_autoscaler_batch_scale.go
```diff
@@ -79,6 +79,7 @@ func (s *batchScaler) Add(st *ScaleTarget) {
 				for {
 					select {
 					case <-after:
+						// TODO: What was the intention here? Close the channel?
```
cory-miller, line 82: Can you help me with this one?

```
controllers/horizontal_runner_autoscaler_batch_scale.go:83:7: ineffectual assignment to after (ineffassign)
                                                after = nil
```

controllers/runner_controller.go
```diff
@@ -834,6 +834,7 @@ func newRunnerPodWithContainerMode(containerMode string, template corev1.Pod, ru
 		if dockerdContainer != nil {
 			template.Spec.Containers = append(template.Spec.Containers[:dockerdContainerIndex], template.Spec.Containers[dockerdContainerIndex+1:]...)
 		}
+		// TODO: What was the intention here?
```
cory-miller, line 837: Can you help me with this one?

```
controllers/runner_controller.go:838:6: SA4000: identical expressions on the left and right side of the '<' operator (staticcheck)
                if runnerContainerIndex < runnerContainerIndex {
```

controllers/runner_controller.go
```diff
@@ -834,6 +834,7 @@ func newRunnerPodWithContainerMode(containerMode string, template corev1.Pod, ru
...
 		if runnerContainerIndex < runnerContainerIndex {
 			runnerContainerIndex--
 		}
```
mumoshu, line 840: @cory-miller Re https://github.com/actions-runner-controller/actions-runner-controller/pull/1794/files#r967413200, it has been there since it's inception in https://github.com/actions-runner-controller/actions-runner-controller/commit/0386c0734cb6fd6ca689ffeb0e76b9ca9cf9512d#diff-f4c239205beda2ef5d9d041809cd93cb32958b6b658fb0b2331c85b853777465R697-R699. It's been unreachable code since then so the first option is to remove this if block entirely.

A better option would be though, to update the code to something like:

```suggestion
		if dockerdContainer != nil {
			template.Spec.Containers = append(template.Spec.Containers[dockerdContainerIndex+1:], template.Spec.Containers[:dockerdContainerIndex]...)
		}
		if dockerdContainerIndex < runnerContainerIndex {
			runnerContainerIndex--
		}
```

I imagine the intention here would be to remove the dockerd sidecar is you're going to use kubernetes container mode(as they are incompatible). If you removed the dockerd sidecar from the containers slice, the runner container's index may change, hence my suggestion.

controllers/runner_controller.go
mumoshu: Good catch! See https://github.com/actions-runner-controller/actions-runner-controller/pull/1794/files#r974803081

controllers/horizontal_runner_autoscaler_batch_scale.go
mumoshu: Ah good catch. This was emitted probably because `after` will never be read after it has been updated to `nil` due to that we call `break batch` immediately after the assignment.
I believe you can just remove it.

CONTRIBUTING.md
```diff
@@ -155,3 +155,7 @@ GINKGO_FOCUS='[It] should create a new Runner resource from the specified templa
 #### Helm Version Bumps
 
 In general we ask you not to bump the version in your PR, the maintainers in general manage the publishing of a new chart.
+
+#### Running linters locally
+
+To run the `golangci-lint` tool locally, we recommend you use `make lint` to run the tool using a Docker container matching the CI version.
```
mumoshu, line 161: 👍

controllers/runner_controller.go
```diff
@@ -832,9 +832,9 @@ func newRunnerPodWithContainerMode(containerMode string, template corev1.Pod, ru
 
 	if containerMode == "kubernetes" {
 		if dockerdContainer != nil {
-			template.Spec.Containers = append(template.Spec.Containers[:dockerdContainerIndex], template.Spec.Containers[dockerdContainerIndex+1:]...)
+			template.Spec.Containers = append(template.Spec.Containers[dockerdContainerIndex+1:], template.Spec.Containers[:dockerdContainerIndex]...)
```
mumoshu, line 835:
```suggestion
			template.Spec.Containers = append(template.Spec.Containers[:dockerdContainerIndex], template.Spec.Containers[dockerdContainerIndex+1:]...)
```

My fault- this was fine. Reversing the append args result in unnecessarily reversing the order of slice elements. Let me revert it.

mumoshu (approved):
LGTM. Thanks a lot for your contribution @cory-miller!!

Issue #1551, opened 2022-06-18
More runner and workflow metrics


ColinHeathman:
**Is your feature request related to a problem? Please describe.**

As an ARC operator, I want to be able to monitor and alert on the status of the runners I manage. It would be convenient if these metrics were bundled as a part of the ARC deployment instead of using another 3rd part adapter

A few things that would be interesting & actionable:

#### runner status & availability
- runner startup times
- runner status (idle, available, offline)

#### workflow status
- queued workflows
- in-progress workflows
- run times
- pass/fail stats

**Describe the solution you'd like**

ARC is well-positioned to aggregate and expose the data for these metrics.

Keeping up-to-date state for running Workflows would probably require a webhook, since the status can change faster than polling can keep up.

eg.
```
# HELP github_workflow_runtime Duration all workflow runs
# TYPE github_workflow_runtime histogram
github_workflow_runtime_bucket{conclusion="success", workflow_id="12345678",le="0.1"} 5
github_workflow_runtime_bucket{conclusion="success", workflow_id="12345678",le="1"} 5
github_workflow_runtime_bucket{conclusion="success", workflow_id="12345678",le="10"} 5
...
github_workflow_runtime_bucket{conclusion="success", workflow_id="12345678",le="+Inf"} 5
github_workflow_runtime_sum{conclusion="success", workflow_id="12345678"} 1000
github_workflow_runtime_count{conclusion="success", workflow_id="12345678"} 5

# HELP github_workflow_conclusion Conclusions of all github workflow
# TYPE github_workflow_conclusion counter
github_workflow_conclusion{conclusion="success", workflow_id="12345678"} 5

# HELP github_workflow_status Live status information
# TYPE github_workflow_status gauge
github_workflow_status{status="in_progress", workflow_id="12345678"} 5
github_workflow_status{status="queued", workflow_id="12345678"} 5

# HELP github_runner_status Live runner status information
# TYPE github_runner_status gauge
github_runner_status{status="idle", runner_id="4321"} 5
github_runner_status{status="active", runner_id="4321"} 5
github_runner_status{status="offline", runner_id="4321"} 5

# HELP github_runner_startup Duration all runner startups
# TYPE github_runner_startup histogram
github_runner_startup_bucket{runner_id="4321",le="0.1"} 5
github_runner_startup_bucket{runner_id="4321",le="1"} 5
github_runner_startup_bucket{runner_id="4321",le="10"} 5
...
[...]

PR #1814, opened 2022-09-19
Add workflow job metrics to Github webhook server

ColinHeathman:
Relates to issue https://github.com/actions-runner-controller/actions-runner-controller/issues/1551

This PR adds these metrics:
```
github_workflow_job_queue_duration_seconds
github_workflow_job_run_duration_seconds
github_workflow_job_conclusions_total
github_workflow_jobs_queued_total
github_workflow_jobs_started_total
github_workflow_jobs_completed_total
github_workflow_job_failures_total
```

### DIFF
--- a/cmd/githubwebhookserver/metrics/metrics.go
+++ b/cmd/githubwebhookserver/metrics/metrics.go
@@ -1,0 +1,126 @@
+// Package metrics provides monitoring of the GitHub related metrics.
+//
+// This depends on the metrics exporter of kubebuilder.
+// See https://book.kubebuilder.io/reference/metrics.html for details.
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+func init() {
+	metrics.Registry.MustRegister(
+		githubWorkflowJobQueueDurationSeconds,
+		githubWorkflowJobRunDurationSeconds,
+		githubWorkflowJobConclusionsTotal,
+		githubWorkflowJobsQueuedTotal,
+		githubWorkflowJobsStartedTotal,
+		githubWorkflowJobsCompletedTotal,
+		githubWorkflowJobFailuresTotal,
+	)
+}
+
+var (
+	runtimeBuckets []float64 = []float64{
+		0.01,
+		0.05,
+		0.1,
+		0.5,
+		1,
+		2,
+		3,
+		4,
+		5,
+		6,
+		7,
+		8,
+		9,
+		10,
+		12,
+		15,
+		18,
+		20,
+		25,
+		30,
+		40,
+		50,
+		60,
+		70,
+		80,
+		90,
+		100,
+		110,
+		120,
+		150,
+		180,
+		210,
+		240,
+		300,
+		360,
+		420,
+		480,
+		540,
+		600,
+		900,
+		1200,
+		1800,
+		2400,
+		3000,
+		3600,
+	}
+)
+
+var (
+	githubWorkflowJobQueueDurationSeconds = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "github_workflow_job_queue_duration_seconds",
+			Help:    "Queue times for workflow jobs in seconds",
+			Buckets: runtimeBuckets,
+		},
+		[]string{"runs_on", "job_name"},
+	)
+	githubWorkflowJobRunDurationSeconds = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "github_workflow_job_run_duration_seconds",
+			Help:    "Run times for workflow jobs in seconds",
+			Buckets: runtimeBuckets,
+		},
+		[]string{"runs_on", "job_name", "job_conclusion"},
+	)
+	githubWorkflowJobConclusionsTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "github_workflow_job_conclusions_total",
+			Help: "Conclusions for tracked workflow jobs",
+		},
+		[]string{"runs_on", "job_name", "job_conclusion"},
+	)
+	githubWorkflowJobsQueuedTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "github_workflow_jobs_queued_total",
+			Help: "Total count of workflow jobs queued (events where job_status=queued)",
+		},
+		[]string{"runs_on", "job_name"},
+	)
+	githubWorkflowJobsStartedTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "github_workflow_jobs_started_total",
+			Help: "Total count of workflow jobs started (events where job_status=in_progress)",
+		},
+		[]string{"runs_on", "job_name"},
+	)
+	githubWorkflowJobsCompletedTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "github_workflow_jobs_completed_total",
+			Help: "Total count of workflow jobs completed (events where job_status=completed)",
+		},
+		[]string{"runs_on", "job_name"},
+	)
+	githubWorkflowJobFailuresTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "github_workflow_job_failures_total",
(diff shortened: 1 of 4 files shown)


ColinHeathman:
I have tested this locally with a basic installation. I am in the process of testing this change in an environment with some live traffic, so the PR will remain a draft until that's completed.

Nitpicks are welcome in the meantime

cmd/githubwebhookserver/metrics/workflow_job_metrics_event_reader.go
```diff
@@ -0,0 +1,133 @@
...
+	switch action := e.GetAction(); action {
+	case "queued":
+		incrementStatusGauge("queued", labels)
+
+		reader.trackedJobs[job_id] = e
+
+	case "in_progress":
+
+		// discard events if the previous event wasn't tracked
+		previousEvent, ok := reader.trackedJobs[job_id]
```
mumoshu, line 74: Does this imply that this feature will work only when you have only one replica of the githubwebhookserver, but not for two or more replicas? (It's a standard webhook server so it is expected to be scaled out to 2 or more replicas whenever the operator decides to do so.

ColinHeathman: Yes, that's what I had in mind here. If that's not an assumption I can make, then there needs to be some way to share state, or `github_workflow_job_status` should be dropped

mumoshu: @ColinHeathman We may probably introduce an optional dependency to an external distributed kv store to track jobs later. Or GitHub could enhance their API and/or the workflow job event payload so that we can use some information included in the event to measure the queue duration. In the meantime though, we should omit this as this isn't going to work universally

ColinHeathman: I've started a discussion [here](https://github.com/orgs/community/discussions/33640) to enhance the API. I believe that is the cleanest path forwards.

@mumoshu Is there another pending reason to add a kv store? I thought about it but I didn't think these metrics would justify the addition.

Alternatively, we could change `github_workflow_job_status` to a counter, rather than a gauge.
That way queue counts could be calculated by Prometheus, with `github_workflow_job_status{job_status="queued"}[1m] - github_workflow_job_status{job_status="in_progress"}[1m]`

I can't think of a way to make `github_workflow_job_queue` work without a kv store though

mumoshu: Thanks for starting the discussion! Let's see how it goes.

> s there another pending reason to add a kv store?

No- this is the first time I thought we might need to add it.

> I thought about it but I didn't think these metrics would justify the addition.

Same- if this isn't a short-term hard requirement, I'd like not to add the kv store as an ARC dependency. Looks too much.

> we could change github_workflow_job_status to a counter, rather than a gauge.

Ah, counts might be good as well! Can we name the metric like `github_workflow_jobs_total(job_status="...")` though? If it's a count, I think that better follows the canonical metric name convention of Prometheus https://prometheus.io/docs/practices/naming/#metric-names

ColinHeathman:
@mumoshu Please take a look at this new implementation.

This version is stateless. I was able to find the information I need for the metrics from the logs for the jobs. This uses the GH client to download those logs and parse them, so there is a bit of a downside there, but I think having this be stateless is much better.

mumoshu:
@ColinHeathman Hey! Thanks a lot for your effort. Sorry to say, but this approach isn't going to scale and that's actually why we moved from the pull-based autoscaling to webhook-based autoscaling.
GitHub API rate limit can be quite insufficient when you're going to run hundreds of jobs per hour. AFAIK it's 1000/hour with PAT, and 15000/hour with GitHub App organizational installation.
You added an API call per every workflow_job event in the last commit. Assuming we might receive 3 workflow job events for queued, in_progress, and completed, this approach might make ARC not scale more than 333 jobs/hour, which is quite low. We do call some APIs in other places for e.g. runner graceful shutdown. So the actual limit would be lower.

mumoshu:
@ColinHeathman IMHO, making it stateful alone is fine, as long as the state is allowed to get lost. That said, I thought we'd be happy with your alternative mentioned in https://github.com/actions-runner-controller/actions-runner-controller/pull/1814#discussion_r975891494, which is to make those counters so that you can still count events per status, while you can also aggregate counts at query time using e.g. Prometheus, which makes it possible to approximately count events per status without introducing a global state to ARC.

mumoshu:
That said, your awesome `fetchAndParseWorkflowJobLogs` and the related code might better fit within another controller/command within ARC(name it e.g. actions-metrics-server) or another application dedicated to monitoring workflow jobs from the information provided by GitHub Actions API, not from ARC.

ColinHeathman:
> @ColinHeathman IMHO, making it stateful alone is fine, as long as the state is allowed to get lost. That said, I thought we'd be happy with your alternative mentioned in https://github.com/actions-runner-controller/actions-runner-controller/pull/1814#discussion_r975891494, which is to make those counters so that you can still count events per status, while you can also aggregate counts at query time using e.g. Prometheus, which makes it possible to approximately count events per status without introducing a global state to ARC. WDYT?

@mumoshu I did make the change to counters here (the gauges are gone) that's well and good.

Runtimes can be measured off of the `action=completed` event, but it's less accurate than the logs. I think losing that accuracy is fine though.

I currently don't see a way to get Queue times without either comparing events or making an API call for the logs. I think maybe that needs to be a separate issue or PR since it's going to be more difficult.

ColinHeathman:
What do you think of the metric names? I want to be sure that they are correct

cmd/githubwebhookserver/metrics/metrics.go
```diff
@@ -0,0 +1,126 @@
...
+			Name:    "github_workflow_job_run_duration_seconds",
+			Help:    "Run times for workflow jobs in seconds",
+			Buckets: runtimeBuckets,
+		},
+		[]string{"runs_on", "job_name", "job_conclusion"},
+	)
+	githubWorkflowJobConclusionsTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "github_workflow_job_conclusions_total",
+			Help: "Conclusions for tracked workflow jobs",
```
mumoshu, line 94:
```suggestion
			Help: "Total count of conclusions for tracked workflow jobs",
```

Probably make it consistent with other metric names would be nicer

mumoshu:
@ColinHeathman Thanks for your confirmation. The metrics names look great to me!

mumoshu:
> Runtimes can be measured off of the action=completed event, but it's less accurate than the logs. I think losing that accuracy is fine though.

I'm fine with that too. We'd better add some comment in code to let us keep remembering that the run_duration metric count is not ging to be the same as workflow_job_completions_total when you have two or more ARC webhook-server pods.

> I currently don't see a way to get Queue times without either comparing events or making an API call for the logs. I think maybe that needs to be a separate issue or PR since it's going to be more difficult.

I agree with you! If it's going to be something like what I've explained in https://github.com/actions-runner-controller/actions-runner-controller/pull/1814#issuecomment-1266264170, that looks good to me.

mumoshu:
@ColinHeathman Hey! Is this ready for review?

ColinHeathman:
Hey @mumoshu, this does work in my cluster. We at BenchSci are currently using this as it is, since we have GH Enterprise and don't need to worry so much about the API rate limit.

I was planning on moving this to the dedicated controller/command/applilcation as you suggested in https://github.com/actions-runner-controller/actions-runner-controller/pull/1814#issuecomment-1266264170, but I haven't had the time to do the rework. I don't think I'll be able to get to it until next month.

Feel free to make changes :) I can try to help anyway I can

mumoshu:
@ColinHeathman Hey! I finally had some time to work on it, and #2057's what I intended. Would you mind taking a look when you have some time?

mumoshu:
@ColinHeathman Hey! I'm closing this in favor of #2057. Thank you again for doing 99% of the work in this pull request! I've featured you in our release note to make your awesome work more visible ❤️ 
https://github.com/actions-runner-controller/actions-runner-controller/pull/2068

mumoshu:
#2057 will be included in the next release of ARC, 0.27.0. Please feel free to give it a shot and report back any issues and ideas if any 🙏

PR #1820, opened 2022-09-20
Create app-version-mapping.md

megamorf:
I had to upgrade an old ARC setup from 0.12.0 today and the two different versions in the GH tags and releases were really confusing. To help others like me, we should add a mapping of the two different versions that are being maintained in the repo.


mumoshu (approved):
Great! Thanks a lot for your contribution.

BTW, I was wondering if we'd better add a similar matrix for k8s versions x ARC versions.
Would you mind if I enhanced this mapping to also include the supported K8s versions in the right-most column?

mumoshu:
@megamorf Thanks for your confirmation. No need to push your branch as I intended to do that in another PR. Sorry for being unclear!

PR #1837, opened 2022-09-22
Encourage folks to write up a proper title for a bug report

mumoshu:
Honestly, I'm a bit tired of seeing issues filed with the default title "Bug" that are seemingly not related to real bugs!
I'm emptying it so that the reporter is more encouraged to write up a sentence to describe the problem and they hopefully notice along the way that there are places to diagnose before considering it as a bug.


toast-gear (approved):

mumoshu:
CleanShot 2022-10-04 at 20 26 25@2x

OMG I wanted GitHub to tell this earlier 😂

PR #1839, opened 2022-09-23
Let it be a bug only when it's reproducible with official runner image

mumoshu:
A custom runner image can break runners and ARC in interesting ways. Probably it's better to clearly state that ARC is not guaranteed to work with every custom runner image in the wild.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -11,7 +11,7 @@ body:
     options:
     - label: I've already read https://github.com/actions-runner-controller/actions-runner-controller/blob/master/TROUBLESHOOTING.md and I'm sure my issue is not covered in the troubleshooting guide.
       required: true
-    - label: I'm not using a custom runner image
+    - label: I'm not using a custom entrypoint in my runner image
       required: true
 - type: input
   id: controller-version
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -7,9 +7,11 @@ body:
   id: read-troubleshooting-guide
   attributes:
     label: Checks
-    description: Please check the boxes below before submitting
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -7,10 +7,12 @@ body:
   id: read-troubleshooting-guide
   attributes:
     label: Checks
-    description: Please check the boxes below before submitting
+    description: Please check all the boxes below before submitting
     options:
     - label: I've already read https://github.com/actions-runner-controller/actions-runner-controller/blob/master/TROUBLESHOOTING.md and I'm sure my issue is not covered in the troubleshooting guide.
       required: true
+    - label: I'm not using a custom runner image
```
toast-gear, line 14: I think this needs wording slightly differently, we expect images built on our base to work fine and I can picture some people thinking custom images built from our images (as per the guide) are "customer" runner images which isn't really what we're trying to capture here right?

mumoshu: I'm skeptical if everyone can be 100% sure if their "customizations" added on top of the base image won't break anything that ARC expects. Perhaps it's easier for us if we communicate this way. On the other hand, perhaps it's nicer if we could capture folks who make only simple customizations like adding pkgs and commands to the base image.
Do you have any idea about what would be a good sentence to capture only the latter?

mumoshu: @toast-gear I missed your edit, sorry! So..

> I'm not using a custom entrypoint in my runner image?

This sounds good!

Can we also add another field to let the user list up all the modifications made to the custom image, which is required when this option is checked?

mumoshu:
```suggestion
    - label: I'm not using a custom entrypoint in my runner image
```

toast-gear (approved):

mumoshu:
Thanks for your review! Merging.

PR #1843, opened 2022-09-23
Add MTU propagation docker-shim also to rootless dind runner images

mumoshu:
#1201 added MTU propagation docker-shim to the standard runner and dind runner.
This adds the same feature to rootless dind runners, by enhancing and adding the docker-shim to the rootless dind runner dockerfile.
I also added a QoL enhancement to the dind runner startup.sh so that it becomes easier to spot an issue like https://github.com/actions-runner-controller/actions-runner-controller/pull/1201#issuecomment-1255901580.

Other two commits are for updating the E2E testdata to the one that I used for testing the feature, and adding a lot of comments to the DockerMTU implementation so that you can understand how it works along with all the MTU-related settings present in ARC.

### DIFF
--- a/runner/docker-shim.sh
+++ b/runner/docker-shim.sh
@@ -2,11 +2,16 @@
 
 set -Eeuo pipefail
 
+DOCKER=/usr/bin/docker
+if [ ! -e $DOCKER ]; then
+  DOCKER=$HOME/bin/docker
+fi
+
 if [[ ${ARC_DOCKER_MTU_PROPAGATION:-false} == true ]] &&
   (($# >= 2)) && [[ $1 == network && $2 == create ]] &&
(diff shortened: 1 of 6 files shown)


runner/docker-shim.sh
```diff
@@ -2,11 +2,16 @@
 
 set -Eeuo pipefail
 
+DOCKER=/usr/bin/docker
+if [ ! -e $DOCKER ]; then
+  DOCKER=$HOME/bin/docker
+fi
```
toast-gear, line 8: could this be simplified with `DOCKER=$(which docker)`?

mumoshu: Maybe not, because it would resolve to `/usr/local/bin/docker` (otherwise this shim won't work) which results in an infinite loop.

toast-gear (approved):

mumoshu:
Thanks for reviewing!

PR #1850, opened 2022-09-24
e2e: Bump runner version to 2.297.0

mumoshu:
Otherwise, every runner brought up by the E2E test needs to auto-update itself before running any jobs, which results in more time until the jobs are run and the E2E test finishes.


toast-gear (approved):

mumoshu:
Thanks!

PR #1759, opened 2022-08-27
Fix runners to do their best to gracefully stop on pod eviction

mumoshu:
Please see the updated README for more details.

~~Perhaps this may already work with RunnerDeployment-managed rootless-dind runners too. I'll give it a shot and update the PR title and README if it works.~~ Worked. This works for rootless-dind pods managed by either RunnerDeployment or RunnerSet.

UPDATE: Added support for rootful dind runners too.
UPDATE: Added support for runners with dind sidecars.

Since this change, the runner which received SIGTERM while still running a job will show log like the below:

```
√ Connected to GitHub

2022-08-28 05:18:11.927  NOTICE --- Executing actions-runner-controller's SIGTERM handler.
2022-08-28 05:18:11.928  NOTICE --- Note that if this takes more time than terminationGracePeriodSeconds, the runner will be forcefully terminated by Kubernetes, which may result in the in-progress workflow job, if any, to fail.
2022-08-28 05:18:11.930  NOTICE --- Ensuring dockerd is still running.
CONTAINER ID   IMAGE     COMMAND   CREATED   STATUS    PORTS     NAMES

# Runner removal

Current runner version: '2.296.0'
2022-08-28 05:18:12Z: Listening for Jobs
Failed: Removing runner from the server
Runner "org-runnerdeploy-tnshq-2k5p2" is still running a job"
2022-08-28 05:18:14.788  NOTICE --- Waiting for RUNNER_GRACEFUL_STOP_TIMEOUT=100 seconds until the runner agent to stop by itself.
2022-08-28 05:18:15Z: Running job: test2
2022-08-28 05:18:55Z: Job test2 completed with result: Succeeded
√ Removed .credentials
√ Removed .runner
Runner listener exit with 0 return code, stop the service, no retry needed.
Exiting runner...
2022-08-28 05:18:57.502  NOTICE --- The runner agent stopped before RUNNER_GRACEFUL_STOP_TIMEOUT=100
2022-08-28 05:18:57.506  NOTICE --- The actions runner process exited.
2022-08-28 05:18:57.508  NOTICE --- Holding on until runner init (pid 16) exits, so that there will hopefully be no zombie processes remaining.
2022-08-28 05:18:57.510  NOTICE --- Graceful stop completed.
2022-08-28 05:18:57.512  NOTICE --- Runner init exited. Exiting this process with code 0 so that the container and the pod is GC'ed Kubernetes soon.
```

[...]

PR #1853, opened 2022-09-25
Add workflow for validating runner scripts with shellcheck

mumoshu:
I am about to revisit #1517, #1454, #1561, and #1560 as a part of our on-going effort to a major enhancement to the runner entrypoints being made in #1759.

This change is a counterpart of #1852. #1852 enables you to easily run shellcheck locally. This enables you to automatically run shellcheck on every pull request.

Currently, any shellcheck error does not result in failing the workflow job. Once we addressed all the shellcheck findings, we can flip the fail_on_error option to true and let jobs start failing on pull requests that introduces invalid or suspicious bash code. The plan is to flip it after merging #1854 and #1759, because all the findings should have been addressed then.


toast-gear (commented):
Is it worth just bolting this on as another job in the existing validate-runners workflow instead of having another workflwo file? The existing validate runners workflow already runs as it checks for a broader set of paths so it's running either way.

mumoshu:
@toast-gear Thanks for the feedback! I intended to limit the scope of the workflow so that the shellcheck is run only when needed. Bolting it on the existing workflow can result in running spellcheck also on md file changes, which didn't make sense to me. That's a small difference anyway so I don't have a strong opinion. What's your preference after seeing my intention?

mumoshu:
@toast-gear Thanks! I fully agree with you now. I've added the last commit accordingly. Also, see that `update-status` is now covered in the `pattern` arg of the shellcheck job step.

toast-gear (approved):

PR #1856, opened 2022-09-25
Fix inability to configure MTU for rootless dind runner

mumoshu:
Follow-up for https://github.com/actions-runner-controller/actions-runner-controller/pull/1644

Also see #1201 for how to configure MTU correctly in general.

### DIFF
--- a/runner/rootless-startup.sh
+++ b/runner/rootless-startup.sh
@@ -11,8 +11,8 @@ fi
 
 if [ -n "${MTU}" ]; then
 jq ".\"mtu\" = ${MTU}" /home/runner/.config/docker/daemon.json > /tmp/.daemon.json && mv /tmp/.daemon.json /home/runner/.config/docker/daemon.json
-# See https://docs.docker.com/engine/security/rootless/
-echo "environment=DOCKERD_ROOTLESS_ROOTLESSKIT_MTU=${MTU}" >> /etc/supervisor/conf.d/dockerd.conf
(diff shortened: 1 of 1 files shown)


runner/rootless-startup.sh
```diff
@@ -11,8 +11,8 @@ fi
 
 if [ -n "${MTU}" ]; then
 jq ".\"mtu\" = ${MTU}" /home/runner/.config/docker/daemon.json > /tmp/.daemon.json && mv /tmp/.daemon.json /home/runner/.config/docker/daemon.json
-# See https://docs.docker.com/engine/security/rootless/
-echo "environment=DOCKERD_ROOTLESS_ROOTLESSKIT_MTU=${MTU}" >> /etc/supervisor/conf.d/dockerd.conf
+# See https://docs.docker.com/engine/security/rootless/ and https://github.com/docker/engine/blob/8955d8da8951695a98eb7e15bead19d402c6eb27/contrib/dockerd-rootless.sh#L13
+echo "DOCKERD_ROOTLESS_ROOTLESSKIT_MTU=${MTU}" >> /etc/environment
```
mumoshu, line 15: - We aren't using supervisord in rootless dind runners so this needs to go into another config file than the supervisor dockerd.conf.
- We're going to use /etc/environment because its seems get loaded by bash globally. The syntax for the environment file is different from the docker.conf so we omit `environment=`

toast-gear (approved):

PR #1859, opened 2022-09-26
Added `DEBIAN_FRONTEND=noninteractive` to `sudo`

Fleshgrinder:
By default `sudo` drops all environment variables and executes its commands with a clean environment. This is by design, but for the `DEBIAN_FRONTEND` environment variable it is not what we want, since it results in installers being interactive. This adds the `env_keep` instruction to `/etc/sudoers` to keep `DEBIAN_FRONTEND` with its `noninteractive` value, and thus pass it on to commands that care about it. Note that this makes no difference in our builds, because we are running them directly as `root`. However, for users of our image this is going to make a difference, since they start out as `runner` and have to use `sudo`.

### DIFF
--- a/runner/actions-runner-dind.dockerfile
+++ b/runner/actions-runner-dind.dockerfile
@@ -54,7 +54,8 @@ RUN adduser --disabled-password --gecos "" --uid 1000 runner \
     && groupadd docker \
     && usermod -aG sudo runner \
     && usermod -aG docker runner \
-    && echo "%sudo   ALL=(ALL:ALL) NOPASSWD:ALL" > /etc/sudoers
+    && echo "%sudo   ALL=(ALL:ALL) NOPASSWD:ALL" > /etc/sudoers \
+    && echo "Defaults env_keep += \"DEBIAN_FRONTEND\"" >> /etc/sudoers
 
 # arch command on OS X reports "i386" for Intel CPUs regardless of bitness
 # Docker download supports arm64 as aarch64 & amd64 / i386 as x86_64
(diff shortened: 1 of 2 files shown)


mumoshu:
@Fleshgrinder Thanks for your contribution!
Just a quick question, but how can I test this?
Should I kubectl-exec into one of my runner pods' runner containers and try running `sudo apt-get install whatever` and see if it skips the prompt now?
Do you have any other tools in your mind that this change might be useful for?

Fleshgrinder:
`dpkg` (the program underlying `apt` and `apt-get`) is by far the most prominent candidate for `DEBIAN_FRONTEND`, yes. You need to perform an installation that would usually ask questions to see the effect of it. Many users execute with `--yes`, which already hides various dialogs (and is dangerous). I do not remember any installation routine that kicks in with `--yes` and no `DEBIAN_FRONTEND` from the top of my head. Maybe something with locales or time zones. The one were we ran into this problem is https://github.com/ilikenwf/apt-fast, which is also used by GitHub in their VMs. You could try this one if you want.

```bash
sudo add-apt-repository ppa:apt-fast/stable
sudo apt-get update
sudo apt-get install apt-fast
```

That said, to see that it is working `sudo env | grep -F DEBIAN_FRONTEND` is already sufficient.

mumoshu:
@Fleshgrinder Thanks a lot for your explanation! Got it. I'll test it and plan when to merge this. We have a lot of ongoing pull requests waiting for reviews now.

DrJume:
Meanwhile you can add

```Dockerfile
RUN sudo bash -c "echo DEBIAN_FRONTEND=noninteractive >> /etc/environment"
```

to set it globally (also available in sudo commands).

mumoshu (approved):
Manually verified this to work on dind sidecar and rootful dind runner. Note that I didn't add support for rootless dind runner as we aren't generally going to allow sudos from workflow job steps on those runners.

So LGTM. Thanks a lot for your contribution @Fleshgrinder!

PR #1862, opened 2022-09-27
e2e: Make docker build timeout longer

mumoshu:
Depending on the host machine spec and load, it can take more time than 5 minutes. This change adhocly increases the timeout by 1.5x to address that.


toast-gear (approved):

mumoshu:
Thanks a lot for your reviews!

PR #1880, opened 2022-09-29
Use `-trimpath` and ldflags `-s -w` build flags

pratikbin:
Use build flags `-trimpath` and ldflags `-s -w` to generate slim binaries, More [here](https://pkg.go.dev/cmd/link) and [here](https://pkg.go.dev/cmd/go)

### Before

image

### After (~48% downsize)
image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,14 +30,14 @@ ARG TARGETPLATFORM TARGETOS TARGETARCH TARGETVARIANT VERSION=dev
 # to avoid https://github.com/moby/buildkit/issues/2334
 # We can use docker layer cache so the build is fast enogh anyway
 # We also use per-platform GOCACHE for the same reason.
-env GOCACHE /build/${TARGETPLATFORM}/root/.cache/go-build
+ENV GOCACHE /build/${TARGETPLATFORM}/root/.cache/go-build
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -30,14 +30,14 @@ ARG TARGETPLATFORM TARGETOS TARGETARCH TARGETVARIANT VERSION=dev
 # to avoid https://github.com/moby/buildkit/issues/2334
 # We can use docker layer cache so the build is fast enogh anyway
 # We also use per-platform GOCACHE for the same reason.
-env GOCACHE /build/${TARGETPLATFORM}/root/.cache/go-build
```
mumoshu, line 33: OMG! How did it ever work with lower-case?!

mumoshu (approved):
LGTM. Thanks a lot for your effort @pratikbin!

pratikbin:
Yay

image

Dockerfile
pratikbin: Docker tries to interpret the first word as instruction unless it's bounded in between here doc

mumoshu: Good to know! Thanks a lot for the explanation ☺️

PR #1910, opened 2022-10-12
Modify startup scripts to support setting docker default cidr range.

claudio-vellage:
Currently it's not easily possible to change the docker network CIDR range. This can lead to issues if there are networks with overlapping CIDR ranges in the same network, i.e. reachable by the same network interface. To avoid connectivity issues, it's possible to change the default CIDR range as follows:

```
{
  "default-address-pools": [{"base":"172.17.0.0/16","size":24}]
}
```

I've parametrized it and added it to the startup scripts, similar to how it's possible to set the MTU and registry mirrors. I would highly appreciate a quick review, let me know if I can assist in any way. This will allow for a much simpler runner setup.

### DIFF
--- a/runner/rootless-startup.sh
+++ b/runner/rootless-startup.sh
@@ -19,8 +19,8 @@ if [ -n "${DOCKER_REGISTRY_MIRROR}" ]; then
 jq ".\"registry-mirrors\"[0] = \"${DOCKER_REGISTRY_MIRROR}\"" /home/runner/.config/docker/daemon.json > /tmp/.daemon.json && mv /tmp/.daemon.json /home/runner/.config/docker/daemon.json
 fi
 
-if [ -n "${DOCKER_DEFAULT_NETWORK_CIDR_RANGE}" ]; then
-jq ".\"default-address-pools\" = {\"base\": \"${DOCKER_DEFAULT_NETWORK_CIDR_RANGE}\", \"size\": 24}" /etc/docker/daemon.json > /tmp/.daemon.json && /tmp/.daemon.json /etc/docker/daemon.json
+if [ -n "${DOCKER_DEFAULT_ADDRESS_POOL_BASE}" ] && [ -n "${DOCKER_DEFAULT_ADDRESS_POOL_SIZE}" ]; then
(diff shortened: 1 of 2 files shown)


runner/rootless-startup.sh
```diff
@@ -18,6 +18,10 @@ fi
 if [ -n "${DOCKER_REGISTRY_MIRROR}" ]; then
 jq ".\"registry-mirrors\"[0] = \"${DOCKER_REGISTRY_MIRROR}\"" /home/runner/.config/docker/daemon.json > /tmp/.daemon.json && mv /tmp/.daemon.json /home/runner/.config/docker/daemon.json
 fi
+
+if [ -n "${DOCKER_DEFAULT_NETWORK_CIDR_RANGE}" ]; then
```
mumoshu, line 22: Hey! Thanks a lot for your contribution.
A quick comment on your awesome work, sorry for being nit-picky but, is that R in CIDR stands for Range already?
Then, can we instead name this after the setting key path, so that it would look like `DOCKER_DEFAULT_ADDRESS_POOL_BASE` paired with `DOCKER_DEFAULT_ADDRESS_POOL_SIZE`?
Naming things that way will also enable us to make those consistent with the existing variable name like `DOCKER_REGISTRY_MIRROR`. WDYT?

claudio-vellage: It makes a lot of sense. I didn't think too much about the naming (too focused on resolving the problem I had) and will change it.

mumoshu: Oops! Sorry about that. I somehow misunderstood that for years 😂 Thank you so much for pointing it out! You saved me.

claudio-vellage:
Ive added the changes, but see the shellcheck is failing, when running locally, I don't see any errors, I assume the test itself is broken.

toast-gear:
@claudio-vellage shellcheck is fixed now, could you integrate master into your branch please so the check can pass

claudio-vellage:
@toast-gear I've rebased and pushed

claudio-vellage:
@mumoshu @toast-gear all checks seem to have passed, would be great if you can do a final review and merge.

claudio-vellage:
@mumoshu @toast-gear can we get this merged, it's exhausting to keep rebasing all the time. It would be appreciated if you can do it from your side and just get those changes merged. It's just a few lines of codes, you might as well close this PR and just add the code yourself. It's blocking one of our CI/CD pipelines and I don't want to have to use a forked version of the code.

claudio-vellage:
I'm opening a new PR, because all this rebasing makes it too complicated.

mumoshu:
@claudio-vellage Sorry for the delay and thank you for your patience. We've finally merged the almost 2months old PR that fixes runner reliability, which resulted in a huge conflict.
We will review and merge all the open PRs in the coming days, and I'll ensure yours is covered.

PR #1913, opened 2022-10-12
chore: add new label to bug form

toast-gear:
We no longer support any webhook other than the workflow job event and the other events will be removed next release https://github.com/actions-runner-controller/actions-runner-controller/issues/1607. As a result we require everyone to move to this event in order to get any support.


mumoshu (approved):
LGTM. Thanks!

PR #1917, opened 2022-10-13
Revamp the contribution guide

cory-miller:
I have added an architecture decision record (ADR) template. This is a process we use to document decisions made so that future users and contributors can understand how we arrived to a certain decision. I have added some guidance in the contribution guide around this.

I have also rearranged the contribution guide in an attempt to make it flow smoothly from the developer perspective (i.e. start at "setting up your dev environment" and end with "send us your PR")

[Rendered](https://github.com/cory-miller/actions-runner-controller/blob/contribution-guidelines/CONTRIBUTING.md)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,85 +1,53 @@
-## Contributing
+# Contribution Guide
 
-### Testing Controller Built from a Pull Request
+- [Contribution Guide](#contribution-guide)
+  - [Welcome](#welcome)
+  - [Before contrbuting code](#before-contrbuting-code)
+  - [How to Contribute a Patch](#how-to-contribute-a-patch)
+    - [Developing the Controller](#developing-the-controller)
+    - [Developing the Runners](#developing-the-runners)
+      - [Tests](#tests)
+      - [Running Ginkgo Tests](#running-ginkgo-tests)
+    - [Running End to End Tests](#running-end-to-end-tests)
+      - [Rerunning a failed test](#rerunning-a-failed-test)
+      - [Testing in a non-kind cluster](#testing-in-a-non-kind-cluster)
+    - [Code conventions](#code-conventions)
+    - [Opening the Pull Request](#opening-the-pull-request)
+  - [Helm Version Changes](#helm-version-changes)
+  - [Testing Controller Built from a Pull Request](#testing-controller-built-from-a-pull-request)
 
-We always appreciate your help in testing open pull requests by deploying custom builds of actions-runner-controller onto your own environment, so that we are extra sure we didn't break anything.
+## Welcome
 
-It is especially true when the pull request is about GitHub Enterprise, both GHEC and GHES, as [maintainers don't have GitHub Enterprise environments for testing](docs/detailed-docs.md#github-enterprise-support).
+This document is the single source of truth for how to contribute to the code base.
+Feel free to browse the [open issues] and file new ones, all feedback is welcome!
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -1,85 +1,53 @@
-## Contributing
+# Contribution Guide
 
-### Testing Controller Built from a Pull Request
+- [Contribution Guide](#contribution-guide)
+  - [Welcome](#welcome)
+  - [Before contrbuting code](#before-contrbuting-code)
```
AvaStancu, line 5:
```suggestion
  - [Before contributing code](#before-contributing-code)
```

CONTRIBUTING.md
```diff
@@ -1,85 +1,53 @@
...
+## Welcome
 
-Depending on what you are patching depends on how you should go about it. Below are some guides on how to test patches locally as well as develop the controller and runners.
+This document is the single source of truth for how to contribute to the code base.
+Feel free to browse the [open issues](https://github.com/actions-runner-controller/actions-runner-controller/issues) or file a new one, all feedback is welcome!
+By reading this guide, we hope to give you all of the information you need to be able to pick up issues, contribute new features, and get your work
+reviewed and merged.
 
-When submitting a PR for a change please provide evidence that your change works as we still need to work on improving the CI of the project. Some resources are provided for helping achieve this, see this guide for details.
+## Before contrbuting code
```
AvaStancu, line 26:
```suggestion
## Before contributing code
```

CONTRIBUTING.md
```diff
@@ -1,85 +1,53 @@
...
-> somewhere under `$HOME`.
-> Otherwise `kind load docker-image` fail while running `docker save`.
-> See https://kind.sigs.k8s.io/docs/user/known-issues/#docker-installed-with-snap for more information.
+We welcome code patches, but to make sure things are well coordinated you should  discuss any significant change before starting the work.
+The maintainers ask that you signal your intention to contribute to the project using the issue tracker. 
+If there is an existing issue that you want to work on, please let us know so we can get it assigned to you.
+If you noticed a bug or want to add a new feature, there are issue templates you can fill out.
 
-To test your local changes against both PAT and App based authentication please run the `acceptance` make target with the authentication configuration details provided:
+When filing a feature request, the maintainers will review the change and give you a decision on whether we are willing to accept the feature into the project.
```
AvaStancu, line 33:
```suggestion
When filling a feature request, the maintainers will review the change and give you a decision on whether we are willing to accept the feature into the project.
```

cory-miller: I meant this as 'to file'. Do you think it should be "filling out" instead?

AvaStancu: No, I am ok with filing too in this case :)

AvaStancu (dismissed):

mumoshu (approved):
LGTM. Thanks a lot for the write-up @cory-miller!

On a side note, I'm wondering if we could also enhance the feature request template to cover the kinds of feature requests that are unlikely to be accepted by the ARC project.

For example, we won't be accepting feature requests to add cloud-specific toolings, like awscli, gcloud-sdk, and so on, directly to our runner images. WDYT about that?

cory-miller:
Thanks for the pointer, I've added a note about that. We are also looking at our Runner image, and that should be coupled with information to provide a custom Runner image which will let users customize their tooling. More on that to come.

mumoshu (approved):
LGTM. Thanks a lot for the enhancements!

PR #1927, opened 2022-10-15
ci: use github-pr-check reporter for shellcheck

toast-gear:
we've set the reporter to a value that doesn't seem to be part of the action https://github.com/reviewdog/action-shellcheck/blob/master/action.yml#L15

```
      Reporter of reviewdog command [github-pr-check,github-pr-review,github-check].
      Default is github-pr-check.
      github-pr-review can use Markdown and add a link to rule page in reviewdog reports.
```


mumoshu (approved):
LGTM. Thanks for fixing it!

PR #1928, opened 2022-10-16
ci: replace set-output

toast-gear:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

set-output is being removed


mumoshu (approved):
Awesome! Thanks for the update 😄

PR #1931, opened 2022-10-17
use v1 policy api

zetaab:
```
❯ helm upgrade -n actions-runner-system snip actions-runner-controller/actions-runner-controller -f manifests/snip/values.yaml --install --version $CHART_VERSION
Error: UPGRADE FAILED: resource mapping not found for name: "snip-pdb" namespace: "actions-runner-system" from "": no matches for kind "PodDisruptionBudget" in version "policy/v1beta1"
ensure CRDs are installed first
```


toast-gear:
Drop the v1beta1 version entirely, v1 was introduced in 1.21 which is old enough at this point for us to removed support for the beta tag.

zetaab:
@toast-gear done

toast-gear (approved):
LGTM

PR #1946, opened 2022-10-23
Updates Makefile so shellcheck is cross-platform.

gwynforthewyn:
Specifically, the makefile was always downloading the linux version of shellcheck no matter the platform. So I moved the OS detection to be early in the makefile and verified that I can now download the darwin version of shellcheck.

I verified that the old behaviour still works by checking out the branch in debian and running `make test`.


mumoshu (approved):
LGTM. Thanks a lot for your contribution @gwynforthewyn!

PR #1968, opened 2022-10-31
admissionWebHooks: fix checking for caBundle

cristicalin:
This PR fixes a check for `caBundle` being present in the configured values. Previous check would always fail due to `.Values.admissionWebHooks` being `{}`.

This also fixes the example in the docs when relying on helm generated tls certificates for the web hook.


mumoshu (approved):
LGTM. Thanks a lot for the fixes @cristicalin ☺️

PR #1971, opened 2022-11-02
Allow to set docker default address pool

claudio-vellage:
Currently it's not easily possible to change the docker network CIDR range. This can lead to issues if there are networks with overlapping CIDR ranges in the same network, i.e. reachable by the same network interface. To avoid connectivity issues, it's possible to change the default CIDR range as follows:

```
{
  "default-address-pools": [{"base":"172.17.0.0/16","size":24}]
}
```

I've parametrized it and added it to the startup scripts, similar to how it's possible to set the MTU and registry mirrors. I would highly appreciate a quick review, let me know if I can assist in any way. This will allow for a much simpler runner setup.

Also check: https://github.com/actions-runner-controller/actions-runner-controller/pull/1910 for more discussion.

### DIFF
--- a/runner/entrypoint-dind-rootless.sh
+++ b/runner/entrypoint-dind-rootless.sh
@@ -18,7 +18,7 @@ echo "DOCKERD_ROOTLESS_ROOTLESSKIT_MTU=${MTU}" | sudo tee -a /etc/environment
 fi
 
 if [ -n "${DOCKER_DEFAULT_ADDRESS_POOL_BASE}" ] && [ -n "${DOCKER_DEFAULT_ADDRESS_POOL_SIZE}" ]; then
-  jq ".\"default-address-pools\" = [{\"base\": \"${DOCKER_DEFAULT_ADDRESS_POOL_BASE}\", \"size\": ${DOCKER_DEFAULT_ADDRESS_POOL_SIZE}}]" /etc/docker/daemon.json > /tmp/.daemon.json && /tmp/.daemon.json /etc/docker/daemon.json
+  jq ".\"default-address-pools\" = [{\"base\": \"${DOCKER_DEFAULT_ADDRESS_POOL_BASE}\", \"size\": ${DOCKER_DEFAULT_ADDRESS_POOL_SIZE}}]" /home/runner/.config/docker/daemon.json > /tmp/.daemon.json && mv /tmp/.daemon.json /home/runner/.config/docker/daemon.json
 fi
 
 if [ -n "${DOCKER_REGISTRY_MIRROR}" ]; then
--- a/runner/entrypoint-dind-rootless.sh
+++ b/runner/entrypoint-dind-rootless.sh
@@ -17,6 +17,10 @@ jq ".\"mtu\" = ${MTU}" /home/runner/.config/docker/daemon.json > /tmp/.daemon.js
 echo "DOCKERD_ROOTLESS_ROOTLESSKIT_MTU=${MTU}" | sudo tee -a /etc/environment
 fi
 
+if [ -n "${DOCKER_DEFAULT_ADDRESS_POOL_BASE}" ] && [ -n "${DOCKER_DEFAULT_ADDRESS_POOL_SIZE}" ]; then
+  jq ".\"default-address-pools\" = [{\"base\": \"${DOCKER_DEFAULT_ADDRESS_POOL_BASE}\", \"size\": ${DOCKER_DEFAULT_ADDRESS_POOL_SIZE}}]" /etc/docker/daemon.json > /tmp/.daemon.json && /tmp/.daemon.json /etc/docker/daemon.json
+fi
+
 if [ -n "${DOCKER_REGISTRY_MIRROR}" ]; then
 jq ".\"registry-mirrors\"[0] = \"${DOCKER_REGISTRY_MIRROR}\"" /home/runner/.config/docker/daemon.json > /tmp/.daemon.json && mv /tmp/.daemon.json /home/runner/.config/docker/daemon.json
 fi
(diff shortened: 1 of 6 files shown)


claudio-vellage:
Please approve and merge this asap, I see the file structure has changed, if this is in the wrong file, it would be highly appreciated if you can just move this to the correct file, otherwise let me know where it needs to go, I really need this feature.

runner/startup.sh
```diff
@@ -32,6 +32,10 @@ if [ -z "${RUNNER_NAME}" ]; then
   exit 1
 fi
 
+if [ -n "${DOCKER_DEFAULT_ADDRESS_POOL_BASE}" ] && [ -n "${DOCKER_DEFAULT_ADDRESS_POOL_SIZE}" ]; then
+  jq ".\"default-address-pools\" = [{\"base\": \"${DOCKER_DEFAULT_ADDRESS_POOL_BASE}\", \"size\": ${DOCKER_DEFAULT_ADDRESS_POOL_SIZE}}]" /etc/docker/daemon.json > /tmp/.daemon.json && /tmp/.daemon.json /etc/docker/daemon.json
+fi
```
mumoshu, line 37: Note that this needs to be in https://github.com/actions-runner-controller/actions-runner-controller/blob/master/runner/entrypoint-dind-rootless.sh and https://github.com/actions-runner-controller/actions-runner-controller/blob/master/runner/entrypoint-dind.sh but https://github.com/actions-runner-controller/actions-runner-controller/blob/master/runner/entrypoint.sh

cc/ @toast-gear- this might answer your question in DM

toast-gear (approved):

runner/entrypoint-dind-rootless.sh
```diff
@@ -17,6 +17,10 @@ jq ".\"mtu\" = ${MTU}" /home/runner/.config/docker/daemon.json > /tmp/.daemon.js
 echo "DOCKERD_ROOTLESS_ROOTLESSKIT_MTU=${MTU}" >> /etc/environment
 fi
 
+if [ -n "${DOCKER_DEFAULT_ADDRESS_POOL_BASE}" ] && [ -n "${DOCKER_DEFAULT_ADDRESS_POOL_SIZE}" ]; then
+  jq ".\"default-address-pools\" = [{\"base\": \"${DOCKER_DEFAULT_ADDRESS_POOL_BASE}\", \"size\": ${DOCKER_DEFAULT_ADDRESS_POOL_SIZE}}]" /etc/docker/daemon.json > /tmp/.daemon.json && /tmp/.daemon.json /etc/docker/daemon.json
```
mumoshu, line 21:
```suggestion
  jq ".\"default-address-pools\" = [{\"base\": \"${DOCKER_DEFAULT_ADDRESS_POOL_BASE}\", \"size\": ${DOCKER_DEFAULT_ADDRESS_POOL_SIZE}}]" /home/runner/.config/docker/daemon.json > /tmp/.daemon.json && mv /tmp/.daemon.json /home/runner/.config/docker/daemon.json
```

Apparent we had a wrong path here and missed `mv` too. Even after those fixes, it doesn't work for rootless dind runner for a reason that I don't know of. This is the error I observed:

```
time="2022-11-04T14:37:28.331425806Z" level=info msg="loading plugin \"io.containerd.internal.v1.opt\"..." type=io.containerd.internal.v1
time="2022-11-04T14:37:28.331485812Z" level=warning msg="failed to load plugin io.containerd.internal.v1.opt" error="mkdir /opt/containerd: permission denied"
```

runner/entrypoint-dind.sh
```diff
@@ -16,6 +16,10 @@ jq ".\"mtu\" = ${MTU}" /etc/docker/daemon.json > /tmp/.daemon.json && mv /tmp/.d
 echo "environment=DOCKERD_ROOTLESS_ROOTLESSKIT_MTU=${MTU}" >> /etc/supervisor/conf.d/dockerd.conf
 fi
 
+if [ -n "${DOCKER_DEFAULT_ADDRESS_POOL_BASE}" ] && [ -n "${DOCKER_DEFAULT_ADDRESS_POOL_SIZE}" ]; then
+  jq ".\"default-address-pools\" = [{\"base\": \"${DOCKER_DEFAULT_ADDRESS_POOL_BASE}\", \"size\": ${DOCKER_DEFAULT_ADDRESS_POOL_SIZE}}]" /etc/docker/daemon.json > /tmp/.daemon.json && /tmp/.daemon.json /etc/docker/daemon.json
```
mumoshu, line 20:
```suggestion
  jq ".\"default-address-pools\" = [{\"base\": \"${DOCKER_DEFAULT_ADDRESS_POOL_BASE}\", \"size\": ${DOCKER_DEFAULT_ADDRESS_POOL_SIZE}}]" /etc/docker/daemon.json > /tmp/.daemon.json && mv /tmp/.daemon.json /etc/docker/daemon.json
```

mumoshu:
@claudio-vellage Hey! Quick question- how did you test your proposed changes in your cluster? I fixed both rootless and rootful dind entrypoints as each missed either or both the `mv` command and the correct daemon.json path. And even after those fixes, it results in my dind daemon never coming up, which makes every workflow job fail due to the missing docker access.

mumoshu (approved):
I think I managed to make it work as probably expected. The last missing piece was that I had missed the trailing "/12" in the base value.
Thanks for your contribution @claudio-vellage!

claudio-vellage:
Thanks, sorry for late reply, it’s quite busy these days, so I appreciate that you did all the heavy lifting. I noticed the file structure changed, but didn’t have time to check where the code went. I’m glad it’s still working as expected.

PR #1977, opened 2022-11-03
Fix permission issue when you use PV for rootless dind cache

mumoshu:
This fixes the said issue I have found while testing #1759.

### DIFF
--- a/runner/entrypoint-dind-rootless.sh
+++ b/runner/entrypoint-dind-rootless.sh
@@ -22,8 +22,8 @@ jq ".\"registry-mirrors\"[0] = \"${DOCKER_REGISTRY_MIRROR}\"" /home/runner/.conf
 fi
 SCRIPT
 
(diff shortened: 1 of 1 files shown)


runner/entrypoint-dind-rootless.sh
```diff
@@ -22,6 +22,17 @@ jq ".\"registry-mirrors\"[0] = \"${DOCKER_REGISTRY_MIRROR}\"" /home/runner/.conf
 fi
 SCRIPT
 
+if [ -f /home/runner/.local ]; then
+  if [ ! -f /home/runner/.local/share ]; then
```
toast-gear, line 26: shouldn't these be checks for directories? `-d`

mumoshu: Good catch! Fixed in 300adcf7ec02b345e5ee173a24fde4f846089c75

toast-gear (approved):

PR #1983, opened 2022-11-04
ci: Fix runner builds but not pushes for forks

mumoshu:
I noticed that our runners workflow is failing on docker-login due to that a pull request workflow job from a fork does not have access to repo secrets.

https://github.com/malachiobadeyi/actions-runner-controller/actions/runs/3390463793/jobs/5634638183

Can we try this, so that hopefully it suppresses docker-login for pull requests from forks?

### DIFF
--- a/.github/workflows/runners.yaml
+++ b/.github/workflows/runners.yaml
@@ -62,6 +62,7 @@ jobs:
           password: ${{ secrets.DOCKER_ACCESS_TOKEN }}
           ghcr_username: ${{ github.actor }}
           ghcr_password: ${{ secrets.GITHUB_TOKEN }}
+        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
 
       - name: Build and Push Versioned Tags
         uses: docker/build-push-action@v3


.github/workflows/runners.yaml
```diff
@@ -62,6 +62,7 @@ jobs:
           password: ${{ secrets.DOCKER_ACCESS_TOKEN }}
           ghcr_username: ${{ github.actor }}
           ghcr_password: ${{ secrets.GITHUB_TOKEN }}
+        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
```
mumoshu, line 65:
```suggestion
        # See below for why we need a temporary envvar for this conditional:
        # https://stackoverflow.com/questions/70249519/how-to-check-if-a-secret-variable-is-empty-in-if-conditional-github-actions
        env:
          NECESSARY_ENVVARS: ${{ secrets.DOCKERHUB_USERNAME }}${{ secrets.DOCKER_ACCESS_TOKEN }}${{ secrets.GITHUB_TOKEN }}
        if: "${{ env.NECESSARY_ENVVARS != '' }}"
```

toast-gear:
Would it make more sense to extend the existing conditionals in teh composite action to check for empty inputs so we are doing an if in 1 place instead of 2? https://github.com/actions-runner-controller/actions-runner-controller/blob/ef19b41a1f412f7f0ce9f380f0ef0558d6056500/.github/actions/setup-docker-environment/action.yaml#L35-L42

mumoshu:
@toast-gear That might be a lot better! Thanks for the feedback ☺️ Will fix it tonight

.github/actions/setup-docker-environment/action.yaml
```diff
@@ -32,14 +32,14 @@ runs:
...
-      if: ${{ github.event_name == 'release' || github.event_name == 'push' && github.ref == 'refs/heads/master' }}
+      if: ${{ github.event_name == 'release' || github.event_name == 'push' && github.ref == 'refs/heads/master' && inputs.password != ''  }}
       uses: docker/login-action@v2
       with:
         username: ${{ inputs.username }}
         password: ${{ inputs.password }}
 
     - name: Login to GitHub Container Registry
-      if: ${{ github.event_name == 'release' || github.event_name == 'push' && github.ref == 'refs/heads/master' }}
+      if: ${{ github.event_name == 'release' || github.event_name == 'push' && github.ref == 'refs/heads/master' && inputs.password != ''  }}
```
toast-gear, line 42: `inputs.password` is for DockerHub, we should check for `inputs.ghcr_password` when checking to see if we should log into the GitHub registry?

mumoshu: Ah, good catch! Fixed in 9831dfe

toast-gear (approved):

PR #1988, opened 2022-11-05
docs: add the limitation to disabling updates

toast-gear:
We get issues and dicussions raised a lot for this limitation, it's documented in GitHub's docs (which we try not to document again in ARC) but it doesn't seem to get seen a lot. We get raised issues / dicussions frequently enough that we should document it here too to cut down on raised issues / dicussions.

https://docs.github.com/en/actions/hosting-your-own-runners/autoscaling-with-self-hosted-runners#controlling-runner-software-updates-on-self-hosted-runners


mumoshu (approved):
LGTM. Thank you so much for the addition!

PR #1992, opened 2022-11-07
fix doc detailed-docs.md

vitaliytv:
helm params


mumoshu (approved):
Good catch. Thanks for fixing it!

Issue #1380, opened 2022-04-24
could not parse webhook


sagi-shimoni:
I'm trying to enable HorizontalRunnerAutoscaler based on github webhooks but it's not working due to webhooks not being parsed correctly.

I checked the logs of pod "actions-runner-controller-github-webhook-server" and I encountered below error.

Installed v0.22.3 via Helm chart.

```
2022-04-24T09:26:05Z	ERROR	controllers.webhookbasedautoscaler	could not parse webhook	{"webhookType": "workflow_job", "payload": "payload=%7B%22action%22%3A%22queued%22%2C%22workflow_job

..................... (sensitive payload info) .....................

"error": "invalid character 'p' looking for beginning of value"}
net/http.HandlerFunc.ServeHTTP
	/usr/local/go/src/net/http/server.go:2047
net/http.(*ServeMux).ServeHTTP
	/usr/local/go/src/net/http/server.go:2425
net/http.serverHandler.ServeHTTP
	/usr/local/go/src/net/http/server.go:2879
net/http.(*conn).serve
	/usr/local/go/src/net/http/server.go:1930
```

mumoshu:
@sagi-shimoni Hey. You seem have configured a wrong content-type for your webhook. Please use application/json.

jonathanbeber:
@mumoshu to avoid errors such this one, is there any plan or an issue (sorry, couldn't find one) about adding the feature of the controller creating the webhook by itself? i.g. was it already discussed and, if not, is it worth to create a new issue with this request?

PR #1995, opened 2022-11-08
improve webhook setting doc

0xF4D3C0D3:
the default value of "Content Type" is `application/x-www-form-urlencoded`. But with this "Content Type" webhook-controller can't parse the event and log like `invalid character 'p' looking for beginning of value`. So it should be `application/json`. I met this issue a while ago and found [this solution](https://github.com/actions-runner-controller/actions-runner-controller/issues/1380#issue-1213634393) I think it would've been better if it had been in the doc.


mumoshu (approved):
Ah, good catch! Thanks for the improvement. LGTM!

PR #2024, opened 2022-11-18
Runner does not need to run privileged if docker is disabled

Pierre-Raffa:


mumoshu:
Hey! This doesn't look right because some folks still use a privileged container without dockerd to run a job step like building a VM image.

Pierre-Raffa:
Then, do you think it would require another parameter to handle my scenario ?

mumoshu:
@Pierre-Raffa Well, could you share an example RunnerDeployment manifest to reproduce your issue? If you just set DockerEnabled to false in the runner spec, didn't to just omit the docker sidecar and produce a pod whose runner container does not have `privileged` set?

mumoshu:
Please read the docs in case you havent done so yet.
https://github.com/actions-runner-controller/actions-runner-controller/blob/master/docs/detailed-docs.md#additional-tweaks

PR #2034, opened 2022-11-23
docs: remove caveat from useRunnerGroupsVisibility

toast-gear:
This cavaet isn't really true anymore since GitHub added the visibility parameter, we had this warning because previously we had to make a bunch of API calls to figure it out.

Compare with whitespace changes hidden


mumoshu (approved):
Good catch! Thanks a lot for the fix ☺️

PR #2036, opened 2022-11-23
ci: multiple ubuntu version

toast-gear:


.github/workflows/runners.yaml
```diff
@@ -77,9 +77,9 @@ jobs:
           tags: |
             ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:v${{ env.RUNNER_VERSION }}-${{ matrix.os-name }}-${{ matrix.os-version }}
             ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:v${{ env.RUNNER_VERSION }}-${{ matrix.os-name }}-${{ matrix.os-version }}-${{ env.sha_short }}
-            ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:latest
-            ghcr.io/${{ github.repository }}/${{ matrix.name }}:latest
+            ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:${{ matrix.os-name }}-${{ matrix.os-version }}-latest
```
mumoshu, line 80:
```suggestion
            ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:${{ matrix.os-name }}-${{ matrix.os-version }}
```

This looks geat overall! Thanks for your efforts. One suggestion for this though- Can we omit the "-latest" suffix for this variant? I think we are basically omitting that suffix also on the existing `v${{ env.RUNNER_VERSION }}-${{ matrix.os-name }}-${{ matrix.os-version }}` image.

.github/workflows/runners.yaml
```diff
@@ -77,9 +77,9 @@ jobs:
           tags: |
             ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:v${{ env.RUNNER_VERSION }}-${{ matrix.os-name }}-${{ matrix.os-version }}
             ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:v${{ env.RUNNER_VERSION }}-${{ matrix.os-name }}-${{ matrix.os-version }}-${{ env.sha_short }}
-            ${{ env.DOCKERHUB_USERNAME }}/${{ matrix.name }}:latest
```
mumoshu, line 80: I believe it's not easy, but can we add another pipeline to keep publishing `latest`, that points to the latest ubuntu 20.04 image for a while and will eventually get updated to 22.04 in the near future?

Without that another pipeline, we end up immediately stopping updates to the `latest` image, which many folks seem to be using today, and those folks would end up seeing a lot of runner auto updates triggered before their runners within runner pods get up and running, which will trigger a storm of questions and bug reports...

mumoshu: > **will eventually get updated to 22.04 in the near future?**

Or maybe we can deprecate the latest tag soon and stop updating it e.g. 1 month before we stop publishing legacy 20.04 images.
The result might be the same. I think stopping updates on the `latest` tag immediately can be a source of confusion and breakage though.

toast-gear: Having had a think I agree with all this. I think we should deprecate the `latest` tag entirely but yeh, we probably should do it after 22.04 has been out for some time. Perhaps we can update the startup.sh to spit out some warnings too and pin an issue to give people the best chance of seeing it before we stop updating it. Should we look to delete the latest tag from DockerHub too as it'll be very misleading the tag is left hanging around once we have stopped using it.

.github/workflows/runners.yaml
toast-gear: We are that's true. Now you've pointed it out it's a bit annoying but I think we probably should stick with not having it for backwards compatibility reasons.

mumoshu (approved):
LGTM. Thank you so much for your contribution as always!

PR #2044, opened 2022-11-25
e2e: Do honor the runner graceful stop timeout also in the dockerd sidecar prestop hook

mumoshu:
The runner graceful stop timeout has never been propagated to the dind sidecar due to configuration error in E2E. This fixes it, so that we can verify that the dind sidecar prestop can respect the graceful stop timeout.

Related to #1759


toast-gear (approved):

mumoshu:
Thanks for reviewing ❤️

PR #2077, opened 2022-12-06
feat: Set runner UID and docker GID to match github actions runner

mumoshu:
This is a successor to #1688


toast-gear (approved):

toast-gear:
Having thought about this do we even need to parameterise this? We're going to keep 20.04 images at the old UID and GID values, shall we just hardcode 22.04 to the new values and be done with it?

mumoshu:
As discussed in Slack, I think the build args would help folks to build custom images with alternative uid/gid for whatever reasons. Although we can't offer user support for custom runner images, making it easy to customize might still be handy!

PR #2093, opened 2022-12-13
Add Repository information to Runner Status

Moser-ss:
This is an extension of the feature introduced in the PR https://github.com/actions/actions-runner-controller/pull/1268 to add `.status.repository` `.status.owner` `.status.repo` `.status.workflow` to enrich the data related to Runner.

The idea is to add more context information to the Runner status. In this way, we can correlate the metrics of the runners with the usage done by the workflows. For example, if a runner is evicted or gets OOM, we can know which repository and workflow caused that issue.
Another use case is to understand if some workflows suffer resource constraints in terms of CPU, like workflows that run tests or build assets. With the new status properties, we can correlate high CPU usage with specific repositories and workflow.

An additional use case is to fine-tune runners for specific workflows. By knowing which workflows are executed in the Runner with time, we can understand resource usage patterns and create new runners for specific workflows


mumoshu:
@Moser-ss Thanks for your PR! Let me rebase and try testing this locally to move this forward.

>  In this way, we can correlate the metrics of the runners with the usage done by the workflows

This is indeed important!

We currently need to correlate runner logs with workflow job logs to see which runner failed due to which job or which job failed on which runner, and correlating two unstructured log streams with the fact that two log streams contains the same job ID isn't that easy.

With this feature, we can just watch runner statuses instead, which would greatly eases troubleshooting.. Did I get it right?

config/crd/bases/actions.summerwind.dev_runners.yaml
```diff
@@ -5196,6 +5208,14 @@ spec:
                   type: string
                 message:
                   type: string
+                owner:
+                  type: string
+                repository:
+                  type: string
+                repo:
+                  type: string
+                workflow:
```
mumoshu: Nit but I would prefer `workflow_name` so that it corresponds with the `name` field of the workflow defintiion

apis/actions.summerwind.net/v1alpha1/runner_types.go
```diff
@@ -316,10 +366,8 @@ func (w *WorkVolumeClaimTemplate) V1VolumeMount(mountPath string) corev1.VolumeM
 // +kubebuilder:printcolumn:JSONPath=".spec.labels",name=Labels,type=string
 // +kubebuilder:printcolumn:JSONPath=".status.phase",name=Status,type=string
 // +kubebuilder:printcolumn:JSONPath=".status.message",name=Message,type=string
-// +kubebuilder:printcolumn:JSONPath=".status.repository",name=Repository,type=string
-// +kubebuilder:printcolumn:JSONPath=".status.owner",name=Owner,type=string
-// +kubebuilder:printcolumn:JSONPath=".status.repo",name=Repo,type=string
-// +kubebuilder:printcolumn:JSONPath=".status.workflow",name=Workflow,type=string
+// +kubebuilder:printcolumn:JSONPath=".status.workflow.repository",name=WF Repo,type=string
```
mumoshu, line 369: Can we keep the number of new columns minimal not to bloat kubectl output?
Here's my suggestion- the repo and the run ID should be enough to find out what workflow run is triggered on the runner.

apis/actions.summerwind.net/v1alpha1/runner_types.go
```diff
@@ -248,10 +248,60 @@ type RunnerStatus struct {
 	// +optional
 	Message string `json:"message,omitempty"`
 	// +optional
+	WorkflowStatus *WorkflowStatus `json:"workflow"`
```
mumoshu, line 251: In your original work, you directly modified the auto-generated CRD. This is how you should add fields in Go, which are then reflected to the (re-)generated CRDs by running `make manifests`!

I've also grouped the all the workflow related status fields into the `status.workflow` field so that it's easy to extract only workflow related statuses for further processing.
For example, you can use `kubectl get runner --watch -o json | jq .status.worfklow` so obtain the workflow status stream only.

charts/actions-runner-controller/crds/actions.summerwind.dev_runners.yaml
```diff
@@ -5245,6 +5231,31 @@ spec:
                     - expiresAt
                     - token
                   type: object
+                workflow:
+                  description: WorkflowStatus contains various information that is propagated from GitHub Actions workflow run environment variables to ease monitoring workflow run/job/steps that are triggerred on the runner.
```
mumoshu, line 5235: FYI, `make manifests` (or controller-tools `controller-gen` used internally by the make target) automatically reflects the Go field comment to this description field.

runner/hooks/job-started.d/update-status
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 set -u
 
-exec update-status Running "Run $GITHUB_RUN_ID from $GITHUB_REPOSITORY" "$GITHUB_REPOSITORY" "$GITHUB_REPOSITORY_OWNER" "$GITHUB_WORKFLOW"
```
mumoshu: I modified your work to not take those as function arguments, and instead refer to envvars presuming they are all optional, so that we won't bloat function parameters and arguments as we add more.

runner/update-status
```diff
@@ -16,18 +16,27 @@ if [[ ${RUNNER_STATUS_UPDATE_HOOK:-false} == true ]]; then
...
-    [[ -n "$repository" ]] && repo=${repository#"$owner/"}
 
     data=$(jq -n --arg phase "$phase" \
       --arg message "$message" \
-      --arg repository "$repository" \
-      --arg owner "$owner" \
-      --arg repo "${repo:-}" \
-      --arg workflow "$workflow" \
-      '.status.phase = $phase | .status.message = $message | .status.repository = $repository | .status.owner = $owner | .status.repo = $repo | .status.workflow = $workflow')
+      --arg workflow_repository "${GITHUB_REPOSITORY:-}" \
```
mumoshu, line 22: I tried to name the arg name after the status field path for consistency. For example, this is `workflow_repository` because the corresponding status field path is `status.workflow.repository`.

All other new args follow the same pattern.

mumoshu:
@Moser-ss I've updated your awesome work with improvements and suggestions. I'd appreciate it if you could confirm all! Thanks in advance for your help.

apis/actions.summerwind.net/v1alpha1/runner_types.go
Moser-ss: Thanks for the explanation, this way looks way better

runner/update-status
Moser-ss: With the new change thes new syntax makes sense

apis/actions.summerwind.net/v1alpha1/runner_types.go
Moser-ss: Ok, I agree

Moser-ss:
@mumoshu  Thanks a lot for the rework; after I saw the fine-tuning you did, this PR looks even better. I like all the changes

Moser-ss:
> With this feature, we can just watch runner statuses instead, which would greatly eases troubleshooting.. Did I get it right?

It is not only about the logs, but is about also about resources, especially when I have OOM or pod evictions. Now when I got an OOM I can go to GitHub find the right workflow run and repository and understand why we have that OOM

mumoshu:
@Moser-ss Makes sense a lot! Thank you so much for your explanation ☺️

Moser-ss:
@mumoshu Any idea when this PR can be merged? do we have any blocker?

mumoshu:
@Moser-ss We had to defer this until we merge a [big PR](#2153) to make testing and conflict resolution easier! Thanks for your patience. Let me merge this now.

mumoshu (approved):
Thanks a lot for your contribution @Moser-ss ☺️

PR #2098, opened 2022-12-16
Update ubuntu 22.04 image with openssh

arielly-parussulo:
Add Openssh client to ubuntu 22.04 runner.

### DIFF
--- a/runner/actions-runner-dind.ubuntu-22.04.dockerfile
+++ b/runner/actions-runner-dind.ubuntu-22.04.dockerfile
@@ -19,6 +19,7 @@ RUN apt-get update -y \
     && apt-get install -y --no-install-recommends \
     curl \
     ca-certificates \
+    openssh-client \
     git \
     git-lfs \
     iptables \


arielly-parussulo:
Hello, how are you doing?
We started to use the ubuntu 22.04 image for our Github Actions hosted-runners and noticed that the ssh client was not installed in that image.

mumoshu:
@arielly-parussulo Hey! Thanks for submitting this. Our plan for 22.04 runner images is to make those dead simple, including only minimum dependencies that are needed in every environment. I think ssh client is needed only when you ssh from the runner into other nodes or whatever running sshd, or cloning private Git repositories via ssh in case https auth isn't possible. Did I get it correctly?

@toast-gear Please correct me if I missed anything!

arielly-parussulo:
> @arielly-parussulo Hey! Thanks for submitting this. Our plan for 22.04 runner images is to make those dead simple, including only minimum dependencies that are needed in every environment. I think ssh client is needed only when you ssh from the runner into other nodes or whatever running sshd, or cloning private Git repositories via ssh in case https auth isn't possible. Did I get it correctly?
> 
> @toast-gear Please correct me if I missed anything!

No problem! I did this Pull Request because some of the developers of my company are doing a git clone operation with SSH. I will roll back to the Ubuntu 20 runner for now and check with my team whether is possible or them to use HTTPS to clone their repositories.
Thanks for the reply!

mumoshu:
@arielly-parussulo Thanks for your response and feedback! We're still figuring out what to be included in the 22.04 runner image. We're are to include anything that can be considered "core" and perhaps `ssh` might be too "core" to be stripped. Please don't hesitate to share your team's feedback to us so that we can reconsider with more confidence!

asardaes:
@mumoshu May I ask if the discussion about "core" dependencies to include was finalized? My team also uses SSH to create commits in a repository different than the one that triggers the workflow. I imagine there are alternatives for us, but it would be good to first know what your team decided.

mumoshu:
@asardaes Hey! We might have two channels of runner images. First, the GH official runner images would be minimal (See https://github.com/actions/runner/blob/main/images/Dockerfile). I don't expect those to have dependencies like `ssh` included.

Second, we might provide "community" runner images with different flavors. What if we had:

- Base runner images: including so-called "common" dependencies like `ssh`
- AWS runner images: including awscli
- GCP runner images: including google cloud-sdk
- ...

asardaes:
But then that's still under discussion, right? Is there a place where that's tracked?

PR #2111, opened 2022-12-29
remove redundant description

dongho-jung:
always thanks for awesome opensource : )


mumoshu (approved):
Good catch...! Thanks for fixing it ☺️

PR #2126, opened 2023-01-09
Add ability to specify annotations in helm chart CRDs manifests

minhthong582000:
This is an PR for the issue i've created https://github.com/actions/actions-runner-controller/issues/2125 to add ability to specify custom annotations to the Helm Chart CRDs manifests.


mumoshu:
@minhthong582000 Hey! Thanks for your contribution. I've rebased your changes to fix conflicts. This should be ready for testing again.

I've also reviewed this. I understand your motivation and the new values and the CRD templates look good. However, my concern is- this seems to break `helm template --include-crds` and `helm upgrade --skip-crds`. The CRDs are now rendered/installed regardless of those flags and that might make this chart unconventional 😢

Do you have any idea on how we could introduce this change without making the chart unconventional?

rufreakde:
I think the cleanest way would be to support helm repository directly.
https://github.com/helm/helm/pull/11022

minhthong582000:
Closed since these changes will make this chart unconventional.

PR #2120, opened 2023-01-06
Update release workflows

Link-:
- resolves https://github.com/github/c2c-actions-runtime/issues/2221
- resolves https://github.com/github/c2c-actions-runtime/issues/2194
- resolves https://github.com/github/c2c-actions-runtime/issues/2212

### Summary

Following the transfer of `actions-runner-controller` to the `actions` organisation, the controller and runner release workflows have been broken.

This PR intends to fix those workflows while introducing the least amount of changes. The fact that packages (all images pushed to GHCR) remain under [the old organisation](https://github.com/orgs/actions-runner-controller/packages) – which is intentional to avoid breaking end-user environments – introduced unnecessary complexity to these workflows.

As we work on improving our posture, this PR remedies the situation and provides a mechanism to publish changes until the permanent fixes are introduced.

The behaviour of these workflows is fully described here: https://github.com/actions-runner-controller/releases

### Caveats

For all of these workflows, the image pushing after build has not been tested (to avoid polluting registries). This needs to be verified after we merge.

### TODO before merge

- [x] Commit all TODO suggestions
- [x] Test & confirm the [publish-chart.yaml workflow](https://github.com/actions/actions-runner-controller/blob/c407b33040afd6ca6ad6347105a30f8a94ac8824/.github/workflows/publish-chart.yaml) is working as expected
- [x] Remove all references to `set-output`: [ref1](https://github.com/actions/actions-runner-controller/blob/c407b33040afd6ca6ad6347105a30f8a94ac8824/.github/workflows/publish-chart.yaml#L102)
- [x] Update `DOCKER_ACCESS_TOKEN` and `DOCKER_USER` secrets in this repo and in actions-runner-controller/releases

PR #2140, opened 2023-01-11
Add job summary to the runners release workflow

Link-:
### Context

- Prevents runner images being pushed to registries when the workflow run is triggered by a PR
- Adds a job summary to the runners release workflow.

This is an addendum to the work in: #2120


mumoshu:
@Link- Thanks as always for your effort!

I see the build is failing with 

```
Run peter-murray/workflow-application-token-action@8e1ba3bf1619726336414f1014e37f17fbadf1db
Error: Error: Input required and not supplied: application_private_key
Error: Failed to initialize GitHub Application connection using provided id and private key
```

It's probably due to that the Actions secrets are not available to pull request workflows. We can't run the subsequent `gh auth login` step and the workflow triggering step without the secrets and workflow-application-token-action outputs... right?

If that's the case, I believe you can turn this workflow into a push-only workflow in this PR and merge. We can then follow up by adding another pull request workflow run run docker builds only within this repo.

Link-:
Excellent @mumoshu that makes things much simpler!

mumoshu (approved):
LGTM. Awesome work!

PR #2145, opened 2023-01-12
fix(e2e): Make runner pod deleter cancellable

mumoshu:
TL;DR; This makes the E2E test more reliable.

This makes it so that the whole test run can be stopped immediately when e.g. there was a test bug or a test config mistake that resulted in it unable to call the necessary K8s API to delete the runner pod.

This should be merged before finishing the work toward the next ARC 0.27.0 release.


Link- (approved):
LGTM - 🚢

mumoshu:
Thanks for reviewing!

PR #2146, opened 2023-01-12
fix(e2e): Use the correct full chart name in test

mumoshu:
The whole E2E test breaks due to the invalid chart name without this fix.
This should be merged before finishing the work towards the next ARC 0.27.0 release.


toast-gear (approved):

mumoshu:
@toast-gear Thanks for reviewing!

PR #2147, opened 2023-01-12
Fix various golangci-lint errors

mumoshu:
that we introduced via controller-runtime upgrade and via the removal of legacy pull-based scale triggers (#2001).

This should address all the golangci-lint findings and make the golangci-lint workflow that is currently [failing on our main branch](https://github.com/actions/actions-runner-controller/actions/runs/3893274099/jobs/6645671769) to pass.

It would be great if we could review and merge this before releasing ARC 0.27.0!


Link-:
LGTM! Thank you for addressing this 🙇

Link- (approved):

PR #2150, opened 2023-01-12
Update controller package names to match the owning API group name

TingluoHuang:
Follow up PR after https://github.com/actions/actions-runner-controller/pull/2105

Rename the existing controller package name to `actionssummerwindnet`, preparation for creating a new controller package under `actions.github.com` API group


cmd/githubwebhookserver/main.go
```diff
@@ -26,8 +26,8 @@ import (
 	"sync"
 	"time"
 
-	actionsv1alpha1 "github.com/actions/actions-runner-controller/apis/actions.summerwind.net/v1alpha1"
```
mumoshu, line 29: Shall we leave this as `acctionsv1alpha1` for consistency with others https://github.com/search?q=repo%3Aactions%2Factions-runner-controller%20actionsv1alpha1&type=code for now?

Also perhaps your eventual goal would be to differentiate this with the new API package comes along with the new runnerscaleset feature?
If that's the case, how about renaming this alias to `summerwindv1alpha` beforehand and use the alias `actionsv1(alpha|beta)1?` for the new API?

Not only do they look consistent, but differentiating aliases like that would make code search easier.

mumoshu (dismissed):
One comment but LGTM overall! Thank you for your effort ☺️

cmd/githubwebhookserver/main.go
TingluoHuang: 👍 I don't know how I missed this.

mumoshu (approved):
LGTM. Thanks!

PR #2155, opened 2023-01-14
Include actions-runner-controller in runner's User-Agent for better telemetry in Actions service.

TingluoHuang:
We want to get a general idea about how many runners on the Actions service are coming from the actions-runner-controller.

- In a future runner release, the runner will read ENV `GITHUB_ACTIONS_RUNNER_EXTRA_USER_AGENT` on startup and add the value of the env as part of the User-Agent header for every HTTP request to Actions service.
- We can check telemetry on the Actions service's backend to see which runner is configured by actions-runner-controller.

With this data, we can better optimize backend performance for customers using the actions-runner-controller. 🙇


mumoshu (approved):
This is absolutely a good idea ❤️ LGTM!

PR #2157, opened 2023-01-16
Add resolve push to registries step

Link-:
This is a fix for the failing workflows. The input push to registries is empty when the workflow is triggered by a release, so we have to provide it with a default value. 

I've added a step to handle this.


mumoshu (approved):
Awesome! Thanks for fixing it so quickly @Link- ☺️

PR #2159, opened 2023-01-16
Fix the workflow by adding the version resolve step

Link-:


rentziass (dismissed):

.github/workflows/publish-arc.yaml
```diff
@@ -71,22 +81,13 @@ jobs:
             echo "RELEASE_TAG_NAME=${{ github.event.inputs.release_tag_name }}" >> $GITHUB_ENV
           fi
 
-      - name: Resolve push to registries
```
mumoshu, line 74: To be extra sure, we still need this step regardless of we set up the `VERSION` env for `make github-release` ☺️

Link-: 100% ! I renamed the variables to reduce redundancy

mumoshu: Ah! It looks even more nicer now!

mumoshu (approved):
LGTM. Thank you!

PR #2160, opened 2023-01-16
chart: Bump chart and app versions for ARC 0.27.0

mumoshu:
This is a routine pull request that we need to submit manually for each new ARC release, to make the default chart installation accommodates the new ARC version via the updated `appVersion`.


Link- (approved):
🚀  :shipit:

PR #2158, opened 2023-01-16
Fix typo in release note for ARC 0.27.0

KENNYSOFT:
Just a minor typo fix for the documentation! Thanks and congratulations on the 0.27.0 release!!

PR #2166, opened 2023-01-17
Fix logFormat comment for each module in Helm chart

KENNYSOFT:
Also a small typo fix in values.yaml but it's about the Helm chart, so I've separated PR from #2158. Thanks!


mumoshu (approved):

PR #2177, opened 2023-01-17
Include extra user-agent for runners created by actions-runner-controller.

TingluoHuang:
Follow-up PR to https://github.com/actions/actions-runner-controller/pull/2155
Both ARC modes will set the same user-agent (actions-runner-controller/0.xx.x) for requests made by the runner inside the pod.


mumoshu (approved):
LGTM. Thanks a lot for the enhancement!

PR #2178, opened 2023-01-18
Fix L0 test to make it more reliable.

TingluoHuang:
There are some L0 tests are randomly failed during CI/PR run, but work fine in our dev environment.
I am going to try improving the tests or fixing the bug the test exposes depending on what I found.

- [x] Test: `It should delete finished EphemeralRunner and create new EphemeralRunner`
- [x] Test: `It should not re-create pod indefinitely`
- [x] Test: `It should create/add all required resources for a new EphemeralRunnerSet (finalizer)`
> Ex: 
>       https://github.com/actions/actions-runner-controller/actions/runs/3942825864/jobs/6746938772
>       https://github.com/actions/actions-runner-controller/actions/runs/3946185966/jobs/6760411672

### DIFF
--- a/controllers/actions.github.com/ephemeralrunner_controller.go
+++ b/controllers/actions.github.com/ephemeralrunner_controller.go
@@ -310,6 +310,7 @@ func (r *EphemeralRunnerReconciler) cleanupRunnerLinkedPods(ctx context.Context,
 	}
 
 	if len(runnerLinkedPodList.Items) == 0 {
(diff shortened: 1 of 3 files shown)


TingluoHuang:
The tests are now passed 4 times in a row, we should be in a better spot.
image

nikola-jokic (approved):

rentziass (approved):

PR #2193, opened 2023-01-20
Refactor actions.Client with options to help extensibility

rentziass:
This is a refactor of `actions.Client` to help extensibility for the upcoming work for supporting proxies and custom TLS. Sorry for the big diff, most of it is test changes to adopt the new API. The notable changes are:

- [change `NewClient` signature](https://github.com/actions/actions-runner-controller/blob/6667f2417b42c6f8a45517becf7a5dbdd7acbadb/github/actions/client.go#L102) to only require the bare minimum a `Client` needs to function correctly, and move every other kind of configuration into a `ClientOption`. This is both to help ease extending the client and to make it easier to rely on `NewClient` within tests: the more logic we add to a client, the more manually instantiating one becomes error prone.
- A client is created with an HTTP client that will be used for all requests (at the moment we are creating one per request)
- to encourage the use of `NewClient`, [`newActionsServer`](https://github.com/actions/actions-runner-controller/blob/6667f2417b42c6f8a45517becf7a5dbdd7acbadb/github/actions/actions_server_test.go#L14-L20) is now available in tests.
- `UpdateRunnerScaleSet` and `UpdateRunnerScaleSet` methods were removed, I couldn't find any uses of these?
- By default logs in tests are silent

### DIFF
--- a/github/actions/client.go
+++ b/github/actions/client.go
@@ -7,7 +7,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"log"
 	"net/http"
 	"net/url"
@@ -118,7 +117,7 @@ func NewClient(ctx context.Context, githubConfigURL string, creds *ActionsAuth, 
 
 	// TODO: this silences retryclient default logger, do we want to provide one
 	// instead? by default retryablehttp logs all requests to stderr
-	retryClient.Logger = log.New(ioutil.Discard, "", log.LstdFlags)
+	retryClient.Logger = log.New(io.Discard, "", log.LstdFlags)
 
 	retryClient.RetryMax = ac.retryMax
 	retryClient.RetryWaitMax = ac.retryWaitMax
--- a/github/actions/client.go
+++ b/github/actions/client.go
@@ -7,6 +7,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
+	"log"
 	"net/http"
 	"net/url"
 	"path"
@@ -62,8 +64,8 @@ type Client struct {
 	ActionsServiceAdminTokenExpiresAt *time.Time
 	ActionsServiceURL                 *string
 
-	RetryMax     *int
-	RetryWaitMax *time.Duration
+	retryMax     int
+	retryWaitMax time.Duration
 
 	creds           *ActionsAuth
 	githubConfigURL string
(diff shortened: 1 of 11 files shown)


github/actions/client_generate_jit_test.go
```diff
@@ -3,73 +3,60 @@ package actions_test
...
-
-		if diff := cmp.Diff(want, got); diff != "" {
-			t.Errorf("GenerateJitRunnerConfig(%v) mismatch (-want +got):\n%s", name, diff)
-		}
+		got, err := client.GenerateJitRunnerConfig(ctx, runnerSettings, 1)
+		require.NoError(t, err)
+		assert.Equal(t, want, got)
 	})
 
 	t.Run("Default retries on server error", func(t *testing.T) {
```
rentziass, line 37: Since the HTTP client is now persisted across requests, should we test that the client itself has retries or do we need to keep testing retries on all calls?

mumoshu: I'd prefer the latter because we can ensure all calls use the retryable client or its equivalent!

github/actions/client.go
```diff
@@ -62,23 +63,66 @@ type Client struct {
...
+		retryWaitMax: 30 * time.Second,
+	}
+
+	for _, option := range options {
+		option(ac)
 	}
 
+	retryClient := retryablehttp.NewClient()
+
+	// TODO: this silences retryclient default logger, do we want to provide one
```
mumoshu, line 118: It would be great if we could selectively enable the logging, for example, let retryablehttp log requests only when the log level is DEBUG or finer.
That way I'd be able to set the log level to DEBUG when I'm curious why job run delays are unusually long and the retrylablehttp logs would reveal that it's due to excessive retries caused by e.g. an actions service incident.

mumoshu: Anyway, we'd better address that in another pull request, if necessary, to avoid scope creeping this one!

mumoshu (approved):
LGTM. Awesome job @rentziass!!

github/actions/client.go
TingluoHuang: I think we might want to have some logging on error cases, especially to log out some HTTP headers so we can check telemetry on the GitHub service.
Ex: `x-github-activity-id`, `ActivityId`, etc.

TingluoHuang: again, it doesn't have to be this PR. 😄

github/actions/client.go
```diff
@@ -260,59 +297,14 @@ func (c *Client) CreateRunnerScaleSet(ctx context.Context, runnerScaleSet *Runne
...
-	}
-	var createdRunnerScaleSet *RunnerScaleSet
-	err = unmarshalBody(resp, &createdRunnerScaleSet)
-	if err != nil {
-		return nil, err
-	}
-	return createdRunnerScaleSet, nil
-}
-
-func (c *Client) UpdateRunnerScaleSet(ctx context.Context, runnerScaleSetId int, runnerScaleSet *RunnerScaleSet) (*RunnerScaleSet, error) {
```
TingluoHuang, line 281: Do we not using `UpdateRunnerScaleSet` anywhere? 😄

TingluoHuang (approved):

nikola-jokic (approved):
Looks great to me!

github/actions/client.go
rentziass: Yeah I was very surprised, but the only use of it was in its own test 😄

github/actions/client.go
rentziass: Thanks both, I'll look at this separately. It's a bit of a pain that retryablehttp wants a different interface for logging, it'd be great not to have to care about it 😄

github/actions/client_generate_jit_test.go
rentziass: But given that the HTTP client is now re-used for the whole lifetime of an actions.Client wouldn't it make sense to have tests that verify that the option of having (or not) retries is observed? The only way to change the HTTP client down the line is for callers of this package to manually change `client.Client`, but then I would argue it's that caller to test responsibility that their override is working as intended. What do you think?

rentziass: Merging now to unlock other work on expanding the client, but definitely interested in this conversation!

PR #2199, opened 2023-01-23
Add support for custom CA in actions.Client

rentziass:
This adds the ability to configure actions.Client to use custom root CAs. Getting said root CAs from a user-provided config map will come in the next PR, I wanted to keep this as small and contained as possible. 
The checked-in certificates/keys are test purposes only, they have been generated on the fly and haven't been used for anything else.

EDIT: I've also checked in the results of `go mod tidy`, looks like we were a bit behind.


github/actions/testdata/rootCA.crt
```diff
@@ -0,0 +1,18 @@
+-----BEGIN CERTIFICATE-----
```
TingluoHuang, line 1: do we want another test cert with CA chains?

TingluoHuang: I assume you generate this with `openssl`?

TingluoHuang: image

github/actions/client.go
```diff
@@ -892,7 +917,7 @@ func (c *Client) getActionsServiceAdminConnection(ctx context.Context, rt *regis
 
 	c.logger.Info("getting Actions tenant URL and JWT", "registrationURL", registrationURL.String())
 
-	resp, err := http.DefaultClient.Do(req)
+	resp, err := c.Do(req)
```
TingluoHuang, line 932: there is another `http.DefaultClient` in the `client.go` you might want to replace with `c.Do`

rentziass: Thank you!

github/actions/testdata/rootCA.crt
rentziass: Yeah I did as you can tell by the very non glamorous issuer 😂 We could test chains too, not sure much would change though as if the certificates are issued properly we're relying on the standard library for certificate parsing. I'll give it a spin!

github/actions/multi_client.go
```diff
@@ -168,3 +169,16 @@ func (m *multiClient) GetClientFromSecret(ctx context.Context, githubConfigURL,
 	auth.AppCreds = &GitHubAppAuth{AppID: parsedAppID, AppInstallationID: parsedAppInstallationID, AppPrivateKey: appPrivateKey}
 	return m.GetClientFor(ctx, githubConfigURL, auth, namespace)
 }
+
+func RootCAsFromConfigMap(configMapData map[string][]byte) (*x509.CertPool, error) {
+	caCertPool := x509.NewCertPool()
```
TingluoHuang: should we default to `x509.SystemCertPool()` so we can get all default certs on the machine?

TingluoHuang:
Do we need to expose an option to allow users to skip SSL validation?
the `actions/runner` allows the user to disable SSL validation via `export GITHUB_ACTIONS_RUNNER_TLS_NO_VERIFY=1`

rentziass:
> Do we need to expose an option to allow users to skip SSL validation?

@TingluoHuang added that [here](https://github.com/actions/actions-runner-controller/pull/2199/commits/b69305e4b0c2606984fcabff088369209595bec5), thank you!

github/actions/client.go
```diff
@@ -121,6 +138,37 @@ func NewClient(ctx context.Context, githubConfigURL string, creds *ActionsAuth,
...
+			// cleanhttp.DefaultPooledTransport()
+			return nil, fmt.Errorf("failed to get http transport from retryablehttp client")
+		}
+
+		if transport.TLSClientConfig == nil {
+			transport.TLSClientConfig = &tls.Config{}
+		}
+		transport.TLSClientConfig.InsecureSkipVerify = true
+		retryClient.HTTPClient.Transport = transport
+	}
```
TingluoHuang:
```suggestion
	if ac.tlsInsecureSkipVerify {
		transport, ok := retryClient.HTTPClient.Transport.(*http.Transport)
		if !ok {
			// this should always be true, because retryablehttp.NewClient() uses
			// cleanhttp.DefaultPooledTransport()
			return nil, fmt.Errorf("failed to get http transport from retryablehttp client")
		}

		if transport.TLSClientConfig == nil {
			transport.TLSClientConfig = &tls.Config{}
		}
		transport.TLSClientConfig.InsecureSkipVerify = true
		retryClient.HTTPClient.Transport = transport
	} else if ac.rootCAs != nil {
		transport, ok := retryClient.HTTPClient.Transport.(*http.Transport)
		if !ok {
			// this should always be true, because retryablehttp.NewClient() uses
			// cleanhttp.DefaultPooledTransport()
			return nil, fmt.Errorf("failed to get http transport from retryablehttp client")
		}

		if transport.TLSClientConfig == nil {
			transport.TLSClientConfig = &tls.Config{}
		}
		transport.TLSClientConfig.RootCAs = ac.rootCAs
		retryClient.HTTPClient.Transport = transport
	}

```

TingluoHuang: what about skip checking `ac.rootCAs` when `ac.tlsInsecureSkipVerify  == true`?

rentziass: I've just refactored that, do you mind taking a look? And also I don't think they should be exclusive, `InsecureSkipVerify` only kicks in if none of the provided CAs failed

TingluoHuang (dismissed):

github/actions/client.go
TingluoHuang: I like your refactor. 😄

TingluoHuang (dismissed):

nikola-jokic (approved):

TingluoHuang (approved):

PR #2208, opened 2023-01-25
chart: Create actionsmetrics.secrets.yaml

dhawalseth:
Adding the missing secret for actions metrics server which is reported in https://github.com/actions/actions-runner-controller/issues/2175

PR #2210, opened 2023-01-25
Update actionsmetrics.deployment.yaml

dhawalseth:
Adding the missing secret for actions metrics server which is reported in https://github.com/actions/actions-runner-controller/issues/2175


mumoshu:
Hi @dhawalseth Thank you for your efforts!
I've seen both this and https://github.com/actions/actions-runner-controller/pull/2208, and I prefer https://github.com/actions/actions-runner-controller/pull/2208 as I presume we're going to deprecate the githubwebhook server component once the new autoscalingrunnerset API becomes fully available.
WDYT?

dhawalseth:
> Hi @dhawalseth Thank you for your efforts! I've seen both this and #2208, and I prefer #2208 as I presume we're going to deprecate the githubwebhook server component once the new autoscalingrunnerset API becomes fully available. WDYT?

Hi @mumoshu , I see, yes in that case lets go with #2208. Thanks for the input.

PR #2215, opened 2023-01-26
Return error for non-existing runner group.

TingluoHuang:
The controller currently crashes when you provide a non-existing runner group.
image

https://github.com/github/c2c-actions-runtime/issues/2268

### DIFF
--- a/github/actions/client_runner_test.go
+++ b/github/actions/client_runner_test.go
@@ -173,3 +173,69 @@ func TestDeleteRunner(t *testing.T) {
 		assert.Equalf(t, actualRetry, expectedRetry, "A retry was expected after the first request but got: %v", actualRetry)
 	})
 }
+
+func TestGetRunnerGroupByName(t *testing.T) {
+	ctx := context.Background()
+	auth := &actions.ActionsAuth{
+		Token: "token",
+	}
+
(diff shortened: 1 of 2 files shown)


github/actions/client_runner_test.go
```diff
@@ -173,3 +173,69 @@ func TestDeleteRunner(t *testing.T) {
...
+			actualRetry++
+		}))
+
+		client, err := actions.NewClient(ctx, server.configURLForOrg("my-org"), auth, actions.WithRetryMax(retryMax), actions.WithRetryWaitMax(retryWaitMax))
+		require.NoError(t, err)
+
+		_, err = client.GetRunnerByName(ctx, runnerGroupName)
+		require.Error(t, err)
+		assert.Equalf(t, actualRetry, expectedRetry, "A retry was expected after the first request but got: %v", actualRetry)
+	})
```
rentziass: Not sure we need this now that the client is reused across all methods? If it's configured with retries all calls will be retried, what do you think?

rentziass (dismissed):
LGTM, just a question about testing retries

github/actions/client_runner_test.go
TingluoHuang: i will remove this test to save some CI time. 😄

rentziass (approved):

PR #2216, opened 2023-01-27
Allow update runner group for AutoScalingRunnerSet

TingluoHuang:
https://github.com/github/c2c-actions-runtime/issues/2267

Support update the deployed `AutoScalingRunnerSet` and move it between different runner groups.

Ex:
```
helm upgrade arc-test --namespace demo ./charts/auto-scaling-runner-set --set runnerGroup=newgroup
```

### DIFF
--- a/controllers/actions.github.com/autoscalingrunnerset_controller.go
+++ b/controllers/actions.github.com/autoscalingrunnerset_controller.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"sort"
 	"strconv"
+	"strings"
 
 	"github.com/actions/actions-runner-controller/apis/actions.github.com/v1alpha1"
 	"github.com/actions/actions-runner-controller/github/actions"
@@ -47,6 +48,7 @@ const (
 	LabelKeyAutoScaleRunnerSetName    = "auto-scale-runner-set-name"
 	autoscalingRunnerSetFinalizerName = "autoscalingrunnerset.actions.github.com/finalizer"
 	runnerScaleSetIdKey               = "runner-scale-set-id"
+	runnerScaleSetRunnerGroupNameKey  = "runner-scale-set-runner-group-name"
 
 	// scaleSetListenerLabel is the key of pod.meta.labels to label
 	// that the pod is a listener application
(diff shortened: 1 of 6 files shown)


controllers/actions.github.com/autoscalingrunnerset_controller_test.go
```diff
@@ -144,10 +145,14 @@ var _ = Describe("Test AutoScalingRunnerSet controller", func() {
...
-					return created.Annotations[runnerScaleSetIdKey], nil
+					if _, ok := created.Annotations[runnerScaleSetRunnerGroupNameKey]; !ok {
+						return "", nil
+					}
+
+					return fmt.Sprintf("%s_%s", created.Annotations[runnerScaleSetIdKey], created.Annotations[runnerScaleSetRunnerGroupNameKey]), nil
 				},
 				autoscalingRunnerSetTestTimeout,
-				autoscalingRunnerSetTestInterval).Should(BeEquivalentTo("1"), "RunnerScaleSet should be created/fetched and update the AutoScalingRunnerSet's annotation")
+				autoscalingRunnerSetTestInterval).Should(BeEquivalentTo("1_testgroup"), "RunnerScaleSet should be created/fetched and update the AutoScalingRunnerSet's annotation")
```
TingluoHuang, line 155: I need to refactor the test class a little bit before adding new tests, which I will do a follow-up PR to make sure the diff of this PR is reasonable.

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -338,17 +347,59 @@ func (r *AutoscalingRunnerSetReconciler) createRunnerScaleSet(ctx context.Contex
...
-	logger.Info("Adding runner scale set ID as an annotation")
-	if err := r.Update(ctx, autoscalingRunnerSet); err != nil {
-		logger.Error(err, "Failed to add runner scale set ID")
+	logger.Info("Adding runner scale set ID and runner group name as an annotation")
+	err = patch(ctx, r.Client, autoscalingRunnerSet, func(obj *v1alpha1.AutoscalingRunnerSet) {
+		obj.Annotations[runnerScaleSetIdKey] = strconv.Itoa(runnerScaleSet.Id)
+		obj.Annotations[runnerScaleSetRunnerGroupNameKey] = runnerScaleSet.RunnerGroupName
+	})
+
+	if err != nil && !kerrors.IsNotFound(err) {
```
nikola-jokic: Maybe pull in the case where this is not found to report it as such. Because if err is not found, then we should not log that we updated with runner scale set id as an annotation

TingluoHuang: updated. 🙇

nikola-jokic (approved):
LGTM

PR #2218, opened 2023-01-27
Add reposity name and full name for prometheus labels in actions metrics

kirillbilchenko:
Add more labels for produced metrics, to be able to build better dashboards, based on repository and organization.


pkg/actionsmetrics/event_reader.go
```diff
@@ -64,6 +64,9 @@ func (reader *EventReader) ProcessWorkflowJobEvent(ctx context.Context, event in
 	runsOn := strings.Join(e.WorkflowJob.Labels, `,`)
 	labels["runs_on"] = runsOn
 	labels["job_name"] = *e.WorkflowJob.Name
+	labels["repository"] = *e.Repo.Name
+	labels["repository_full_name"] = *e.Repo.FullName
+	labels["organization"] = *e.Org.Name
```
mumoshu: I was a bit concerned about the increase in cardinality. However, presuming the total number of time series would be `# of repos * # of jobs` anyway, these new labels won't contribute to high cardinality.. (Please correct me if I'm getting it wrong

pkg/actionsmetrics/event_reader.go
```diff
@@ -85,6 +88,9 @@ func (reader *EventReader) ProcessWorkflowJobEvent(ctx context.Context, event in
 			reader.Log.Info("reading workflow_job logs",
 				"job_name", *e.WorkflowJob.Name,
 				"job_id", fmt.Sprint(*e.WorkflowJob.ID),
+				"repository", *e.Repo.Name,
+				"repository_full_name", *e.Repo.FullName,
+				"organization", *e.Org.Name,
```
mumoshu: I think this isn't safe because `Org` can be nil for repositories owned by a `User`. Let's add a nil check and add this label only when the workflow job is coming from a repo owned by an org!

kirillbilchenko: Updated

pkg/actionsmetrics/event_reader.go
kirillbilchenko: Updated the code, but yes cardinality most probably will increase but not significantly

mumoshu:
@kirillbilchenko Thanks for your support! I've also added my own commit for even more extensive nil checks. [e65904d](https://github.com/actions/actions-runner-controller/pull/2218/commits/e65904d350ec44d9b388fab46ab32878f9db589c)

mumoshu (approved):
LGTM. Thank you for your contribution @kirillbilchenko ☺️

mumoshu:
Apparently, we had to include those new labels into the metrics definitions. See #2225

PR #2219, opened 2023-01-27
Remove network requests from actions.NewClient

rentziass:
This PR allows us to create new clients without making any HTTP requests. This is needed for allowing us to more effectively cache clients and update them when needed.

Since we were extracting auth logic from NewClient we refactored all methods in Client using that logic by extracting into two http.Request constructors.

This also introduces GitHubConfig in replacement of config url used as a string to simplify logic around hosts, hosted, etc.

/cc @nikola-jokic


nikola-jokic (dismissed):
LGTM

github/actions/client.go
```diff
@@ -161,48 +161,94 @@ func NewClient(ctx context.Context, githubConfigURL string, creds *ActionsAuth,
...
+	body = trimByteOrderMark(body)
+	resp.Body = io.NopCloser(bytes.NewReader(body))
+	return resp, nil
+}
 
-	ac.mu.Lock()
-	defer ac.mu.Unlock()
-	ac.ActionsServiceAdminToken = adminConnInfo.AdminToken
-	ac.ActionsServiceAdminTokenExpiresAt, err = actionsServiceAdminTokenExpiresAt(*adminConnInfo.AdminToken)
+func (c *Client) NewGitHubAPIRequest(ctx context.Context, method, path string, body io.Reader) (*http.Request, error) {
```
TingluoHuang, line 187: `NewGitHubAPIRequest` or `newGitHubAPIRequest`? do we call this method publically?

github/actions/client.go
```diff
@@ -161,48 +161,94 @@ func NewClient(ctx context.Context, githubConfigURL string, creds *ActionsAuth,
...
+	if c.userAgent != "" {
+		req.Header.Set("User-Agent", c.userAgent)
+	}
+
+	return req, nil
 }
 
-func (c *Client) GetRunnerScaleSet(ctx context.Context, runnerScaleSetName string) (*RunnerScaleSet, error) {
-	u := fmt.Sprintf("%s/%s?name=%s&api-version=6.0-preview", *c.ActionsServiceURL, scaleSetEndpoint, runnerScaleSetName)
+func (c *Client) NewActionsServiceRequest(ctx context.Context, method, path string, body io.Reader) (*http.Request, error) {
```
TingluoHuang, line 201: same, `NewActionsServiceRequest` or `newActionsServiceRequest`?

github/actions/client.go
rentziass: Not at the moment, no. The idea behind having these two exported was that it can build a request if you want to do something that's not covered by the client directly (which now that I type it out I'm not sure it's a good idea anymore)

github/actions/client.go
```diff
@@ -587,24 +567,18 @@ func (c *Client) doSessionRequest(ctx context.Context, method, url string, reque
...
-	u := fmt.Sprintf("%s/%s/%d/acquirejobs?api-version=6.0-preview", *c.ActionsServiceURL, scaleSetEndpoint, runnerScaleSetId)
+	path := fmt.Sprintf("/%s/%d/acquirejobs", scaleSetEndpoint, runnerScaleSetId)
 
 	body, err := json.Marshal(requestIds)
 	if err != nil {
 		return nil, err
 	}
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, u, bytes.NewBuffer(body))
+	req, err := c.NewActionsServiceRequest(ctx, http.MethodPost, path, bytes.NewBuffer(body))
```
TingluoHuang: `http.NewRequestWithContext` since we are passing the `messageQueueAccessToken` token?

rentziass: Whew yes good catch (one more reason to have an independent client with message queue things :) ). How is this working like that though?

rentziass: Just curious, how come `GetAcquirableJobs` uses `c.ActionsServiceAdminToken` and `AcquireJobs` uses the message queue token?

TingluoHuang: Everyone with `ActionsServiceAdminToken` can list acquirable jobs via `GetAcquirableJobs`, but only the active listener can `POST` to acquire the job via `AcquireJobs`

rentziass: Ah gotcha, thanks :)

nikola-jokic (approved):

PR #2222, opened 2023-01-28
Fix helm charts when pass values file. 

TingluoHuang:
Apparently, pass value via `--set foo=bar` and `--file values.yaml` will end up with a different data type. 😢 

Ex:
`--set foo=5` -> `typeOf(foo)=int64`

`--file values.yaml` with `foo: 5` in the yaml -> `typeOf(foo)=float64`

### DIFF
--- a/charts/.ci/ct-config.yaml
+++ b/charts/.ci/ct-config.yaml
@@ -1,4 +1,5 @@
 # This file defines the config for "ct" (chart tester) used by the helm linting GitHub workflow
+all: true
 lint-conf: charts/.ci/lint-config.yaml
 chart-repos:
   - jetstack=https://charts.jetstack.io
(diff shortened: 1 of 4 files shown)


charts/.ci/ct-config.yaml
```diff
@@ -1,4 +1,5 @@
 # This file defines the config for "ct" (chart tester) used by the helm linting GitHub workflow
+all: true
```
TingluoHuang, line 2: Our `AutoScalingRunnerSet` depends on the `actions-runner-controller-2` chart being installed.
The CT will try to install charts that only have changes by default which means the `AutoScalingRunnerSet` will fail to install since the CRDs are not there.
I am updating the `ct-config` to install all charts by default.

nikola-jokic (dismissed):

TingluoHuang:
~~Still have some bugs... need to keep digging in.~~

@nikola-jokic 

Fixed, with L0 test added. 🙇

rentziass (approved):

PR #2223, opened 2023-01-28
Delete RunnerScaleSet on service when AutoScalingRunnerSet is deleted.

TingluoHuang:
On the `AutoScalingRunnerSet` deletion, after cleanup of all resources of the `AutoScalingRunnerSet`, the controller should delete the `RunnerScaleSet` on the Actions service as well since it was created by the controller.

### DIFF
--- a/controllers/actions.github.com/autoscalingrunnerset_controller.go
+++ b/controllers/actions.github.com/autoscalingrunnerset_controller.go
@@ -379,13 +379,18 @@ func (r *AutoscalingRunnerSetReconciler) updateRunnerScaleSetRunnerGroup(ctx con
 		return ctrl.Result{}, err
 	}
 
-	runnerGroup, err := actionsClient.GetRunnerGroupByName(ctx, autoscalingRunnerSet.Spec.RunnerGroup)
-	if err != nil {
-		logger.Error(err, "Failed to get runner group by name", "runnerGroup", autoscalingRunnerSet.Spec.RunnerGroup)
-		return ctrl.Result{}, err
+	runnerGroupId := 1
+	if len(autoscalingRunnerSet.Spec.RunnerGroup) > 0 {
+		runnerGroup, err := actionsClient.GetRunnerGroupByName(ctx, autoscalingRunnerSet.Spec.RunnerGroup)
+		if err != nil {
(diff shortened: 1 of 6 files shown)


controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -154,7 +160,7 @@ func (r *AutoscalingRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl
 
 	// Make sure the runner group of the scale set is up to date
 	currentRunnerGroupName, ok := autoscalingRunnerSet.Annotations[runnerScaleSetRunnerGroupNameKey]
-	if !ok || !strings.EqualFold(currentRunnerGroupName, autoscalingRunnerSet.Spec.RunnerGroup) {
+	if !ok || (len(autoscalingRunnerSet.Spec.RunnerGroup) > 0 && !strings.EqualFold(currentRunnerGroupName, autoscalingRunnerSet.Spec.RunnerGroup)) {
```
TingluoHuang, line 163: `autoscalingRunnerSet.Spec.RunnerGroup` can be empty, which means we are using the `default` runner group, so we don't need to do any updates.

nikola-jokic: Do we need to check if annotation exists? The only way we can get to this part is that we created a scale-set and in that method, we are adding the annotation. So the only way to hit this part is by having scale set created, and with that, the annotation: https://github.com/actions/actions-runner-controller/blob/d0c6a96fca1589233c7de751fb4bc5703e430475/controllers/actions.github.com/autoscalingrunnerset_controller.go#L359

TingluoHuang: I will add a check since technically user can remove it by accident.

TingluoHuang: Actually, should the `!ok||` take care the case?

rentziass: Yeah, I think we're good as is.

rentziass (approved):

nikola-jokic (approved):

PR #2225, opened 2023-01-30
feat(actionsmetrics): Add owner and workflow_name labels to workflow job metrics

### LINKED ISSUE #2176: Add more labels to metrics-server metrics

The [metrics](https://github.com/actions/actions-runner-controller/blob/0324658a3fad79237b5da9c21a3ce6f6cc9a4429/pkg/actionsmetrics/metrics.go) emitted from the metrics-server currently only include `job` and `runs_on` labels.

I think `repository` is required at a minimum for this to be useful, and probably `organisation` too?

[edit]
Actually `workflow_name` is probably required as well
Although workflow name has only [just](https://github.com/google/go-github/pull/2630) been added to the go-github library!
[/edit]

### Why is this needed?

If I have 2 Github repos (`A` and `B`) and they both have a workflow jobs called `test` and `build` (for example) the metrics aren't differentiated, I have no idea how many test jobs ran for repo A vs repo B, only how many ran for all of them.

We would get metrics like:

```
github_workflow_jobs_completed_total{job_name="test",runs_on="ubuntu-latest"} 123
github_workflow_jobs_completed_total{job_name="build",runs_on="ubuntu-latest"} 123
```

To be useful we'd want:

```
github_workflow_jobs_completed_total{organisation="foo",repository="A",job_name="test",runs_on="ubuntu-latest"} 23
github_workflow_jobs_completed_total{organisation="bar,repository="B",job_name="test",runs_on="ubuntu-latest"} 100
github_workflow_jobs_completed_total{organisation="foo",repository="A",job_name="build",runs_on="ubuntu-latest"} 23
github_workflow_jobs_completed_total{organisation="bar",repository="B",job_name="build",runs_on="ubuntu-latest"} 100
```

hamishforbes:
Fixes #2176

### DIFF
--- a/pkg/actionsmetrics/event_reader.go
+++ b/pkg/actionsmetrics/event_reader.go
@@ -79,14 +79,34 @@ func (reader *EventReader) ProcessWorkflowJobEvent(ctx context.Context, event in
 			labels["repository_full_name"] = *n
 			keysAndValues = append(keysAndValues, "repository_full_name", *n)
 		}
+
+		if e.Repo.Owner != nil {
+			if l := e.Repo.Owner.Login; l != nil {
+				labels["owner"] = *l
+				keysAndValues = append(keysAndValues, "owner", *l)
+			}
+		}
 	}
 
+	var org string
 	if e.Org != nil {
 		if n := e.Org.Name; n != nil {
-			labels["organization"] = *e.Org.Name
+			org = *n
 			keysAndValues = append(keysAndValues, "organization", *n)
 		}
 	}
+	labels["organization"] = org
+
+	var wn string
+	if e.WorkflowJob != nil {
+		if n := e.WorkflowJob.WorkflowName; n != nil {
+			wn = *n
+			keysAndValues = append(keysAndValues, "workflow_name", *n)
+		}
+	}
+	labels["workflow_name"] = wn
+
+	log := reader.Log.WithValues(keysAndValues...)
 
 	// switch on job status
 	switch action := e.GetAction(); action {
@@ -102,14 +122,10 @@ func (reader *EventReader) ProcessWorkflowJobEvent(ctx context.Context, event in
 
 		parseResult, err := reader.fetchAndParseWorkflowJobLogs(ctx, e)
 		if err != nil {
-			reader.Log.Error(err, "reading workflow job log")
+			log.Error(err, "reading workflow job log")
 			return
 		} else {
-			reader.Log.WithValues("job_name", *e.WorkflowJob.Name, "job_id", fmt.Sprint(*e.WorkflowJob.ID), "repository", *e.Repo.Name, "repository_full_name", *e.Repo.FullName)
-			if len(*e.Org.Name) > 0 {
-				reader.Log.WithValues("organization", *e.Org.Name)
-			}
-			reader.Log.Info("reading workflow_job logs")
+			log.Info("reading workflow_job logs")
 		}
 
 		githubWorkflowJobQueueDurationSeconds.With(labels).Observe(parseResult.QueueTime.Seconds())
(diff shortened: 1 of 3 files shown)


pkg/actionsmetrics/metrics.go
```diff
@@ -72,55 +72,56 @@ var (
...
+		commonLabels,
 	)
 	githubWorkflowJobRunDurationSeconds = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
 			Name:    "github_workflow_job_run_duration_seconds",
 			Help:    "Run times for workflow jobs in seconds",
 			Buckets: runtimeBuckets,
 		},
-		[]string{"runs_on", "job_name", "job_conclusion"},
+		append(commonLabels, "job_conclusion"),
```
mumoshu: I believe this usage of `append` isn't safe in Go and can produce unexpected results for subsequent `append(commonLabels, ...)` calls, because append can mutate the slice.

hamishforbes: Ah interesting, sounds like that's more of an issue when using the colon slice notation but yeah seems sketchy!
Would you prefer to just use a literal and duplicate the common labels for each metric or try and retain the re-usable common label definition?

mumoshu: I do like the idea of explicitly extracting and defining commonLabels!
How about adding a function to create a new copy of commonLabels, add some extra labels, and return the resulting slice? That way, we can repeat the function calls in places we currently refer to `commonLabels`.

mumoshu: something lile:

```
func metricLabels(extras ...string) {
    return append(append([[]string{}, commonLabels...), extras...)
}
```

so that we can rewrite this line to
```suggestion
		metricLabels("job_conclusion"),
```

mumoshu:
I recently merged #2218, and apparently, it contained half of this pull request, resulting in a conflict. Let me try resolving it.

pkg/actionsmetrics/event_reader.go
```diff
@@ -64,6 +64,9 @@ func (reader *EventReader) ProcessWorkflowJobEvent(ctx context.Context, event in
 	runsOn := strings.Join(e.WorkflowJob.Labels, `,`)
 	labels["runs_on"] = runsOn
 	labels["job_name"] = *e.WorkflowJob.Name
+	labels["organization"] = *e.Repo.Owner.Login
```
mumoshu: @hamishforbes I realized #2218 added the organization label with another value of `*e.Org.Name`. If we call it "organization", that might be the correct value.

However, I do think it might be a good idea to be able to see `*e.Repo.Owner.Login` via labels when the workflow job event is coming from a runner associated with a repository owned by a user, not an organization.

That said, would you like to include `*e.Repo.Owner.Login` as another label, like "owner"?

mumoshu:
@hamishforbes Hey! I've addressed my own comments in the lastest commit. The label is now `owner` and we use `metricLabels` function to copy commonLabels and add some custom labels. WDYT?

hamishforbes:
> @hamishforbes Hey! I've addressed my own comments in the lastest commit. The label is now `owner` and we use `metricLabels` function to copy commonLabels and add some custom labels. WDYT?

Looks good to me!

I do think `branch` as an optional label would also be good, but that can be another PR/Issue.

mumoshu:
@hamishforbes Thanks for your review! Re `branch`, it turned out that github-go does not expose workflow job's "head_branch" which is what I believe needed for our use case.
I'll try to submit a tiny pull request to github-go. Once it gets merged, we are good to go!

mumoshu:
Fixed some panics related to the logger and the metrics labels for the case we feed workflow jobs from a repository instead of an org.

mumoshu (approved):
Looks like it's working well for me! Let me merge this and keep testing locally.
Thank you so much for your effort @hamishforbes!

PR #2234, opened 2023-01-31
Allow provide pre-defined kubernetes secret when helm-install AutoScalingRunnerSet

TingluoHuang:
In case the customer has created a secret with all the required data, we will just reference it instead of creating a new one.

### DIFF
--- a/charts/auto-scaling-runner-set/values.yaml
+++ b/charts/auto-scaling-runner-set/values.yaml
@@ -13,6 +13,14 @@ githubConfigSecret:
 
   ### GitHub PAT Configuration
   github_token: ""
+## If you have a pre-define Kubernetes secret in the same namespace the auto-scaling-runner-set is going to deploy,
+## you can also reference it via `githubConfigSecret: pre-defined-secret`.
+## You need to make sure your predefined secret has all the required secret data set properly.
+##   For a pre-defined secret using GitHub PAT, the secret needs to be created like this:
+##   > kubectl create secret generic pre-defined-secret --namespace=my_namespace --from-literal=github_token='ghp_your_pat'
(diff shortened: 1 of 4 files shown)


rentziass (commented):
Do we require the secret to follow a certain structure? Do we have a way to tell if we're dealing with a PAT or a GitHub app? (There's a good chance I'm missing something)

TingluoHuang:
> Do we require the secret to follow a certain structure? Do we have a way to tell if we're dealing with a PAT or a GitHub app? (There's a good chance I'm missing something)

@rentziass we do, I am not quite sure how to enforce it in the helm level, so just left a comment in the [values.yaml](https://github.com/actions/actions-runner-controller/pull/2234/files#diff-7183f737213e8105334652650f1e926a560f6941b6d09fbda892a47aa4e80b8bR18)

charts/auto-scaling-runner-set/values.yaml
```diff
@@ -13,6 +13,10 @@ githubConfigSecret:
 
   ### GitHub PAT Configuration
   github_token: ""
+## If you have a pre-define kubernetes secret in the same namespace the auto-scaling-runner-set is going to deploy,
+## you can also reference it via `githubConfigSecret: predefinedSecret`.
+## You need to make sure your predefined secret has all required secret data set properly, ex, github_token/github_app_id/github_app_installation_id/github_app_private_key
```
Link-: nit: maybe we add an example for how to create the secret with the appropriate structure in our docs and reference it here? OR - just add the kubectl command to create the secret with the adequate structure

Link- (dismissed):
:shipit:

Link- (dismissed):

johnsudol (approved):

PR #2237, opened 2023-02-01
Add Identifier to actions.Client

rentziass:
This adds an Identifier method to actions.Client to be able to uniquely identify clients and improve caching them.

### DIFF
--- a/github/actions/client.go
+++ b/github/actions/client.go
@@ -170,15 +170,15 @@ func NewClient(githubConfigURL string, creds *ActionsAuth, options ...ClientOpti
 // change warrants creating a new client. Any changes to Client that would
 // require a new client should be reflected here.
 func (c *Client) Identifier() string {
-	identifier := fmt.Sprintf("configURL:%s,", c.config.ConfigURL.String())
+	identifier := fmt.Sprintf("configURL:%q,", c.config.ConfigURL.String())
 
 	if c.creds.Token != "" {
-		identifier += fmt.Sprintf("token:%s", c.creds.Token)
+		identifier += fmt.Sprintf("token:%q", c.creds.Token)
 	}
 
 	if c.creds.AppCreds != nil {
 		identifier += fmt.Sprintf(
-			"appID:%d,installationID:%d,key:%s",
+			"appID:%q,installationID:%q,key:%q",
 			c.creds.AppCreds.AppID,
 			c.creds.AppCreds.AppInstallationID,
 			c.creds.AppCreds.AppPrivateKey,
--- a/github/actions/client.go
+++ b/github/actions/client.go
@@ -165,6 +165,29 @@ func NewClient(githubConfigURL string, creds *ActionsAuth, options ...ClientOpti
 	return ac, nil
 }
 
+// Identifier returns a string to help identify a client uniquely.
+// This is used for caching client instances and understanding when a config
+// change warrants creating a new client. Any changes to Client that would
+// require a new client should be reflected here.
+func (c *Client) Identifier() string {
+	identifier := fmt.Sprintf("configURL:%s,", c.config.ConfigURL.String())
+
+	if c.creds.Token != "" {
+		identifier += fmt.Sprintf("token:%s", c.creds.Token)
+	}
+
+	if c.creds.AppCreds != nil {
+		identifier += fmt.Sprintf(
+			"appID:%d,installationID:%d,key:%s",
+			c.creds.AppCreds.AppID,
+			c.creds.AppCreds.AppInstallationID,
+			c.creds.AppCreds.AppPrivateKey,
+		)
+	}
+
+	return uuid.NewMD5(uuid.NameSpaceOID, []byte(identifier)).String()
+}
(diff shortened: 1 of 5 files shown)


github/actions/config_test.go
```diff
@@ -91,18 +91,19 @@ func TestGitHubConfig(t *testing.T) {
...
-		"https://github.com/",
-		"https://github.com",
-		"https://github.com/some/random/path",
-	}
+	t.Run("when given an invalid URL", func(t *testing.T) {
+		invalidURLs := []string{
+			"https://github.com/",
+			"https://github.com",
+			"https://github.com/some/random/path",
+		}
```
JoannaaKL, line 100: 👍

JoannaaKL (dismissed):

github/actions/multi_client.go
```diff
@@ -105,16 +96,6 @@ func (m *multiClient) GetClientFor(ctx context.Context, githubConfigURL string,
...
-		githubConfigURL,
-		&creds,
-		WithUserAgent(m.userAgent),
-		WithLogger(m.logger),
-	)
-	if err != nil {
-		return nil, err
-	}
-
 	m.clients[key] = &actionsClientWrapper{
```
nikola-jokic: Maybe we can include in this PR not to have client wrapper
wrapper has a mutex that should guard refresh, but itt is never accessed since the client itself is guarding against refresh right? We don't really need a wrapper as far as I can tell

github/actions/client.go
```diff
@@ -165,6 +165,29 @@ func NewClient(githubConfigURL string, creds *ActionsAuth, options ...ClientOpti
 	return ac, nil
 }
 
+// Identifier returns a string to help identify a client uniquely.
+// This is used for caching client instances and understanding when a config
+// change warrants creating a new client. Any changes to Client that would
+// require a new client should be reflected here.
+func (c *Client) Identifier() string {
+	identifier := fmt.Sprintf("configURL:%s,", c.config.ConfigURL.String())
```
nikola-jokic: Suggestion (ignore me if you don't agree :relaxed:)
Should we use %q  for strings? It would not add any value for key since it is being hashed but it might be easier for debugging to have more visible separation of individual parts

rentziass: Actually very nice, thank you!

github/actions/multi_client.go
rentziass: LOVE IT ❤️

nikola-jokic (approved):
LGTM

github/actions/client.go
```diff
@@ -165,6 +165,29 @@ func NewClient(githubConfigURL string, creds *ActionsAuth, options ...ClientOpti
...
+	if c.creds.AppCreds != nil {
+		identifier += fmt.Sprintf(
+			"appID:%q,installationID:%q,key:%q",
+			c.creds.AppCreds.AppID,
+			c.creds.AppCreds.AppInstallationID,
+			c.creds.AppCreds.AppPrivateKey,
+		)
+	}
+
+	return uuid.NewMD5(uuid.NameSpaceOID, []byte(identifier)).String()
```
TingluoHuang, line 188: can we use `SHA256` instead of `MD5`?
the GitHub/MSFT security audit won't be happy if we use `MD5`, you will need to file some exemption every time.

rentziass: We're not really encrypting anything with it, how about the vanilla [`NewUUID`](https://pkg.go.dev/github.com/google/UUID#NewUUID)?

github/actions/multi_client.go
```diff
@@ -77,34 +63,6 @@ func (m *multiClient) GetClientFor(ctx context.Context, githubConfigURL string,
...
-
-	clientWrapper, has := m.clients[key]
-	if has {
-		m.logger.Info("using cache client", "githubConfigURL", githubConfigURL, "namespace", namespace)
-		return clientWrapper.client, nil
-	}
-
-	m.logger.Info("creating new client", "githubConfigURL", githubConfigURL, "namespace", namespace)
-
 	client, err := NewClient(
```
TingluoHuang, line 66: should we only create a new client on a cache miss?

github/actions/client.go
TingluoHuang: I know, that's what I mean by file exemption about `we don't use it for encrypting`.

github/actions/multi_client.go
nikola-jokic: Then the identity of a client should be moved to some other entity which might not be ideal...
The client should know how to identify itself, and since NewClient does not issue any other calls when instantiated, it should not be expensive anymore

rentziass: We use a client.Identifier now as a caching key so we need the "potential" new client to understand if we have one like it already in the cache, but creating a new client is "free" now (aka no network calls) until you use to make some calls.

github/actions/client.go
rentziass: `NewUUID` it is, I'm convinced, thanks 😄

PR #2238, opened 2023-02-01
Update configuring-windows-runners.md

ealasgarov:
insure latest runner version is installed


ealasgarov (commented):
insure the latest runner version is installed

PR #2239, opened 2023-02-01
Avoid deleting scale set if annotation is not parsable or if it does not exist

nikola-jokic:
If autoscaling runner set is applied without secret, the controller will not be able to create scale set.
It does not annotate the resource with an ID.

If you try to manually delete the autoscaling runner set, the controller will get stuck reporting an error trying to parse scale set id.
This change skips the deletion if annotation contains invalid ID or if it does not exist. In that case, deletion of a resource will proceed and cleanup would be successful

### DIFF
--- a/controllers/actions.github.com/autoscalingrunnerset_controller.go
+++ b/controllers/actions.github.com/autoscalingrunnerset_controller.go
@@ -412,8 +412,11 @@ func (r *AutoscalingRunnerSetReconciler) deleteRunnerScaleSet(ctx context.Contex
 	logger.Info("Deleting the runner scale set from Actions service")
 	runnerScaleSetId, err := strconv.Atoi(autoscalingRunnerSet.Annotations[runnerScaleSetIdKey])
 	if err != nil {
-		logger.Error(err, "Failed to parse runner scale set ID")
-		return err
+		// If the annotation is not set correctly, or if it does not exist, we are going to get stuck in a loop trying to parse the scale set id.
+		// If the configuration is invalid (secret does not exist for example), we never get to the point to create runner set. But then, manual cleanup
+		// would get stuck finalizing the resource trying to parse annotation indefinitely
+		logger.Info("autoscaling runner set does not have annotation describing scale set id. Skip deletion")
+		return nil
 	}
 
 	actionsClient, err := r.actionsClientFor(ctx, autoscalingRunnerSet)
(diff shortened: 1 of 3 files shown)


controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -412,8 +412,11 @@ func (r *AutoscalingRunnerSetReconciler) deleteRunnerScaleSet(ctx context.Contex
 	logger.Info("Deleting the runner scale set from Actions service")
 	runnerScaleSetId, err := strconv.Atoi(autoscalingRunnerSet.Annotations[runnerScaleSetIdKey])
 	if err != nil {
-		logger.Error(err, "Failed to parse runner scale set ID")
-		return err
+		// If the annotation is not set correctly, or if it does not exist, we are going to get stuck in a loop trying to parse the scale set id.
+		// If the configuration is invalid (secret does not exist for example), we never get to the point to create runner set. But then, manual cleanup
+		// would get stuck finalizing the resource trying to parse annotation indefinitely
+		logger.Info("autoscaling runner set does not have annotation describing scale set id. Skip deletion")
```
TingluoHuang: do we want to add an L0 for this case?
should we trace the `err` out?

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -412,9 +412,13 @@ func (r *AutoscalingRunnerSetReconciler) deleteRunnerScaleSet(ctx context.Contex
...
 	if err != nil {
-		logger.Error(err, "Failed to parse runner scale set ID")
-		return err
+		// If the annotation is not set correctly, or if it does not exist, we are going to get stuck in a loop trying to parse the scale set id.
+		// If the configuration is invalid (secret does not exist for example), we never get to the point to create runner set. But then, manual cleanup
+		// would get stuck finalizing the resource trying to parse annotation indefinitely
+		logger.Info("autoscaling runner set does not have annotation describing scale set id. Skip deletion", "err", err.Error())
+		return nil
 	}
+	fmt.Println("Well, that is wierd")
```
rentziass: A debug survivor? :)

nikola-jokic: hahaha yeah, sorry for that :smile:

rentziass:
```suggestion
```

rentziass (dismissed):

rentziass (approved):

PR #2241, opened 2023-02-01
Use UUID v5 for client identifiers

rentziass:


JoannaaKL (dismissed):
I like using the strongest possible encryption algorithm! 💪

TingluoHuang (approved):

PR #2244, opened 2023-02-02
ADR: automate runner updates

rentziass:
[Rendered](https://github.com/actions/actions-runner-controller/blob/rentziass/auto-runner-update/adrs/2023-02-02-automate-runner-updates.md)


adrs/2023-02-02-automate-runner-updates.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Context
+
+When a new [runner](https://github.com/actions/runner) version is released, new
+images need to be built in
+[actions-runner-controller/releases](https://github.com/actions-runner-controller/releases).
+This is currently started by the
+[release-runners](https://github.com/actions/actions-runner-controller/blob/master/.github/workflows/release-runners.yaml)
+workflow, although this only starts when the set of file containing the runner
+version is updated (and this is currently done manually).
```
Link-, line 13: One thing I don't like at all about our current runner release process is the need to update all these files: https://github.com/actions/actions-runner-controller/pull/2182 

I'm not sure why all these values are needed. I'm adding this here for full awareness.

Link-: Could we also add the alternatives and why we dismissed them?

adrs/2023-02-02-automate-runner-updates.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Decision
+
+We can have another workflow running on a cadence (hourly seems sensible) and checking for new runner
+releases, creating a PR with the updated version if there is one. Once that PR
+is merged, the existing workflow will pick things up.
```
Link-: Could we also add the alternatives and why we dismissed them?

Link- (changes requested):
Small nits

adrs/2023-02-02-automate-runner-updates.md
rentziass: These changes are needed because what actions-runner-controller/releases does is pull this ARC repo and [use these Dockerfiles to build images](https://github.com/actions-runner-controller/releases/blob/main/.github/workflows/release-runners.yaml#L100-L122).

adrs/2023-02-02-automate-runner-updates.md
rentziass: On it, thanks 👍

adrs/2023-02-02-automate-runner-updates.md
Link-: Oh sorry I wasn't clear, what I meant to say is: I'm not confident that this should be the way to go moving forward. We should be able to pass all these arguments at runtime! I know how they're used because I migrated the pipelines :)

rentziass: Ah gotcha, it seemed weird 😄 ehm you're right, it doesn't look like they're needed, we're passing the [runner version](https://github.com/actions-runner-controller/releases/blob/main/.github/workflows/release-runners.yaml#L109) as an ARG anyway. Files that need updating then:
- `.github/workflows/release-runners.yaml`
- `Makefile`
- `runner/Makefile`
- `test/e2e/e2e_test.go`

Link- (approved):
We're aligned, thank you for drafting this!

PR #2247, opened 2023-02-03
Add new workflow to automate runner updates

rentziass:
## New workflow: `update-runners`

This adds a new workflow that polls actions/runner releases hourly, and in case there is a new version will open a pull request that, if merged, will trigger the usual [`release-runners`](.github/workflows/release-runners.yaml) workflow. The PR will contain changes to:
- `runner/VERSION`
- `runner/Makefile`
- `Makefile`
- `test/e2e/e2e_test.go`

## Changes to `release-runners` and existing Dockerfiles

`release-runners` has been changed to read its runner version from `runner/VERSION`: this is because automating a workflow definition requires a special PAT with `workflow` permission. As a side effect this also makes version changes a bit more "intentional" imo.

This is also removing default values for `RUNNER_VERSION` from the various Dockerfiles as we're providing this as [build args anyway](https://github.com/actions-runner-controller/releases/blob/main/.github/workflows/release-runners.yaml#L109).

Below is a sample screenshots of what the automated PR will look like:
Screenshot 2023-02-03 at 16 30 09


.github/workflows/release-runners.yaml
```diff
@@ -3,23 +3,19 @@ name: Runners
...
   # are available to the workflow run
   push:
     branches:
       - 'master'
     paths:
-      - 'runner/**'
-      - '!runner/Makefile'
-      - '.github/workflows/runners.yaml'
-      - '!**.md'
+      - 'runner/VERSION'
```
Link-, line 12: This doesn't factor in the changes (updated) to the container-hooks version or the docker version

Link- (changes requested):
I like this approach ✨ thank you for putting it together

.github/workflows/release-runners.yaml
rentziass: We weren't before either? Adding this workflow file to `paths`.

nikola-jokic (dismissed):

.github/workflows/release-runners.yaml
rentziass: Unless you meant updating those automatically too, but discussing approaches yesterday it seems we're very intentional with releasing new container hooks versions and they're not necessarily tied to runner releases.

Link-: Nevermind this comment, I didn't see this line 🤦‍♂️ we're all good here

`- '.github/workflows/release-runners.yaml'`

.github/workflows/update-runners.yaml
```diff
@@ -0,0 +1,107 @@
...
+        env:
+          LATEST_VERSION: ${{ needs.check_versions.outputs.latest_version }}
+        run: |
+          PR_NAME="Update runner to version ${LATEST_VERSION}"
+
+          result=$(gh pr list --search "$PR_NAME" --json number --jq ".[].number" --limit 1)
+          if [ -z "$result" ]
+          then
+          echo "No existing PRs found, setting output with pr_name=$PR_NAME"
+          echo pr_name=$PR_NAME >> $GITHUB_OUTPUT
```
Link-: nit: indentation for readability

.github/workflows/update-runners.yaml
```diff
@@ -0,0 +1,107 @@
...
+          else
+          echo "Found a PR with title '$PR_NAME' already existing: ${{ github.server_url }}/${{ github.repository }}/pull/$result"
```
Link-: nit: indentation for readability

.github/workflows/update-runners.yaml
```diff
@@ -0,0 +1,107 @@
+# This workflows polls releases from actions/runner and in case of a new one it
+# updates files containing runner version and opens a pull request.
+name: Update runners
+
+on:
+  schedule:
+    # run hourly
+    - cron: "0 * * * *"
```
Link-: I'm thinking out loud here - running this hourly will be quite noisy here: https://github.com/actions/actions-runner-controller/actions 

Maybe we should rethink this frequency

Link- (dismissed):

.github/workflows/update-runners.yaml
rentziass: Do we want to try daily first? I'm down with that, I had started with hourly only to make this solution more compelling against the real-time solution :)

.github/workflows/update-runners.yaml
```diff
@@ -5,7 +5,7 @@ name: Update runners
 on:
   schedule:
     # run hourly
```
nikola-jokic: Update comment above :relaxed:

.github/workflows/update-runners.yaml
```diff
@@ -0,0 +1,107 @@
+# This workflows polls releases from actions/runner and in case of a new one it
+# updates files containing runner version and opens a pull request.
+name: Update runners
+
+on:
+  schedule:
+    # run hourly
```
nikola-jokic:
```suggestion
    # run daily
```

nikola-jokic (approved):

Link- (approved):

PR #2252, opened 2023-02-06
Added ability to configure log level from chart values

AvaStancu:


charts/actions-runner-controller-2/templates/deployment.yaml
```diff
@@ -48,6 +48,9 @@ spec:
         {{- with .Values.imagePullSecrets }}
         - "--auto-scaler-image-pull-secrets={{ include "actions-runner-controller-2.imagePullSecretsNames" . }}"
         {{- end }}
+        {{- if .Values.logging }}
```
nikola-jokic: Should we expose all flags supported and just apply them all? Or at least, all that arc-2 app supports? For example, disable webhook etc

AvaStancu: Maybe, we can discuss this but if we decide to do it, those can be added in a separate PR

charts/actions-runner-controller-2/values.yaml
```diff
@@ -62,4 +62,9 @@ affinity: {}
 # Leverage a PriorityClass to ensure your pods survive resource shortages
 # ref: https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/
 # PriorityClass: system-cluster-critical
-priorityClassName: ""
\ No newline at end of file
+priorityClassName: ""
+
+logging:
```
nikola-jokic: To anticipate other changes to the manager, maybe we should introduce a top level struct like following:
```yaml
manager:
  flags:
    logLevel: "debug"
```
What do you think? :relaxed:

rentziass (dismissed):
:shipit: 🚀

charts/actions-runner-controller-2/templates/deployment.yaml
```diff
@@ -48,6 +48,9 @@ spec:
         {{- with .Values.imagePullSecrets }}
         - "--auto-scaler-image-pull-secrets={{ include "actions-runner-controller-2.imagePullSecretsNames" . }}"
         {{- end }}
+        {{- if and .Values.manager .Values.manager.flags }}
```
TingluoHuang: do you need to check both?

TingluoHuang: `if and .Values.manager .Values.manager.flags` vs `with .Values.manager.flags.logLevel`

TingluoHuang:
```
{{- with .Values.manager.flags.logLevel }}
  - "--log-level={{ . }}"
{{- end }}
```

AvaStancu: I put it in case a user might delete (intentionally or by mistake) the manager section or the flags section so we don't get an error. If `with` does basically the same thing, I am ok to add `with`

AvaStancu: Seems to be doing the same thing, updated, thanks!

charts/actions-runner-controller-2/values.yaml
```diff
@@ -62,4 +62,10 @@ affinity: {}
...
 # PriorityClass: system-cluster-critical
-priorityClassName: ""
\ No newline at end of file
+priorityClassName: ""
+
+manager:
+  flags:
+    # Log level can be set here with one of the following values: "debug", "info", "warn", "error".
+    # Defaults to "debug".
+    logLevel: "debug"
```
TingluoHuang: do we need this nested layer?

TingluoHuang: What about just the following?

```suggestion
# Log level can be set here with one of the following values: "debug", "info", "warn", "error".
# Defaults to "debug".
logLevel: "debug"
```

AvaStancu: My first proposal was
```
logging:
    logLevel:
```
So that in case anything else logging related appears, it can be put under one single `logging` section.
Nikola had the idea of tying this to a `flags` section, under `manager` and since we don't have anything else `manager` related in this doc, that is fine with me, to create this section and have also `flags` under it that can include several other flags in the future and be easy to spot by a future reader.

TingluoHuang: most settings in the `values.yaml` file is already targeted to the `manager`, so I am not sure how much value the extra `manager` will provide.
Also, the `logLevel: "debug"` is how the existing ARC used and I didn't find any complaints about it. 😄

nikola-jokic: I suggested moving it into the manager because those settings targetted to the manager are related to the deployment, and not the manager app itself.

The idea was to have something that would separate manager options passed at the startup from the ones related to the manager image, number of replicas etc.

And especially because there are many flags registered on the manager that could be used.  But if you both disagree, please ignore me :smile:

AvaStancu: If there are manager deployment details and manager app details, having a `manager` level without another specification like `manager-app` is not enough. To simplify, I removed `manager` but left `flags` since we might still add more flags to this file, it's ok to start with this section already.

nikola-jokic (approved):
LGTM

PR #2255, opened 2023-02-06
EphemeralRunner: On cleanup, if pod is pending, delete from service

nikola-jokic:
If pod is unable to start due to image pull issue, the pod will get stuck in Pending phase. It would require manual intervention to delete that ephemeral runner.

This change removes runner from the service if pod is pending (never actually started).

### DIFF
--- a/controllers/actions.github.com/ephemeralrunner_controller.go
+++ b/controllers/actions.github.com/ephemeralrunner_controller.go
@@ -271,7 +271,7 @@ func (r *EphemeralRunnerReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 func (r *EphemeralRunnerReconciler) cleanupRunnerFromService(ctx context.Context, ephemeralRunner *v1alpha1.EphemeralRunner, log logr.Logger) (ctrl.Result, error) {
 	actionsError := &actions.ActionsError{}
 	err := r.deleteRunnerFromService(ctx, ephemeralRunner, log)
-	if err != nill {
+	if err != nil {
 		if errors.As(err, &actionsError) &&
 			actionsError.StatusCode == http.StatusBadRequest &&
 			strings.Contains(actionsError.ExceptionName, "JobStillRunningException") {
@@ -284,7 +284,7 @@ func (r *EphemeralRunnerReconciler) cleanupRunnerFromService(ctx context.Context
 	}
 
 	log.Info("Successfully removed runner registration from service")
-	err := patch(ctx, r.Client, ephemeralRunner, func(obj *v1alpha1.EphemeralRunner) {
+	err = patch(ctx, r.Client, ephemeralRunner, func(obj *v1alpha1.EphemeralRunner) {
 		controllerutil.RemoveFinalizer(obj, ephemeralRunnerActionsFinalizerName)
 	})
 	if err != nil {
--- a/controllers/actions.github.com/ephemeralrunner_controller.go
+++ b/controllers/actions.github.com/ephemeralrunner_controller.go
@@ -243,6 +243,13 @@ func (r *EphemeralRunnerReconciler) cleanupResources(ctx context.Context, epheme
 	switch {
 	case err == nil:
 		if pod.ObjectMeta.DeletionTimestamp.IsZero() {
+			if pod.Status.Phase == corev1.PodPending && ephemeralRunner.Status.RunnerId > 0 {
+				// In case kubelet could not pull the image, it will have an exponential back-off.
+				// If someone manually tries to delete the runner, it will not get deleted from the service.
+				if err := r.deleteRunnerFromService(ctx, ephemeralRunner, log); err != nil {
+					return false, fmt.Errorf("failed to remove the runner from service: %v", err)
+				}
+			}
 			log.Info("Deleting the runner pod")
 			if err := r.Delete(ctx, pod); err != nil && !kerrors.IsNotFound(err) {
 				return false, fmt.Errorf("failed to delete pod: %v", err)
--- a/controllers/actions.github.com/ephemeralrunner_controller.go
+++ b/controllers/actions.github.com/ephemeralrunner_controller.go
@@ -44,7 +44,7 @@ const (
 	EphemeralRunnerContainerName = "runner"
 
 	ephemeralRunnerFinalizerName        = "ephemeralrunner.actions.github.com/finalizer"
-	ephemeralRunnerActionsFinalizerName = "ephemeralrunner.actions.github.com/actions"
+	ephemeralRunnerActionsFinalizerName = "ephemeralrunner.actions.github.com/runner-registration-finalizer"
 )
 
 // EphemeralRunnerReconciler reconciles a EphemeralRunner object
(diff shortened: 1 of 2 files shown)


TingluoHuang:
Will this cause ARC stuck in a loop of creating/deleting runners on the service?

nikola-jokic:
Yes :relaxed: . So basically, if the issue is on the container registry, the back-off will eventually pull it.
If the issue is some typo in the `image` field, you would need to update `AutoscalingRunnerSet` anyway.
The update will take care of deletion, but if someone manually tries to delete it (maybe by accident or image pull back-off tops at 5 minutes so to speed it up), then the runner will stay registered.

TingluoHuang:
@nikola-jokic The service will return 429 if you configure too many runners within a short time window.

controllers/actions.github.com/ephemeralrunner_controller.go
```diff
@@ -246,6 +246,13 @@ func (r *EphemeralRunnerReconciler) cleanupResources(ctx context.Context, epheme
 	switch {
 	case err == nil:
 		if pod.ObjectMeta.DeletionTimestamp.IsZero() {
+			if pod.Status.Phase == corev1.PodPending && ephemeralRunner.Status.RunnerId > 0 {
```
TingluoHuang: I think we would just go with a second finalizer on the ephemeralrunner to handle removing the runner from the service.

controllers/actions.github.com/ephemeralrunner_controller.go
```diff
@@ -43,7 +43,8 @@ const (
 	// It represents the name of the container running the self-hosted runner image.
 	EphemeralRunnerContainerName = "runner"
 
-	ephemeralRunnerFinalizerName = "ephemeralrunner.actions.github.com/finalizer"
+	ephemeralRunnerFinalizerName        = "ephemeralrunner.actions.github.com/finalizer"
+	ephemeralRunnerActionsFinalizerName = "ephemeralrunner.actions.github.com/actions"
```
TingluoHuang: `ephemeralrunner.actions.github.com/runnerRegistrationFinalizer` ?

controllers/actions.github.com/ephemeralrunner_controller.go
```diff
@@ -80,6 +81,40 @@ func (r *EphemeralRunnerReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 			return ctrl.Result{}, nil
 		}
 
+		if controllerutil.ContainsFinalizer(ephemeralRunner, ephemeralRunnerActionsFinalizerName) {
+			if ephemeralRunner.Status.Phase == corev1.PodSucceeded {
```
TingluoHuang: this feels like a weak assumption.

controllers/actions.github.com/ephemeralrunner_controller.go
```diff
@@ -80,6 +81,40 @@ func (r *EphemeralRunnerReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+
+			actionsError := &actions.ActionsError{}
+			if errors.As(err, &actionsError) &&
+				actionsError.StatusCode == http.StatusBadRequest &&
+				strings.Contains(actionsError.ExceptionName, "JobStillRunningException") {
+				log.Info("Runner is still running the job. Re-queue in 30 seconds")
+				return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
+			}
+			log.Error(err, "Failed finalize deleting runner from the service")
+			return ctrl.Result{}, err
```
TingluoHuang: Can we warp these into some method for better readability? similar to `r.cleanupResources()`

controllers/actions.github.com/ephemeralrunner_controller.go
TingluoHuang: add a new `Deleted` to `ephemeralRunner.Status` or something to indicate it has been deleted?

controllers/actions.github.com/ephemeralrunner_controller.go
nikola-jokic: How do you feel about: `ephemeralrunner.actions.github.com/runner-registration-finalizer`

controllers/actions.github.com/ephemeralrunner_controller.go
nikola-jokic: So the PodSucceeded phase can only be set in case we checked the pod, checked with the service and everything is fine. Otherwise, ephemeralRunner can never reach the PodSucceeded phase. That is why I wanted to avoid another check since we are checking before we are setting the phase to this one

controllers/actions.github.com/ephemeralrunner_controller.go
```diff
@@ -236,6 +267,29 @@ func (r *EphemeralRunnerReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+
+	actionsError := &actions.ActionsError{}
+	if errors.As(err, &actionsError) &&
+		actionsError.StatusCode == http.StatusBadRequest &&
+		strings.Contains(actionsError.ExceptionName, "JobStillRunningException") {
+		log.Info("Runner is still running the job. Re-queue in 30 seconds")
+		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
+	}
+	log.Error(err, "Failed clean up runner from the service")
+	return ctrl.Result{}, err
```
TingluoHuang:
```suggestion
	actionsError := &actions.ActionsError{}
	err := r.deleteRunnerFromService(ctx, ephemeralRunner, log)
	if err != nill {
	   if errors.As(err, &actionsError) &&
	      actionsError.StatusCode == http.StatusBadRequest &&
	      strings.Contains(actionsError.ExceptionName, "JobStillRunningException") {
	        log.Info("Runner is still running the job. Re-queue in 30 seconds")
		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
	   }
	   
	   log.Error(err, "Failed clean up runner from the service")
           return ctrl.Result{}, err
	}
        
       	log.Info("Successfully removed runner registration from service")
	err := patch(ctx, r.Client, ephemeralRunner, func(obj *v1alpha1.EphemeralRunner) {
		controllerutil.RemoveFinalizer(obj, ephemeralRunnerActionsFinalizerName)
	})
	if err != nil {
		return ctrl.Result{}, err
	}
	
	log.Info("Successfully removed runner registration finalizer")
	return ctrl.Result{}, nil
```

You might want to fix the format a little bit. 😄

controllers/actions.github.com/ephemeralrunner_controller.go
```diff
@@ -80,6 +81,23 @@ func (r *EphemeralRunnerReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+			case corev1.PodSucceeded:
+				// deleted by the runner set, we can just remove finalizer without API calls
+				err := patch(ctx, r.Client, ephemeralRunner, func(obj *v1alpha1.EphemeralRunner) {
+					controllerutil.RemoveFinalizer(obj, ephemeralRunnerActionsFinalizerName)
+				})
+				if err != nil {
+					log.Error(err, "Failed to update ephemeral runner without runner registration finalizer")
+					return ctrl.Result{}, err
+				}
+				return ctrl.Result{}, nil
```
TingluoHuang, line 96:
```suggestion
                                 log.Info("Successfully removed runner registration finalizer")
				return ctrl.Result{}, nil
```

TingluoHuang (approved):

PR #2257, opened 2023-02-06
Add options to multi client

rentziass:
This was added while working on TLS, and this being a necessity (for proxy work too) I wanted to avoid pushing another giant PR 😄

### DIFF
--- a/github/actions/multi_client.go
+++ b/github/actions/multi_client.go
@@ -11,8 +11,8 @@ import (
 )
 
 type MultiClient interface {
-	GetClientFor(ctx context.Context, githubConfigURL string, creds ActionsAuth, namespace string) (ActionsService, error)
-	GetClientFromSecret(ctx context.Context, githubConfigURL, namespace string, secretData KubernetesSecretData) (ActionsService, error)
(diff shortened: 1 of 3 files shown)


nikola-jokic (approved):
LGTM

github/actions/multi_client.go
```diff
@@ -66,8 +66,10 @@ func (m *multiClient) GetClientFor(ctx context.Context, githubConfigURL string,
 	client, err := NewClient(
 		githubConfigURL,
 		&creds,
-		WithUserAgent(m.userAgent),
-		WithLogger(m.logger),
+		append([]ClientOption{
+			WithUserAgent(m.userAgent),
+			WithLogger(m.logger),
+		}, options...)...,
```
TingluoHuang, line 72: do we need these many `...`? 😄

nikola-jokic: You have to break them to variadic arguments by doing that
Only thing that can be done is to split it as:
```go
options := []ClientOption{
			WithUserAgent(m.userAgent),
			WithLogger(m.logger),
		}, opts...)
```
And then call it with `options...` 
I'm okay with both :smile:

rentziass: The idea was that that block wouldn't be changed very easily 😄 but yeah, what Nikola said :)

Issue #2264, opened 2023-02-07
Restricting the RBAC rules on secrets


wherka-ama:
### What would you like added?

*A possibility to restrict the permission to read the secrets from the cluster-wide, to a namespace-scoped role*

### Why is this needed?

Working on a shared cluster in a corporate environment is very common. It is about reducing the cost of ownership and maintenance. As it stands the ARC controller grants itself a very generous permissions - including the possibility to  see all the secrets from all the namespaces on the entire cluster. 
This is very excessive and unnecessary. Any respectful security analyst and/or SRE would reject the request of installing such workload on a shared cluster.

### Additional context

Here is a short recipe on how to test this access:

```
helm repo add actions-runner-controller https://actions-runner-controller.github.io/actions-runner-controller
helm repo update
helm upgrade --install --namespace actions-runner-system --create-namespace\
  --set=authSecret.create=true\
  --set=authSecret.github_token="ghp_XXX"\
  --wait actions-runner-controller actions-runner-controller/actions-runner-controller

kubectl auth can-i get secret -n kube-system --as=system:serviceaccount:actions-runner-system:actions-runner-controller
yes
```

PR #2265, opened 2023-02-07
Restricting the RBAC rules on secrets

wherka-ama:
The PR is aiming at reducing the scope of the permissions for secrets.

As it stands the cluster-wide permissions are granted by default, therefore the service account used by the actions runner controller can see all the secrets from the entire cluster - even if it's not a desired behaviour.
We can easily imagine the shared cluster where we are only interested in a specific namespace. Whether we want it or not, the namespace admin can see all the secrets since we bind the ClusterRole using the ClusterRoleBinding is is not namespace-scoped.

See #2264 

Fix here is relatively simple and it doesn't involve any code level change:
1. Extracting the rules related to secrets and forming a dedicated set of role/binding
2. Reacting on `scope.singleNamespace` and switching between cluster-wide and namespace wide roles(`ClusterRole/ClusteRoleBinding` vs `Role/RoleBinding`) depending on the value. By default the current behaviour is preserved i.e. cluster-wide scope, otherwise it's limited to a namespace.

### DIFF
--- a/acceptance/deploy.sh
+++ b/acceptance/deploy.sh
@@ -61,6 +61,9 @@ if [ "${tool}" == "helm" ]; then
     flags+=( --set githubWebhookServer.imagePullSecrets[0].name=${IMAGE_PULL_SECRET})
     flags+=( --set actionsMetricsServer.imagePullSecrets[0].name=${IMAGE_PULL_SECRET})
   fi
+  if [ "${WATCH_NAMESPACE}" != "" ]; then
+    flags+=( --set watchNamespace=${WATCH_NAMESPACE} --set singleNamespace=true)
+  fi
   if [ "${CHART_VERSION}" != "" ]; then
     flags+=( --version ${CHART_VERSION})
   fi
--- a/charts/actions-runner-controller/templates/manager_role.yaml
+++ b/charts/actions-runner-controller/templates/manager_role.yaml
@@ -250,14 +250,6 @@ rules:
   - patch
   - update
   - watch
-- apiGroups:
-  - ""
-  resources:
-  - secrets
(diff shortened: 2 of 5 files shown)


mumoshu (approved):
I've verified this to work with or without the watch/singleNamespace. Awesome work, @wherka-ama!

wherka-ama:
> I've verified this to work with or without the watch/singleNamespace. Awesome work, @wherka-ama!

@mumoshu : thanks a lot for your review and merging this PR! I really appreciate it. 

Just one more question: I'm assuming that the `gha-runner-scale-set-controller` v0.3.0+ would also come with a similar restriction. Is that correct? I'm asking as the 0.3.0 still holds the role with excessive permissions. I've tested it with the following [this procedure](https://github.com/actions/actions-runner-controller/tree/master/docs/preview/gha-runner-scale-set-controller#install-actions-runner-controller)
I just need a confirmation. I believe the respective changes were applied by  @TingluoHuang. Just wanted to be sure.

TingluoHuang:
@wherka-ama if you try out the helm chart from the latest master, you should be able to see the permission has been trimmed down for `gha-runner-scale-set-controller `

wherka-ama:
> @wherka-ama if you try out the helm chart from the latest master, you should be able to see the permission has been trimmed down for `gha-runner-scale-set-controller `

@TingluoHuang : thanks for the response. Indeed, I can see the state of the chart in the repository is exactly how we were expecting it - thanks a lot for your hard work and for being so receptive!

My only concern was when we would be able to pull it from the official helm package repo. I'm not really trying to rush things, but it would be nice to know roughly when it would be released :-)

PR #2268, opened 2023-02-09
Add support for self-signed CA certificates

rentziass:
This allows users to use self-signed CA certificates contained in a config map within autoscaling runner sets via `githubServerTLS.`. These certificates will be used by Actions client when talking to Actions, and will be exposed in the listener pod via an environment variable (to avoid copying the configmap around). If `githubServerTLS.runnerMountPath` is set, the helm chart will try to create a volume, mount it on the runner at the specified path and set `NODE_EXTRA_CA_CERTS` and `RUNNER_UPDATE_CA_CERTS` (unless otherwise already specified by the template provided by the user).

In order to read the configmap the user references we've added a new permission to the manager cluster role, something to keep track of in #2275 since we only need these if `githubServerTLS` is configured.


apis/actions.github.com/v1alpha1/autoscalinglistener_types.go
```diff
@@ -54,11 +54,13 @@ type AutoscalingListenerSpec struct {
 
 	// Required
 	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
+
+	// +optional
+	GitHubServerTLS *GitHubServerTLSConfig `json:"githubServerTLS,omitempty"`
```
rentziass, line 62: Do we need this on the listener spec too? Can we just rely on the scale set the listener belongs to?

github/actions/testserver/server.go
```diff
@@ -0,0 +1,115 @@
+package testserver
```
rentziass: Extracted the actions server into its own package so it can be used in all tests, didn't use it everywhere yet to try and keep this PR sub 1000 lines 😓

apis/actions.github.com/v1alpha1/autoscalinglistener_types.go
nikola-jokic: I think we do :relaxed: 
The ownership is one-directional, meaning each higher level resource should pass down information needed by the lower level resource to do its thing
The listener here should not know anything about the resource owning it. It just works toward having the cluster state meet the spec.

github/actions/testserver/server.go
nikola-jokic: :+1: !

cmd/githubrunnerscalesetlistener/main.go
```diff
@@ -155,3 +158,38 @@ func validateConfig(config *RunnerScaleSetListenerConfig) error {
 
 	return nil
 }
+
+func actionsClientFromConfig(config RunnerScaleSetListenerConfig, creds *actions.ActionsAuth, options ...actions.ClientOption) (*actions.Client, error) {
```
nikola-jokic: Since this is always building a new actions client, can we name it like `newActionsClientFromConfig`?

apis/actions.github.com/v1alpha1/autoscalinglistener_types.go
TingluoHuang: or feed in the content of the cert as env to the pod?

TingluoHuang: I think the configMap may scope to a namespace, which means the listener might not have access to it.

rentziass: @TingluoHuang there's a risk that configmap contains multiple certs though no? Would it be safer to copy it like we do with the secret?

TingluoHuang: I was thinking about `GitHubServerTLS string` being just the content of the certs, then you don't need to copy the configMap across namespaces. 
Maybe my thought is to naive. 😄

rentziass: I mean it would certainly be easier! I'll put something together so we can see how it looks.

rentziass: Seem to work well, as long as the variable is < 1mb. I'll go with that thanks!

TingluoHuang: Chrome rejects SSL certs that are> 64KB... I think the 1MB limit should be fine. 👍  😄

Link-: @rentziass shall we document this limit with a comment maybe?

apis/actions.github.com/v1alpha1/ephemeralrunnerset_types.go
```diff
@@ -26,6 +26,9 @@ type EphemeralRunnerSetSpec struct {
 	Replicas int `json:"replicas,omitempty"`
 
 	EphemeralRunnerSpec EphemeralRunnerSpec `json:"ephemeralRunnerSpec,omitempty"`
+
+	// +optional
+	GitHubServerTLS *GitHubServerTLSConfig `json:"githubServerTLS,omitempty"`
```
TingluoHuang: should this also live inside `EphemeralRunnerSpec` just like the proxy?

charts/gha-runner-scale-set/values.yaml
```diff
@@ -44,6 +44,36 @@ githubConfigSecret:
 
 # runnerGroup: "default"
 
+## Custom CA certificates can be provided by bundling them in a configmap and
+## referring it in githubServerTLS.certConfigMapRef, for example:
+#
+# githubServerTLS:
+#   certConfigMapRef: your-configmap-name
+#
+## If you do need custom CA certificates do remember that the runner (and your
```
TingluoHuang: should we do this under `ephemeral_runner`, just like we do it for proxy?

charts/gha-runner-scale-set/values.yaml
```diff
@@ -44,6 +44,36 @@ githubConfigSecret:
 
 # runnerGroup: "default"
 
+## Custom CA certificates can be provided by bundling them in a configmap and
+## referring it in githubServerTLS.certConfigMapRef, for example:
+#
+# githubServerTLS:
+#   certConfigMapRef: your-configmap-name
```
TingluoHuang: do we need a nested map? or just `githubServerTLSCertConfigMapRef: your-configmap-name`

github/actions/multi_client.go
```diff
@@ -142,14 +142,14 @@ func (m *multiClient) GetClientFromSecret(ctx context.Context, githubConfigURL,
 	return m.GetClientFor(ctx, githubConfigURL, auth, namespace, options...)
 }
 
-func RootCAsFromConfigMap(configMapData map[string][]byte) (*x509.CertPool, error) {
+func RootCAsFromConfigMap(configMapData map[string]string) (*x509.CertPool, error) {
 	caCertPool, err := x509.SystemCertPool()
```
TingluoHuang: is there any way we can trace out that the current client is using a proxy or CACert? those would be super helpful during debug

controllers/actions.github.com/ephemeralrunnerset_controller.go
```diff
@@ -468,6 +468,37 @@ func (r *EphemeralRunnerSetReconciler) actionsClientFor(ctx context.Context, rs
...
+
+		certs, err := actions.RootCAsFromConfigMap(rootCAsConfigMap.Data)
+		if err != nil {
+			return nil, fmt.Errorf(
+				"failed to get certificates from configmap %s: %w",
+				tlsConfig.RootCAsConfigMapRef,
+				err,
+			)
+		}
+
```
TingluoHuang, line 508: can we trace out that we are creating the client with a CACert?
does it make sense to do similar things to the proxy, and trace out that the client is using a proxy?
you can do a follow-up PR if you think the PR is getting bigger and bigger. 😄

controllers/actions.github.com/autoscalinglistener_controller.go
```diff
@@ -424,6 +424,35 @@ func (r *AutoscalingListenerReconciler) createListenerPod(ctx context.Context, a
...
+				err,
+			)
+		}
+
+		var certString string
+		for _, cert := range rootCAsConfigMap.Data {
+			certString += cert
+		}
+
+		newPod.Spec.Containers[0].Env = append(newPod.Spec.Containers[0].Env, corev1.EnvVar{
```
TingluoHuang: do we want to move this up before calling `newScaleSetListenerPod` and make it similar to how you handle the proxy env?

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -494,7 +495,23 @@ func (r *AutoscalingRunnerSetReconciler) actionsClientFor(ctx context.Context, a
 		return nil, fmt.Errorf("failed to find GitHub config secret: %w", err)
 	}
 
-	var opts []actions.ClientOption
+	opts, err := r.actionsClientOptionsFor(ctx, autoscalingRunnerSet)
+	if err != nil {
+		return nil, err
```
TingluoHuang: do we want to wrap the error? 😄

apis/actions.github.com/v1alpha1/ephemeralrunnerset_types.go
rentziass: YES thank you 🙌 there's a bunch of changes here that predated the proxy work, expect this answer a lot, thank you for pointing these out 🙇

charts/gha-runner-scale-set/values.yaml
rentziass: I think there's an argument for leaving this to the pod spec, how and where you want to mount certificates in the container changes with the image being used. And depending on the image and actions being used users may need to set specific env vars too?

charts/gha-runner-scale-set/values.yaml
rentziass: To be honest I've used the existing spec for all this without giving it enough though, you have a good point for sure. Much like proxy it's a discussion of convenience (no nesting) vs leaving the door open to other ways to provide certificates? If we want to go the way we went for proxy then it should probably stay as is, but open for discussion. What do you think? @actions/actions-runtime any thoughts?

github/actions/multi_client.go
rentziass: We could add these to the client logger when applying options perhaps? Probably best in a separate PR right after this one to save a diff?

controllers/actions.github.com/ephemeralrunnerset_controller.go
rentziass: same as https://github.com/actions/actions-runner-controller/pull/2268#discussion_r1123112719

controllers/actions.github.com/autoscalinglistener_controller.go
rentziass: Pre proxy code, on it 🙌

apis/actions.github.com/v1alpha1/autoscalinglistener_types.go
```diff
@@ -57,6 +57,9 @@ type AutoscalingListenerSpec struct {
 
 	// +optional
 	Proxy *ProxyConfig `json:"proxy,omitempty"`
+
+	// +optional
+	GitHubServerTLS *GitHubServerTLSConfig `json:"githubServerTLS,omitempty"`
```
Link-, line 62: What's the rational for calling it `GitHubServerTLS` as opposed to something more generic?

charts/gha-runner-scale-set-controller/crds/actions.github.com_ephemeralrunnersets.yaml
```diff
@@ -47,7 +47,7 @@ spec:
                     githubServerTLS:
                       properties:
                         certConfigMapRef:
-                          description: Required
+                          description: Required, the summed size of the values of the configmap must be less than 1048576 bytes as these will be mounted as an environment variable on the listener pod.
```
Link-:
```suggestion
                          description: Required, the summed size of the values of the configmap must be less than 1048576 bytes (~1 MB) as these will be mounted as an environment variable on the listener pod.
```

charts/gha-runner-scale-set-controller/crds/actions.github.com_ephemeralrunnersets.yaml
```diff
@@ -4270,6 +4270,12 @@ spec:
                         - containers
                       type: object
                   type: object
+                githubServerTLS:
+                  properties:
+                    certConfigMapRef:
+                      description: Required, the summed size of the values of the configmap must be less than 1048576 bytes as these will be mounted as an environment variable on the listener pod.
```
Link-:
```suggestion
                      description: Required, the summed size of the values of the configmap must be less than 1048576 bytes (~1 MB) as these will be mounted as an environment variable on the listener pod.
```

charts/gha-runner-scale-set-controller/crds/actions.github.com_ephemeralrunnersets.yaml
Link-: Makes it easier to understand at a glance

charts/gha-runner-scale-set/values.yaml
Link-: yeah this is going to be very tricky for runners. I think this should be something that end users have to deal with on a case by case basis. Trust stores vary greatly from one distribution to another

apis/actions.github.com/v1alpha1/autoscalinglistener_types.go
rentziass: This was already part of the spec in other types, it's being discussed [here](https://github.com/actions/actions-runner-controller/pull/2268#discussion_r1123110932) if you don't mind moving the conversation there 🙇

charts/gha-runner-scale-set-controller/crds/actions.github.com_ephemeralrunnersets.yaml
rentziass: Updated at the source and regenerated manifests to propagate the change 👍

charts/gha-runner-scale-set/values.yaml
TingluoHuang: I am fine with what you have (nested), the only other thing I can think of for the future is to support SSL Client cert...

charts/gha-runner-scale-set/values.yaml
TingluoHuang: I still think we might want to do some work to get the runner to work with the self-sign cert out of the box, similar to the proxy setting. For advance users, they can always do whatever they want. I could be wrong, btw. 😄

Link-: I think tackling our own runner image is a good compromise while adding documentation for how it can be done for other images

rentziass: Even with our own runner image mounting the configmap is only part of the picture, specific actions can require other customisations from the user. Personally I think documentation can go a long way, other than that what we can do is if `image == ghcr.io/actions/actions-runner` we mount certificates under `/usr/local/share/ca-certificates/`?

rentziass: And set `NODE_EXTRA_CA_CERTS` env var to cover first party node actions?

rentziass: `NODE_EXTRA_CA_CERTS` [takes a single file](https://nodejs.org/api/cli.html#node_extra_ca_certsfile), and by using configmaps we could potentially have multiple. Should we change the spec to point to a [configmap key](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#configmapkeyselector-v1-core) instead?

charts/gha-runner-scale-set/values.yaml
```diff
@@ -44,6 +44,36 @@ githubConfigSecret:
...
+#
+## If you do need custom CA certificates do remember that the runner (and your
+## actions) will probably need access to them as well. Below is an example
+## configuration for the default runner:
+#
+# template:
+#   spec:
+#     containers:
```
nikola-jokic, line 59: Can we document this to be an action done by the init container?

charts/gha-runner-scale-set/values.yaml
nikola-jokic: Or we can have exported constant that we document saying: this key is special to us. Basically, similar to the runner being a special name for the EphemeralRunner, or `dind` being a special name.

Link-: > other than that what we can do is if image == ghcr.io/actions/actions-runner we mount certificates under /usr/local/share/ca-certificates/

Maybe, always mount if the field is populated without checking the image? Because what if rename the image or publish somewhere else or ... 

Also, I think just adding the files will not do, we need to run `update-ca-certificates` at some point right?

Link-: Node... *sigh*

rentziass: We've decided to change the spec for `githubServerTLS` to take a direct reference to a configmap key (leaves the door open for other methods down the line) and a `runnerMountPath` defaulting to `/usr/local/share/ca-certificates/` if not specified. When creating runner pods we'll respect that and set `NODE_EXTRA_CA_CERTS` to point to the same path.

apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
```diff
@@ -80,7 +82,98 @@ type AutoscalingRunnerSetSpec struct {
...
-	RootCAsConfigMapRef string `json:"certConfigMapRef,omitempty"`
+	CertificateFrom *TLSCertificateSource `json:"certificateFrom,omitempty"`
+
+	// +optional
+	// +kubebuilder:default:=/usr/local/share/ca-certificates/
+	RunnerMountPath string `json:"runnerMountPath,omitempty"`
+}
+
+func (c *GitHubServerTLSConfig) ToCertPool(keyFetcher func(name, key string) ([]byte, error)) (*x509.CertPool, error) {
+	if c == nil {
```
nikola-jokic: Should we guard against the nil receiver? I feel like it should be callers responsibility to check if GitHubServerTLSConfig is set before calling this method

charts/gha-runner-scale-set-controller/templates/manager_role.yaml
```diff
@@ -146,6 +146,13 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
```
TingluoHuang, line 153: do you need to have both `list` and `watch`? maybe `get` is enough?

nikola-jokic: if we are not owning configmap or watching on it, as far as I know, get should be enough

rentziass: I'll try going back to `get`, I remember getting errors from the controller but maybe it was unrelated.

charts/gha-runner-scale-set/values.yaml
```diff
@@ -47,6 +47,20 @@ githubConfigSecret:
 ## name of the runner scale set to create.  Defaults to the helm release name
 # runnerScaleSetName: ""
 
+## A self-signed CA certificate for communication with the GitHub server can be
+## provided using a config map key selector. # If provided, the certificate will
+## also be mounted on runner pods under the # path specified with
+## `runnerMountPath` (defaults to # /usr/local/share/ca-certificates/ if left
+## blank) and NODE_EXTRA_CA_CERTS # environment variable will be set to the path
+## to that certificate to help with # NodeJS-based actions. Example configuration:
```
Link-: nit: do we want to cleanup these unnecessary `#` ?

rentziass: The idea for these comments in this file is that uncommenting the whole block gives you valid configuration + retains the comment.

Link-: Even for the second one in the middle of the sentence? 🤔

rentziass: Oooooh my bad 🤦

charts/gha-runner-scale-set-controller/templates/manager_role.yaml
rentziass: Client needs both because it caches by default. should we maybe also revisit as part of the work on permissions?

apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
```diff
@@ -80,7 +81,39 @@ type AutoscalingRunnerSetSpec struct {
...
+	}
+
+	if c.CertificateFrom.ConfigMapKeyRef == nil {
+		return nil, fmt.Errorf("configMapKeyRef not specified")
+	}
+
+	cert, err := keyFetcher(c.CertificateFrom.ConfigMapKeyRef.Name, c.CertificateFrom.ConfigMapKeyRef.Key)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"failed to fetch key %s in %s: %w",
```
nikola-jokic:
```suggestion
			"failed to fetch key %q in configmap %q: %w",
```
nit
It might be cleaner to see: 
```bash
failed to fetch key "my-key" with name "my-cm": ...
```

nikola-jokic: ignore if you think this is not useful :smile:

rentziass: +1 for `%q`, not sure I get `with name`? `in configmap %q` perhaps?

charts/gha-runner-scale-set/values.yaml
```diff
@@ -47,6 +47,20 @@ githubConfigSecret:
 ## name of the runner scale set to create.  Defaults to the helm release name
 # runnerScaleSetName: ""
 
+## A self-signed CA certificate for communication with the GitHub server can be
+## provided using a config map key selector. If provided, the certificate will
+## also be mounted on runner pods under the path specified with
+## `runnerMountPath` (defaults to /usr/local/share/ca-certificates/ if left
+## blank) and NODE_EXTRA_CA_CERTS environment variable will be set to the path
```
TingluoHuang:
```suggestion
## `runnerMountPath` (defaults to /usr/local/share/ca-certificates/)
## and NODE_EXTRA_CA_CERTS environment variable will be set to the path
```

apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
```diff
@@ -80,7 +81,39 @@ type AutoscalingRunnerSetSpec struct {
...
+	if err != nil {
+		return nil, fmt.Errorf(
+			"failed to fetch key %q in configmap %q: %w",
+			c.CertificateFrom.ConfigMapKeyRef.Key,
+			c.CertificateFrom.ConfigMapKeyRef.Name,
+			err,
+		)
+	}
+
+	pool := x509.NewCertPool()
```
TingluoHuang: should we use `x509.SystemCertPool()` as our base?

TingluoHuang: nvm, looks like we merged with the system cert pool in `RootCAsFromConfigMap`

cmd/githubrunnerscalesetlistener/main.go
```diff
@@ -160,6 +162,16 @@ func validateConfig(config *RunnerScaleSetListenerConfig) error {
 }
 
 func newActionsClientFromConfig(config RunnerScaleSetListenerConfig, creds *actions.ActionsAuth, options ...actions.ClientOption) (*actions.Client, error) {
+	if config.ServerRootCA != "" {
+		pool := x509.NewCertPool()
```
TingluoHuang: should we use the `SystemCertPool` as default?

github/actions/multi_client.go
```diff
@@ -142,14 +142,14 @@ func (m *multiClient) GetClientFromSecret(ctx context.Context, githubConfigURL,
 	return m.GetClientFor(ctx, githubConfigURL, auth, namespace, options...)
 }
 
-func RootCAsFromConfigMap(configMapData map[string][]byte) (*x509.CertPool, error) {
+func RootCAsFromConfigMap(configMapData map[string]string) (*x509.CertPool, error) {
```
TingluoHuang: Does this use anywhere?

apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
TingluoHuang: actually, we might want to double-check to make sure we merge the cert with the system cert pool, I couldn't find where we use `RootCAsFromConfigMap`

charts/gha-runner-scale-set/templates/autoscalingrunnerset.yaml
```diff
@@ -21,6 +21,18 @@ spec:
...
+  {{- if .Values.githubServerTLS }}
+  githubServerTLS:
+    {{- if .Values.githubServerTLS.certificateFrom }}
+    {{- with .Values.githubServerTLS.certificateFrom }}
+    certificateFrom:
+      configMapKeyRef:
+        name: {{ .configMapKeyRef.name }}
+        key: {{ .configMapKeyRef.key }}
+    {{ end }}
+    {{ end }}
```
TingluoHuang:
```suggestion
    {{- with .Values.githubServerTLS.certificateFrom }}
    certificateFrom:
      configMapKeyRef:
        name: {{ .configMapKeyRef.name }}
        key: {{ .configMapKeyRef.key }}
    {{- end }}
```

`with` should check empty for you already.

charts/gha-runner-scale-set/templates/autoscalingrunnerset.yaml
```diff
@@ -21,6 +21,18 @@ spec:
...
+  {{ end }}
```
TingluoHuang:
```suggestion
  {{- end }}
```

charts/gha-runner-scale-set/templates/autoscalingrunnerset.yaml
```diff
@@ -103,10 +115,14 @@ spec:
...
-      {{ .Values.template.spec.containers | toYaml | nindent 6 }}
+      {{- include "gha-runner-scale-set.default-mode-runner-containers" . | nindent 6 }}
       {{- end }}
-      {{- if or .Values.template.spec.volumes (eq .Values.containerMode.type "dind") (eq .Values.containerMode.type "kubernetes") }}
+      {{- $tlsConfig := (default (dict) .Values.githubServerTLS) }}
+      {{- if or .Values.template.spec.volumes (eq .Values.containerMode.type "dind") (eq .Values.containerMode.type "kubernetes") $tlsConfig.runnerMountPath }}
       volumes:
+        {{- if $tlsConfig.runnerMountPath }}
+          {{- include "gha-runner-scale-set.tls-volume" $tlsConfig | nindent 6 }}
+        {{ end }}
```
TingluoHuang:
```suggestion
        {{- end }}
```

charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -229,8 +246,20 @@ env:
   - name: RUNNER_WAIT_FOR_DOCKER_IN_SECONDS
     value: "120"
     {{- end }}
+    {{- if $setNodeExtraCaCerts }}
+  - name: NODE_EXTRA_CA_CERTS
+    value: {{ clean (print $tlsConfig.runnerMountPath "/" $tlsConfig.certificateFrom.configMapKeyRef.key) }}
```
TingluoHuang, line 251: do we need to add `"` around the string in case the path contains space?

charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -257,11 +289,17 @@ volumeMounts:
     mountPath: /certs/client
     readOnly: true
     {{- end }}
+    {{- if $mountGitHubServerTLS }}
+  - name: github-server-tls-cert
+    mountPath: {{ clean (print $tlsConfig.runnerMountPath "/" $tlsConfig.certificateFrom.configMapKeyRef.key) }}
+    subPath: {{ $tlsConfig.certificateFrom.configMapKeyRef.key }}
```
TingluoHuang, line 295: same, do we need to `"` around those paths?

charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -306,13 +356,28 @@ env:
   - name: ACTIONS_RUNNER_REQUIRE_JOB_CONTAINER
     value: "true"
     {{- end }}
+    {{- if $setNodeExtraCaCerts }}
+  - name: NODE_EXTRA_CA_CERTS
+    value: {{ clean (print $tlsConfig.runnerMountPath "/" $tlsConfig.certificateFrom.configMapKeyRef.key) }}
```
TingluoHuang, line 361: `"`?

charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -325,6 +390,83 @@ volumeMounts:
   - name: work
     mountPath: /actions-runner/_work
     {{- end }}
+    {{- if $mountGitHubServerTLS }}
+  - name: github-server-tls-cert
+    mountPath: {{ clean (print $tlsConfig.runnerMountPath "/" $tlsConfig.certificateFrom.configMapKeyRef.key) }}
+    subPath: {{ $tlsConfig.certificateFrom.configMapKeyRef.key }}
```
TingluoHuang, line 396: `"`?

charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -325,6 +390,83 @@ volumeMounts:
...
 {{- end }}
-{{- end }}
\ No newline at end of file
+{{- end }}
+
+{{- define "gha-runner-scale-set.default-mode-runner-containers" -}}
+{{- $tlsConfig := (default (dict) .Values.githubServerTLS) }}
+{{- range $i, $container := .Values.template.spec.containers -}}
+- name: {{ $container.name }}
+  {{- if eq $container.name "runner" -}}
```
TingluoHuang: do we want to check `{{- if eq $container.name "runner" -}}` first and ignore other container?

ex:
```yaml
{{- if ne $container.name "runner" -}}
- {{ $container | toYaml | nindent 2 }}
{{- else }}
- name: {{ $container.name }}
  .....
{{- end }}
```

charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -325,6 +390,83 @@ volumeMounts:
...
+        {{- range $envKey, $envVal := $env }}
+          {{- if ne $envKey "name" }}
+      {{ $envKey }}: {{ $envVal | toYaml | nindent 10 }}
+          {{- end }}
+        {{- end }}
+      {{- end }}
+    {{- end }}
+    {{- if $setNodeExtraCaCerts }}
+    - name: NODE_EXTRA_CA_CERTS
+      value: {{ clean (print $tlsConfig.runnerMountPath "/" $tlsConfig.certificateFrom.configMapKeyRef.key) }}
```
TingluoHuang, line 439: `"`?

charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -325,6 +390,83 @@ volumeMounts:
...
+        {{- range $mountKey, $mountVal := $volMount }}
+          {{- if ne $mountKey "name" }}
+      {{ $mountKey }}: {{ $mountVal | toYaml | nindent 10 }}
+          {{- end }}
+        {{- end }}
+      {{- end }}
+    {{- end }}
+    {{- if $mountGitHubServerTLS }}
+    - name: github-server-tls-cert
+      mountPath: {{ clean (print $tlsConfig.runnerMountPath "/" $tlsConfig.certificateFrom.configMapKeyRef.key) }}
```
TingluoHuang, line 465: `"`?

charts/gha-runner-scale-set/templates/_helpers.tpl
nikola-jokic: We should put quotes around names as well, because filename can also contain spaces

charts/gha-runner-scale-set/values.yaml
rentziass: I realise now this comment does not apply anymore, needs to be updated. If `runnerMountPath` is not provided we don't touch the runner pod spec at all.

apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
rentziass: `RootCAsFromConfigMap` can go now (we now only take in one certificate, it has no reason for existing). I would make `x509.SystemCertPool` the default in base in the client and we just add to it, so to avoid another "proxy from environment" situation in the future :lolsob:

cmd/githubrunnerscalesetlistener/main.go
rentziass: I would centralize this in the actions client itself so we don't allow ourselves (me) to forget about it in different places.

charts/gha-runner-scale-set/templates/_helpers.tpl
rentziass: Resolving all these as spaces are not allowed in config map keys (must satisfy `'[-._a-zA-Z0-9]+'`).

controllers/actions.github.com/ephemeralrunner_controller.go
```diff
@@ -698,6 +698,31 @@ func (r *EphemeralRunnerReconciler) actionsClientFor(ctx context.Context, runner
 		opts = append(opts, actions.WithProxy(proxyFunc))
 	}
 
+	tlsConfig := runner.Spec.GitHubServerTLS
```
nikola-jokic, line 716: Nit: extract it similar to `func (r *AutoscalingRunnerSetReconciler) actionsClientOptionsFor(...)`

TingluoHuang (approved):

PR #2275, opened 2023-02-12
ADR for Limit cluster role permission on Secerts.

TingluoHuang:
https://github.com/github/c2c-actions-runtime/issues/2296

https://github.com/actions/actions-runner-controller/issues/2264

### DIFF
--- a/adrs/2023-02-10-limit-manager-role-permission.md
+++ b/adrs/2023-02-10-limit-manager-role-permission.md
@@ -1,0 +1,95 @@
+# ADR 0007: Limit Permissions for Service Accounts in Actions-Runner-Controller
+**Date**: 2023-02-10
+
+**Status**: Pending
+
+## Context
+
+- `actions-runner-controller` is a Kubernetes CRD (with controller) built using https://github.com/kubernetes-sigs/controller-runtime
+
+- [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime) has a default cache based k8s API client.Reader to make query k8s API server more efficiency. 
+
+- The cache-based API client requires cluster scope `list` and `watch` permission for any resource the controller may query.
+
+- This documentation only scopes to the AutoscalingRunnerSet CRD and its controller.
+
+## Service accounts and their role binding in actions-runner-controller
+
+There are 3 service accounts involved for a working `AutoscalingRunnerSet` based `actions-runner-controller`
+
+1. Service account for each Ephemeral runner Pod
+
+This should have the lowest privilege (not any `RoleBinding` nor `ClusterRoleBinding`) by default, in the case of `containerMode=kubernetes`, it will get certain write permission with `RoleBinding` to limit the permission to a single namespace.
+
+> References:
+> - ./charts/gha-runner-scale-set/templates/no_permission_serviceaccount.yaml
+> - ./charts/gha-runner-scale-set/templates/kube_mode_role.yaml
+> - ./charts/gha-runner-scale-set/templates/kube_mode_role_binding.yaml
+> - ./charts/gha-runner-scale-set/templates/kube_mode_serviceaccount.yaml
+
+2. Service account for AutoScalingListener Pod
+
+This has a `RoleBinding` to a single namespace with a `Role` that has permission to `PATCH` `EphemeralRunnerSet` and `EphemeralRunner`.
+
+3. Service account for the controller manager
+
+Since the CRD controller is a singleton installed in the cluster that manages the CRD across multiple namespaces by default, the service account of the controller manager pod has a `ClusterRoleBinding` to a `ClusterRole` with broader permissions.
+
+The current `ClusterRole` has the following permissions:
+
+- Get/List/Create/Delete/Update/Patch/Watch on `AutoScalingRunnerSets` (with `Status` and `Finalizer` sub-resource)
+- Get/List/Create/Delete/Update/Patch/Watch on `AutoScalingListeners` (with `Status` and `Finalizer` sub-resource)
+- Get/List/Create/Delete/Update/Patch/Watch on `EphemeralRunnerSets` (with `Status` and `Finalizer` sub-resource)
+- Get/List/Create/Delete/Update/Patch/Watch on `EphemeralRunners` (with `Status` and `Finalizer` sub-resource)
+
+- Get/List/Create/Delete/Update/Patch/Watch on `Pods` (with `Status` sub-resource)
+- **Get/List/Create/Delete/Update/Patch/Watch on `Secrets`**
+- Get/List/Create/Delete/Update/Patch/Watch on `Roles`
+- Get/List/Create/Delete/Update/Patch/Watch on `RoleBindings`
+- Get/List/Create/Delete/Update/Patch/Watch on `ServiceAccounts`
+
+> Full list can be found at: https://github.com/actions/actions-runner-controller/blob/facae69e0b189d3b5dd659f36df8a829516d2896/charts/actions-runner-controller-2/templates/manager_role.yaml
+
+## Limit cluster role permission on Secrets
+
+The cluster scope `List` `Secrets` permission might be a blocker for adopting `actions-runner-controller` for certain customers as they may have certain restriction in their cluster that simply doesn't allow any service account to have cluster scope `List Secrets` permission. 
+
+To help these customers and improve security for `actions-runner-controller` in general, we will try to limit the `ClusterRole` permission of the controller manager's service account down to the following:
+
+- Get/List/Create/Delete/Update/Patch/Watch on `AutoScalingRunnerSets` (with `Status` and `Finalizer` sub-resource)
+- Get/List/Create/Delete/Update/Patch/Watch on `AutoScalingListeners` (with `Status` and `Finalizer` sub-resource)
+- Get/List/Create/Delete/Update/Patch/Watch on `EphemeralRunnerSets` (with `Status` and `Finalizer` sub-resource)
+- Get/List/Create/Delete/Update/Patch/Watch on `EphemeralRunners` (with `Status` and `Finalizer` sub-resource)
+
+- List/Watch on `Pods`
+- List/Watch on `Roles`
+- List/Watch on `RoleBindings`
+- List/Watch on `ServiceAccounts`
+
+> We will change the default cache-based client to bypass cache on reading `Secrets`, so we can eliminate the need for `List` and `Watch` `Secrets` permission in cluster scope.
+
+Introduce a new `Role` per `AutoScalingRunnerSet` installation and `RoleBinding` the `Role` with the controller's service account in the namespace that each `AutoScalingRunnerSet` deployed with the following permission.
+
+- Get/Create/Delete/Update/Patch on `Secrets`
+- Get/Create/Delete/Update/Patch on `Pods`
+- Get/Create/Delete/Update/Patch on `Roles`
+- Get/Create/Delete/Update/Patch on `RoleBindings`
+- Get/Create/Delete/Update/Patch on `ServiceAccounts`
+
+The `Role` and `RoleBinding` creation will happen during `helm install demo oci://ghcr.io/actions/actions-runner-controller-charts/gha-runner-scale-set` to grant the controller's service account required permissions to operate in the namespace the `AutoScalingRunnerSet` deployed.
+
+## Install ARC to only watch/react resources in a single namespace
+
+In case the user doesn't want to have any `ClusterRole`, they can choose to install the `actions-runner-controller` in a mode that only requires a `Role` with `RoleBinding` in a particular namespace.
+
+In this mode, the `actions-runner-controller` will only be able to watch the `AutoScalingRunnerSet` resource in a single namespace.
+
+If you want to deploy multiple `AutoScalingRunnerSet` into different namespaces, you will need to install `actions-runner-controller` in this mode multiple times as well and have each installation watch the namespace you want to deploy an `AutoScalingRunnerSet`
+
+You will install `actions-runner-controller` with something like `helm install arc --set watchSingleNamespace=TestNamespace oci://ghcr.io/actions/actions-runner-controller-charts/gha-runner-scale-set-controller` (the `TestNamespace` namespace needs to be created first).
+
+You will deploy the `AutoScalingRunnerSet` with something like `helm install demo --namespace TestNamespace oci://ghcr.io/actions/actions-runner-controller-charts/gha-runner-scale-set`
+
+In this mode, you will end up with a manager `Role` that has all Get/List/Create/Delete/Update/Patch/Watch permissions on resources we need, and a `RoleBinding` to bind the `Role` with the controller `ServiceAccount` in the watched single namespace, ex: `TestNamespace` in the above example.
+
+The downside of this mode is when you have multiple controllers deployed, they will still use the same version of the CRD. So you will need to make sure every controller you deployed has to be the same version as each other.


chrispat (dismissed):
This looks good to me based on my current knowledge of Kubernetes RBAC.

nikola-jokic (dismissed):
LGTM

adrs/2023-02-10-limit-manager-role-permission.md
```diff
@@ -0,0 +1,79 @@
...
+- Get/List/Create/Delete/Update/Patch/Watch on `EphemeralRunners` (with `Status` and `Finalizer` sub-resource)
+
+- List/Watch on `Pods`
+- List/Watch on `Roles`
+- List/Watch on `RoleBindings`
+- List/Watch on `ServiceAccounts`
+
+> We will change the default cache-based client to bypass cache on reading `Secrets`, so we can eliminate the need for `List` and `Watch` `Secrets` permission in cluster scope.
+
+Introduce a new `Role` for the namespace that an `AutoScalingRunnerSet` deployed with the following permission
```
mumoshu: Just to be extra sure, you don't need `Role` to make the permissions namespace-restricted. Binding a `ClusterRole` using `ClusterRoleBinding` results in a cluster-wide access while binding it using `RoleBinding` results in a namespace-restricted access.

So if we are to support two use-cases, allow a single instance of ARC a cluster-wide access for operability reasons, or allow the instance a namespace-wide access only for security reasons, the chart can be authored to create ClusterRoleBinding anyway and conditionally create either ClusterRoleBinding or RoleBinding depending on the desired config.

It's fine if you are explicitly using Role for the sake of e.g. clarity, though!

adrs/2023-02-10-limit-manager-role-permission.md
```diff
@@ -0,0 +1,79 @@
...
+- **Get/List/Create/Delete/Update/Patch/Watch on `Secrets`**
+- Get/List/Create/Delete/Update/Patch/Watch on `Roles`
+- Get/List/Create/Delete/Update/Patch/Watch on `RoleBindings`
+- Get/List/Create/Delete/Update/Patch/Watch on `ServiceAccounts`
+
+> Full list can be found at: https://github.com/actions/actions-runner-controller/blob/facae69e0b189d3b5dd659f36df8a829516d2896/charts/actions-runner-controller-2/templates/manager_role.yaml
+
+## Limit cluster role permission on Secrets
+
+The cluster scope `List` `Secrets` permission might be a blocker for adopting `actions-runner-controller` for certain customers as they may have certain restriction in their cluster that simply doesn't allow any service account to have cluster scope `List Secrets` permission.
```
mumoshu, line 55: This is just "my two cents" right now but anyway- another possibility that isn't covered in the current proposal looks to be about introducing a "single namespace" mode to the new ScaleSet-based ARC and using RoleBinding instead of ClusterRoleBinding.

That would allow us to still leverage the list and watch APIs for performance.

> Note that controller-runtime has the non-cached and the cached client. The former does not depend on the list and watch APIs while the latter scales better.

Also, note that the legacy ARC had `singleNamespace=true` and `watchNamespace=NS_NAME` chart values to enable the single namespace mode. The legacy chart is [hard-coded to use ClusterRoleBinding anyway
(https://github.com/actions/actions-runner-controller/blob/master/charts/actions-runner-controller/templates/manager_role_binding.yaml) so we might better use RoleBinding instead for the single namespace mode.

Perhaps the same approach would work for the new ARC too?

adrs/2023-02-10-limit-manager-role-permission.md
TingluoHuang: Sorry, I was mean to be a `ClusterRole` but defer the `RoleBinding` to helm install time, updated the doc.

adrs/2023-02-10-limit-manager-role-permission.md
TingluoHuang: I am trying to see if we eliminate the need for List secrets in cluster scope, how much do we need to support the watch single namespace installation?

TingluoHuang: do you have an idea of the scenarios that the current users try to install ARC with `singleNamespace=true`?

adrs/2023-02-10-limit-manager-role-permission.md
```diff
@@ -0,0 +1,79 @@
...
```
wherka-ama, line 55: I'm not representing all the customers who would love to use ARC, but they are constrained by tighter security rules, but I can add my 5 cents.

This section starts with "The cluster scope `List` `Secrets` permission.. ".  I don't think `List` or in fact `Watch` on the cluster level is a problem here. I'm a lot more concerned about all the rest of them: `Get/Create/Delete/Update/Patch`. There are the ones which can create havoc on the entire cluster.

adrs/2023-02-10-limit-manager-role-permission.md
```diff
@@ -0,0 +1,79 @@
...
+
+Introduce a new `ClusterRole` for the namespace that each `AutoScalingRunnerSet` deployed with the following permission
+
+- Get/Create/Delete/Update/Patch on `Secrets`
+- Get/Create/Delete/Update/Patch on `Pods`
+- Get/Create/Delete/Update/Patch on `Roles`
+- Get/Create/Delete/Update/Patch on `RoleBindings`
+- Get/Create/Delete/Update/Patch on `ServiceAccounts`
+
+The `RoleBinding` for the new `ClusterRole` will happen during `helm install demo oci://ghcr.io/actions/actions-runner-controller-charts/auto-scaling-runner-set` to grant the controller's service account required permissions to operate in the namespace the `AutoScalingRunnerSet` deployed.
```
wherka-ama: If we ensure that the `RoleBinding` is used here by default, then it's definitely acceptable from our side. However, I'm wondering why do we favour `ClusterRole` here if it's never going to be granted on the cluster level. Is it done just for the sake of simplicity i.e. to avoid explicitly specifying the namespace in the role definition? If that's the case it's fine to me.

Besides, I'm not sure if the statement: "`ClusterRole` for the namespace" is correct. ClusterRole is not really scoped to any namespace per its definition. It can only be scoped during the binding.

See:
https://kubernetes.io/docs/reference/access-authn-authz/rbac/#clusterrole-example

TingluoHuang: The idea is to have a single `ClusterRole` and multiple `RoleBinding` to bind the cluster role into each namespace.
Otherwise, we will need to create a `Role` in each namespace.
Again, sorry for the confusion... I will keep updating the doc.

adrs/2023-02-10-limit-manager-role-permission.md
TingluoHuang: @wherka-ama do you concern about the permission that ARC itself has or the permission each runner pod has? The ARC is open-sourced, so you can audit what operation it's doing with its permission.

adrs/2023-02-10-limit-manager-role-permission.md
toast-gear: It was so people could deploy multiple controllers allowing them to work around rate limit restrictions by using multiple PATs / GitHub Apps. It preceded the introduction of caching so possibly isn't as required for large installations as it once was when first introduced but we wouldn't know.

TingluoHuang: Thanks for the context @toast-gear, I would avoid introducing the single namespace installation for the new `AutoScalingRunnerSet` mode for now since the new mode suppose to handle a much higher load without consuming all your GitHub API rate limit. 🙇 
We can always revisit this base on the customer's ask. 😄

adrs/2023-02-10-limit-manager-role-permission.md
mumoshu: Got it! Thanks for your confirmation ☺️

adrs/2023-02-10-limit-manager-role-permission.md
mumoshu: Thanks for sharing @toast-gear! That matches my memory too.

@TingluoHuang To be clear, although the single namespace mode has been added for scalability, the idea here is it could be leveraged to enhance the security posture of the legacy and new ARC by two means-

First, it can be configured to use the namespaced controller-runtime client(which is backed by the namespaced List and Watch API calls).

Second, we can use role bindings instead of cluster role bindings for tighter access control.

This way, you have layered security- the potential attacker needs to break two things instead of one or zero in a relevant attack scenario before they can grab random secrets contained in the cluster.

Code auditing (https://github.com/actions/actions-runner-controller/pull/2275#discussion_r1106255247 ) is important too so that you can see "how" secure ARC is. On the other hand, I believe layered security is about making ARC "more" secure.

We can probably continue the discussion about this "what API verbs on which namespaces to support in ARC" thing in @wherka-ama's thread below? https://github.com/actions/actions-runner-controller/pull/2275#discussion_r1106160657

adrs/2023-02-10-limit-manager-role-permission.md
mumoshu: ⭐ I presume ARC should have two types of installation, cluster-wide and namespace-isolated.
Using ClusterRole universally and combining it with either RoleBinding or ClusterRoleBinding depending on the desired config would enable users to choose the right balance between security and operability while keeping the chart DRY.

TingluoHuang: @mumoshu I will add a new section in the doc to talk about that. 🙇

adrs/2023-02-10-limit-manager-role-permission.md
wherka-ama: > do you have an idea of the scenarios that the current users try to install ARC with `singleNamespace=true`?

I might give you some examples from our own backyard.

Without going into details I can tell you some of our clusters are treated as shared resources. They serve multiple purposes and  used by hundreds of people daily throughout their DEV activities. I reckon that's a fairly standard situation in corporate world.

In these clusters individual teams can only get namespace admin and anything beyond needs to be coordinated and tightly controlled. Things like CRDs are not something taken lightly.. and rightly so. The moment we push a new version and sticking with old controllers, suddenly all hell break loose. These CRDs are not guaranteed to be backward/forward compatible, so we would need to roll-out all the upgrades at the same time. In terms of logistics it's not great. But that's yet another story.

If there is a proper way of deploying ARC in a single namespace, it could greatly help with the adoption of this solution in our company. Otherwise the risk of escalating the permissions on the entire cluster just for the sake of deploying ARC and amount of paperwork to go through it may simply be questioned.

adrs/2023-02-10-limit-manager-role-permission.md
wherka-ama: > @wherka-ama do you concern about the permission that ARC itself has or the permission each runner pod has? The ARC is open-sourced, so you can audit what operation it's doing with its permission.

@TingluoHuang my concern is not really about ARC itself. I'm perfectly happy with the fact it's open. Let's not change that. I love open source and the fact I can easily refer to source code when the docs or the behaviour of the software don't make perfect sense right away. However, the provenance aka supply chain needs to be assured to build that trust. From GO sources, via binaries, to docker image there is a long way and we need to be sure that this chain is never tampered with. I hope we will not need to worry about these things here. Just stating the obvious.

My concern is not about ARC's internal logic - more about the doors which are opened for it and not secured which may lead to use them for malicious purposes.  It is that indirect access that gives us headache and principle of least privilege, which is not necessarily followed religiously in ARC. 

You can imagine the case where the admin user installs ARC with all its ClusterRoles, ClusterRoleBindings and binds them to a specific service account which is typically used to run the business of ARC manager. 
As soon as one can use such service account - completely independently of ARC - it's as easy as starting the POD with with curl or a kubectl inside, pointing to certificate automatically mounted for a service account in question and completely impersonating it. You can only imagine what kind of operation you could perform on it's behalf. Starting from reading all the secrets on the cluster, then modifying them or deleting, because you can, going further to granting yourself new clusterroles and binding them on the cluster level.. that's pretty much unrestricted access to the entire cluster. 
It's not very hard to imagine someone using a simple recipe like this:
https://kubernetes.io/docs/tasks/run-application/access-api-from-pod/#accessing-the-api-from-within-a-pod

adrs/2023-02-10-limit-manager-role-permission.md
wherka-ama: > Thanks for sharing @toast-gear! That matches my memory too.
> 
> @TingluoHuang To be clear, although the single namespace mode has been added for scalability, the idea here is it could be leveraged to enhance the security posture of the legacy and new ARC by two means-
> 
> First, it can be configured to use the namespaced controller-runtime client(which is backed by the namespaced List and Watch API calls).
> 
> Second, we can use role bindings instead of cluster role bindings for tighter access control.
> 
> This way, you have layered security- the potential attacker needs to break two things instead of one or zero in a relevant attack scenario before they can grab random secrets contained in the cluster.
> 
> Code auditing ([#2275 (comment)](https://github.com/actions/actions-runner-controller/pull/2275#discussion_r1106255247) ) is important too so that you can see "how" secure ARC is. On the other hand, I believe layered security is about making ARC "more" secure.
> 
> We can probably continue the discussion about this "what API verbs on which namespaces to support in ARC" thing in @wherka-ama's thread below? [#2275 (comment)](https://github.com/actions/actions-runner-controller/pull/2275#discussion_r1106160657)

@mumoshu : I couldn't agree more with your statement here:
"This way, you have layered security- the potential attacker needs to break two things instead of one or zero in a relevant attack scenario before they can grab random secrets contained in the cluster."

Thanks a lot for recognising the need to work on that subject. We really appreciate that!

adrs/2023-02-10-limit-manager-role-permission.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Install ARC to only watch/react resources in a single namespace
+
+In case the user doesn't want to have any `ClusterRole`, they can choose to install the `actions-runner-controller` in a mode that only requires a `Role` with `RoleBinding` in a particular namespace.
```
mumoshu, line 83: @wherka-ama I think this is what you exactly wanted! I'd appreciate your confirmation when you have some time.

adrs/2023-02-10-limit-manager-role-permission.md
```diff
@@ -0,0 +1,93 @@
...
+
+> We will change the default cache-based client to bypass cache on reading `Secrets`, so we can eliminate the need for `List` and `Watch` `Secrets` permission in cluster scope.
+
+Introduce a new `ClusterRole` for the controller and `RoleBinding` it to the namespace that each `AutoScalingRunnerSet` deployed with the following permission.
+
+- Get/Create/Delete/Update/Patch on `Secrets`
+- Get/Create/Delete/Update/Patch on `Pods`
+- Get/Create/Delete/Update/Patch on `Roles`
+- Get/Create/Delete/Update/Patch on `RoleBindings`
+- Get/Create/Delete/Update/Patch on `ServiceAccounts`
```
mumoshu: @TingluoHuang Although it might further increase the complexity of a whole ARC deployment, can we perhaps enhance the `auto-scaling-runner-set` chart to additionally support creating a dedicated `Role` with these permissions that is then bound by the mentioned `RoleBinding`?

Combined with "Install ARC to only watch/react resources in a single namespace" below, that would enable users to omit ClusterRoles for both manager and autoscaling runner sets. WDYT?

adrs/2023-02-10-limit-manager-role-permission.md
wherka-ama: @mumoshu : that's indeed in line with my expectations! Thank you very much for taking my voice into account. I really appreciate it. 

Looking forward to seeing the implementation.

Cheers

adrs/2023-02-10-limit-manager-role-permission.md
TingluoHuang: that's actually where I started. 😄 
make auto-scaling-runner-set to create `role` with `rolebinding`, I like it now since it further trims the usage for `ClusterRole`
I will update the ADR with this info. 🙇

adrs/2023-02-10-limit-manager-role-permission.md
wherka-ama: @mumoshu , @TingluoHuang : I've just checked a new release - 0.2.0 as per [latest announcements](https://github.com/actions/actions-runner-controller/blob/master/docs/preview/gha-runner-scale-set-controller/README.md):

`helm pull oci://ghcr.io/actions/actions-runner-controller-charts/gha-runner-scale-set-controller     --version 0.2.0 --untar`

I can still see the ClusterRole with ClusterRoleBinding which contains the rules for secrets. There are some improvements of course, like changes around [pv and pvc](https://github.com/actions/actions-runner-controller/pull/2260/files#diff-3b0c76ac2a8b78135a6787687e6d89a69bf9344b1c90c581df6d6d4a45ae10a8)  to name a few, but I'm afraid the secrets are still a bone of contention :/

Nevertheless, thanks a lot for this batch of improvements! We'll wait for the next version to address issues around secrets.

Cheers!

TingluoHuang: @wherka-ama once this ADR is reviewed and merged, I will work on implementing those improvements, which should be included in 0.3.0. 🤞

adrs/2023-02-10-limit-manager-role-permission.md
mumoshu: Awesome! I very much appreciate your work @TingluoHuang!

mumoshu (approved):
LGTM. I very much appreciate your work- Looking forward to seeing enhancements based on this!
The new ARC will be far more secure than any time before ❤️

adrs/2023-02-10-limit-manager-role-permission.md
wherka-ama: @TingluoHuang : great work! Thank you very much indeed ❤️

Please do let me know if there is anything in the pipeline for an early preview. I'd be happy to cast an eye and share my thoughts if that's helpful.

nikola-jokic (approved):
LGTM!

adrs/2023-02-10-limit-manager-role-permission.md
```diff
@@ -0,0 +1,95 @@
...
+- Get/List/Create/Delete/Update/Patch/Watch on `EphemeralRunners` (with `Status` and `Finalizer` sub-resource)
+
+- List/Watch on `Pods`
+- List/Watch on `Roles`
+- List/Watch on `RoleBindings`
+- List/Watch on `ServiceAccounts`
+
+> We will change the default cache-based client to bypass cache on reading `Secrets`, so we can eliminate the need for `List` and `Watch` `Secrets` permission in cluster scope.
+
+Introduce a new `Role` per `AutoScalingRunnerSet` installation and `RoleBinding` the `Role` with the controller's service account in the namespace that each `AutoScalingRunnerSet` deployed with the following permission.
```
rentziass, line 71: In https://github.com/actions/actions-runner-controller/pull/2268/ we're adding a new `list/watch` for `configmaps` in case the user configures the autoscaling runner set to use self-signed certificates. Should we add that here and possibly add that only _if_ the scale set needs it?

TingluoHuang: do you need `List` and `Watch`? or `Get` is enough?

adrs/2023-02-10-limit-manager-role-permission.md
TingluoHuang: Once I finish all the changes, you can definitely try it out by building the controller image yourself and using the helm chart from the latest source code. 🙇

adrs/2023-02-10-limit-manager-role-permission.md
TingluoHuang: I will update the section later after confirming the required permission with you in a follow-up PR. 😄

PR #2276, opened 2023-02-13
Remove list/watch secrets permission from the manager cluster role.

TingluoHuang:
The first part of the prototype changes for https://github.com/actions/actions-runner-controller/pull/2275

Changes:
- Remove `list/watch` permission on secrets for cluster scope
- Disable cache on secrets for the default manager client, so it won't complain about not having `list/watch` permission for secrets in cluster scope.
- Requeue reconciliation after any secret operations.

### DIFF
--- a/controllers/actions.github.com/autoscalinglistener_controller.go
+++ b/controllers/actions.github.com/autoscalinglistener_controller.go
@@ -86,7 +86,7 @@ func (r *AutoscalingListenerReconciler) Reconcile(ctx context.Context, req ctrl.
 		}
 		if !done {
 			log.Info("Waiting for resources to be deleted before removing finalizer")
-			return ctrl.Result{}, nil
+			return ctrl.Result{Requeue: true}, nil
 		}
 
 		log.Info("Removing finalizer")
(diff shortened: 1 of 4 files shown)


controllers/actions.github.com/autoscalinglistener_controller.go
```diff
@@ -86,7 +86,7 @@ func (r *AutoscalingListenerReconciler) Reconcile(ctx context.Context, req ctrl.
 		}
 		if !done {
 			log.Info("Waiting for resources to be deleted before removing finalizer")
-			return ctrl.Result{}, nil
+			return ctrl.Result{Requeue: true}, nil
```
TingluoHuang, line 89: we need to requeue ourselves for every place that we rely on the Secret is updated/created/deleted

TingluoHuang:
The failed PR check is due to https://github.com/github/c2c-actions-runtime/issues/2361

nikola-jokic (dismissed):
LGTM

controllers/actions.github.com/autoscalinglistener_controller.go
rentziass: Sounds really easy to forget 😬 do we have an easy way to ensure that doesn't happen?

rentziass (dismissed):

controllers/actions.github.com/autoscalinglistener_controller.go
nikola-jokic: As far as I know, we can't :disappointed: 
This is artificially enqueueing the same event, so we get triggered again
Since we are not watching on those resources, we would have to trigger ourselves each time we do some operation...

nikola-jokic (dismissed):
LGTM

controllers/actions.github.com/autoscalinglistener_controller.go
TingluoHuang: Our controller L0 test will catch it when you forgot requeue, I was able to use L0 to help me found that you added a new secret for proxy in listener and I need to requeue on that secret creation.

nikola-jokic (dismissed):

nikola-jokic (approved):

PR #2279, opened 2023-02-13
Adding parameter to configure the runner set name.

chrispat:
Currently the runner set name is always the same as the helm release name.  However, there are scenarios where one might want to have the same runner names across multiple runner groups and organizations.  Given the helm release still has to be unique per namespace it would be useful to optionally provide a runner set name.

### DIFF
--- a/controllers/actions.github.com/autoscalinglistener_controller.go
+++ b/controllers/actions.github.com/autoscalinglistener_controller.go
@@ -99,6 +99,7 @@ func (r *AutoscalingListenerReconciler) Reconcile(ctx context.Context, req ctrl.
 		}
 
 		log.Info("Successfully removed finalizer after cleanup")
+		return ctrl.Result{}, nil
 	}
 
 	if !controllerutil.ContainsFinalizer(autoscalingListener, autoscalingListenerFinalizerName) {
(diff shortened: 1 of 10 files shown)


controllers/actions.github.com/autoscalinglistener_controller.go
```diff
@@ -99,6 +99,7 @@ func (r *AutoscalingListenerReconciler) Reconcile(ctx context.Context, req ctrl.
 		}
 
 		log.Info("Successfully removed finalizer after cleanup")
+		return ctrl.Result{}, nil
```
TingluoHuang, line 102: noticed we forgot to return early during test locally.

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -128,6 +129,7 @@ func (r *AutoscalingRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl
 		}
 
 		log.Info("Successfully removed finalizer after cleanup")
+		return ctrl.Result{}, nil
```
TingluoHuang, line 127: noticed we forgot to return early during test locally.

rentziass (approved):

PR #2283, opened 2023-02-14
Added workflow to be triggered via rest api dispatch in e2e test

AvaStancu:
Part of [Linux VM testing](https://github.com/github/c2c-actions-runtime/issues/2195) - added before the rest of the steps as we'll use this workflow ID in the rest API /dispatches call

### DIFF
--- a/.github/workflows/e2e-test-dispatch-workflow.yaml
+++ b/.github/workflows/e2e-test-dispatch-workflow.yaml
@@ -1,0 +1,16 @@
+name: ARC-REUSABLE-WORKFLOW
+on:
+  workflow_dispatch:
+    inputs:
+      date_time:
+        description: 'Datetime for runner name uniqueness, format: %Y-%m-%d-%H-%M-%S-%3N, example: 2023-02-14-13-00-16-791'
+        required: true
+jobs:
+  arc-runner-job:
+    strategy:
+      fail-fast: false
+      matrix:
+        job: [1, 2, 3]
+    runs-on: arc-runner-${{ inputs.date_time }}
+    steps:
+      - run: echo "Hello World!" >> $GITHUB_STEP_SUMMARY


.github/workflows/e2e-test-dispatch-workflow.yaml
```diff
@@ -0,0 +1,16 @@
+name: ARC-REUSABLE-WORKFLOW
```
Link-, line 1: This project's been using this style for naming workflows:

https://github.com/actions/actions-runner-controller/blob/1701d34799bcda7498cd1a58fff8c8ab41f70cdb/.github/workflows/publish-arc.yaml#LL1C18-L1C18

Maybe we continue using it? The name will appear in here: https://github.com/actions/actions-runner-controller/actions

.github/workflows/e2e-test-dispatch-workflow.yaml
```diff
@@ -0,0 +1,16 @@
...
+      date_time:
+        description: 'Datetime for runner name uniqueness'
+        required: true
+jobs:
+  arc-runner-job:
+    strategy:
+      fail-fast: false
+      matrix:
+        job: [1, 2, 3]
+    runs-on: arc-runner-${{ inputs.date_time }}
```
Link-, line 14: Is the `date_time` the best suffix to guarantee determinism here?

.github/workflows/e2e-test-dispatch-workflow.yaml
```diff
@@ -0,0 +1,16 @@
+name: ARC-REUSABLE-WORKFLOW
+on:
+  workflow_dispatch:
+    inputs:
+      date_time:
+        description: 'Datetime for runner name uniqueness'
```
Link-: It'll be helpful to add either a comment or preferably include an example of the Date time format in the description

.github/workflows/e2e-test-dispatch-workflow.yaml
AvaStancu: I see different styles there as well. Here is how it will look like with these 2, which personally I prefer as they are since we do not have a pattern, it helps me see them easier albeit for now
Screenshot 2023-02-14 at 14 42 41

.github/workflows/e2e-test-dispatch-workflow.yaml
AvaStancu: If you will have another suggestion I'd be happy to add it, for now this is the name of one of the inputs passed in the dispatches call via the go tests 
Preview: `jsonStr := []byte(fmt.Sprintf(`{"ref":"master", "inputs":{"date_time":"%s"}}`, dateTime))`

.github/workflows/e2e-test-dispatch-workflow.yaml
Link-: Yeah you're right, we should do better :/ I'll leave it up to you, I'm not sure adding a 3rd style is the best way to go though

.github/workflows/e2e-test-dispatch-workflow.yaml
Link-: Is there a situation where 2 tests could run at exactly the same time & attempt to create runner scale sets with the same name for example? What's the granularity of this date_time field? Is it a timestamp? in milliseconds / micro / seconds?

Link- (approved):
🚢

PR #2285, opened 2023-02-14
Avastancu/arc e2e test linux vm

AvaStancu:
Second PR to cover the base for [Linux VM testing batch](https://github.com/github/c2c-actions-runtime/issues/2195)
First PR was [this one ](https://github.com/actions/actions-runner-controller/pull/2283) to add a workflow and then in this 2nd PR use its ID for the manual trigger in the go tests.
Workflow `e2e-test-linux-vm.yaml` is set to be triggered on `workflow_dispatch` right now in order to test it manually without having to do any "fake" push. Once confirmed it works properly (permissions ok, no other bug) I will set it to be triggered on push.
The action included covers the helm installation and also runs the go tests. Can be reused if needed.


.github/actions/e2e-arc-test/action.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+inputs:
+  github-token:
+    description: 'A Github PAT'
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - name: Helm installation
```
TingluoHuang: our hosted runner doesn't have helm already installed?

.github/actions/e2e-arc-test/action.yaml
```diff
@@ -0,0 +1,44 @@
+name: 'E2E ARC Test Action'
+description: 'Includes common arc installation, setup and test file run'
+
+inputs:
+  github-token:
+    description: 'A Github PAT'
```
Link-: Are we only accepting GitHub Personal Access Tokens here? What if we use a token from a GitHub App?

.github/actions/e2e-arc-test/action.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - name: Helm installation
+      run: |
+        curl https://baltocdn.com/helm/signing.asc | gpg --dearmor | sudo tee /usr/share/keyrings/helm.gpg > /dev/null
+        sudo apt-get install apt-transport-https --yes
+        echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/helm.gpg] https://baltocdn.com/helm/stable/debian/ all main" | sudo tee /etc/apt/sources.list.d/helm-stable-debian.list
+        sudo apt-get update
+        sudo apt-get install helm
+      shell: bash
+    - name: Install ARC
+      run: helm install arc --namespace "arc-systems" --create-namespace oci://ghcr.io/actions/actions-runner-controller-charts/actions-runner-controller-2 --version 0.1.0
```
Link-: Should we be hardcoding the version used or do we want to parametrise this?

.github/actions/e2e-arc-test/action.yaml
```diff
@@ -0,0 +1,44 @@
...
+      run: echo "DATE_TIME=$(date +'%Y-%m-%d-%H-%M-%S-%3N')" >> $GITHUB_ENV
+      shell: bash
+    - name: Install runners
+      run: |
+          helm install "arc-runner-${{ env.DATE_TIME }}" \
+          --namespace "arc-runners" \
+          --create-namespace \
+          --set githubConfigUrl="https://github.com/actions/actions-runner-controller" \
+          --set githubConfigSecret.github_token="${{ inputs.github-token }}" \
+          oci://ghcr.io/actions/actions-runner-controller-charts/auto-scaling-runner-set --version 0.1.0 \
```
Link-: Same here for the chart version

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -0,0 +1,28 @@
...
+  workflow_dispatch:
+  
+jobs:
+  setup-steps:
+    runs-on: [ubuntu-latest]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.18
```
Link-: Go is pre-installed on the runners, no?

https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - name: Install latest version of Kind
+        run: go install sigs.k8s.io/kind@latest
```
Link-: So is Kind

https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - name: Create Kind cluster
+        run: |
+          PATH=$(go env GOPATH)/bin:$PATH
+          kind create cluster --name e2e-test
+      - name: Get Token
+        id: get_workflow_token
+        uses: peter-murray/workflow-application-token-action@v1
```
Link-: I would pin this to a specific SHA instead of using v1. Tags in GitHub are not immutable (can be deleted / re-created), a fix is on the way but until then it's best to pin to a specific version

test_e2e_arc/arc_jobs_test.go
```diff
@@ -0,0 +1,132 @@
+package e2e_dev_arc
```
Link-: nit: a more descriptive package name? we've archived `dev_arc` already no?

test_e2e_arc/arc_jobs_test.go
```diff
@@ -0,0 +1,132 @@
...
+			t.Fatal("Expected pods count did not match available pods count before job run.")
+		}
+	},
+	)
+	t.Run("Get available pods during job run", func(t *testing.T) {
+		c := http.Client{}
+		dateTime := os.Getenv("DATE_TIME")
+		// We are triggering manually a workflow that already exists in the repo.
+		// This workflow is expected to spin up a number of runner pods matching the runners value set in podCountsByType.
+		url := "https://api.github.com/repos/actions/actions-runner-controller/actions/workflows/48368242/dispatches"
```
Link-: Why don't we use: https://github.com/cli/go-gh instead of hardcoding the workflow ID here? You can trigger a workflow run based on the workflow name or better yet, a workflow filename: https://cli.github.com/manual/gh_workflow_run

AvaStancu: Each one of ID, name, path will be hardcoded and any of these can change. If the name of the workflow changes, the ID and path stay the same. If the path changes, the ID will change. I will try to add the path for this reason and for readability! Thanks!

.github/actions/e2e-arc-test/action.yaml
AvaStancu: Not sure, was following the commands from here https://github.com/actions/actions-runner-controller/tree/master/docs/preview/actions-runner-controller-2
Can't we just take the `latest` instead of specifying the version?

.github/actions/e2e-arc-test/action.yaml
AvaStancu: Same as above, can we use the latest?

test_e2e_arc/arc_jobs_test.go
AvaStancu: Updated, thanks!

.github/actions/e2e-arc-test/action.yaml
AvaStancu: No, it was an outdated description, thanks for catching this!

.github/actions/e2e-arc-test/action.yaml
AvaStancu: Seems it does, thanks!

.github/workflows/e2e-test-linux-vm.yaml
AvaStancu: Yes, was not aware, thank you!

.github/workflows/e2e-test-linux-vm.yaml
AvaStancu: 🥳

test_e2e_arc/arc_jobs_test.go
AvaStancu: Updated 🎊

.github/workflows/e2e-test-linux-vm.yaml
AvaStancu: There are pros and cons to both but I'm ok to add the SHA as well for now, updated

.github/actions/e2e-arc-test/action.yaml
AvaStancu: @nikola-jokic had a good observation, we should use the local charts to test the code before publishing a new version, updated

.github/actions/e2e-arc-test/action.yaml
AvaStancu: Updated

.github/actions/e2e-arc-test/action.yaml
Link-: Yeah I agree, that's why making this an input parameter might be a good idea. We can then instruct it to be `local` or specify a release version `x.x.x` depending on the usecase

Link- (approved):
🚢

PR #2286, opened 2023-02-14
Add support for proxy

rentziass:
This adds support for proxy configuration in the autoscaling runner set chart, which you can use by setting any (or all) of these parameters when creating a runner set:

```bash
helm install "${INSTALLATION_NAME}" \
    --namespace "${NAMESPACE}" \
    --create-namespace \
    --set githubConfigUrl="${GITHUB_CONFIG_URL}" \
    --set githubConfigSecret.github_token="${GITHUB_PAT}" \
    --set proxy.http.url="http://proxy:3128" \ # proxy settings start here
    --set proxy.http.credentialSecretRef="proxy-auth" \
    --set proxy.https.url="http://proxy:3128" \
    --set proxy.https.credentialSecretRef="proxy-auth" \
    --set proxy.noProxy="{example.com,example.org}" \
    ./charts/auto-scaling-runner-set
```

The controller will use those proxy settings in its actions.Client when talking to GitHub APIs, and `http_proxy`, `https_proxy` and `no_proxy` environment variables will be added to the listener and ephemeral runners pods that belong to that runner set. If your proxy lives outside Kubernetes (or has no access to k8s API) you **will need** to add Kubernetes API address to `proxy.noProxy` or the listener will fail with `Service Unavailable` errors (if proxy env vars are set, by default Go http clients use them, so these applies to both calls to GitHub API and calls to k8s API in the case of the listener).

## Testing
I've tested this with a runner set with no proxy settings, one using a proxy outside of k8s and one using a proxy inside of k8s. The scenarios involving proxies used Squid with the `rentziass/squid` image (which uses `http_admin` : `password1234` for auth). If anyone wants to play around with this/spot problems with how this was tested see details below. To be able to run tests against the same repo (aka same config url) and avoiding conflicts of proxy/no proxy the namespace was nuked between each scenario.

<details>
<summary><h3>No proxy</h3></summary>
Install arc:

```bash
helm install arc \
    --namespace "${NAMESPACE}" \
    --create-namespace \
    --set image.repository=rentziass/actions-runner-controller \
    --set image.tag=dev \
    --set image.pullPolicy=Always \
    ./charts/actions-runner-controller-2
```

Create runner set:

```bash
helm install "no-proxy-arc" \
    --namespace "${NAMESPACE}" \
    --create-namespace \
    --set githubConfigUrl="${GITHUB_CONFIG_URL}" \
    --set githubConfigSecret.github_token="${GITHUB_PAT}" \
    --set runnerGroup="" \
    ./charts/auto-scaling-runner-set
```

Run a workflow with `runs-on: no-proxy-arc`.

</details>

<details>
<summary><h3>Proxy outside k8s</h3></summary>
Start Squid on the host using Docker:

```bash
docker run -d \
    --name squid \
    --publish 3128:3128 \
    -e TZ=UTC \
    rentziass/squid
```

Install arc:

```bash
helm install arc \
    --namespace "${NAMESPACE}" \
    --create-namespace \
    --set image.repository=rentziass/actions-runner-controller \
    --set image.tag=dev \
    --set image.pullPolicy=Always \
    ./charts/actions-runner-controller-2
```

Create a secret to use to authenticate the proxy:

```bash
kubectl create secret generic proxy-auth \
    --namespace=$NAMESPACE \
    --from-literal=username=http_admin \
    --from-literal=password='password1234'
```

Create runner set:

```bash
helm install "proxy-outside-arc" \
    --namespace "${NAMESPACE}" \
    --create-namespace \
    --set githubConfigUrl="${GITHUB_CONFIG_URL}" \
    --set githubConfigSecret.github_token="${GITHUB_PAT}" \
    --set proxy.http.url="http://host.minikube.internal:3128" \  # if you're using minikube, this corresponds to your host
    --set proxy.http.credentialSecretRef="proxy-auth" \
    --set proxy.https.url="http://host.minikube.internal:3128" \
    --set proxy.https.credentialSecretRef="proxy-auth" \
    --set proxy.noProxy="{example.com,example.org,10.96.0.1:443}" \ # swap 10.96.0.1:443 with the address of k8s api in your cluster
    --set runnerGroup="" \
    ./charts/auto-scaling-runner-set
```

Run a workflow with `runs-on: proxy-outside-arc`.

</details>

<details>
<summary><h3>Proxy inside k8s</h3></summary>
Using this manifest (called `squid.yaml` in following commands):

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: squid
  labels:
    app: squid
spec:
  replicas: 1
  selector:
    matchLabels:
      app: squid
  template:
    metadata:
      labels:
        app: squid
    spec:
      containers:
      - name: squid
        image: rentziass/squid
        ports:
        - containerPort: 3128
---
apiVersion: v1
kind: Service
metadata:
  name: squid
spec:
  selector:
    app: squid
  ports:
    - protocol: TCP
      port: 3128
      targetPort: 3128
```

Create a deployment and a service for squid with

```bash
kubectl apply -f squid.yaml
```

Install arc:

```bash
helm install arc \
    --namespace "${NAMESPACE}" \
    --create-namespace \
    --set image.repository=rentziass/actions-runner-controller \
    --set image.tag=dev \
    --set image.pullPolicy=Always \
    ./charts/actions-runner-controller-2
```

Create a secret to use to authenticate the proxy:

```bash
kubectl create secret generic proxy-auth \
    --namespace=$NAMESPACE \
    --from-literal=username=http_admin \
    --from-literal=password='password1234'
```

Create runner set:

```bash
helm install "proxy-inside-arc" \
    --namespace "${NAMESPACE}" \
    --create-namespace \
    --set githubConfigUrl="${GITHUB_CONFIG_URL}" \
    --set githubConfigSecret.github_token="${GITHUB_PAT}" \
    --set proxy.http.url="http://10.99.193.57:3128" \ # swap 10.99.193.57 with the address of `squid` service in your cluster
    --set proxy.http.credentialSecretRef="proxy-auth" \
    --set proxy.https.url="http://10.99.193.57:3128" \
    --set proxy.https.credentialSecretRef="proxy-auth" \
    --set proxy.noProxy="{example.com,example.org}" \
    --set runnerGroup="" \
    ./charts/auto-scaling-runner-set
```

Run a workflow with `runs-on: proxy-inside-arc`.

</details>

/cc @nikola-jokic


github/actions/client.go
```diff
@@ -259,7 +269,7 @@ func (c *Client) NewActionsServiceRequest(ctx context.Context, method, path stri
 	}
 
 	req.Header.Set("Content-Type", "application/json")
-	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.ActionsServiceAdminToken))
+	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", c.ActionsServiceAdminToken))
```
TingluoHuang: why do we need to change this?

apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
```diff
@@ -80,6 +85,124 @@ type ProxyConfig struct {
 
 	// +optional
 	HTTPS *ProxyServerConfig `json:"https,omitempty"`
+
+	// +optional
+	NoProxy []string `json:"noProxy,omitempty"`
+}
+
+func (c *ProxyConfig) ProxyFunc(secretFetcher func(string) (*corev1.Secret, error)) (func(*http.Request) (*url.URL, error), error) {
```
TingluoHuang: feels like these methods should live in some http client.go files instead of the types.go

controllers/actions.github.com/autoscalinglistener_controller_test.go
```diff
@@ -15,7 +15,7 @@ import (
 	. "github.com/onsi/gomega"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	actionsv1alpha1 "github.com/actions/actions-runner-controller/apis/actions.github.com/v1alpha1"
```
TingluoHuang, line 18: I would avoid changing this in this PR to reduce the amount of PR diff. 😄

controllers/actions.github.com/autoscalinglistener_controller.go
```diff
@@ -345,7 +346,24 @@ func (r *AutoscalingListenerReconciler) createServiceAccountForListener(ctx cont
...
 
 func (r *AutoscalingListenerReconciler) createListenerPod(ctx context.Context, autoscalingRunnerSet *v1alpha1.AutoscalingRunnerSet, autoscalingListener *v1alpha1.AutoscalingListener, serviceAccount *corev1.ServiceAccount, secret *corev1.Secret, logger logr.Logger) (ctrl.Result, error) {
-	newPod := r.resourceBuilder.newScaleSetListenerPod(autoscalingListener, serviceAccount, secret)
+	var envs []corev1.EnvVar
+	if autoscalingListener.Spec.Proxy != nil {
+		proxyEnv, err := autoscalingListener.Spec.Proxy.EnvVars(func(s string) (*corev1.Secret, error) {
+			var secret corev1.Secret
+			err := r.Get(ctx, types.NamespacedName{Name: s, Namespace: autoscalingRunnerSet.Namespace}, &secret)
+			if err != nil {
+				return nil, err
```
TingluoHuang: should we pass the logger into these places so we can log extra info about each error?

github/actions/client.go
rentziass: We don't, sorry 😅 that's a leftover from a bug that left scratching our heads for a while.

apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
rentziass: We can type the function returned here in client, that's a good shout. This feels like a method for ProxyConfig though, it knows everything it needs about itself already and personally it doesn't feel right to couple a client with a k8s type. What do you think?

controllers/actions.github.com/autoscalinglistener_controller_test.go
rentziass: Yup, reverting 👍

apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
rentziass: Also worth noting the returned func isn't a client specific detail but it's the type [http.Transport](https://pkg.go.dev/net/http#Transport) expects

controllers/actions.github.com/autoscalinglistener_controller_test.go
rentziass: Diff only got 7 lines smaller 😭

controllers/actions.github.com/autoscalinglistener_controller.go
rentziass: I've added logging in place where we have a logger already, otherwise I've checked the other instances: the error returned is bubbled up and logged immediately from the caller

charts/auto-scaling-runner-set/templates/autoscalingrunnerset.yaml
```diff
@@ -12,6 +12,23 @@ spec:
...
+    {{- if .Values.proxy.https }}
+    https:
+      url: {{ .Values.proxy.https.url }}
+      credentialSecretRef: {{ .Values.proxy.https.credentialSecretRef }}
+    {{ end }}
+    {{- if and .Values.proxy.noProxy (kindIs "slice" .Values.proxy.noProxy) }}
+    noProxy: {{ .Values.proxy.noProxy | toYaml | nindent 6}}
+    {{ end }}
+  {{ end }}
+
```
rentziass, line 31: This gets rendered to 

```yaml
spec:
  # ...
  proxy:
    http:
      url: http://proxy:3128
      credentialSecretRef: proxy-auth

    https:
      url: http://proxy:3128
      credentialSecretRef: proxy-auth

    noProxy:
      - example.com
      - example.org
```

TingluoHuang: do you mind adding an L0 test for this in `charts/auto-scaling-runner-set/tests/template_test.go`?

apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
```diff
@@ -80,6 +85,124 @@ type ProxyConfig struct {
...
+	config := &httpproxy.Config{
+		NoProxy: strings.Join(c.NoProxy, ","),
+	}
+
+	if c.HTTP != nil {
+		u, err := url.Parse(c.HTTP.Url)
+		if err != nil {
+			return nil, err
```
TingluoHuang: do you mind wrapping all errors to provide extra info, ex: `fmt.Errorf("failed to parse http proxy url: %v", err)`?

apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
```diff
@@ -80,6 +85,124 @@ type ProxyConfig struct {
...
+		return config.ProxyFunc()(req.URL)
+	}
+
+	return proxyFunc, nil
+}
+
+func (c *ProxyConfig) EnvVars(secretFetcher func(string) (*corev1.Secret, error)) ([]corev1.EnvVar, error) {
+	var vars []corev1.EnvVar
+
+	if c.HTTP != nil {
```
TingluoHuang: is it possible to put the common code between `EnvVars` and `ProxyFunc` into some shared function?

charts/auto-scaling-runner-set/templates/autoscalingrunnerset.yaml
TingluoHuang: do you also need to update the `values.yaml` to provide some example input behind comments?

apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
```diff
@@ -80,6 +85,124 @@ type ProxyConfig struct {
...
+		vars = append(vars, corev1.EnvVar{
+			Name:  "no_proxy",
+			Value: strings.Join(c.NoProxy, ","),
+		})
+	}
+
+	return vars, nil
 }
 
 type ProxyServerConfig struct {
```
TingluoHuang, line 179: I know this is kind of too late for this type of suggestion, but stilling want to bring it up...
What if we change `ProxyServerConfig` to be
```go
type ProxyServerConfig struct {
  // +optional
  AnonymousUrl string `json:"anonymousUrl,omitempty"`

  // +optional
  AuthenticatedUrlSecretRef string `json:"AuthenticatedUrlSecretRef,omitempty"`
}
```

And we ask the customer to provide the wellformed auth URL as the data in the secret.

I was trying to think if the customer already has an auth proxy, what kind of secret they may already have, can we reuse that?

Sorry about the extra thoughts.

controllers/actions.github.com/autoscalinglistener_controller.go
```diff
@@ -345,7 +346,25 @@ func (r *AutoscalingListenerReconciler) createServiceAccountForListener(ctx cont
...
+				return nil, fmt.Errorf("failed to get secret %s: %w", s, err)
+			}
+			return &secret, nil
+		})
+		if err != nil {
+			logger.Error(err, "Unable to get proxy env vars")
+			return reconcile.Result{}, fmt.Errorf("failed to get proxy env vars: %w", err)
+		}
+
+		envs = append(envs, proxyEnv...)
```
TingluoHuang: do we need to use secret for the env, since it may contain a username/password about the proxy auth?

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -495,7 +494,31 @@ func (r *AutoscalingRunnerSetReconciler) actionsClientFor(ctx context.Context, a
...
 	}
 
-	return r.ActionsClient.GetClientFromSecret(ctx, autoscalingRunnerSet.Spec.GitHubConfigUrl, autoscalingRunnerSet.Namespace, configSecret.Data)
+	var opts []actions.ClientOption
+	if autoscalingRunnerSet.Spec.Proxy != nil {
+		proxyFunc, err := autoscalingRunnerSet.Spec.Proxy.ProxyFunc(func(s string) (*corev1.Secret, error) {
+			var secret corev1.Secret
+			err := r.Get(ctx, types.NamespacedName{Namespace: autoscalingRunnerSet.Namespace, Name: s}, &secret)
+			if err != nil {
+				return nil, fmt.Errorf("failed to get secret %s: %w", s, err)
```
TingluoHuang:
```suggestion
				return nil, fmt.Errorf("failed to get proxy secret %s: %w", s, err)
```

controllers/actions.github.com/ephemeralrunner_controller.go
```diff
@@ -557,8 +557,28 @@ func (r *EphemeralRunnerReconciler) updateStatusWithRunnerConfig(ctx context.Con
...
+	var envs []corev1.EnvVar
+	if runner.Spec.Proxy != nil {
+		proxyEnvs, err := runner.Spec.Proxy.EnvVars(func(s string) (*corev1.Secret, error) {
+			var secret corev1.Secret
+			err := r.Client.Get(ctx, types.NamespacedName{
+				Name:      s,
+				Namespace: runner.Namespace,
+			}, &secret)
+			if err != nil {
+				return nil, fmt.Errorf("failed to get secret %s: %w", s, err)
```
TingluoHuang:
```suggestion
				return nil, fmt.Errorf("failed to get proxy secret %s: %w", s, err)
```

controllers/actions.github.com/ephemeralrunner_controller.go
```diff
@@ -632,7 +652,31 @@ func (r *EphemeralRunnerReconciler) actionsClientFor(ctx context.Context, runner
...
 	}
 
-	return r.ActionsClient.GetClientFromSecret(ctx, runner.Spec.GitHubConfigUrl, runner.Namespace, secret.Data)
+	var opts []actions.ClientOption
+	if runner.Spec.Proxy != nil {
+		proxyFunc, err := runner.Spec.Proxy.ProxyFunc(func(s string) (*corev1.Secret, error) {
+			var secret corev1.Secret
+			err := r.Get(ctx, types.NamespacedName{Namespace: runner.Namespace, Name: s}, &secret)
+			if err != nil {
+				return nil, err
```
TingluoHuang:
```suggestion
				return nil, fmt.Errorf("failed to get proxy secret %s: %w", s, err)
```

controllers/actions.github.com/ephemeralrunner_controller.go
```diff
@@ -632,7 +652,31 @@ func (r *EphemeralRunnerReconciler) actionsClientFor(ctx context.Context, runner
...
+			}
+
+			return &secret, nil
+		})
+		if err != nil {
+			return nil, err
```
TingluoHuang:
```suggestion
			return nil, fmt.Errorf("failed to get proxy func: %w", err)
```

controllers/actions.github.com/ephemeralrunner_controller.go
```diff
@@ -557,8 +557,28 @@ func (r *EphemeralRunnerReconciler) updateStatusWithRunnerConfig(ctx context.Con
...
+			if err != nil {
+				return nil, fmt.Errorf("failed to get secret %s: %w", s, err)
+			}
+			return &secret, nil
+		})
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to get proxy env vars: %w", err)
+		}
+
+		envs = append(envs, proxyEnvs...)
```
TingluoHuang: I was thinking the same question about using secret for the env here as well, since the proxy URL may contains a username/password

apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
rentziass: I actually like this, and if there's a time to make this kind of changes is now 👍 I genuinely don't know what would be more comfortable/less confusing from a user perspective. Just putting both cases here to see them next to each other

```go
type ProxyServerConfig struct {
  // Required
  Url string `json:"url,omitempty"`

  // +optional
  CredentialSecretRef string `json:"credentialSecretRef,omitempty"`
}

// vs

type ProxyServerConfig struct {
  // +optional
  AnonymousUrl string `json:"anonymousUrl,omitempty"`

  // +optional
  AuthenticatedUrlSecretRef string `json:"AuthenticatedUrlSecretRef,omitempty"`
}
```

I like that in the second case we don't deal with building the URL ourselves, although the first one seems less confusing to me 🤔

controllers/actions.github.com/autoscalinglistener_controller.go
rentziass: Eventually we do yes, to build the URL and set it as an env var (or read the authenticated URL directly if we change the spec). It's pretty standard to surface secrets as env vars in pods though, the only way around this would be putting the secret in a file and mount that inside a volume.

charts/auto-scaling-runner-set/templates/autoscalingrunnerset.yaml
rentziass: Will do as soon as we're set on the spec 👌

apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
```diff
@@ -80,6 +85,124 @@ type ProxyConfig struct {
...
+	// +optional
+	NoProxy []string `json:"noProxy,omitempty"`
+}
+
+func (c *ProxyConfig) ProxyFunc(secretFetcher func(string) (*corev1.Secret, error)) (func(*http.Request) (*url.URL, error), error) {
+	config := &httpproxy.Config{
+		NoProxy: strings.Join(c.NoProxy, ","),
+	}
+
+	if c.HTTP != nil {
```
AvaStancu, line 99: this HTTPProxy configuration along with the HTTPSProxy configuration could be extracted in a common function to avoid duplication?

apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
fhammerl: Isn't the 2nd definition contained in the first one? I mean, can't one just use `Url` as either `AnonymousUrl` or `AuthenticatedUrlSecretRef` and leave `CredentialSecretRef` empty?

rentziass: We decided to go with the current spec.

controllers/actions.github.com/autoscalinglistener_controller.go
TingluoHuang: do we need to put the `http(s)_proxy` env with auth into a secret we create and reference it via the following?
```
    - name: https_proxy
      valueFrom:
        secretKeyRef:
          name: proxyAuthUrl
          key: https_proxy
```

Otherwise, I think anyone with `Get` pod permission can get the proxy auth by reading the env, will that be a problem?

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: jokicnikola07/actions-runner-controller
+  newName: summerwind/actions-runner-controller
```
fhammerl, line 7: TODO: figure out newName value

controllers/actions.github.com/autoscalinglistener_controller.go
rentziass: Ha I see what you mean now, yeah that sounds about right 🤦 . Thanks for pointing that out!

fhammerl: To piggyback on that, if `CredentialSecretRef`, or even its value, changes, we'll have to be able to propagate that change all the way down to the runner pods.

apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
```diff
@@ -80,6 +86,126 @@ type ProxyConfig struct {
...
+	}
+
+	proxyFunc := func(req *http.Request) (*url.URL, error) {
+		return config.ProxyFunc()(req.URL)
+	}
+
+	return proxyFunc, nil
+}
+
+func (c *ProxyConfig) EnvVars(secretFetcher func(string) (*corev1.Secret, error)) ([]corev1.EnvVar, error) {
```
TingluoHuang: is this still used?

rentziass: Nope, you got here before the cleanup 😂

apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
```diff
@@ -80,6 +86,94 @@ type ProxyConfig struct {
...
+		config.HTTPSProxy = u.String()
+	}
+
+	return config, nil
+}
+
+func (c *ProxyConfig) ToSecretData(secretFetcher func(string) (*corev1.Secret, error)) (map[string][]byte, error) {
+	config, err := c.toHTTPProxyConfig(secretFetcher)
+	if err != nil {
+		return nil, err
```
TingluoHuang, line 155:
```suggestion
		return nil, fmt.Errorf("fail to create HTTP proxy config: %w", err)
```

apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
```diff
@@ -80,6 +86,94 @@ type ProxyConfig struct {
...
+	data["https_proxy"] = []byte(config.HTTPSProxy)
+	data["no_proxy"] = []byte(config.NoProxy)
+
+	return data, nil
+}
+
+func (c *ProxyConfig) ProxyFunc(secretFetcher func(string) (*corev1.Secret, error)) (func(*http.Request) (*url.URL, error), error) {
+	config, err := c.toHTTPProxyConfig(secretFetcher)
+	if err != nil {
+		return nil, err
```
TingluoHuang, line 169:
```suggestion
		return nil, fmt.Errorf("fail to create HTTP proxy config: %w", err)
```

apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
rentziass: We wrap these in `toHTTPProxyConfig` and with the callers already, happy to wrap here too just wondering if at some point these will actually make it harder to get to the actual error?

controllers/actions.github.com/autoscalinglistener_controller.go
```diff
@@ -345,7 +380,43 @@ func (r *AutoscalingListenerReconciler) createServiceAccountForListener(ctx cont
...
+		httpURL := corev1.EnvVar{
+			Name: "http_proxy",
+			ValueFrom: &corev1.EnvVarSource{
+				SecretKeyRef: &corev1.SecretKeySelector{
+					LocalObjectReference: corev1.LocalObjectReference{Name: proxyListenerSecretName(autoscalingListener)},
+					Key:                  "http_proxy",
+				},
+			},
+		}
+		envs = append(envs, httpURL)
```
TingluoHuang: should we add each env as needed?
ex:
```
if autoscalingListener.Spec.Proxy.HTTP != nil {
  envs = append(envs, httpURL)
}
```

rentziass: Yeah I think that would help debugging issues, right now it works as in the env var would be set to `""` if absent but it may confuse to see that being set if you didn't provide it. I'll change it for the runner too, thanks!

controllers/actions.github.com/ephemeralrunnerset_controller.go
```diff
@@ -122,6 +122,24 @@ func (r *EphemeralRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+		proxySecret := new(corev1.Secret)
+		if err := r.Get(ctx, types.NamespacedName{Namespace: ephemeralRunnerSet.Namespace, Name: proxyEphemeralRunnerSetSecretName(ephemeralRunnerSet)}, proxySecret); err != nil {
+			if !kerrors.IsNotFound(err) {
+				log.Error(err, "Unable to get ephemeralRunnerSet proxy secret", "namespace", ephemeralRunnerSet.Namespace, "name", proxyEphemeralRunnerSetSecretName(ephemeralRunnerSet))
+				return ctrl.Result{}, err
+			}
+
+			// Create a compiled secret for the runner pods in the runnerset namespace
+			log.Info("Creating a ephemeralRunnerSet proxy secret for the runner pods")
+			if err := r.createProxySecret(ctx, ephemeralRunnerSet, log); err != nil {
```
TingluoHuang, line 136: should we make AutoScalingRunnerSet_controller to create/cleanup this secret?

the EphemeralRunnerSet will get recreated every time there is a pod template change.

apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
TingluoHuang: I think it's fine for here since there is only one `return nil, err` in this method, so it's easy fo figure out where is the error coming from. 🙇

apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
TingluoHuang: same. should be fine with the existing code.

controllers/actions.github.com/ephemeralrunnerset_controller.go
rentziass: When talking about this with @nikola-jokic we decided to go this way for that very reason, even considering the extreme of creating one secret per runner but that was introducing way too overhead and duplication. We went this way to decouple the two controllers, otherwise the other extreme is using one secret for all (listener and runners): a lot easier to reason about but very coupled. I trust your and Nikola's opinions more, so up to you really!

TingluoHuang: I am fine with the current approach. 👍

controllers/actions.github.com/ephemeralrunnerset_controller.go
```diff
@@ -290,6 +335,47 @@ func (r *EphemeralRunnerSetReconciler) createEphemeralRunners(ctx context.Contex
...
+	})
+	if err != nil {
+		return fmt.Errorf("failed to convert proxy config to secret data: %w", err)
+	}
+
+	runnerPodProxySecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      proxyEphemeralRunnerSetSecretName(ephemeralRunnerSet),
+			Namespace: ephemeralRunnerSet.Namespace,
+			Labels:    map[string]string{
```
TingluoHuang, line 352: do we want to fix this TODO?

controllers/actions.github.com/ephemeralrunnerset_controller.go
```diff
@@ -290,6 +335,47 @@ func (r *EphemeralRunnerSetReconciler) createEphemeralRunners(ctx context.Contex
...
+				// TODO: figure out autoScalingRunnerSet name and set it as a label for this secret
+				// "auto-scaling-runner-set-namespace": ephemeralRunnerSet.Namespace,
+				// "auto-scaling-runner-set-name": ephemeralRunnerSet.Name,
+			},
+		},
+		Data: proxySecretData,
+	}
+
+	// create runnerPodProxySecret
```
TingluoHuang: this comment seems off a little bit.

controllers/actions.github.com/ephemeralrunnerset_controller.go
rentziass: Thanks, I forgot this was a thing :( never again PRs this size, feel free to quote me on that 😄

controllers/actions.github.com/ephemeralrunnerset_controller.go
```diff
@@ -349,17 +349,14 @@ func (r *EphemeralRunnerSetReconciler) createProxySecret(ctx context.Context, ep
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      proxyEphemeralRunnerSetSecretName(ephemeralRunnerSet),
 			Namespace: ephemeralRunnerSet.Namespace,
-			Labels:    map[string]string{
-				// TODO: figure out autoScalingRunnerSet name and set it as a label for this secret
-				// "auto-scaling-runner-set-namespace": ephemeralRunnerSet.Namespace,
-				// "auto-scaling-runner-set-name": ephemeralRunnerSet.Name,
+			Labels: map[string]string{
+				"auto-scaling-runner-set-namespace": ephemeralRunnerSet.Namespace,
+				"auto-scaling-runner-set-name":      ephemeralRunnerSet.Name,
```
TingluoHuang: is this label right? `auto-scaling-runner-set-name` is set to an `ephemeral-runner-set-name`

rentziass: 🤦 that's why it was a todo, we have another issue to deal with labels to enable us to group all resources belonging to a scale set, reverting the todo.

TingluoHuang: 👍

TingluoHuang (approved):

PR #2287, opened 2023-02-15
doc: Fix chart name for helm commands in docs

mumoshu:
Currently, some helm command examples do not work due to inconsistency between the repository name specified in the helm-repo-add and helm-upgrade commands.

`installing-arc.md` and `quickstard.md` use `actions-runner-controller` as the repo name:

```
helm repo add actions-runner-controller https://actions-runner-controller.github.io/actions-runner-controller
```

while some helm upgrade commands refers to the repo name as `actions`.

This fixes helm-upgrade command examples so that they refer to `actions-runner-controller`.

Note that we've been using `actions-runner-controller` since before the repo transfer and I assume many folks are still referring to the chart repo name as `actions-runner-controller` in their local and CI helm installation. Probably that justifies fixing helm-upgrade examples rather than fixing helm-repo-add to use `actions`.

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -67,7 +67,7 @@ To fix this, you may either:
    # With helm, you'd set `webhookPort` to the port number of your choice
    # See https://github.com/actions/actions-runner-controller/pull/1410/files for more information
    helm upgrade --install --namespace actions-runner-system --create-namespace \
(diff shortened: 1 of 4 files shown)


Link- (approved):
Thank you for fixing this @mumoshu 🙏

mumoshu:
Thanks for reviewing @Link-! Merging.

PR #2291, opened 2023-02-15
Fix helm chart when containerMode.type=dind.

TingluoHuang:
The chart is not rendered properly when `--set containerMode.type=dind`

It was rendered to:
```
     containers:
      - name: runner
        command: [/actions-runner/run.sh]
        env:
        volumeMounts:
          - name: work
            mountPath: /actions-runner/_work
          - name: dind-cert
            mountPath: /certs/client
```

Notice the `env` section is empty due to the bug.

### DIFF
--- a/charts/auto-scaling-runner-set/templates/_helpers.tpl
+++ b/charts/auto-scaling-runner-set/templates/_helpers.tpl
@@ -208,18 +208,18 @@ env:
           {{- end }}
         {{- end }}
       {{- end }}
-      {{- if $setDockerHost }}
+    {{- end }}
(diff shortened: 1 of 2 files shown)


charts/auto-scaling-runner-set/templates/_helpers.tpl
```diff
@@ -208,18 +208,18 @@ env:
           {{- end }}
```
TingluoHuang, line 208: hide whitespace when reviewing.

TLDR, the bug is `{{- end }}` get added in the wrong line.

rentziass (approved):

PR #2153, opened 2023-01-13
Introduce new preview auto-scaling mode for ARC.

TingluoHuang:
As mentioned in https://github.com/actions/actions-runner-controller/discussions/2072, we are trying to introduce a new auto-scaling mode to ARC and hope to improve the following areas:

- No longer require `cert-manager` as a pre-requirement for using ARC.
- Reliable scale down to zero runner pod and scale up based on job demands.
- Reduce API requests to `api.github.com` in order to avoid rate-limiting.
- No longer pass your GitHub admin privilege token to the runner pod for runner registration.
- Max flexibility on customizing your runner pod template.
- Unify ARC installation process to helm charts.

The preview auto-scaling mode introduced in this PR requires certain new APIs on GitHub.com. 
Those new APIs are behind a feature flag which is `OFF` right now, we will run through a set of testing before turning on the feature flag for everyone.

**TLDR: You can't use this new auto-scaling mode now with your GitHub enterprise/organization/repository**

PR #2293, opened 2023-02-16
Fix manager crashloopback for ARC deployments without scaleset-related controllers

mumoshu:
I've realized that ARC's controller manager has been unable to start correctly with the current Helm chart since #2153, due to incompatibility between the current Helm chart (missing new, scaleset related actions.github.com_* CRDs) and the controller-manager's entrypoint (main.go, which requires the scaleset related CRDs on startup).

I'd like to modify the main.go so that it doesn't require the new CRDs when ARC is deployed via our current chart (=actions-runner-controller chart, not actions-runner-controller-2 chart).

ARC now enables the new scale-set related controllers only when the CONTROLLER_MANAGER_POD_NAME envvar is provided. This way we can fix the issue without changing the current or the new charts. This change is backward-compatible for both the private testers and the current users of ARC.


TingluoHuang (approved):

mumoshu:
@TingluoHuang Thanks for your review! Merging.

PR #2295, opened 2023-02-16
Add actions runner CRDs Helm Chart

minhthong582000:
This is a PR for the issue i've created https://github.com/actions/actions-runner-controller/issues/2125 to add ability to specify custom annotations to the Helm Chart CRDs manifests.

After some discussions in the issue above, we decided to create a new helm chart to handle the lifecycle of the CRDs.

I also created a small script that helps us generate the CRDs templates so we don't have to copy and paste them if there're any changes in the future. And a github actions workflow to publish the helm chart.

### DIFF
--- a/.github/workflows/arc-crds-publish-chart.yaml
+++ b/.github/workflows/arc-crds-publish-chart.yaml
@@ -1,0 +1,73 @@
+name: (crds) Publish Helm Charts
+
+# Revert to https://github.com/actions-runner-controller/releases#releases
+# for details on why we use this approach
+on:
+  push:
+    branches:
+    - master
+    paths:
+    - 'charts/**'
+    - '.github/workflows/arc-crds-publish-chart.yaml'
+    - '!charts/actions-runner-controller/**'
+    - '!charts/gha-runner-scale-set-controller/**'
+    - '!charts/gha-runner-scale-set/**'
+    - '!**.md'
+
+env:
+  HELM_VERSION: v3.8.0
+
+permissions:
+  packages: write
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
+
+jobs:
+  publish-helm-chart-actions-runner-controller-crds:
+    if: ${{ inputs.publish_actions_runner_controller_crds_chart == true }}
+    name: Publish Helm chart for actions-runner-controller-crds
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          # If inputs.ref is empty, it'll resolve to the default branch
+          ref: ${{ inputs.ref }}
+
+      - name: Resolve parameters
+        id: resolve_parameters
+        run: |
+          resolvedRef="${{ inputs.ref }}"
+          if [ -z "$resolvedRef" ]
+          then
+            resolvedRef="${{ github.ref }}"
+          fi
+          echo "INFO: Resolving short SHA for $resolvedRef"
+          echo "short_sha=$(git rev-parse --short $resolvedRef)" >> $GITHUB_OUTPUT
+          echo "INFO: Normalizing repository name (lowercase)"
+          echo "repository_owner=$(echo ${{ github.repository_owner }} | tr '[:upper:]' '[:lower:]')" >> $GITHUB_OUTPUT
+
+      - name: Set up Helm
+        # Using https://github.com/Azure/setup-helm/releases/tag/v3.5
+        uses: azure/setup-helm@5119fcb9089d432beecbf79bb2c7915207344b78
+        with:
+          version: ${{ env.HELM_VERSION }}
+
+      - name: Publish new helm chart for gha-runner-scale-set-controller
+        run: |
+          echo ${{ secrets.GITHUB_TOKEN }} | helm registry login ghcr.io --username ${{ github.actor }} --password-stdin
+          ACTIONS_RUNNER_CONTROLLER_CRDS_CHART_VERSION_TAG=$(cat charts/actions-runner-controller-crds/Chart.yaml | grep version: | cut -d " " -f 2)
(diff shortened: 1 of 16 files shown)


minhthong582000:
Hi @mumoshu, I've some questions regarding the github actions pipeline to push the helm chart.

Currently, I've created the pipeline base on this [workflow (publish-arc)](https://github.com/actions/actions-runner-controller/blob/master/.github/workflows/publish-arc.yaml) since the [publish-chart workflow](https://github.com/actions/actions-runner-controller/blob/master/.github/workflows/publish-chart.yaml) has some hard-coded variables which make the workflow only work with changes to `github-actions-controller` helm chart (please correct me if I'm wrong).

Is the [publish-arc workflow](https://github.com/actions/actions-runner-controller/blob/master/.github/workflows/publish-arc.yaml) just a temporary solution for testing purpose and I shouldn't rely on this workflow ?

If yes, do you want me to add some additional logic to make [publish-chart workflow](https://github.com/actions/actions-runner-controller/blob/master/.github/workflows/publish-chart.yaml) able to push any charts under `charts` folder ?

And It would be great if you have any better idea on how we could handle the release pipeline.

Thank you.

mumoshu:
Hey @minhthong582000, thanks a lot for your contribution and sorry for the radio silence!
I've been discussing this with the GitHub Actions team and have not yet been settled on how we handle this.
I love having this because I agree having flexibility in how CRDs are tweaked is a must-have in a certain environment.

> Is the [publish-arc workflow](https://github.com/actions/actions-runner-controller/blob/master/.github/workflows/publish-arc.yaml) just a temporary solution for testing purpose and I shouldn't rely on this workflow ?

I believe it's a temporary solution until we fully migrate to the new autoscaling-runner-set based solution- I presume this temporary solution might be kept alive for months as we'd need to wait until everyone migrates to the new solution before removing the legacy chart.

> If yes, do you want me to add some additional logic to make [publish-chart workflow](https://github.com/actions/actions-runner-controller/blob/master/.github/workflows/publish-chart.yaml) able to push any charts under charts folder ?

Thanks for your suggestion! Could you hold on a bit more? I got to think this can be merged into the main chart so that you can run `helm upgrade --install --skip-crds --set crds.render=true --set crds.keep=true --set crds.annotations=...`.

This way, we won't need another workflow or modify the workflow to get the chart working on ArgoCD...
WDYT?

minhthong582000:
> Thanks for your suggestion! Could you hold on a bit more? I got to think this can be merged into the main chart so that you can run `helm upgrade --install --skip-crds --set crds.render=true --set crds.keep=true --set crds.annotations=...`.
> 
> This way, we won't need another workflow or modify the workflow to get the chart working on ArgoCD... WDYT?

So you mean we'll have 2 `crds` folder in the main chart, one with helm template and the other is not.

```
├── chart
│   ├── actions-runner-controller
│   │   ├── crds
│   │   │   ├── templates
│   │   │   |   ├── crds
```

I think this way can solve the problem with the crds annotations but may create some confusion for who look at the chart for the first time. For a temporary solution, yes, but I'm still happy to wait for the complete actions workflow.

I will leave the final decision to you and the Github Actions team.

Thanks a lot for your response @mumoshu !

Shrooblord:
> Thanks for your suggestion! Could you hold on a bit more? I got to think this can be merged into the main chart so that you can run `helm upgrade --install --skip-crds --set crds.render=true --set crds.keep=true --set crds.annotations=...`.

Hello @mumoshu! What's the status on this PR / feature? In my production environment I'm trying to migrate from a Helm installation to a Kustomize-managed one, i.e.
1. import the Helm chart 'locally' through a `helm template > {chart-name}.yaml` command
2. use that as the base for a Kustomize setup
3. modify some resources through Kustomize as I see fit
4. deploy through `kustomize build (...) | kubectl apply -f -`

in the same way that ArgoCD handles Helm charts internally.

However, since my current deployment is managed by Helm, I need to `helm uninstall` the currently installed charts at some point. My preference would be to add the `helm.sh/resource-policy: keep` annotation to all the chart resources first e.g. by applying the following Kustomize patch to all the chart's resources in the cluster:
```
- op: add
  path: /metadata/annotations
  value:
    helm.sh/resource-policy: keep
```

However, [as mentioned in the main Helm repo](https://github.com/helm/helm/issues/10890), adding this annotation manually and/or through external tools like Kustomize means the annotation will get ignored. We figured out this is because, to determine the state of the cluster, Helm actually only looks at the Secrets that it creates that are called `sh.helm.release.v{#}.{chartname-randomstring}.v{#}`, which turn out to be the entire Helm chart fully rendered out and gzip'd and then converted to base64... kinda cool process... but also gets in the way for operations like the one I described above!

----

I realise we could 'just' uninstall the Helm chart and then reinstall it through the Kustomize route with minimal downtime, but optimal would be zero downtime. Besides, if I am not fully aware of which Secrets for example are being managed by the Helm chart and thus do not backup the correct resources before uninstalling, then my workloads will lose access to their DBs!! etc. etc.

Would love to know what my options are for this project moving forward :)

----

Chart owners appear to be adding this functionality manually:
- https://github.com/hashicorp/vault-secrets-operator/pull/169/files
- https://github.com/dgraph-io/charts/pull/71/files
- https://github.com/bitnami/charts/blob/063463f42c06881ec457b8951cadd4c68e724cd7/bitnami/redis/values.yaml#L54C1-L54C18

mdgreenwald:
@mumoshu, @toast-gear, @nikola-jokic Breaking the CRDs into their own helm chart would be extremely helpful. For example, this is exactly how the Karpenter project handles their own CRDs with regards to helm. [[Link]](https://karpenter.sh/preview/upgrading/upgrade-guide/#crd-upgrades)

Is there any update on this effort? If someone were to open a new updated PR, would it be reviewed and adopted?

mdgreenwald:
@mumoshu, @toast-gear, @nikola-jokic

I created an idea here as a feature request: [https://github.com/actions/actions-runner-controller/discussions/3887](https://github.com/actions/actions-runner-controller/discussions/3887)

PR #2302, opened 2023-02-20
fix: case-insensitive webhook label matching

tufitko:
reopen #1500

Github definitely uses case-insensitive label matching. I didn't find any info in docs, but I tested it.

I had two workflows, the one with `[self-hosted,linux]`, and the second one with `[self-hosted,Linux]` labels.
I also had configured ARC with webhook scaling and `[self-hosted,Linux]` labels. 
The first workflow didn't scale up runners, but GitHub scheduled it to runner (after second workflow).


mumoshu (dismissed):
Awesome work! Thank you for your detailed testing and contribution @tufitko!

mumoshu:
Addressed the golangci-lint findings.

mumoshu (approved):

PR #2303, opened 2023-02-20
Use DOCKER_IMAGE_NAME instead of NAME to avoid conflict.

TingluoHuang:
Why build ARC under WSL, the env `NAME` is set to the machine's name and causes the `Makefile` can't set the docker image name properly.

I am updating the Makefile to use `DOCKER_IMAGE_NAME` instead of `NAME` to avoid potential collision of the env.

The change is pulled from https://github.com/actions/actions-runner-controller/pull/2279 to make the review more focused on both PRs.


mumoshu (approved):
Good to know that it doesn't work well on WSL! LGTM :+1:

PR #2307, opened 2023-02-22
Disable metrics serving in proxy tests

rentziass:
Plus a couple of other fixes to two flaky tests.

### DIFF
--- a/controllers/actions.github.com/ephemeralrunner_controller_test.go
+++ b/controllers/actions.github.com/ephemeralrunner_controller_test.go
@@ -3,6 +3,7 @@ package actionsgithubcom
 import (
 	"context"
 	"encoding/base64"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"strings"
@@ -476,7 +477,7 @@ var _ = Describe("EphemeralRunner", func() {
 					})
 					err = k8sClient.Status().Update(ctx, pod)
 					Expect(err).To(BeNil(), "Failed to update pod status")
-					return false, nil
+					return false, fmt.Errorf("pod haven't failed for 5 times.")
 				},
 				timeout,
 				interval,
--- a/controllers/actions.github.com/ephemeralrunner_controller_test.go
+++ b/controllers/actions.github.com/ephemeralrunner_controller_test.go
@@ -3,7 +3,6 @@ package actionsgithubcom
 import (
 	"context"
 	"encoding/base64"
-	"fmt"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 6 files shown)


controllers/actions.github.com/ephemeralrunner_controller_test.go
```diff
@@ -477,7 +476,7 @@ var _ = Describe("EphemeralRunner", func() {
 					})
 					err = k8sClient.Status().Update(ctx, pod)
 					Expect(err).To(BeNil(), "Failed to update pod status")
-					return false, fmt.Errorf("pod haven't failed for 5 times.")
+					return false, nil
```
TingluoHuang: why do we need to change this?

rentziass: To keep the `Eventually` loop going, otherwise returning that error makes it stop immediately. From `Eventually` docs:

>If multiple values are returned by the function, Eventually will pass the first value to the matcher and require that all others are zero-valued.  This allows you to pass Eventually a function that returns a value and an error - a common pattern in Go.

TingluoHuang: I didn't find anything indicating that `return error` will make `Eventually` stop immediately. 😕

rentziass: It's this bit `require that all others are zero-valued` meaning that if we have a function like this returning `(bool, error)` any non-zero value returned for error (zero value for error is nil) will stop the loop.

The [full doc](https://pkg.go.dev/github.com/onsi/gomega#Eventually) piece of interest:
> If multiple values are returned by the function, Eventually will pass the first value to the matcher and require that all others are zero-valued.  This allows you to pass Eventually a function that returns a value and an error - a common pattern in Go.

>For example, consider a method that returns a value and an error:
>
>	func FetchFromDB() (string, error)
>
>Then
>
>	Eventually(FetchFromDB).Should(Equal("got it"))
>
>will pass only if and when the returned error is nil *and* the returned string satisfies the matcher.

rentziass: For posterity, this passes:

```golang
func Test(t *testing.T) {
	g := gomega.NewWithT(t)
	i := 0
	g.Eventually(func() (int, error) {
		if i == 0 {
			i++
			return 0, errors.New("some error")
		}
		return i, nil
	}).Should(gomega.Equal(1))
}
```

🤦 , reverted.

TingluoHuang (approved):

PR #2322, opened 2023-02-25
Upgrade Docker to latest version

milas:
* Upgrade to Docker Engine v23.0.1
* Upgrade to Docker Compose v2.16.0

* Adjust download location of Docker Compose to use CLI plugins directory and symlink to `/usr/bin` for compatibility
   * This enables usage as `docker compose`, so you can specify both Docker flags (e.g. `--context`) as well as Compose flags
* Rename `CHANNEL` to `DOCKER_CHANNEL` for clarity
* Remove extraneous comment & arg

### DIFF
--- a/.github/workflows/release-runners.yaml
+++ b/.github/workflows/release-runners.yaml
@@ -17,7 +17,7 @@ env:
   PUSH_TO_REGISTRIES: true
   TARGET_ORG: actions-runner-controller
   TARGET_WORKFLOW: release-runners.yaml
-  DOCKER_VERSION: 20.10.21
+  DOCKER_VERSION: 23.0.1
   RUNNER_CONTAINER_HOOKS_VERSION: 0.2.0
 
 jobs:
--- a/runner/Makefile
+++ b/runner/Makefile
@@ -8,7 +8,7 @@ TARGETPLATFORM ?= $(shell arch)
 
 RUNNER_VERSION ?= 2.301.1 
 RUNNER_CONTAINER_HOOKS_VERSION ?= 0.2.0
-DOCKER_VERSION ?= 20.10.21
+DOCKER_VERSION ?= 23.0.1
 
 # default list of platforms for which multiarch image is built
 ifeq (${PLATFORMS}, )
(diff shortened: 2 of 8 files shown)


mumoshu:
@milas Hey! Thanks for your contribution ☺️ 
This looks almost good. However I'd appreciate it if you could separate pull requests so that each pull request does one thing, which eases reviewing and merging.

Here are my comments:

- I think we might need to keep the docker engine version, as we're trying to match the version numbers to those used in the actions VM images https://github.com/actions/runner-images/blob/bcb2886e5b0fb66c491f66640461128755c06885/images/linux/Ubuntu2204-Readme.md?plain=1#L76-L77
- docker-compose version bump seems good as the actions runner VM seems to be using the same version
- How did you test these changes? You touched six variants of runners images, and that means we might want someone to test these six variants by deploying runners and letting some jobs run on those runners to make sure we don't break anything. Have you tested it? Or do you want me to test it out?

milas:
> I think we might need to keep the docker engine version, as we're trying to match the version numbers to those used in the actions VM images

I will switch this to the latest 20.10.x. FYI that currently the sidecar variants default to `docker:dind` tag, which points to 23.x as of a few weeks ago. It might be better to change this to `20.10-dind` for now if you want to keep it aligned to the upstream runner images

Also the rootless images are using the helper script from `get.docker.com`, so will install v23.x: https://github.com/actions/actions-runner-controller/blob/ec3afef00d918f88f26dd57fa5818610ccb519c0/runner/actions-runner-dind-rootless.ubuntu-22.04.dockerfile#L111-L114

> How did you test these changes?

The Docker Compose path change for the plugin is self-testing - the command (in the pattern of others in the file) does `&& which docker-compose && docker compose version` to ensure that the symlink is in the path and that the plugin can be executed.

My usage has mostly been on the default Ubuntu 22.04 image (using sidecar), but I've built out the other images and done basic verification (e.g. running them locally with `--entrypoint=/bin/bash` and verifying that Docker components are working / the expected versions)

milas:
Closing this PR, see:
* #2326 
* #2327 
* #2328

PR #2324, opened 2023-02-25
Use Docker socket via shared emptyDir instead of TCP/mTLS

### LINKED ISSUE #893: Cannot Creating a new builder instance in [Set up Docker Buildx]
**Describe the bug**
* Action which works correctly on hosted github runners does not work in self-hosted version
* 
**Checks**

- [x] My actions-runner-controller version (v0.x.y) does support the feature
- [ ] I'm using an unreleased version of the controller I built from HEAD of the default branch

**To Reproduce**

```yaml
      - name: Checkout
        uses: actions/checkout@v2
      - name: Set up Docker Buildx
        uses: docker/setup-buildx-action@v1
```
```
  /usr/local/bin/docker buildx create --name builder-3367d142-667f-46da-9e5a-56a8706f3c86 --driver docker-container --buildkitd-flags --allow-insecure-entitlement security.insecure --allow-insecure-entitlement network.host --use
  error: could not create a builder instance with TLS data loaded from environment. Please use `docker context create <context-name>` to create a context for current environment and then create a builder instance with `docker buildx create <context-name>`
  Error: The process '/usr/local/bin/docker' failed with exit code 1
```

**Expected behavior**
It will work same as in hosted github runners

**Environment (please complete the following information):**
 - Controller Version [e.g. 0.18.2] app.kubernetes.io/version=0.20.2
 - Deployment Method [e.g. Helm and Kustomize ]: helm 
 - Helm Chart Version [e.g. 0.11.0, if applicable]: helm.sh/chart=actions-runner-controller-0.13.2

Helm values yaml:

```
# helm upgrade --install --namespace actions-runner-system --create-namespace actions-runner-controller actions-runner-controller/actions-runner-controller -f ~/Desktop/grid/infra/staging/gh.yaml
authSecret:
  create: true
  <redacted>

scope:
  singleNamespace: true

githubWebhookServer:
  enabled: true
  secret:
    create: true
    name: "github-webhook-server"
    github_webhook_secret_token: "<redacted>"

metrics:
  serviceMonitor: true
```

milas:
By default, the `docker:dind` entrypoint will auto-generate mTLS certs and run with TCP on `0.0.0.0`. This is handy for accessing the running Docker Engine remotely by then publishing the ports. For the runner, we don't need (or want) that behavior, so a Unix socket lets us rely on filesystem permissions.

This also has the benefit of eliminating the need for mTLS, which will speed up Pod start slightly (no need to generate CA & client certs), and will fix #893 and generally improve compatibility with apps that interact with the Docker API without requiring a custom Docker context to be initialized.

### DIFF
--- a/controllers/actions.summerwind.net/runner_controller.go
+++ b/controllers/actions.summerwind.net/runner_controller.go
@@ -30,6 +30,7 @@ import (
 	"github.com/go-logr/logr"
 
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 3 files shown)


milas:
I'm successfully running jobs using this without needing explicit Docker context setup.

Sample: https://github.com/milas/rock5-toolchain/actions/runs/4270921744/jobs/7435031799

controllers/actions.summerwind.net/runner_controller.go
```diff
@@ -1001,6 +1002,35 @@ func newRunnerPodWithContainerMode(containerMode string, template corev1.Pod, ru
...
+
+		// this must match a GID for the user in the runner image
+		// default matches GitHub Actions infra (and default runner images
+		// for actions-runner-controller) so typically should not need to be
+		// overridden
+		if ok, _ := envVarPresent("DOCKER_GROUP_GID", dockerdContainer.Env); !ok {
+			dockerdContainer.Env = append(dockerdContainer.Env,
+				corev1.EnvVar{
+					Name: "DOCKER_GROUP_GID",
+					Value: "121",
```
mumoshu, line 1019: If we go this way, we might need to make this configurable because folks can use another gid for the docker user if they use a custom runner image(and we don't forbid the use of a custom runner image...

mumoshu: I just realized you already addressed that by checking for the existence of DOCKER_GROUP_GID and only use the hard-coded 121 when the envvar does not exist!
Good work 👍

milas: Yeah, this is unfortunately still not the most elegant, but I couldn't figure out a cleaner way to handle this, as we do need to pass a GID so that the socket is usable.

mumoshu:
@milas Hey! Sorry for the radio silence. Testing this locally and fixing failing unit tests took a bit more time than I expected.

I think this is now ready to be merged, and I'll merge this in a day or two.
After that, our automated runner image release process would get triggered, and the `latest` tag will contain this change.

mumoshu (approved):

mumoshu:
Merged. Thank you so much for your contribution, @milas!

PR #2326, opened 2023-02-26
Install Docker Compose v2 as a Docker CLI plugin

milas:
Docker Compose v2.x is designed to run primarily as a Docker CLI plugin. This has several advantages, such as allowing consistent handling of connection-related flags, e.g. it's possible to run a command like `docker --context=foo compose ...` this way.

Adjust the install logic so that it's installed into the expected CLI plugin path (`/usr/libexec/docker/cli-plugins` for standard installs and `~/.docker/cli-plugins` for rootless).

A symlink from `docker-compose` is also included to avoid breaking compatibility. Compose v2 also supports being invoked this way, and this is how it was previously being used in this image.


mumoshu (approved):
I finally had time to test this, and it worked flawlessly! Thank you so much for your awesome work, @milas!

PR #2328, opened 2023-02-26
Upgrade to Docker Engine v20.10.23

milas:
The latest stable Docker Engine is v23.x, but upstream GitHub Actions images are still on 20.10.x, so keeping consistency here.


milas:
Copying comment from #2322 over here about the versioning in general:

> FYI that currently the sidecar variants default to `docker:dind` tag, which points to 23.x as of a few weeks ago. It might be better to change this to `20.10-dind` for now if you want to keep it aligned to the upstream runner images
>
> Also the rootless images are using the helper script from `get.docker.com`, so will install v23.x: https://github.com/actions/actions-runner-controller/blob/ec3afef00d918f88f26dd57fa5818610ccb519c0/runner/actions-runner-dind-rootless.ubuntu-22.04.dockerfile#L111-L114

mumoshu (dismissed):
LGTM. Thank you so much for your contribution @milas ☺️

mumoshu (approved):

PR #2333, opened 2023-02-27
Rename actions-runner-controller-2 and auto-scaling-runner-set helm charts

Link-:
### Context

> There are only two hard things in Computer Science: cache invalidation and naming things.
> -- Phil Karlton

We initially published the new auto-scaling mode's helm chart under the name `actions-runner-controller-2` for the lack of a better name. Now, we have a better name 😄

As such, we're renaming & republishing the helm charts as follows:
- `actions-runner-controller-2` -> `gha-runner-scale-set-controller`
- `auto-scaling-runner-set` -> `gha-runner-scale-set`

Since the new scaling mode will rely entirely on the newly introduced concept of runner scale sets, it made sense to use the same concept in the name of the controller and the starter helm charts.

### Reviewer attention

The renaming covers:
- Helm charts
- Makefile
- Publishing workflows
- Docs


.github/workflows/publish-arc2.yaml
```diff
@@ -19,12 +19,12 @@ on:
         type: boolean
         default: false
       publish_actions_runner_controller_2_chart:
-        description: 'Publish new helm chart for actions-runner-controller-2'
```
TingluoHuang, line 22: do you want to also rename the workflow file? `publish-arc2.yaml`

Link-: Yep good point, let me fix it

Link-: This should be fixed now

.github/workflows/publish-runner-scale-set.yaml
```diff
@@ -87,8 +87,8 @@ jobs:
           build-args: VERSION=${{ inputs.release_tag_name }}
           push: ${{ inputs.push_to_registries }}
           tags: |
-            ghcr.io/${{ steps.resolve_parameters.outputs.repository_owner }}/actions-runner-controller-2:${{ inputs.release_tag_name }}
-            ghcr.io/${{ steps.resolve_parameters.outputs.repository_owner }}/actions-runner-controller-2:${{ inputs.release_tag_name }}-${{ steps.resolve_parameters.outputs.short_sha }}
+            ghcr.io/${{ steps.resolve_parameters.outputs.repository_owner }}/gha-runner-scale-set-controller:${{ inputs.release_tag_name }}
+            ghcr.io/${{ steps.resolve_parameters.outputs.repository_owner }}/gha-runner-scale-set-controller:${{ inputs.release_tag_name }}-${{ steps.resolve_parameters.outputs.short_sha }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
```
TingluoHuang, line 94: you might want to update line 97, it still mentions `blob/main/.github/workflows/publish-arc2.yaml`

TingluoHuang (dismissed):

rentziass (approved):
LGTM! 🚀

PR #2337, opened 2023-02-28
Refactor main.go and introduce make run-scaleset to be able to run manager locally

nikola-jokic:
This change should not affect how the application operates right now.

The intention of this change is to make it easier to run the controller locally, without spinning up the cluster, so it is easier and faster to debug.

### DIFF
--- a/charts/gha-runner-scale-set-controller/templates/deployment.yaml
+++ b/charts/gha-runner-scale-set-controller/templates/deployment.yaml
@@ -54,10 +54,8 @@ spec:
         command:
         - "/manager"
         env:
-        - name: CONTROLLER_MANAGER_POD_NAME
-          valueFrom:
-            fieldRef:
-              fieldPath: metadata.name
+        - name: CONTROLLER_MANAGER_CONTAINER_IMAGE
+          value: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
         - name: CONTROLLER_MANAGER_POD_NAMESPACE
           valueFrom:
             fieldRef:
--- a/main.go
+++ b/main.go
@@ -17,7 +17,6 @@ limitations under the License.
 package main
 
 import (
-	"context"
 	"flag"
 	"fmt"
 	"os"
@@ -33,9 +32,7 @@ import (
 	"github.com/actions/actions-runner-controller/github/actions"
 	"github.com/actions/actions-runner-controller/logging"
 	"github.com/kelseyhightower/envconfig"
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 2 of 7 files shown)


main.go
```diff
@@ -315,6 +309,18 @@ func main() {
 				os.Exit(1)
 			}
 		}
+
+		if !disableAdmissionWebhook {
```
TingluoHuang: do you want to merge this with line 298?

main.go
```diff
@@ -327,6 +333,10 @@ func main() {
 	// However, doing so would affect either private ARC testers or current ARC users
 	// who run ARC without those variabls.
 	mgrPodName := os.Getenv("CONTROLLER_MANAGER_POD_NAME")
+	mgrPodNamespace := os.Getenv("CONTROLLER_MANAGER_POD_NAMESPACE")
+
+	listenerImage := os.Getenv("SCALE_SET_LISTENER_IMAGE") // default
```
TingluoHuang: maybe we just go with `mgrPodImage := os.Getenv("CONTROLLER_MANAGER_POD_NAMESPACE_IMAGE")`?

main.go
nikola-jokic: Yep, my bad... Thanks! :+1:

main.go
nikola-jokic: Should we set that env in helm then when we apply the manager? At that point, we know what the image would be right?

TingluoHuang: we should. 
A more general question is what would be the best way to pass the listener image and image pull secret into the controller? 
the current approach was added as a short-term solution. 😄

charts/gha-runner-scale-set-controller/templates/deployment.yaml
```diff
@@ -54,10 +54,8 @@ spec:
         command:
         - "/manager"
         env:
-        - name: CONTROLLER_MANAGER_POD_NAME
-          valueFrom:
-            fieldRef:
-              fieldPath: metadata.name
+        - name: CONTROLLER_MANAGER_CONTAINER_IMAGE
+          value: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
         - name: CONTROLLER_MANAGER_POD_NAMESPACE
```
TingluoHuang, line 59: do we want to change the `CONTROLLER_MANAGER_POD_NAMESPACE` to also read from `{{ .Release.Namespace }}`?

main.go
```diff
@@ -314,95 +305,73 @@ func main() {
...
-		}
-
-		var mgrContainer *corev1.Container
-		for _, container := range mgrPod.Spec.Containers {
-			if container.Name == "manager" {
-				mgrContainer = &container
-				break
-			}
-		}
+	managerImage := os.Getenv("CONTROLLER_MANAGER_CONTAINER_IMAGE")
```
TingluoHuang: should we wrap all of these into `if autoScalingRunnerSetOnly {}`? otherwise, the old `actions-runner-controller` helm chart installation won't start the controller correctly.

nikola-jokic: Yes definitely, good call!

charts/gha-runner-scale-set-controller/templates/deployment.yaml
nikola-jokic: It feels safer to me to reference a field if we can

TingluoHuang (approved):

PR #2338, opened 2023-02-28
Change e2e config url

AvaStancu:
Since we are using the old actions-runner-controller org Github App for JWT generation, the config url for arc runner registration should be a repo in the old org.
The custom action was also built with the scope of being reusable, so the url will be passed via the workflow that calls it, to make this customizable too.

### DIFF
--- a/.github/actions/e2e-arc-test/action.yaml
+++ b/.github/actions/e2e-arc-test/action.yaml
@@ -5,6 +5,9 @@ inputs:
   github-token:
     description: 'JWT generated with Github App inputs'
     required: true
+  config-url:
+    description: "Url of repo or org where runners will be installed"
+    required: true
 
 runs:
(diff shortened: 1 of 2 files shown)


.github/actions/e2e-arc-test/action.yaml
```diff
@@ -5,6 +5,9 @@ inputs:
   github-token:
     description: 'JWT generated with Github App inputs'
     required: true
+  config-url:
+    description: "Url of repo or org where runners will be installed"
```
Link-:
```suggestion
    description: "URL of the repo, org or enterprise where the runner scale sets will be registered"
```

Link- (changes requested):
Small change to the description

Link- (approved):

PR #2341, opened 2023-03-01
Split listener pod label to avoid long names issue

nikola-jokic:
All other resources are having labels split on namespace and the name.

Split for the listener pod will allow for the longer autoscaling runner set `NamespacedName` combination (Label restriction is 63 characters, so right now, the combination of the namespace and the name must be < 63)

### DIFF
--- a/controllers/actions.github.com/resourcebuilder.go
+++ b/controllers/actions.github.com/resourcebuilder.go
@@ -14,8 +14,15 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// secret constants
 const (
 	jitTokenKey = "jitToken"
+)
+
+// labels applied to resources
+const (
+	LabelKeyAutoScaleRunnerSetName      = "auto-scale-runner-set-name"
+	LabelKeyAutoScaleRunnerSetNamespace = "auto-scale-runner-set-namespace"
 )
 
 type resourceBuilder struct{}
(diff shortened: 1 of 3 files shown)


controllers/actions.github.com/resourcebuilder.go
```diff
@@ -133,7 +130,10 @@ func (b *resourceBuilder) newScaleSetListenerPod(autoscalingListener *v1alpha1.A
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      autoscalingListener.Name,
 			Namespace: autoscalingListener.Namespace,
-			Labels:    newLabels,
+			Labels: map[string]string{
+				"auto-scaling-runner-set-namespace": autoscalingListener.Spec.AutoscalingRunnerSetNamespace,
```
TingluoHuang: Do we want to include listener in the label name?

nikola-jokic: Not sure, because the listener name is the name of the pod, you can easily search for that pod without passing the `-l`

controllers/actions.github.com/resourcebuilder.go
```diff
@@ -133,7 +130,10 @@ func (b *resourceBuilder) newScaleSetListenerPod(autoscalingListener *v1alpha1.A
...
+				"auto-scaling-runner-set-name":      autoscalingListener.Spec.AutoscalingRunnerSetName,
```
TingluoHuang: we have a key for `auto-scale-runner-set-name`, (`LabelKeyAutoScaleRunnerSetName`). we might need to make sure all labels match each other.

nikola-jokic: I'll extract it
I saw in other parts that we are using string literal, so just made it like that, but great suggestion!

TingluoHuang (approved):

PR #2342, opened 2023-03-01
Update trigger events for validate-chart

Link-:
### Context

Currently this workflow only runs when we push changes of the helm charts to the master branch. I think it's valuable to run these tests on the PR as well.

I've added trigger event to run this workflow on PRs.


nikola-jokic (dismissed):
LGTM

nikola-jokic (approved):

PR #2346, opened 2023-03-01
Upgrading & pinning action versions

Link-:
Upgrading setup-helm action to remove the set-output deprecation notice & pinning the version for added security.


nikola-jokic (approved):
LGTM

PR #2366, opened 2023-03-06
Apply proxy settings from environment in listener

rentziass:
We are correctly setting up environment variables on the listener pod to configure proxy behaviour but we're just ignoring them at the moment. These reinstates those in client configuration.

### DIFF
--- a/cmd/githubrunnerscalesetlistener/main.go
+++ b/cmd/githubrunnerscalesetlistener/main.go
@@ -19,6 +19,8 @@ package main
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"net/url"
 	"os"
 	"os/signal"
 	"syscall"
@@ -28,6 +30,7 @@ import (
 	"github.com/actions/actions-runner-controller/logging"
 	"github.com/go-logr/logr"
 	"github.com/kelseyhightower/envconfig"
+	"golang.org/x/net/http/httpproxy"
 )
 
 type RunnerScaleSetListenerConfig struct {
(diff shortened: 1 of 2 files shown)


cmd/githubrunnerscalesetlistener/main.go
```diff
@@ -155,3 +156,12 @@ func validateConfig(config *RunnerScaleSetListenerConfig) error {
 
 	return nil
 }
+
+func newActionsClientFromConfig(config RunnerScaleSetListenerConfig, creds *actions.ActionsAuth, options ...actions.ClientOption) (*actions.Client, error) {
```
rentziass, line 162: This isn't of much use as of yet, but makes the creation of the client testable, and it's reused in https://github.com/actions/actions-runner-controller/pull/2268

cmd/githubrunnerscalesetlistener/main.go
```diff
@@ -84,9 +87,7 @@ func run(rc RunnerScaleSetListenerConfig, logger logr.Logger) error {
 		}
 	}
 
-	actionsServiceClient, err := actions.NewClient(
-		rc.ConfigureUrl,
-		creds,
+	actionsServiceClient, err := newActionsClientFromConfig(rc, creds,
 		actions.WithUserAgent(fmt.Sprintf("actions-runner-controller/%s", build.Version)),
 		actions.WithLogger(logger),
 	)
```
TingluoHuang:
```suggestion
	actionsServiceClient, err := newActionsClientFromConfig(
	        rc,
	        creds,
		actions.WithUserAgent(fmt.Sprintf("actions-runner-controller/%s", build.Version)),
		actions.WithLogger(logger),
	)
```

TingluoHuang (dismissed):

TingluoHuang (approved):

PR #2374, opened 2023-03-07
Support the controller to watching a single namespace.

TingluoHuang:
Part 3 of https://github.com/actions/actions-runner-controller/pull/2275

- Introduce `watchSingleName` to `gha-runner-scale-set-controller` to allow the controller only watch resources in a single namespace.

### DIFF
--- a/charts/gha-runner-scale-set/templates/_helpers.tpl
+++ b/charts/gha-runner-scale-set/templates/_helpers.tpl
@@ -476,25 +476,46 @@ volumeMounts:
   {{- end }}
 {{- end }}
 {{- if eq $searchControllerDeployment 1 }}
-  {{- $counter := 0 }}
+  {{- $multiNamespacesCounter := 0 }}
+  {{- $singleNamespaceCounter := 0 }}
   {{- $controllerDeployment := dict }}
+  {{- $singleNamespaceControllerDeployments := dict }}
   {{- $managerServiceAccountName := "" }}
   {{- range $index, $deployment := (lookup "apps/v1" "Deployment" "" "").items }}
-    {{- range $key, $val := $deployment.metadata.labels }}
-      {{- if and (eq $key "app.kubernetes.io/part-of") (eq $val "gha-runner-scale-set-controller") }}
-        {{- $counter = add $counter 1 }}
-        {{- $controllerDeployment = $deployment }}
+    {{- if kindIs "map" $deployment.metadata.labels }}
+      {{- if eq (get $deployment.metadata.labels "app.kubernetes.io/part-of") "gha-runner-scale-set-controller" }}
+        {{- if hasKey $deployment.metadata.labels "actions.github.com/controller-watch-single-namespace" }}
+          {{- $singleNamespaceCounter = add $singleNamespaceCounter 1 }}
+          {{- $_ := set $singleNamespaceControllerDeployments (get $deployment.metadata.labels "actions.github.com/controller-watch-single-namespace") $deployment}}
+        {{- else }}
+          {{- $multiNamespacesCounter = add $multiNamespacesCounter 1 }}
+          {{- $controllerDeployment = $deployment }}
+        {{- end }}
       {{- end }}
     {{- end }}
   {{- end }}
-  {{- if lt $counter 1 }}
-    {{- fail "No gha-runner-scale-set-controller deployment found using label (app.kubernetes.io/part-of=gha-runner-scale-set-controller), consider setting controllerServiceAccount.name in values.yaml to be explicit if you think the discovery is wrong." }}
+  {{- if and (eq $multiNamespacesCounter 0) (eq $singleNamespaceCounter 0) }}
+    {{- fail "No gha-runner-scale-set-controller deployment found using label (app.kubernetes.io/part-of=gha-runner-scale-set-controller). Consider setting controllerServiceAccount.name in values.yaml to be explicit if you think the discovery is wrong." }}
   {{- end }}
-  {{- if gt $counter 1 }}
-    {{- fail "More than one gha-runner-scale-set-controller deployment found using label (app.kubernetes.io/part-of=gha-runner-scale-set-controller), consider setting controllerServiceAccount.name in values.yaml to be explicit if you think the discovery is wrong." }}
+  {{- if and (gt $multiNamespacesCounter 0) (gt $singleNamespaceCounter 0) }}
+    {{- fail "Found both gha-runner-scale-set-controller installed with flags.watchSingleNamespace set and unset in cluster, this is not supported. Consider setting controllerServiceAccount.name in values.yaml to be explicit if you think the discovery is wrong." }}
   {{- end }}
-  {{- with $controllerDeployment.metadata }}
-    {{- $managerServiceAccountName = (get $controllerDeployment.metadata.labels "actions.github.com/controller-service-account-name") }}
+  {{- if gt $multiNamespacesCounter 1 }}
+    {{- fail "More than one gha-runner-scale-set-controller deployment found using label (app.kubernetes.io/part-of=gha-runner-scale-set-controller). Consider setting controllerServiceAccount.name in values.yaml to be explicit if you think the discovery is wrong." }}
+  {{- end }}
+  {{- if eq $multiNamespacesCounter 1 }}
+    {{- with $controllerDeployment.metadata }}
+      {{- $managerServiceAccountName = (get $controllerDeployment.metadata.labels "actions.github.com/controller-service-account-name") }}
+    {{- end }}
+  {{- else if gt $singleNamespaceCounter 0 }}
+    {{- if hasKey $singleNamespaceControllerDeployments .Release.Namespace }}
+      {{- $controllerDeployment = get $singleNamespaceControllerDeployments .Release.Namespace }}
+      {{- with $controllerDeployment.metadata }}
+        {{- $managerServiceAccountName = (get $controllerDeployment.metadata.labels "actions.github.com/controller-service-account-name") }}
+      {{- end }}
+    {{- else }}
+      {{- fail "No gha-runner-scale-set-controller deployment that watch this namespace found using label (actions.github.com/controller-watch-single-namespace). Consider setting controllerServiceAccount.name in values.yaml to be explicit if you think the discovery is wrong." }}
+    {{- end }}
   {{- end }}
   {{- if eq $managerServiceAccountName "" }}
     {{- fail "No service account name found for gha-runner-scale-set-controller deployment using label (actions.github.com/controller-service-account-name), consider setting controllerServiceAccount.name in values.yaml to be explicit if you think the discovery is wrong." }}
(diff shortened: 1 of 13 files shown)


main.go
```diff
@@ -149,13 +152,27 @@ func main() {
...
 	ctrl.SetLogger(log)
 
+	managerNamespace := ""
+	var newCache cache.NewCacheFunc
+
 	if autoScalingRunnerSetOnly {
 		// We don't support metrics for AutoRunnerScaleSet for now
 		metricsAddr = "0"
+
+		managerNamespace = os.Getenv("CONTROLLER_MANAGER_POD_NAMESPACE")
```
nikola-jokic, line 162: I don't know if this would be a good idea, but I'll just ask for discussion:
Should we default to the `default` namespace if no namespace is provided, like k8s does? It seems more dangerous in case someone makes a mistake, but it also aligns well with the way k8s treats the empty value

main.go
```diff
@@ -126,6 +128,7 @@ func main() {
 	flag.DurationVar(&syncPeriod, "sync-period", 1*time.Minute, "Determines the minimum frequency at which K8s resources managed by this controller are reconciled.")
 	flag.Var(&commonRunnerLabels, "common-runner-labels", "Runner labels in the K1=V1,K2=V2,... format that are inherited all the runners created by the controller. See https://github.com/actions/actions-runner-controller/issues/321 for more information")
 	flag.StringVar(&namespace, "watch-namespace", "", "The namespace to watch for custom resources. Set to empty for letting it watch for all namespaces.")
+	flag.StringVar(&watchSingleNamespace, "watch-single-namespace", "", "Restrict to watch for custom resources in a single namespace.")
```
nikola-jokic, line 131: Isn't this fixed with `watch-namespace` flag?

charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -476,25 +476,46 @@ volumeMounts:
...
   {{- $controllerDeployment := dict }}
+  {{- $singleNamespaceControllerDeployments := dict }}
   {{- $managerServiceAccountName := "" }}
   {{- range $index, $deployment := (lookup "apps/v1" "Deployment" "" "").items }}
-    {{- range $key, $val := $deployment.metadata.labels }}
-      {{- if and (eq $key "app.kubernetes.io/part-of") (eq $val "gha-runner-scale-set-controller") }}
-        {{- $counter = add $counter 1 }}
-        {{- $controllerDeployment = $deployment }}
+    {{- if kindIs "map" $deployment.metadata.labels }}
+      {{- if eq (get $deployment.metadata.labels "app.kubernetes.io/part-of") "gha-runner-scale-set-controller" }}
```
nikola-jokic, line 486: Is this a new value that we should set for the part-of?
The [ADR](https://github.com/actions/actions-runner-controller/blob/master/adrs/2022-12-05-adding-labels-k8s-resources.md#controller-manager) says just "actions-runner-controller"
Should we update the ADR?

rentziass: Good shout, the ADR pre dates the rename.

main.go
rentziass: This env var gets its value from 

```
    - name: CONTROLLER_MANAGER_POD_NAMESPACE
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.namespace
```

so it depends on where arc is installed?

charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -476,25 +476,46 @@ volumeMounts:
...
+          {{- $multiNamespacesCounter = add $multiNamespacesCounter 1 }}
+          {{- $controllerDeployment = $deployment }}
+        {{- end }}
       {{- end }}
     {{- end }}
   {{- end }}
-  {{- if lt $counter 1 }}
-    {{- fail "No gha-runner-scale-set-controller deployment found using label (app.kubernetes.io/part-of=gha-runner-scale-set-controller), consider set controllerServiceAccount.name in values.yaml to be explicitly if you think the discovery is wrong." }}
+  {{- if and (eq $multiNamespacesCounter 0) (eq $singleNamespaceCounter 0) }}
+    {{- fail "No gha-runner-scale-set-controller deployment found using label (app.kubernetes.io/part-of=gha-runner-scale-set-controller). Consider set controllerServiceAccount.name in values.yaml to be explicitly if you think the discovery is wrong." }}
```
rentziass:
```suggestion
    {{- fail "No gha-runner-scale-set-controller deployment found using label (app.kubernetes.io/part-of=gha-runner-scale-set-controller). Consider setting controllerServiceAccount.name in values.yaml to be explicit if you think the discovery is wrong." }}
```

charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -476,25 +476,46 @@ volumeMounts:
...
+  {{- end }}
+  {{- if and (gt $multiNamespacesCounter 0) (gt $singleNamespaceCounter 0) }}
+    {{- fail "Found both gha-runner-scale-set-controller installed with flags.watchSingleNamespace set and unset in cluster, this is not supported. Consider set controllerServiceAccount.name in values.yaml to be explicitly if you think the discovery is wrong." }}
```
rentziass:
```suggestion
    {{- fail "Found both gha-runner-scale-set-controller installed with flags.watchSingleNamespace set and unset in cluster, this is not supported. Consider setting controllerServiceAccount.name in values.yaml to be explicit if you think the discovery is wrong." }}
```

main.go
nikola-jokic: Yes, but I am not sure if it is going to receive `""` if no namespace is set :relaxed: I'll play around to make sure, because I am not sure how it would look like if `metadata.namespace` is not set

charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -476,25 +476,46 @@ volumeMounts:
...
   {{- end }}
-  {{- if gt $counter 1 }}
-    {{- fail "More than one gha-runner-scale-set-controller deployment found using label (app.kubernetes.io/part-of=gha-runner-scale-set-controller), consider set controllerServiceAccount.name in values.yaml to be explicitly if you think the discovery is wrong." }}
+  {{- if gt $multiNamespacesCounter 1 }}
+    {{- fail "More than one gha-runner-scale-set-controller deployment found using label (app.kubernetes.io/part-of=gha-runner-scale-set-controller). Consider set controllerServiceAccount.name in values.yaml to be explicitly if you think the discovery is wrong." }}
```
rentziass:
```suggestion
    {{- fail "More than one gha-runner-scale-set-controller deployment found using label (app.kubernetes.io/part-of=gha-runner-scale-set-controller). Consider setting controllerServiceAccount.name in values.yaml to be explicit if you think the discovery is wrong." }}
```

charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -476,25 +476,46 @@ volumeMounts:
...
+      {{- $managerServiceAccountName = (get $controllerDeployment.metadata.labels "actions.github.com/controller-service-account-name") }}
+    {{- end }}
+  {{- else if gt $singleNamespaceCounter 0 }}
+    {{- if hasKey $singleNamespaceControllerDeployments .Release.Namespace }}
+      {{- $controllerDeployment = get $singleNamespaceControllerDeployments .Release.Namespace }}
+      {{- with $controllerDeployment.metadata }}
+        {{- $managerServiceAccountName = (get $controllerDeployment.metadata.labels "actions.github.com/controller-service-account-name") }}
+      {{- end }}
+    {{- else }}
+      {{- fail "No gha-runner-scale-set-controller deployment that watch this namespace found using label (actions.github.com/controller-watch-single-namespace). Consider set controllerServiceAccount.name in values.yaml to be explicitly if you think the discovery is wrong." }}
```
rentziass: Maybe these should live in a variable 😂 

```suggestion
      {{- fail "No gha-runner-scale-set-controller deployment that watch this namespace found using label (actions.github.com/controller-watch-single-namespace). Consider setting controllerServiceAccount.name in values.yaml to be explicit if you think the discovery is wrong." }}
```

charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -512,28 +533,49 @@ volumeMounts:
...
+          {{- $multiNamespacesCounter = add $multiNamespacesCounter 1 }}
+          {{- $controllerDeployment = $deployment }}
+        {{- end }}
       {{- end }}
     {{- end }}
   {{- end }}
-  {{- if lt $counter 1 }}
-    {{- fail "No gha-runner-scale-set-controller deployment found using label (app.kubernetes.io/part-of=gha-runner-scale-set-controller), consider set controllerServiceAccount.name to be explicitly if you think the discovery is wrong." }}
+  {{- if and (eq $multiNamespacesCounter 0) (eq $singleNamespaceCounter 0) }}
+    {{- fail "No gha-runner-scale-set-controller deployment found using label (app.kubernetes.io/part-of=gha-runner-scale-set-controller). Consider set controllerServiceAccount.name in values.yaml to be explicitly if you think the discovery is wrong." }}
```
rentziass:
```suggestion
    {{- fail "No gha-runner-scale-set-controller deployment found using label (app.kubernetes.io/part-of=gha-runner-scale-set-controller). Consider setting controllerServiceAccount.name in values.yaml to be explicit if you think the discovery is wrong." }}
```

charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -512,28 +533,49 @@ volumeMounts:
...
+  {{- end }}
+  {{- if and (gt $multiNamespacesCounter 0) (gt $singleNamespaceCounter 0) }}
+    {{- fail "Found both gha-runner-scale-set-controller installed with flags.watchSingleNamespace set and unset in cluster, this is not supported. Consider set controllerServiceAccount.name in values.yaml to be explicitly if you think the discovery is wrong." }}
```
rentziass:
```suggestion
    {{- fail "Found both gha-runner-scale-set-controller installed with flags.watchSingleNamespace set and unset in cluster, this is not supported. Consider setting controllerServiceAccount.name in values.yaml to be explicit if you think the discovery is wrong." }}
```

charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -512,28 +533,49 @@ volumeMounts:
...
   {{- end }}
-  {{- if gt $counter 1 }}
-    {{- fail "More than one gha-runner-scale-set-controller deployment found using label (app.kubernetes.io/part-of=gha-runner-scale-set-controller), consider set controllerServiceAccount.name to be explicitly if you think the discovery is wrong." }}
+  {{- if gt $multiNamespacesCounter 1 }}
+    {{- fail "More than one gha-runner-scale-set-controller deployment found using label (app.kubernetes.io/part-of=gha-runner-scale-set-controller). Consider set controllerServiceAccount.name in values.yaml to be explicitly if you think the discovery is wrong." }}
```
rentziass:
```suggestion
    {{- fail "More than one gha-runner-scale-set-controller deployment found using label (app.kubernetes.io/part-of=gha-runner-scale-set-controller). Consider setting controllerServiceAccount.name in values.yaml to be explicit if you think the discovery is wrong." }}
```

charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -512,28 +533,49 @@ volumeMounts:
...
+      {{- $managerServiceAccountNamespace = (get $controllerDeployment.metadata.labels "actions.github.com/controller-service-account-namespace") }}
+    {{- end }}
+  {{- else if gt $singleNamespaceCounter 0 }}
+    {{- if hasKey $singleNamespaceControllerDeployments .Release.Namespace }}
+      {{- $controllerDeployment = get $singleNamespaceControllerDeployments .Release.Namespace }}
+      {{- with $controllerDeployment.metadata }}
+        {{- $managerServiceAccountNamespace = (get $controllerDeployment.metadata.labels "actions.github.com/controller-service-account-namespace") }}
+      {{- end }}
+    {{- else }}
+      {{- fail "No gha-runner-scale-set-controller deployment that watch this namespace found using label (actions.github.com/controller-watch-single-namespace). Consider set controllerServiceAccount.name in values.yaml to be explicitly if you think the discovery is wrong." }}
```
rentziass:
```suggestion
      {{- fail "No gha-runner-scale-set-controller deployment that watch this namespace found using label (actions.github.com/controller-watch-single-namespace). Consider setting controllerServiceAccount.name in values.yaml to be explicit if you think the discovery is wrong." }}
```

charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -512,28 +533,49 @@ volumeMounts:
...
+    {{- end }}
   {{- end }}
   {{- if eq $managerServiceAccountNamespace "" }}
-    {{- fail "No service account namespace found for gha-runner-scale-set-controller deployment using label (actions.github.com/controller-service-account-namespace), consider set controllerServiceAccount.name to be explicitly if you think the discovery is wrong." }}
+    {{- fail "No service account namespace found for gha-runner-scale-set-controller deployment using label (actions.github.com/controller-service-account-namespace), consider set controllerServiceAccount.name in values.yaml to be explicitly if you think the discovery is wrong." }}
```
rentziass:
```suggestion
    {{- fail "No service account namespace found for gha-runner-scale-set-controller deployment using label (actions.github.com/controller-service-account-namespace), consider setting controllerServiceAccount.name in values.yaml to be explicit if you think the discovery is wrong." }}
```

main.go
rentziass: It should never be empty though, you can't create a resource outside a namespace?

main.go
TingluoHuang: we can't really watch a single namespace, we always need to watch 2 namespaces, the controller namespace (for listener creation) + the AutoScalingRunnerSet namespace.

nikola-jokic (approved):

PR #2377, opened 2023-03-08
Allow webhook server env to be set individually

jonathanwiemers:
Similar to https://github.com/actions/actions-runner-controller/pull/1565

### Problem
We are using a CSI Provder to sync secrets between AWS and EKS. The way that secrets are bound into the controller deployment doesn't really fit our workflow, so I'd be happy to specify my own environment variables from secrets, kind of the same way as I can specify additional volumes and mounts. 

This PR would allow something like this: 
```yaml
githubWebhookServer:
  env:
    - name: GITHUB_WEBHOOK_SECRET_TOKEN
      valueFrom:
        secretKeyRef:
          key: GITHUB_WEBHOOK_SECRET_TOKEN
          name: prod-gha-controller-webhook-token
          optional: true
```

The last reference in the env variables should be the one, that's taken by k8s to populate the env variable.

### DIFF
--- a/acceptance/deploy.sh
+++ b/acceptance/deploy.sh
@@ -83,6 +83,10 @@ if [ "${tool}" == "helm" ]; then
     flags+=( --set actionsMetricsServer.secret.create=true)
     flags+=( --set actionsMetricsServer.secret.github_token=${WEBHOOK_GITHUB_TOKEN})
   fi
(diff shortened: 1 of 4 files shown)


jonathanwiemers:
@mumoshu any objections?

mumoshu (approved):
Sorry it took some time for me to verify this doesn't break anything and the added feature works! It's all green now.
Thank you for your contribution @jonathanwiemers!

PR #2378, opened 2023-03-08
Build local image and load to kind cluster

AvaStancu:


.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -9,16 +9,33 @@ on:
 
 env:
   TARGET_ORG: actions-runner-controller
+  CLUSTER_NAME: e2e-test
+  RUNNER_VERSION: 2.302.1
```
Link-, line 13: What happens when we release a new runner version?

Link-: I think it'll be good to make this an input parameter that is optional and if not provided, we read the version from this file: https://github.com/actions/actions-runner-controller/blob/master/runner/VERSION 

We will always update this file with every new runner version release, it's used to publish the new runner images

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -9,16 +9,33 @@ on:
...
+            TAG=$(echo "0.0.$GITHUB_SHA")
+            echo "TAG=$TAG" >> $GITHUB_ENV
+            echo "IMAGE=$(echo "$IMAGE_REPO:$TAG)" >> $GITHUB_ENV
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+        with:
+          version: latest
+      - name: Docker Build Test Image
+        run: |
+            docker buildx build --build-arg RUNNER_VERSION=$RUNNER_VERSION --build-arg TAG=$TAG -t $IMAGE . --load
```
Link-: Do we need these flags?
```
	export DOCKER_CLI_EXPERIMENTAL=enabled
	export DOCKER_BUILDKIT=1
```
Source: https://github.com/actions/actions-runner-controller/blob/a462ecbe7956f1bd4edd838cdf476dba35b49667/Makefile#L202

.github/workflows/e2e-test-linux-vm.yaml
TingluoHuang: will [.github/workflows/update-runners.yaml](https://github.com/actions/actions-runner-controller/pull/2378/files#diff-f88bf13f3a64d7d08ec66bd4fd5d0967dab39a858f1b2a3f16af72324f57198e) take care of runner update?

Link-: Yes, I just saw it's already done that way

.github/workflows/e2e-test-linux-vm.yaml
AvaStancu: For our use case I do not see the need but I am not expert so could be missing something

TingluoHuang: I think we should include them to make the workflow align with the makefile.

Link- (approved):
Let's ship it, thank you!

PR #2382, opened 2023-03-09
Surface EphemeralRunnerSet stats to AutoscalingRunnerSet

nikola-jokic:
EphemeralRunnerSet splits runners according to the state they are in.

This change aims to make those numbers visible to the autoscaling runner set status.

Fixes https://github.com/github/c2c-actions-runtime/issues/2245


Link- (approved):
No comments from me, great PR 🚀

TingluoHuang (approved):

PR #2384, opened 2023-03-09
Update image references to Github Container Registry (GHCR)

scruplelesswizard:
Updates existing container image references from `registry.hub.docker.com/summerwind/` to `ghcr.io/actions/actions-runner-controller/`

As the Makefile requires some restructuring for use with the `{registry}/{owner}/{repo}` syntax it will be submitted in a separate PR.


Link-:
Thanks @chaosaffe - we are still publishing to those targets for backward compatibility, so it's safe to pull from them. We'll address this PR when we're ready to sunset these targets.

PR #2391, opened 2023-03-10
Add labels to autoscaling runner set subresources to allow easier inspection

nikola-jokic:
Fixes https://github.com/github/c2c-actions-runtime/issues/2175

Addressing ADR: https://github.com/actions/actions-runner-controller/blob/master/adrs/2022-12-05-adding-labels-k8s-resources.md

### DIFF
--- a/charts/gha-runner-scale-set/templates/_helpers.tpl
+++ b/charts/gha-runner-scale-set/templates/_helpers.tpl
@@ -40,7 +40,7 @@ helm.sh/chart: {{ include "gha-runner-scale-set.chart" . }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
-app.kubernetes.io/part-of: gha-runner-scale-set-controller
+app.kubernetes.io/part-of: gha-runner-scale-set
 {{- end }}
 
 {{/*
--- a/charts/gha-runner-scale-set/templates/_helpers.tpl
+++ b/charts/gha-runner-scale-set/templates/_helpers.tpl
@@ -40,6 +40,7 @@ helm.sh/chart: {{ include "gha-runner-scale-set.chart" . }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
+app.kubernetes.io/part-of: gha-runner-scale-set-controller
 {{- end }}
 
 {{/*
--- a/controllers/actions.github.com/autoscalinglistener_controller.go
+++ b/controllers/actions.github.com/autoscalinglistener_controller.go
@@ -523,8 +523,8 @@ func (r *AutoscalingListenerReconciler) createProxySecret(ctx context.Context, a
 			Name:      proxyListenerSecretName(autoscalingListener),
 			Namespace: autoscalingListener.Namespace,
 			Labels: map[string]string{
-				"auto-scaling-runner-set-namespace": autoscalingListener.Spec.AutoscalingRunnerSetNamespace,
-				"auto-scaling-runner-set-name":      autoscalingListener.Spec.AutoscalingRunnerSetName,
+				LabelKeyGithubScaleSetName:      autoscalingListener.Spec.AutoscalingRunnerSetName,
+				LabelKeyGithubScaleSetNamespace: autoscalingListener.Spec.AutoscalingRunnerSetNamespace,
 			},
 		},
 		Data: data,
--- a/controllers/actions.github.com/autoscalinglistener_controller.go
+++ b/controllers/actions.github.com/autoscalinglistener_controller.go
@@ -523,8 +523,8 @@ func (r *AutoscalingListenerReconciler) createProxySecret(ctx context.Context, a
 			Name:      proxyListenerSecretName(autoscalingListener),
 			Namespace: autoscalingListener.Namespace,
 			Labels: map[string]string{
-				LabelKeyGithubScaleSetName:      autoscalingListener.Spec.AutoscalingRunnerSetName,
 				LabelKeyGithubScaleSetNamespace: autoscalingListener.Spec.AutoscalingRunnerSetNamespace,
+				LabelKeyGithubScaleSetName:      autoscalingListener.Spec.AutoscalingRunnerSetName,
 			},
 		},
 		Data: data,
--- a/controllers/actions.github.com/autoscalinglistener_controller.go
+++ b/controllers/actions.github.com/autoscalinglistener_controller.go
@@ -523,8 +523,8 @@ func (r *AutoscalingListenerReconciler) createProxySecret(ctx context.Context, a
 			Name:      proxyListenerSecretName(autoscalingListener),
 			Namespace: autoscalingListener.Namespace,
 			Labels: map[string]string{
-				LabelKeyGithubScaleSetNamespace: autoscalingListener.Spec.AutoscalingRunnerSetNamespace,
-				LabelKeyGithubScaleSetName:      autoscalingListener.Spec.AutoscalingRunnerSetName,
+				LabelKeyGitHubScaleSetNamespace: autoscalingListener.Spec.AutoscalingRunnerSetNamespace,
+				LabelKeyGitHubScaleSetName:      autoscalingListener.Spec.AutoscalingRunnerSetName,
 			},
 		},
 		Data: data,
--- a/controllers/actions.github.com/autoscalingrunnerset_controller.go
+++ b/controllers/actions.github.com/autoscalingrunnerset_controller.go
@@ -364,12 +364,16 @@ func (r *AutoscalingRunnerSetReconciler) createRunnerScaleSet(ctx context.Contex
 	if autoscalingRunnerSet.Annotations == nil {
 		autoscalingRunnerSet.Annotations = map[string]string{}
 	}
+	if autoscalingRunnerSet.Labels == nil {
+		autoscalingRunnerSet.Labels = map[string]string{}
+	}
 
-	logger.Info("Adding runner scale set ID, name and runner group name as an annotation")
+	logger.Info("Adding runner scale set ID, name and runner group name as an annotation and url labels")
 	if err = patch(ctx, r.Client, autoscalingRunnerSet, func(obj *v1alpha1.AutoscalingRunnerSet) {
 		obj.Annotations[runnerScaleSetNameAnnotationKey] = runnerScaleSet.Name
 		obj.Annotations[runnerScaleSetIdAnnotationKey] = strconv.Itoa(runnerScaleSet.Id)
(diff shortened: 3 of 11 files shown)


controllers/actions.github.com/resourcebuilder_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"testing"
+
+	"github.com/actions/actions-runner-controller/apis/actions.github.com/v1alpha1"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestLabelPropagation(t *testing.T) {
```
nikola-jokic, line 14: Not sure how much value is added through this test, but I left it here since I already wrote it :relaxed: 
Let me know if anyone thinks this is useless. It is basically just something to confirm they are passed from one resource to another...

controllers/actions.github.com/resourcebuilder.go
```diff
@@ -19,14 +20,92 @@ const (
...
 const (
-	LabelKeyAutoScaleRunnerSetName      = "auto-scaling-runner-set-name"
-	LabelKeyAutoScaleRunnerSetNamespace = "auto-scaling-runner-set-namespace"
+	// Kubernetes labels
+	LabelKeyKubernetesPartOf    = "app.kubernetes.io/part-of"
+	LabelKeyKubernetesComponent = "app.kubernetes.io/component"
+	LabelKeyKubernetesVersion   = "app.kubernetes.io/version"
+
+	// Github labels
+	LabelKeyGithubScaleSetName      = "actions.github.com/scale-set-name"
```
TingluoHuang: I would name is as `runner-scale-set-name` to match the real name.
you will need to modify the E2E test workflow since it was looking at the listener pod via the label.

charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -40,6 +40,7 @@ helm.sh/chart: {{ include "gha-runner-scale-set.chart" . }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
+app.kubernetes.io/part-of: gha-runner-scale-set-controller
```
TingluoHuang:
```suggestion
app.kubernetes.io/part-of: gha-runner-scale-set
```

TingluoHuang: I am using the `app.kubernetes.io/part-of: gha-runner-scale-set-controller` to find the controller deployment during delay role binding.

controllers/actions.github.com/autoscalinglistener_controller.go
```diff
@@ -523,8 +523,8 @@ func (r *AutoscalingListenerReconciler) createProxySecret(ctx context.Context, a
 			Name:      proxyListenerSecretName(autoscalingListener),
 			Namespace: autoscalingListener.Namespace,
 			Labels: map[string]string{
-				"auto-scaling-runner-set-namespace": autoscalingListener.Spec.AutoscalingRunnerSetNamespace,
```
TingluoHuang, line 526: make sure you double-check all the existing code that uses the old name via string search and update them properly.

controllers/actions.github.com/autoscalinglistener_controller.go
```diff
@@ -523,8 +523,8 @@ func (r *AutoscalingListenerReconciler) createProxySecret(ctx context.Context, a
...
-				"auto-scaling-runner-set-name":      autoscalingListener.Spec.AutoscalingRunnerSetName,
+				LabelKeyGithubScaleSetName:      autoscalingListener.Spec.AutoscalingRunnerSetName,
+				LabelKeyGithubScaleSetNamespace: autoscalingListener.Spec.AutoscalingRunnerSetNamespace,
```
TingluoHuang:
```suggestion
				LabelKeyGithubScaleSetNamespace: autoscalingListener.Spec.AutoscalingRunnerSetNamespace,
				LabelKeyGithubScaleSetName:      autoscalingListener.Spec.AutoscalingRunnerSetName,
```

Flip the order back might reduce the diff. 😄

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -154,7 +153,7 @@ func (r *AutoscalingRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl
 	}
 
 	// Make sure the runner group of the scale set is up to date
-	currentRunnerGroupName, ok := autoscalingRunnerSet.Annotations[runnerScaleSetRunnerGroupNameKey]
+	currentRunnerGroupName, ok := autoscalingRunnerSet.Annotations[LabelKeyGithubRunnerGroupName]
```
TingluoHuang: change all `Github` -> `GitHub`? 😄

controllers/actions.github.com/resourcebuilder.go
```diff
@@ -337,6 +399,7 @@ func (b *resourceBuilder) newEphemeralRunnerPod(ctx context.Context, runner *v1a
 	for k, v := range runner.Spec.PodTemplateSpec.Labels {
 		labels[k] = v
 	}
+	labels["actions-ephemeral-runner"] = string(corev1.ConditionTrue)
```
TingluoHuang, line 418: do you want to move this to some constant you created?

controllers/actions.github.com/resourcebuilder.go
```diff
@@ -19,14 +20,92 @@ const (
...
+	LabelKeyGithubEnterprise,
+	LabelKeyGithubOrganization,
+	LabelKeyGithubRepository,
+}
+
+const labelValueKubernetesPartOf = "actions-runner-controller"
+
 type resourceBuilder struct{}
 
+func (b *resourceBuilder) newAutoScalingListener(autoscalingRunnerSet *v1alpha1.AutoscalingRunnerSet, ephemeralRunnerSet *v1alpha1.EphemeralRunnerSet, namespace, image string, imagePullSecrets []corev1.LocalObjectReference) (*v1alpha1.AutoscalingListener, error) {
```
TingluoHuang: do you mind moving the method back to the line it was on to reduce the diff?

charts/gha-runner-scale-set/templates/_helpers.tpl
nikola-jokic: Oh I see, this is my mistake

controllers/actions.github.com/resourcebuilder.go
nikola-jokic: Sure, I wanted to move stuff around so it is easier to find, but that might be something for another PR to make this one easier to review :relaxed:

controllers/actions.github.com/resourcebuilder.go
nikola-jokic: The reason why I changed this is to conform to our [ADR](https://github.com/actions/actions-runner-controller/blob/master/adrs/2023-04-14-adding-labels-k8s-resources.md#listener)
If we move back to `runner-scale-set-name`, ideally I'd like to update ADR to specify that :relaxed:

controllers/actions.github.com/autoscalingrunnerset_controller.go
nikola-jokic: I'll rename it :relaxed: Thanks for the suggestion!

controllers/actions.github.com/resourcebuilder.go
nikola-jokic: This label is used only in a single place, so I did not want to have it as a constant.
There are other labels similar to this and I left them as string literals

controllers/actions.github.com/autoscalinglistener_controller.go
nikola-jokic: did a global search I think everything is okay here

controllers/actions.github.com/resourcebuilder.go
TingluoHuang: that's fine, make sure to update the E2E workflow yaml, the workflow is still using the old table to assert the controller is up and running...

controllers/actions.github.com/resourcebuilder.go
```diff
@@ -149,21 +182,40 @@ func (b *resourceBuilder) newScaleSetListenerPod(autoscalingListener *v1alpha1.A
...
+		LabelKeyKubernetesComponent:     "runner-set",
+		LabelKeyKubernetesVersion:       autoscalingRunnerSet.Labels[LabelKeyKubernetesVersion],
+		LabelKeyGitHubScaleSetName:      autoscalingRunnerSet.Name,
+		LabelKeyGitHubScaleSetNamespace: autoscalingRunnerSet.Namespace,
+		LabelKeyGitHubEnterprise:        githubConfig.Enterprise,
+		LabelKeyGitHubOrganization:      githubConfig.Organization,
+		LabelKeyGitHubRepository:        githubConfig.Repository,
+	}
+
+	newAnnotations := map[string]string{
```
nikola-jokic, line 204: Should we inherit all annotations from the autoscaling runner set?

TingluoHuang: I would start with NO, we can added if we need in the future.

controllers/actions.github.com/resourcebuilder.go
```diff
@@ -285,14 +337,25 @@ func (b *resourceBuilder) newAutoScalingListener(autoscalingRunnerSet *v1alpha1.
...
-				LabelKeyAutoScaleRunnerSetName:      autoscalingRunnerSet.Name,
-				LabelKeyRunnerSpecHash:              autoscalingRunnerSet.ListenerSpecHash(),
+				LabelKeyGitHubScaleSetNamespace: autoscalingRunnerSet.Namespace,
+				LabelKeyGitHubScaleSetName:      autoscalingRunnerSet.Name,
+				LabelKeyKubernetesPartOf:        labelValueKubernetesPartOf,
+				LabelKeyKubernetesComponent:     "runner-scale-set-listener",
+				LabelKeyKubernetesVersion:       autoscalingRunnerSet.Labels[LabelKeyKubernetesVersion],
+				LabelKeyGitHubEnterprise:        githubConfig.Enterprise,
+				LabelKeyGitHubOrganization:      githubConfig.Organization,
+				LabelKeyGitHubRepository:        githubConfig.Repository,
```
TingluoHuang, line 353: do we also want to add these labels to the AutoScalingRunnerSet resource?

nikola-jokic: Good suggestion! I'll add them

TingluoHuang: you get a linter error with the latest change. 😄

nikola-jokic: fixed :smile:

TingluoHuang (approved):

PR #2393, opened 2023-03-10
Create separate chart validation workflow for gha-* charts.

TingluoHuang:
Few reasons:
- We want to make the helm chart validation test to validate against the latest controller image build from code, not the released controller image.
- `gha-runner-scale-set` chart depends on `gha-runner-scale-set-controller` chart is installed, the default behavior of `CT` is install `gha-runner-scale-set-controller`, uninstall `gha-runner-scale-set-controller`, install `gha-runner-scale-set`, and uninstall `gha-runner-scale-set`, I need to change the behavior to not uninstall the chart after installing.
- Better triggering filter to avoid unnecessary runs.
- Long term, we might need to evaluate whether we still need to have this workflow for the `gha-*` charts since we have L0 tests cover the chart render already.


Link-:
Can we add a little bit more context (for posterity) in the PR description about why we decided to split the workflow?

.github/workflows/validate-gha-chart.yaml
```diff
@@ -0,0 +1,117 @@
...
+      - '!**.md'
+  push:
+    paths:
+      - 'charts/**'
+      - '.github/workflows/validate-gha-chart.yaml'
+      - '!charts/actions-runner-controller/**'
+      - '!**.md'
+  workflow_dispatch:
+env:
+  KUBE_SCORE_VERSION: 1.10.0
```
nikola-jokic: Should we bump this to the latest version?
```suggestion
  KUBE_SCORE_VERSION: 1.16.1
```

.github/workflows/validate-gha-chart.yaml
```diff
@@ -0,0 +1,117 @@
...
+          echo 'deb [trusted=yes] https://repo.goreleaser.com/apt/ /' | sudo tee /etc/apt/sources.list.d/goreleaser.list
+          sudo apt update
+          sudo apt install goreleaser
+          git clone https://github.com/helm/chart-testing
+          cd chart-testing
+          unset CT_CONFIG_DIR
+          goreleaser build --clean --skip-validate 
+          ./dist/chart-testing_linux_amd64_v1/ct version
+          echo 'Adding ct directory to PATH...'
+          echo "$RUNNER_TEMP/chart-testing/dist/chart-testing_linux_amd64_v1" >> "$GITHUB_PATH"
```
rentziass, line 77: We can probably swap all that with
```suggestion
          go install github.com/helm/chart-testing/v3/ct@main
```

Provided `ubuntu-latest` ships with Go already.

.github/workflows/validate-gha-chart.yaml
```diff
@@ -0,0 +1,117 @@
...
+        if: steps.list-changed.outputs.changed == 'true'
+
+      - name: Load Image to Kind Cluster
+        if: steps.list-changed.outputs.changed == 'true'
+        run: kind load docker-image test-arc:dev --name chart-testing
+
+      - name: Run chart-testing (install)
+        if: steps.list-changed.outputs.changed == 'true'
+        run: |
+          ct install --config charts/.ci/ct-config-gha.yaml
```
rentziass: This is the nit of the century, but can we agree if we want or not new line chars at the end of files? 😄

rentziass:
Thank you soooo much for doing this 🙇

.github/workflows/validate-gha-chart.yaml
TingluoHuang: I tried locally, but the binary from `go install github.com/helm/chart-testing/v3/ct@main` still doesn't support `--skip-clean-up`, did I miss something? 😕

.github/workflows/validate-gha-chart.yaml
TingluoHuang: probably related to my VSCode setup with actions service code styling. 😄 
any suggestion?

.github/workflows/validate-gha-chart.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+      - name: Run chart-testing (lint)
+        run: |
+          ct lint --config charts/.ci/ct-config-gha.yaml
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+        if: steps.list-changed.outputs.changed == 'true'
+        with:
+          version: latest
```
Link-, line 99: Can we pin the version to avoid weird problems in the future?

.github/workflows/validate-gha-chart.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+      - name: Docker Build Test Image
+        if: steps.list-changed.outputs.changed == 'true'
+        run: |
+            DOCKER_CLI_EXPERIMENTAL=enabled DOCKER_BUILDKIT=1 docker buildx build --build-arg TAG=dev -t test-arc:dev . --load
```
Link-: Same comment as here: https://github.com/actions/actions-runner-controller/pull/2395#discussion_r1133840117

.github/workflows/validate-gha-chart.yaml
```diff
@@ -0,0 +1,117 @@
...
+        run: |
+          ct lint --config charts/.ci/ct-config-gha.yaml
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+        if: steps.list-changed.outputs.changed == 'true'
+        with:
+          version: latest
+
+      - name: Docker Build Test Image
```
Link-: nit: either title case everywhere or sentence case everywhere for step / job names :D

.github/workflows/validate-gha-chart.yaml
```diff
@@ -0,0 +1,117 @@
...
+        if: steps.list-changed.outputs.changed == 'true'
+        run: |
+            DOCKER_CLI_EXPERIMENTAL=enabled DOCKER_BUILDKIT=1 docker buildx build --build-arg TAG=dev -t test-arc:dev . --load
+
+      - name: Create kind cluster
+        uses: helm/kind-action@v1.4.0
+        if: steps.list-changed.outputs.changed == 'true'
+
+      - name: Load Image to Kind Cluster
```
Link-: nit: Title case or sentence case

Link- (commented):
Thank you for putting this together

.github/workflows/validate-gha-chart.yaml
nikola-jokic: I'd say with new line at the end to avoid UI showing :no_entry: :relaxed:

.github/workflows/validate-gha-chart.yaml
TingluoHuang: Since this is a validation workflow that only build docker image, I would stay on `latest` to test out docker buildx and report bug to them as needed.

.github/workflows/validate-gha-chart.yaml
TingluoHuang: I don't plan to give the workflow any write permission to GITHUB_TOKEN since it will run as part of PR.

Link-: @TingluoHuang you don't have to. The action will use our cache service https://github.com/moby/buildkit#github-actions-cache-experimental not packages.

TingluoHuang: Changing. 😄

Link-:
image

Link- (approved):

PR #2395, opened 2023-03-11
Update E2E test workflow.

TingluoHuang:
I am planning to expand the e2e workflow to have multiple jobs, and each job will cover a particular setup scenario.
ex:
- Default
- Proxy
- Cert
- Watch single namespace
- etc

To enable that I am pulling the `helm install` part out of the composite action and putting the part that builds the image and setup kind into the composite action.


.github/actions/setup-arc-e2e/action.yaml
```diff
@@ -0,0 +1,55 @@
...
+    description: 'Token to use for configure ARC'
+    value: ${{steps.config-token.outputs.token}}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v2
+      with:
+        version: latest
```
nikola-jokic: Should we pin the action version or buildx version?

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -5,47 +5,73 @@ on:
...
+          export ARC_NAME="${{ steps.install_arc.outputs.ARC_NAME }}"
+          go test ./test_e2e_arc -v
+
+      - name: Uninstall gha-runner-scale-set
+        if: steps.install_arc.outcome == 'success'
+        run: |
+          helm uninstall ${{ steps.install_arc.outputs.ARC_NAME }} --namespace arc-runners
+          sleep 30
+
+      - name: Dump gha-runner-scale-set-controller logs
```
nikola-jokic, line 123: Should we dump the log before uninstall?

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -5,47 +5,73 @@ on:
...
+        if: steps.install_arc_controller.outcome == 'success'
+        run: |
+          kubectl logs deployment/arc-gha-runner-scale-set-controller -n arc-systems
```
nikola-jokic, line 126:
```suggestion
          kubectl logs deployments/arc-gha-runner-scale-set-controller -n arc-systems
```
I think it should be deployments, this might cause error later

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -5,47 +5,73 @@ on:
...
+          ARC_NAME=arc-runner-${{github.job}}-$(date +'%M-%S')-$(($RANDOM % 100 + 1))
+          helm install "$ARC_NAME" \
+          --namespace "arc-runners" \
+          --create-namespace \
+          --set githubConfigUrl="${{ env.TEST_CONFIG_URL }}" \
+          --set githubConfigSecret.github_token="${{ steps.setup.outputs.token }}" \
+          ./charts/gha-runner-scale-set \
+          --debug
+          echo "ARC_NAME=$ARC_NAME" >> $GITHUB_OUTPUT
+          sleep 30
```
rentziass: Can we maybe use something like `kubectl wait --for-condition=ready pod -n arc-systems {a label to get the controller}`?

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -5,47 +5,73 @@ on:
...
+        run: |
+          export GITHUB_TOKEN="${{ steps.setup.outputs.token }}"
+          export ARC_NAME="${{ steps.install_arc.outputs.ARC_NAME }}"
+          go test ./test_e2e_arc -v
+
+      - name: Uninstall gha-runner-scale-set
+        if: steps.install_arc.outcome == 'success'
+        run: |
+          helm uninstall ${{ steps.install_arc.outputs.ARC_NAME }} --namespace arc-runners
+          sleep 30
```
rentziass: Same with `kubectl wait`

test_e2e_arc/arc_jobs_test.go
```diff
@@ -92,11 +92,11 @@ func TestARCJobs(t *testing.T) {
 	)
 	t.Run("Get available pods during job run", func(t *testing.T) {
```
rentziass, line 93: Unrelated to your change, but I think these `t.Run` should be flattened into one, in case `Get available pods before job run` for example there's no need to run the full test. (could be done in a separate PR)

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -5,47 +5,73 @@ on:
...
   workflow_dispatch:
 
 env:
   TARGET_ORG: actions-runner-controller
-  CLUSTER_NAME: e2e-test
-  RUNNER_VERSION: 2.302.1
-  IMAGE_REPO: "test/test-image"
+  IMAGE_NAME: "arc-test-image"
+  IMAGE_VERSION: "test"
+  TEST_CONFIG_URL: "https://github.com/actions-runner-controller/arc_e2e_test_dummy"
```
Link-: I think there's value in making those workflow_dispatch inputs so that we can override their values without creating a PR

.github/actions/setup-arc-e2e/action.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+    - name: Docker build test image
+      shell: bash
+      run: |
+          export DOCKER_IMAGE_NAME=${{inputs.docker-image-name}}
+          export VERSION=${{inputs.docker-image-tag}}
+          export IMG_RESULT=load
+          make docker-buildx
```
Link-: While I think it's important for us to stay consistent, I really think we should optimise this step & cache some of the image layers. This test will run frequently and this will be the biggest bottleneck in the flow

If you use something like this: https://github.com/actions-runner-controller/releases/blob/222166eca744b5a519d65428cef8f7eb471fb3e6/.github/workflows/publish-canary.yaml#L63

We'll be able to speed up our runtime substantially

TingluoHuang: I don't want to give these test workflows any `write` permission to the registry.

.github/actions/setup-arc-e2e/action.yaml
TingluoHuang: Should we be aware of if the latest is broken?
We don't restrict the `buildx` version in the makefile, why do we need to restrict it in the workflow?

.github/workflows/e2e-test-linux-vm.yaml
TingluoHuang: it was working fine with my syntax. 😄 
https://github.com/actions/actions-runner-controller/actions/runs/4395784129/jobs/7697812090#step:8:2

nikola-jokic: Yeah, that one is confusing me
Because kubectl autocomplete completes to deployments :relaxed:

.github/actions/setup-arc-e2e/action.yaml
Link-: @TingluoHuang you don't have to. The action will use our cache service https://github.com/moby/buildkit#github-actions-cache-experimental not packages.

.github/workflows/e2e-test-linux-vm.yaml
TingluoHuang: We don't uninstall the controller.
We uninstall the `gha-runner-scale-set` and then dump the controller log, so we can also get the log for how the controller handles the `AutoScalingRunnerSet` deletion.

.github/actions/setup-arc-e2e/action.yaml
TingluoHuang: Updating...

.github/workflows/e2e-test-linux-vm.yaml
TingluoHuang: `kubectl wait` won't work for creating a new resource, it will error if the resource hasn't been fully created.
I will add a short 5 seconds of sleep to give the cluster a little moment.

controllers/actions.github.com/ephemeralrunnerset_controller_test.go
```diff
@@ -1141,12 +1141,13 @@ var _ = Describe("Test EphemeralRunnerSet controller with custom root CA", func(
...
-		updatedRunnerSet := new(actionsv1alpha1.EphemeralRunnerSet)
-		err = k8sClient.Get(ctx, client.ObjectKey{Namespace: ephemeralRunnerSet.Namespace, Name: ephemeralRunnerSet.Name}, updatedRunnerSet)
+		currentRunnerSet := new(actionsv1alpha1.EphemeralRunnerSet)
+		err = k8sClient.Get(ctx, client.ObjectKey{Namespace: ephemeralRunnerSet.Namespace, Name: ephemeralRunnerSet.Name}, currentRunnerSet)
 		Expect(err).NotTo(HaveOccurred(), "failed to get EphemeralRunnerSet")
 
+		updatedRunnerSet := currentRunnerSet.DeepCopy()
 		updatedRunnerSet.Spec.Replicas = 0
-		err = k8sClient.Update(ctx, updatedRunnerSet)
+		err = k8sClient.Patch(ctx, updatedRunnerSet, client.MergeFrom(currentRunnerSet))
```
TingluoHuang, line 1150: change to patch to avoid conflict on updating the outdate object.

.github/workflows/e2e-test-linux-vm.yaml
TingluoHuang: updated.

.github/actions/setup-arc-e2e/action.yaml
nikola-jokic: I mean for makefile, it will just use the buildx version you have installed on your system right?
and as long as we are able to build our image with some specific version and run it, it should not matter if the latest buildx version is breaking our build
It is like, if the go build breaks on the next go version, we would not care about it until we bump the version up:
https://github.com/actions/actions-runner-controller/blob/bd9f32e3540663360cf47f04acad26e6010f772e/.github/workflows/validate-arc.yaml#L37

TingluoHuang: Ok, I am using the same version as `publish-runner-scale-set.yaml` now.

Link- (approved):
LGTM

PR #2398, opened 2023-03-13
Allow customization of admission webhook timeout

zanehala-wk:
This allows users to specify the timeout for the admission webhooks, while defaulting to the normal Kubernetes default of 10s if not set.

I've run into this issue a few times while attempting to deploy, 10s sometimes seems a bit too short and the deploy fails.


mumoshu (approved):
I've tested this locally and it worked flawlessly! Thank you so much for your contribution @zanehala-wk ☺️

PR #2407, opened 2023-03-14
Introduce ADR change for adding labels to our resources

nikola-jokic:
Addressing a change raised by the comment: https://github.com/actions/actions-runner-controller/pull/2374#discussion_r1133733573

Since we moved ADRs here, we forgot to update the numbers within the files. This change fixes the numbering

### DIFF
--- a/adrs/2022-10-17-runner-image.md
+++ b/adrs/2022-10-17-runner-image.md
@@ -20,11 +20,11 @@ The latest Dockerfile can be found at: https://github.com/actions/runner/blob/ma
 
 # Context
 
-user can bring their own runner images, the contract we have are:
+users can build their own runner images, the contract we require is:
 
-- It must have a runner binary under /actions-runner (/actions-runner/run.sh exists)
-- The WORKDIR is set to /actions-runner
-- If the user inside the container is root, the ENV RUNNER_ALLOW_RUNASROOT should be set to 1
+- It must have a runner binary under `/actions-runner` i.e. `/actions-runner/run.sh` exists
+- The WORKDIR is set to `/actions-runner`
+- If the user inside the container is root, the environment variable `RUNNER_ALLOW_RUNASROOT` should be set to 1
 
 The existing ARC runner images will not work with the new ARC mode out-of-box for the following reason:
 
--- a/adrs/2022-10-17-runner-image.md
+++ b/adrs/2022-10-17-runner-image.md
@@ -20,19 +20,19 @@ The latest Dockerfile can be found at: https://github.com/actions/runner/blob/ma
 
 # Context
 
-user can bring their own runner images, the contract we have are:
+users can bring their own runner images, the contract we require is:
 
-- It must have a runner binary under /actions-runner (/actions-runner/run.sh exists)
-- The WORKDIR is set to /actions-runner
-- If the user inside the container is root, the ENV RUNNER_ALLOW_RUNASROOT should be set to 1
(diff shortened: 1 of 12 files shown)


adrs/2022-10-17-runner-image.md
```diff
@@ -19,6 +21,7 @@ The latest Dockerfile can be found at: https://github.com/actions/runner/blob/ma
 # Context
 
 user can bring their own runner images, the contract we have are:
```
Link-:
```suggestion
users can build their own runner images, the contract we require is:
```

adrs/2022-10-17-runner-image.md
```diff
@@ -19,6 +21,7 @@ The latest Dockerfile can be found at: https://github.com/actions/runner/blob/ma
...
+
 - It must have a runner binary under /actions-runner (/actions-runner/run.sh exists)
```
Link-:
```suggestion
- It must have a runner binary under `/actions-runner` i.e. `/actions-runner/run.sh` exists
```

adrs/2022-10-17-runner-image.md
```diff
@@ -19,6 +21,7 @@ The latest Dockerfile can be found at: https://github.com/actions/runner/blob/ma
...
 - The WORKDIR is set to /actions-runner
```
Link-:
```suggestion
- The WORKDIR is set to `/actions-runner`
```

adrs/2022-10-17-runner-image.md
```diff
@@ -19,6 +21,7 @@ The latest Dockerfile can be found at: https://github.com/actions/runner/blob/ma
...
 - If the user inside the container is root, the ENV RUNNER_ALLOW_RUNASROOT should be set to 1
```
Link-:
```suggestion
- If the user inside the container is root, the environment variable `RUNNER_ALLOW_RUNASROOT` should be set to 1
```

Link- (approved):
Thank you for doing the hard work and cleaning up our ADRs 🙏

charts/gha-runner-scale-set-controller/templates/_helpers.tpl
```diff
@@ -39,7 +39,7 @@ helm.sh/chart: {{ include "gha-runner-scale-set-controller.chart" . }}
 {{- if .Chart.AppVersion }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
-app.kubernetes.io/part-of: {{ .Chart.Name }}
+app.kubernetes.io/part-of: gha-runner-scale-set-controller
```
TingluoHuang, line 42: why do we need to change this?

nikola-jokic: We don't, it was just something I thought it would be nice when looking at the https://github.com/actions/actions-runner-controller/blob/34f38788292d888b830e69245f0033b252dea351/charts/gha-runner-scale-set/templates/_helpers.tpl#L447

I thought it would be easier and more safe. Since we use constant in `gha-runner-scale-set`, I wanted to make it constant in `gha-runner-scale-set-controller` :relaxed: 
Easier to search for where this value is coming from, but I can revert it back

What do you think

TingluoHuang: 👍

TingluoHuang (approved):

PR #2412, opened 2023-03-14
Adding e2e workflows to test dind, kube mode and proxy

TingluoHuang:
Adding E2E test to cover key container setup scenarios:
- DinD mode
- Kubernetes mode
- Auth proxy mode
- Anonymous proxy mode

Will try to figure out how to better organize the YAML in a follow-up PR.

### DIFF
--- a/.github/workflows/e2e-test-linux-vm.yaml
+++ b/.github/workflows/e2e-test-linux-vm.yaml
@@ -506,7 +506,7 @@ jobs:
             --namespace=arc-runners \
             --from-literal=username=github \
             --from-literal=password='actions'
-          MINIKUBE_HOST_ADDRESS=$(minikube -p actions-dev ssh cat /etc/hosts | grep host.minikube.internal | awk '{ print $1 }')
+          MINIKUBE_HOST_ADDRESS=$(minikube ssh cat /etc/hosts | grep host.minikube.internal | awk '{ print $1 }')
           ARC_NAME=arc-runner-${{github.job}}-$(date +'%M-%S')-$(($RANDOM % 100 + 1))
           helm install "$ARC_NAME" \
           --namespace "arc-runners" \
@@ -617,7 +617,7 @@ jobs:
             --name squid \
             --publish 3128:3128 \
             ubuntu/squid:latest
-          MINIKUBE_HOST_ADDRESS=$(minikube -p actions-dev ssh cat /etc/hosts | grep host.minikube.internal | awk '{ print $1 }')
+          MINIKUBE_HOST_ADDRESS=$(minikube ssh cat /etc/hosts | grep host.minikube.internal | awk '{ print $1 }')
           ARC_NAME=arc-runner-${{github.job}}-$(date +'%M-%S')-$(($RANDOM % 100 + 1))
           helm install "$ARC_NAME" \
           --namespace "arc-runners" \
--- a/.github/workflows/e2e-test-linux-vm.yaml
+++ b/.github/workflows/e2e-test-linux-vm.yaml
@@ -112,6 +112,7 @@ jobs:
         run: |
           export GITHUB_TOKEN="${{ steps.setup.outputs.token }}"
           export ARC_NAME="${{ steps.install_arc.outputs.ARC_NAME }}"
+          export WORKFLOW_FILE="arc-test-workflow.yaml"
           go test ./test_e2e_arc -v
 
       - name: Uninstall gha-runner-scale-set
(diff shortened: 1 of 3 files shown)


nikola-jokic (dismissed):
LGTM

test_e2e_arc/arc_jobs_test.go
```diff
@@ -93,16 +94,23 @@ func TestARCJobs(t *testing.T) {
...
 		c := http.Client{}
 		targetArcName := os.Getenv("ARC_NAME")
+		require.NotEmpty(t, targetArcName, "ARC_NAME environment variable is required for this test to run. (e.g. arc-e2e-test)")
+
+		targetWorkflow := os.Getenv("WORKFLOW_FILE")
+		require.NotEmpty(t, targetWorkflow, "WORKFLOW_FILE environment variable is required for this test to run. (e.g. e2e_test.yml)")
+
+		ght := os.Getenv("GITHUB_TOKEN")
+		require.NotEmpty(t, ght, "GITHUB_TOKEN environment variable is required for this test to run.")
+
```
Link-, line 104: I was going to add this myself but let's avoid conflicts since you're already here 😄 

How about adding another variable along the lines of `TEST_UUID` that we can provide as input to the workflow, and we spit it out on the other end. This way we can match the test here with the run in the other repo.

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -225,4 +227,434 @@ jobs:
...
+            sleep 1
+          done
+          kubectl wait --timeout=30s --for=condition=ready pod -n arc-systems -l auto-scaling-runner-set-name=$ARC_NAME
+          kubectl get pod -n arc-systems
+
+      - name: Test ARC scales pods up and down
+        run: |
+          export GITHUB_TOKEN="${{ steps.setup.outputs.token }}"
+          export ARC_NAME="${{ steps.install_arc.outputs.ARC_NAME }}"
+          export WORKFLOW_FILE="arc-test-dind-workflow.yaml"
```
Link-: I suggest exporting this variable to GITHUB_ENV so that we can use in the job summary, or make it an output or extract it to the job level

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -225,4 +227,434 @@ jobs:
...
+      - name: Uninstall gha-runner-scale-set
+        if: always() && steps.install_arc.outcome == 'success'
+        run: |
+          helm uninstall ${{ steps.install_arc.outputs.ARC_NAME }} --namespace arc-runners
+          kubectl wait --timeout=10s --for=delete AutoScalingRunnerSet -n demo -l app.kubernetes.io/instance=${{ steps.install_arc.outputs.ARC_NAME }}
+
+      - name: Dump gha-runner-scale-set-controller logs
+        if: always() && steps.install_arc_controller.outcome == 'success'
+        run: |
+          kubectl logs deployment/arc-gha-runner-scale-set-controller -n arc-systems
```
Link-: I suggest adding a step after this with a job summary. This will require `${{ env.WORKFLOW_FILE }}` to be defined and available

```
- name: Job summary
  if: always()
  run: |
    cat <<-EOF > $GITHUB_STEP_SUMMARY
    | **Outcome**    | ${{ steps.install_arc_controller.outcome }}  |
    |----------------|--------------------------------------------- |
    | **References** | [Test workflow runs](https://github.com/actions-runner-controller/arc_e2e_test_dummy/actions/workflows/${{ env.WORKFLOW_FILE }}) |
    EOF
```

test_e2e_arc/arc_jobs_test.go
TingluoHuang: we can use something like `run-name: Run test workflow on ${{ inputs.arc_name }}` on the other side.

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -225,4 +356,377 @@ jobs:
...
+          kubectl get pod -n arc-systems
+          kubectl describe deployment arc-gha-runner-scale-set-controller -n arc-systems
+
+      - name: Install gha-runner-scale-set
+        id: install_arc
+        run: |
+          echo "Install openebs/dynamic-localpv-provisioner"
+          helm repo add openebs https://openebs.github.io/charts
+          helm repo update
+          helm install openebs openebs/openebs -n openebs --create-namespace
```
Link-, line 433: Is this working for you here? It fails in a codespace:

```
Warning  FailedMount  9s (x10 over 4m19s)  kubelet            MountVolume.SetUp failed for volume "udev" : hostPath type check failed: /run/udev is not a directory
```

The fix is:

```
minikube start --mount-string="/run/udev:/run/udev" --mount
```

test_e2e_arc/arc_jobs_test.go
Link-: Sounds like a good solution

.github/workflows/e2e-test-linux-vm.yaml
TingluoHuang: codespace is already inside a container, so everything around the `mount` become different. 😄

Link- (approved):
🚀  Let's ship it, we can keep improving as we go

PR #2414, opened 2023-03-15
Fix helm chart rendering errors.

TingluoHuang:
Fixes:
- Replace `-}}` with `}}` in the place where trimming trailing whitespace doesn't make sense.
- Make sure to have `toYaml` when rewriting the object back to YAML.
- Adding L0 tests to cover more scenarios.

### DIFF
--- a/charts/gha-runner-scale-set/templates/_helpers.tpl
+++ b/charts/gha-runner-scale-set/templates/_helpers.tpl
@@ -75,19 +75,15 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
 
 {{- define "gha-runner-scale-set.dind-init-container" -}}
-{{- range $i, $val := .Values.template.spec.containers -}}
-{{- if eq $val.name "runner" -}}
+{{- range $i, $val := .Values.template.spec.containers }}
+  {{- if eq $val.name "runner" }}
 image: {{ $val.image }}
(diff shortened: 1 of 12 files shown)


charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -75,12 +75,12 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
 
 {{- define "gha-runner-scale-set.dind-init-container" -}}
-{{- range $i, $val := .Values.template.spec.containers -}}
-{{- if eq $val.name "runner" -}}
+{{- range $i, $val := .Values.template.spec.containers }}
+{{- if eq $val.name "runner" }}
 image: {{ $val.image }}
 {{- if $val.imagePullSecrets }}
 imagePullSecrets:
```
nikola-jokic: I think we have an error here
imagePullSecrets exist on a `pod.spec` level, not on the `pod.spec.containers` level

nikola-jokic: Follow up: It should be safe just to remove it
If anyone wants to add it to the `template.spec`, it will be injected, so we should be fine :relaxed:

nikola-jokic (approved):
LGTM

PR #2416, opened 2023-03-16
Update the values.yaml sample for improved clarity

Link-:
### Context

Moving the template.spec comment up so that it's clearer what's going to be changed. I also moved containerMode up because depending on the value set there, the template.spec will be updated.

### DIFF
--- a/charts/gha-runner-scale-set/values.yaml
+++ b/charts/gha-runner-scale-set/values.yaml
@@ -68,16 +68,20 @@ githubConfigSecret:
 #       key: ca.pem
 #   runnerMountPath: /usr/local/share/ca-certificates/
 
-## template is the PodSpec for each runner Pod
-template:
-  spec:
-    containers:
-    - name: runner
-      image: ghcr.io/actions/actions-runner:latest
-      command: ["/home/runner/run.sh"]
-
 containerMode:
   type: ""  ## type can be set to dind or kubernetes
+  ## the following is required when containerMode.type=kubernetes
+  kubernetesModeWorkVolumeClaim:
+    accessModes: ["ReadWriteOnce"]
+    # For local testing, use https://github.com/openebs/dynamic-localpv-provisioner/blob/develop/docs/quickstart.md to provide dynamic provision volume with storageClassName: openebs-hostpath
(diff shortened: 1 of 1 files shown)


charts/gha-runner-scale-set/values.yaml
```diff
@@ -68,16 +68,20 @@ githubConfigSecret:
...
-  spec:
-    containers:
-    - name: runner
-      image: ghcr.io/actions/actions-runner:latest
-      command: ["/home/runner/run.sh"]
-
 containerMode:
   type: ""  ## type can be set to dind or kubernetes
+  ## the following is required when containerMode.type=kubernetes
+  kubernetesModeWorkVolumeClaim:
```
nikola-jokic: While we are improving, should we comment this out since the default mode is "", this will not have any effect

charts/gha-runner-scale-set/values.yaml
```diff
@@ -68,16 +68,20 @@ githubConfigSecret:
...
+    accessModes: ["ReadWriteOnce"]
+    # For local testing, use https://github.com/openebs/dynamic-localpv-provisioner/blob/develop/docs/quickstart.md to provide dynamic provision volume with storageClassName: openebs-hostpath
+    storageClassName: "dynamic-blob-storage"
+    resources:
+      requests:
+        storage: 1Gi
+
+## template is the PodSpec for each runner Pod
+template:
+  ## template.spec will be modified if you change the container mode
```
nikola-jokic, line 84: If we document the expansion for `dind`, should we also document it for `kubernetes` mode?

charts/gha-runner-scale-set/values.yaml
Link-: Yep, I like this idea

charts/gha-runner-scale-set/values.yaml
Link-: Isn't it documented below it?

nikola-jokic: Oh sorry, I did not see it...

nikola-jokic (approved):
Nice improvement!

TingluoHuang (approved):

PR #2418, opened 2023-03-16
Ignore extra dind container when contaerinMode.type=dind.

TingluoHuang:
When customers provide `contaerinMode.type=dind`, we will skip any container named `dind`.

### DIFF
--- a/.github/workflows/e2e-test-linux-vm.yaml
+++ b/.github/workflows/e2e-test-linux-vm.yaml
@@ -439,7 +439,9 @@ jobs:
           --set githubConfigUrl="https://github.com/${{ steps.resolved_inputs.outputs.TARGET_ORG }}/${{steps.resolved_inputs.outputs.TARGET_REPO}}" \
           --set githubConfigSecret.github_token="${{ steps.setup.outputs.token }}" \
           --set containerMode.type="kubernetes" \
+          --set containerMode.kubernetesModeWorkVolumeClaim.accessModes={"ReadWriteOnce"} \
           --set containerMode.kubernetesModeWorkVolumeClaim.storageClassName="openebs-hostpath" \
+          --set containerMode.kubernetesModeWorkVolumeClaim.resources.requests.storage="1Gi" \
           ./charts/gha-runner-scale-set \
           --debug
           echo "ARC_NAME=$ARC_NAME" >> $GITHUB_OUTPUT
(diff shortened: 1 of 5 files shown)


apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
```diff
@@ -248,7 +248,6 @@ type AutoscalingRunnerSetStatus struct {
 }
 
 func (ars *AutoscalingRunnerSet) ListenerSpecHash() string {
-	type listenerSpec = AutoscalingRunnerSetSpec
```
TingluoHuang, line 251: unused type, golang-linter find this.

charts/gha-runner-scale-set/values.yaml
```diff
@@ -86,11 +86,11 @@ template:
   ## template:
   ##   spec:
   ##     initContainers:
-  ##     - name: initExternalsInternalVolume
```
TingluoHuang, line 89: update the comments to match what gets generated by the helm chart.

nikola-jokic (dismissed):
LGTM

charts/gha-runner-scale-set/values.yaml
nikola-jokic: We should avoid examples with camel case in the future because it is not acceptable by k8s anyway

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -439,7 +439,9 @@ jobs:
           --set githubConfigUrl="https://github.com/${{ steps.resolved_inputs.outputs.TARGET_ORG }}/${{steps.resolved_inputs.outputs.TARGET_REPO}}" \
           --set githubConfigSecret.github_token="${{ steps.setup.outputs.token }}" \
           --set containerMode.type="kubernetes" \
+          --set containerMode.kubernetesModeWorkVolumeClaim.accessModes={"ReadWriteOnce"} \
```
TingluoHuang, line 442: some defaults value get removed from https://github.com/actions/actions-runner-controller/pull/2416/files

nikola-jokic: That is why k8s mode failed without being modified here... Got it!

nikola-jokic (approved):

PR #2423, opened 2023-03-17
Add new workflows

rentziass:
This groups all go related workflows/jobs under one workflow, and it's an attempt at simplifying things a little and remove some of the duplication we had


TingluoHuang (dismissed):

.github/workflows/go.yaml
```diff
@@ -0,0 +1,91 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v4
+        with:
+          go-version-file: 'go.mod'
+          cache: false
+      - name: Install govulncheck
+        run: go install golang.org/x/vuln/cmd/govulncheck@latest
+      - name: Check vulnerabilities
+        run: govulncheck ./...
```
Link-: Why do we need a vulnerability check if we're using GitHub Advanced Security? Does this run against a database we don't support?

rentziass: It just works in a different way. GitHub Advanced Security looks at dependencies, this goes into your code and follows all calls and warns if any of the code you actually use has a vulnerability. And by doing so it also catches things we don't: you'll see we have no advisories currently, but if you try and run this on ARC with go < 1.20.2 you'll get a vulnerability reported (the reason why we upgraded to 1.20 here).

rentziass:
```
Using go1.20.1 and govulncheck@v0.0.0 with
vulnerability data from https://vuln.go.dev/ (last modified 13 Mar 23 20:29 UTC).

Scanning your code and 637 packages across 76 dependent modules for known vulnerabilities...
Your code is affected by 1 vulnerability from the Go standard library.

Vulnerability #1: GO-2023-1621
  The ScalarMult and ScalarBaseMult methods of the P256 Curve may
  return an incorrect result if called with some specific
  unreduced scalars (a scalar larger than the order of the curve).
  This does not impact usages of crypto/ecdsa or crypto/ecdh.

  More info: https://pkg.go.dev/vuln/GO-2023-1621

  Standard library
    Found in: crypto/internal/nistec@go1.20.1
    Fixed in: crypto/internal/nistec@go1.20.2

    Call stacks in your code:
      github/actions/testserver/server.go:94:50: github.com/actions/actions-runner-controller/github/actions/testserver.DefaultActionsToken calls github.com/golang-jwt/jwt/v4.ParseRSAPrivateKeyFromPEM, which eventually calls crypto/internal/nistec.P256Point.ScalarBaseMult
      pkg/githubwebhookdeliveryforwarder/githubwebhookdelivery.go:201:13: github.com/actions/actions-runner-controller/pkg/githubwebhookdeliveryforwarder.server.Errorf calls fmt.Fprintf, which eventually calls crypto/internal/nistec.P256OrdInverse
      pkg/githubwebhookdeliveryforwarder/githubwebhookdelivery.go:201:13: github.com/actions/actions-runner-controller/pkg/githubwebhookdeliveryforwarder.server.Errorf calls fmt.Fprintf, which eventually calls crypto/internal/nistec.P256Point.ScalarMult
```

Link- (dismissed):
Thank you 🚀

Link-:
Can we also remove this workflow as part of this work? https://github.com/actions/actions-runner-controller/blob/master/.github/workflows/golangci-lint.yaml

rentziass:
@Link- sure, it's already gone 👍

nikola-jokic (approved):
LGTM!

Link- (approved):

PR #2427, opened 2023-03-20
Noting type none containerMode

joshjohanning:
When [containerMode](https://github.com/actions/actions-runner-controller/blob/f49d08e4bc6348c0041e2f0f0046058dc76a2ce7/charts/gha-runner-scale-set/tests/values_extra_containers.yaml#L46) is left blank, it defaults to `none`. 

Currently, the values file seems to indicate that it can only be `kubernetes` or `dind`, which is slightly confusing.


Link-:
Thank you @joshjohanning for your contribution. We already addressed this in: https://github.com/actions/actions-runner-controller/pull/2442

I'm closing this PR

joshjohanning:
Nice!

PR #2430, opened 2023-03-20
Update e2e workflow

TingluoHuang:
Changes:
- Move most repeat code behind `.github/actions/setup-arc-e2e` and `.github/actions/execute-assert-arc-e2e`, so extending E2E tests in the future should be easy.
- For workflow run triggered by the E2E test, assert the run is successfully completed.

https://github.com/actions/actions-runner-controller/actions/runs/4475143286

https://github.com/actions-runner-controller/arc_e2e_test_dummy/actions

### DIFF
--- a/.github/actions/execute-assert-arc-e2e/action.yaml
+++ b/.github/actions/execute-assert-arc-e2e/action.yaml
@@ -1,0 +1,160 @@
+name: 'Execute and Assert ARC E2E Test Action'
+description: 'Queue E2E test workflow and assert workflow run result to be succeed'
+
+inputs:
+  auth-token:
+    description: 'GitHub access token to queue workflow run'
+    required: true
+  repo-owner:
+    description: "The repository owner name that has the test workflow file, ex: actions"
+    required: true
+  repo-name:
+    description: "The repository name that has the test workflow file, ex: test"
+    required: true
+  workflow-file:
+    description: 'The file name of the workflow yaml, ex: test.yml'
+    required: true
+  arc-name:
+    description: 'The name of the configured gha-runner-scale-set'
+    required: true
+  arc-namespace:
+    description: 'The namespace of the configured gha-runner-scale-set'
+    required: true
+  arc-controller-namespace:
+    description: 'The namespace of the configured gha-runner-scale-set-controller'
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - name: Queue test workflow
+      shell: bash
+      id: queue_workflow
+      run: |
+        queue_time=`date +%FT%TZ`
+        echo "queue_time=$queue_time" >> $GITHUB_OUTPUT
+        curl -X POST https://api.github.com/repos/${{inputs.repo-owner}}/${{inputs.repo-name}}/actions/workflows/${{inputs.workflow-file}}/dispatches \
+        -H "Accept: application/vnd.github.v3+json" \
+        -H "Authorization: token ${{inputs.auth-token}}" \
+        -d '{"ref": "main", "inputs": { "arc_name": "${{inputs.arc-name}}" } }'
+
+    - name: Fetch workflow run & job ids
+      uses: actions/github-script@v6
+      id: query_workflow
+      with:
+        script: |
+          // Try to find the workflow run triggered by the previous step using the workflow_dispatch event.
+          // - Find recently create workflow runs in the test repository
+          // - For each workflow run, list its workflow job and see if the job's labels contain `inputs.arc-name`
+          // - Since the inputs.arc-name should be unique per e2e workflow run, once we find the job with the label, we find the workflow that we just triggered.
+          function sleep(ms) {
+            return new Promise(resolve => setTimeout(resolve, ms))
+          }
+          const owner = '${{inputs.repo-owner}}'
+          const repo = '${{inputs.repo-name}}'
+          const workflow_id = '${{inputs.workflow-file}}'
+          let workflow_run_id = 0
+          let workflow_job_id = 0
+          let workflow_run_html_url = ""
+          let count = 0
+          while (count++<12) {
+            await sleep(10 * 1000);
+            let listRunResponse = await github.rest.actions.listWorkflowRuns({
+              owner: owner,
+              repo: repo,
+              workflow_id: workflow_id,
+              created: '>${{steps.queue_workflow.outputs.queue_time}}'
+            })
+            if (listRunResponse.data.total_count > 0) {
+              console.log(`Found some new workflow runs for ${workflow_id}`)
+              for (let i = 0; i<listRunResponse.data.total_count; i++) {
+                let workflowRun = listRunResponse.data.workflow_runs[i]
+                console.log(`Check if workflow run ${workflowRun.id} is triggered by us.`)
+                let listJobResponse = await github.rest.actions.listJobsForWorkflowRun({
+                  owner: owner,
+                  repo: repo,
+                  run_id: workflowRun.id
+                })
+                console.log(`Workflow run ${workflowRun.id} has ${listJobResponse.data.total_count} jobs.`)
+                if (listJobResponse.data.total_count > 0) {
+                  for (let j = 0; j<listJobResponse.data.total_count; j++) {
+                    let workflowJob = listJobResponse.data.jobs[j]
+                    console.log(`Check if workflow job ${workflowJob.id} is triggered by us.`)
+                    console.log(JSON.stringify(workflowJob.labels));
+                    if (workflowJob.labels.includes('${{inputs.arc-name}}')) {
+                      console.log(`Workflow job ${workflowJob.id} (Run id: ${workflowJob.run_id}) is triggered by us.`)
+                      workflow_run_id = workflowJob.run_id
+                      workflow_job_id = workflowJob.id
+                      workflow_run_html_url = workflowRun.html_url
+                      break
+                    }
+                  }
+                }
+
+                if (workflow_job_id > 0) {
+                  break;
+                }
+              }
+            }
+
+            if (workflow_job_id > 0) {
+              break;
+            }
+          }
+          if (workflow_job_id == 0) {
+            core.setFailed(`Can't find workflow run and workflow job triggered to 'runs-on ${{inputs.arc-name}}'`)
+          } else {
+            core.setOutput('workflow_run', workflow_run_id);
+            core.setOutput('workflow_job', workflow_job_id);
+            core.setOutput('workflow_run_url', workflow_run_html_url);
+          }
+
+    - name: Generate summary about the triggered workflow run
+      shell: bash
+      run: | 
+        cat <<-EOF > $GITHUB_STEP_SUMMARY
+        | **Triggered workflow run** |
+        |:--------------------------:|
+        | ${{steps.query_workflow.outputs.workflow_run_url}} |
+        EOF
+
+    - name: Wait for workflow to finish successfully
+      uses: actions/github-script@v6
+      with:
+        script: |
+          // Wait 5 minutes and make sure the workflow run we triggered completed with result 'success'
+          function sleep(ms) {
+            return new Promise(resolve => setTimeout(resolve, ms))
+          }
+          const owner = '${{inputs.repo-owner}}'
+          const repo = '${{inputs.repo-name}}'
+          const workflow_run_id = ${{steps.query_workflow.outputs.workflow_run}}
+          const workflow_job_id = ${{steps.query_workflow.outputs.workflow_job}}
+          let count = 0
+          while (count++<10) {
+            await sleep(30 * 1000);
+            let getRunResponse = await github.rest.actions.getWorkflowRun({
+              owner: owner,
+              repo: repo,
+              run_id: workflow_run_id
+            })
+            console.log(`${getRunResponse.data.html_url}: ${getRunResponse.data.status} (${getRunResponse.data.conclusion})`);
+            if (getRunResponse.data.status == 'completed') {
+              if ( getRunResponse.data.conclusion == 'success') {
+                console.log(`Workflow run finished properly.`)
+                return
+              } else {
+                core.setFailed(`The triggered workflow run finish with result ${getRunResponse.data.conclusion}`)
+                return
+              }
+            }
+          }
+          core.setFailed(`The triggered workflow run didn't finish properly using ${{inputs.arc-name}}`)
+
+    - name: Gather logs and cleanup
+      shell: bash
+      if: always()
+      run: |
+        helm uninstall ${{ inputs.arc-name }} --namespace ${{inputs.arc-namespace}} --debug
+        kubectl wait --timeout=10s --for=delete AutoScalingRunnerSet -n ${{inputs.arc-name}} -l app.kubernetes.io/instance=${{ inputs.arc-name }}
+        kubectl logs deployment/arc-gha-runner-scale-set-controller -n ${{inputs.arc-controller-namespace}}
--- a/.github/actions/setup-arc-e2e/action.yaml
+++ b/.github/actions/setup-arc-e2e/action.yaml
@@ -2,20 +2,20 @@ name: 'Setup ARC E2E Test Action'
 description: 'Build controller image, create kind cluster, load the image, and exchange ARC configure token.'
 
 inputs:
-  github-app-id:
+  app-id:
     description: 'GitHub App Id for exchange access token'
     required: true
-  github-app-pk:
+  app-pk:
     description: "GitHub App private key for exchange access token"
     required: true
-  github-app-org:
-    description: 'The organization the GitHub App has installed on'
-    required: true
-  docker-image-name:
+  image-name:
     description: "Local docker image name for building"
     required: true
-  docker-image-tag:
+  image-tag:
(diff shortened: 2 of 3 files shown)


.github/actions/execute-assert-arc-e2e/action.yaml
```diff
@@ -0,0 +1,151 @@
...
+  github-repo-owner:
+    description: "The repository owner name that has the test workflow file, ex: actions"
+    required: true
+  github-repo-name:
+    description: "The repository name that has the test workflow file, ex: test"
+    required: true
+  github-workflow-file:
+    description: 'The file name of the workflow yaml, ex: test.yml'
+    required: true
+  test-arc-name:
```
Link-: nit: is `test` really needed as part of the name? I also feel the name is not very indicative of what this is

.github/actions/execute-assert-arc-e2e/action.yaml
```diff
@@ -0,0 +1,151 @@
...
+  github-auth-token:
+    description: 'GitHub access token to queue workflow run'
+    required: true
+  github-repo-owner:
+    description: "The repository owner name that has the test workflow file, ex: actions"
+    required: true
+  github-repo-name:
+    description: "The repository name that has the test workflow file, ex: test"
+    required: true
+  github-workflow-file:
```
Link-: nit: I'm not really sure why we need `github` as a prefix to any of these

.github/actions/execute-assert-arc-e2e/action.yaml
```diff
@@ -0,0 +1,151 @@
...
+      id: queue_workflow
+      run: |
+        queue_time=`date +%FT%TZ`
+        echo "queue_time=$queue_time" >> $GITHUB_OUTPUT
+        curl -X POST https://api.github.com/repos/${{inputs.github-repo-owner}}/${{inputs.github-repo-name}}/actions/workflows/${{inputs.github-workflow-file}}/dispatches \
+        -H "Accept: application/vnd.github.v3+json" \
+        -H "Authorization: token ${{inputs.github-auth-token}}" \
+        -d '{"ref": "main", "inputs": { "arc_name": "${{inputs.test-arc-name}}" } }'
+
+    - name: Wait test workflow run to start
```
Link-:
```suggestion
    - name: Fetch workflow run & job ids
```

.github/actions/execute-assert-arc-e2e/action.yaml
```diff
@@ -0,0 +1,151 @@
...
+      if: always()
+      run: |
+        helm uninstall ${{ inputs.test-arc-name }} --namespace ${{inputs.test-arc-namespace}} --debug
+        kubectl wait --timeout=10s --for=delete AutoScalingRunnerSet -n ${{inputs.test-arc-name}} -l app.kubernetes.io/instance=${{ inputs.test-arc-name }}
+        kubectl logs deployment/arc-gha-runner-scale-set-controller -n ${{inputs.test-arc-controller-namespace}}
+        cat <<-EOF > $GITHUB_STEP_SUMMARY
+        | **Triggered workflow run** |
+        |:--------------------------:|
+        | ${{steps.query_workflow.outputs.workflow_run_url}} |
+        EOF
```
Link-: Can we split this into its own step? We want the job summary even if some of the commands above fail

.github/actions/execute-assert-arc-e2e/action.yaml
```diff
@@ -0,0 +1,151 @@
...
+        curl -X POST https://api.github.com/repos/${{inputs.github-repo-owner}}/${{inputs.github-repo-name}}/actions/workflows/${{inputs.github-workflow-file}}/dispatches \
+        -H "Accept: application/vnd.github.v3+json" \
+        -H "Authorization: token ${{inputs.github-auth-token}}" \
+        -d '{"ref": "main", "inputs": { "arc_name": "${{inputs.test-arc-name}}" } }'
+
+    - name: Wait test workflow run to start
+      uses: actions/github-script@v6
+      id: query_workflow
+      with:
+        script: |
```
Link-, line 45: nit: Can we add one or two liner explaining the logic behind this script? It'll be easier to understand what's happening at a glance instead of diving into the code

.github/actions/execute-assert-arc-e2e/action.yaml
```diff
@@ -0,0 +1,151 @@
...
+          core.setFailed(`The triggered workflow run didn't finish properly using ${{inputs.test-arc-name}}`)
+
+    - name: Gather logs and cleanup
+      shell: bash
+      if: always()
+      run: |
+        helm uninstall ${{ inputs.test-arc-name }} --namespace ${{inputs.test-arc-namespace}} --debug
+        kubectl wait --timeout=10s --for=delete AutoScalingRunnerSet -n ${{inputs.test-arc-name}} -l app.kubernetes.io/instance=${{ inputs.test-arc-name }}
+        kubectl logs deployment/arc-gha-runner-scale-set-controller -n ${{inputs.test-arc-controller-namespace}}
+        cat <<-EOF > $GITHUB_STEP_SUMMARY
```
Link-: Same comment about separating the summary from the cleanup

.github/actions/execute-assert-arc-e2e/action.yaml
```diff
@@ -0,0 +1,151 @@
+name: 'Execute and Assert ARC E2E Test Action'
+description: 'Queue E2E test workflow and assert workflow run result to be succeed'
+
+inputs:
+  github-auth-token:
```
Link-: Why does this action expect a token while the other expects an app id and a pk? Can we standardise the input?

.github/actions/setup-arc-e2e/action.yaml
```diff
@@ -8,24 +8,54 @@ inputs:
...
+    required: true
+  github-target-repo-default:
+    description: "The default test repository for ARC e2e test"
+    required: true
+  github-target-org-overwrite:
+    description: "Overwrite the default test organization for ARC e2e test"
+    required: true
+  github-target-repo-overwrite:
+    description: "Overwrite the default test repository for ARC e2e test"
+    required: true
```
Link-: 😄 I'm not sure what's happening here.. Is this the best way forward?

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -23,7 +23,7 @@ env:
   TARGET_REPO: arc_e2e_test_dummy
   IMAGE_NAME: "arc-test-image"
   IMAGE_VERSION: "dev"
-  
+
```
Link-, line 20: I've been thinking, sometimes we will want to run a single test but not the others. Can we add for each test boolean input that is "True" by default? This way, if we want to have a custom run with a subset of the tests, we have the option to do it

Link- (changes requested):

.github/actions/execute-assert-arc-e2e/action.yaml
TingluoHuang: I copied from the setup-arc-e2e action. 😄 
I can remove it if you want.

.github/actions/execute-assert-arc-e2e/action.yaml
TingluoHuang: it's the name of the test arc we set up, any suggestions on the name? just `arc-name`?

.github/actions/execute-assert-arc-e2e/action.yaml
TingluoHuang: the setup action takes the appId/pk and outputs a auth token, which gets used for the rest of the workflow.
Do you want me to make this action also take appid/pk and generate another token it can use?

.github/workflows/e2e-test-linux-vm.yaml
TingluoHuang: These tests are running in parallel and they are all critical scenarios for ARC, do you have any case that we only want to run part of it?

.github/actions/setup-arc-e2e/action.yaml
TingluoHuang: if we want to somehow merge inputs... or we just don't allow points to a different test repo, since pointing to different repo most likely doesn't work (you don't have the APP installed).

.github/actions/execute-assert-arc-e2e/action.yaml
Link-: I'll leave it up to you, it's not adding any value to be honest

.github/actions/execute-assert-arc-e2e/action.yaml
Link-: I think having the workflow determine which authentication method to use is better. Our actions can just use a token, without necessarily know whether it's a GITHUB_TOKEN, a PAT or an app installation token. WDYT?

.github/workflows/e2e-test-linux-vm.yaml
Link-: Correct and for PRs that option will not be available, they'll all run by default. The inputs will be handy If this e2e workflow is triggered manually. What if we want to troubleshoot a test? Like the Kubernetes mode tests for example, it'll be handy to be able to trigger only one of the bunch

.github/actions/execute-assert-arc-e2e/action.yaml
Link-: I was trying to think of a better name but everything goes back to `runner-scale-set-name` which long and doesn't say much. I think `arc-name` is fine or in our quick start guide we used: `installation_name` 🤷‍♂️

.github/actions/execute-assert-arc-e2e/action.yaml
TingluoHuang: will remove them.

.github/actions/execute-assert-arc-e2e/action.yaml
TingluoHuang: i will go with `arc-name`

.github/workflows/e2e-test-linux-vm.yaml
TingluoHuang: I want to avoid extra inputs and complicate `if` conditions to make the workflow more unreadable.
Since these jobs do not depend on each other, those extra jobs won't affect your ability to debug a certain job.

.github/actions/execute-assert-arc-e2e/action.yaml
TingluoHuang: that would mean every job need to have an extra step to use `peter-murray/workflow-application-token-action@8e1ba3bf1619726336414f1014e37f17fbadf1db` to fetch the token first.

I am trying to make each job's YAML as tiny as possible and only have diff around `helm install arc`

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -32,28 +23,14 @@ jobs:
...
-
       - uses: ./.github/actions/setup-arc-e2e
         id: setup
         with:
-          github-app-id: ${{secrets.ACTIONS_ACCESS_APP_ID}}
-          github-app-pk: ${{secrets.ACTIONS_ACCESS_PK}}
-          github-app-org: ${{steps.resolved_inputs.outputs.TARGET_ORG}}
-          docker-image-name: ${{env.IMAGE_NAME}}
-          docker-image-tag: ${{env.IMAGE_VERSION}}
+          app-id: ${{secrets.ACTIONS_ACCESS_APP_ID}}
```
Link-, line 29:
```suggestion
          app-id: ${{secrets. E2E_TESTS_ACCESS_APP_ID}}
```

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -32,28 +23,14 @@ jobs:
...
+          app-pk: ${{secrets.ACTIONS_ACCESS_PK}}
```
Link-, line 30:
```suggestion
          app-pk: ${{secrets. E2E_TESTS_ACCESS_PK}}
```

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -145,28 +106,14 @@ jobs:
...
-
       - uses: ./.github/actions/setup-arc-e2e
         id: setup
         with:
-          github-app-id: ${{secrets.ACTIONS_ACCESS_APP_ID}}
-          github-app-pk: ${{secrets.ACTIONS_ACCESS_PK}}
-          github-app-org: ${{steps.resolved_inputs.outputs.TARGET_ORG}}
-          docker-image-name: ${{env.IMAGE_NAME}}
-          docker-image-tag: ${{env.IMAGE_VERSION}}
+          app-id: ${{secrets.ACTIONS_ACCESS_APP_ID}}
```
Link-:
```suggestion
          app-id: ${{secrets. E2E_TESTS_ACCESS_APP_ID}}
```

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -145,28 +106,14 @@ jobs:
...
+          app-pk: ${{secrets.ACTIONS_ACCESS_PK}}
```
Link-:
```suggestion
          app-pk: ${{secrets. E2E_TESTS_ACCESS_PK}}
```

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -260,28 +191,14 @@ jobs:
...
-
       - uses: ./.github/actions/setup-arc-e2e
         id: setup
         with:
-          github-app-id: ${{secrets.ACTIONS_ACCESS_APP_ID}}
-          github-app-pk: ${{secrets.ACTIONS_ACCESS_PK}}
-          github-app-org: ${{steps.resolved_inputs.outputs.TARGET_ORG}}
-          docker-image-name: ${{env.IMAGE_NAME}}
-          docker-image-tag: ${{env.IMAGE_VERSION}}
+          app-id: ${{secrets.ACTIONS_ACCESS_APP_ID}}
```
Link-:
```suggestion
          app-id: ${{secrets.E2E_TESTS_ACCESS_APP_ID}}
```

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -260,28 +191,14 @@ jobs:
...
+          app-pk: ${{secrets.ACTIONS_ACCESS_PK}}
```
Link-:
```suggestion
          app-pk: ${{secrets.E2E_TESTS_ACCESS_PK}}
```

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -374,28 +275,14 @@ jobs:
...
-
       - uses: ./.github/actions/setup-arc-e2e
         id: setup
         with:
-          github-app-id: ${{secrets.ACTIONS_ACCESS_APP_ID}}
-          github-app-pk: ${{secrets.ACTIONS_ACCESS_PK}}
-          github-app-org: ${{steps.resolved_inputs.outputs.TARGET_ORG}}
-          docker-image-name: ${{env.IMAGE_NAME}}
-          docker-image-tag: ${{env.IMAGE_VERSION}}
+          app-id: ${{secrets.ACTIONS_ACCESS_APP_ID}}
```
Link-, line 281:
```suggestion
          app-id: ${{secrets.E2E_TESTS_ACCESS_APP_ID}}
```

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -374,28 +275,14 @@ jobs:
...
+          app-pk: ${{secrets.ACTIONS_ACCESS_PK}}
```
Link-, line 282:
```suggestion
          app-pk: ${{secrets.E2E_TESTS_ACCESS_PK}}
```

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -496,28 +367,14 @@ jobs:
...
```
Link-:
```suggestion
          app-id: ${{secrets.E2E_TESTS_ACCESS_APP_ID}}
          app-pk: ${{secrets.E2E_TESTS_ACCESS_PK}}
```

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -621,28 +462,14 @@ jobs:
...
```
Link-:
```suggestion
          app-id: ${{secrets.E2E_TESTS_ACCESS_APP_ID}}
          app-pk: ${{secrets.E2E_TESTS_ACCESS_PK}}
```

.github/workflows/e2e-test-linux-vm.yaml
Link-:
```suggestion
          app-id: ${{secrets.E2E_TESTS_ACCESS_APP_ID}}
          app-pk: ${{secrets.E2E_TESTS_ACCESS_PK}}
```

.github/workflows/e2e-test-linux-vm.yaml
Link-:
```suggestion
          app-id: ${{secrets.E2E_TESTS_ACCESS_APP_ID}}
          app-pk: ${{secrets.E2E_TESTS_ACCESS_PK}}
```

Link- (approved):
🚀

PR #2433, opened 2023-03-22
Fix helm uninstall cleanup by adding finalizers and cleaning them from the controller

nikola-jokic:
To lower the amount of permissions required for the manager to operate, manager role binding is introduced when the autoscaling runner set is applied.

But when doing helm uninstall, role and role bindings are deleted usually before the `AutoscalingRunnerSet` can clean up resources, causing errors.

This change adds finalizers to resources applied by helm, and extends the manager cluster role, so it can clean up Role in the last step of clean-up.

We added finalizers to the following resources in `gha-runner-scale-set` in AutoscalingRunnerSet namespace:
- Manager Role
- Manager RoleBinding
- Kubernetes mode Role
- Kubernetes mode RoleBinding
- Kubernetes mode ServiceAccount
- No permission ServiceAccount

Finalizer name is changed for the GitHub secret in `gha-runner-scale-set` from `actions.github.com/secret-protection` to `actions.github.com/cleanup-protection`

The order in which resources are being patched to remove finalizers:
1. Kubernetes mode RoleBinding
1. Kubernetes mode Role
1. Kubernetes mode ServiceAccount
1. No permission ServiceAccount 
1. GitHub secret
1. Manager RoleBinding
1. Manager Role

Note: only resources that are specified in `AutoscalingRunnerSet.ObjectMeta.Annotations` are going to be patched.


controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -484,6 +656,13 @@ func (r *AutoscalingRunnerSetReconciler) deleteRunnerScaleSet(ctx context.Contex
 		return err
 	}
 
+	err = patch(ctx, r.Client, autoscalingRunnerSet, func(obj *v1alpha1.AutoscalingRunnerSet) {
```
nikola-jokic, line 544: Had to clean this up now that we clear permissions. We get client from secret, and we keep reconciling which can return an error trying to fetch the client.
So once the scale set is deleted, we remove the annotation signalling this does not have to be done any more

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -306,6 +320,160 @@ func (r *AutoscalingRunnerSetReconciler) deleteEphemeralRunnerSets(ctx context.C
...
+		requeue, err := r.cleanupContainerModeKubernetesPermissions(ctx, autoscalingRunnerSet, logger)
+		if err != nil {
+			return true, err
+		}
+		if requeue {
+			return true, nil
+		}
+	}
+
+	secret := new(corev1.Secret)
```
nikola-jokic: This one is a best effort.
If githubSecret is used as a string, we won't be able to fetch it by name
I can improve this by querying by label, but not sure if that is something that we need to do (if that is the case, the secret is already created before, so not sure if we should even delete it)

charts/gha-runner-scale-set-controller/templates/manager_cluster_role.yaml
```diff
@@ -126,6 +126,9 @@ rules:
   verbs:
   - list
   - watch
+  - get
+  - delete
+  - patch
```
TingluoHuang: why do we need these cluster-wide permission?

nikola-jokic: The role binding is the last one to get deleted
Once we get to that step, we don't have enough permissions to remove it (We removed Role in the step before)
Same problem would be if we delete first the role binding, then we won't be able to delete a role

It seemed more secure to be able to get/patch/delete role bindings than roles, so I picked that one

charts/gha-runner-scale-set/templates/autoscalingrunnerset.yaml
```diff
@@ -11,6 +11,11 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "gha-runner-scale-set.labels" . | nindent 4 }}
+  annotations:
```
TingluoHuang, line 15: should we just put all resource names with the cleanup finalizer as annotations, so you don't need to re-construct the name in the resourceBuilder?

charts/gha-runner-scale-set/templates/githubsecret.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
   labels:
     {{- include "gha-runner-scale-set.labels" . | nindent 4 }}
   finalizers:
-    - actions.github.com/secret-protection
+    - actions.github.com/cleanup-protection
```
TingluoHuang, line 10: what about `actions.github.com/defer-deletion`, a more golang style name? 😄

charts/gha-runner-scale-set-controller/templates/manager_cluster_role.yaml
TingluoHuang: do we need permission on the `rolebindings`? or a subresource `rolebindings/finalizers` will be enough?

charts/gha-runner-scale-set/templates/githubsecret.yaml
nikola-jokic: As much as I'd love to use go terminology, example from docs provide a common example being `kubernetes.io/pv-protection` so I'd prefer to stick with that :smile:

charts/gha-runner-scale-set-controller/templates/manager_cluster_role.yaml
nikola-jokic: Sadly no... I tried but since the finalizer is part of the `ObjectMeta`, we need patch on the actual resource in order to do this operation

charts/gha-runner-scale-set/templates/autoscalingrunnerset.yaml
nikola-jokic: I definitely like this suggestion! I'll work on it!

I thought a lot about how this would impact someone who prefer to use kustomize for example, but I think it won't be a big problem while it can eliminate naming dependency
Awesome suggestion!

charts/gha-runner-scale-set/templates/autoscalingrunnerset.yaml
```diff
@@ -12,6 +12,21 @@ metadata:
...
+    {{- $containerMode := .Values.containerMode }}
+    {{- if (and $containerMode $containerMode.type) }}
+    actions.github.com/mode: {{ $containerMode.type }}
+    {{- end }}
+    {{- if not (kindIs "string" .Values.githubConfigSecret) }}
+    actions.github.com/cleanup-github-secret-name: {{ include "gha-runner-scale-set.githubsecret" . }}
+    {{- end }}
+    {{- if and $containerMode (eq $containerMode.type "kubernetes") (not .Values.template.spec.serviceAccountName) }}
+    actions.github.com/cleanup-manager-role-binding: {{ include "gha-runner-scale-set.managerRoleBindingName" . }}
+    actions.github.com/cleanup-manager-role-name: {{ include "gha-runner-scale-set.managerRoleName" . }}
```
TingluoHuang, line 21: should the manager-role and manager-role-binding always get cleanup regardless of the container mode?

charts/gha-runner-scale-set/templates/manager_role.yaml
TingluoHuang, line 1: do we need to do anything for those single_namespace_role/rolebinding?

charts/gha-runner-scale-set/templates/autoscalingrunnerset.yaml
nikola-jokic: Yes, good catch! My mistake when I wanted to refactor, I accidentally pulled it in :disappointed:

charts/gha-runner-scale-set/templates/kube_mode_serviceaccount.yaml
TingluoHuang, line 1: what about the `no_permission_serviceaccount.yaml`?
My naive thought was we need to add finializer to all resources we created as part of `gha-runner-scale-set`, and remove those finializer in a reasonable order during cleanup.

nikola-jokic: So the reason why I did not add no-permission service account is because it really does not have any permissions
So `helm uninstall` will clean it up, but since we did not gain any permissions by it, it should be okay for it to be cleaned up without the controller having to worry about it
That was my thought process, but please correct me if I am wrong

charts/gha-runner-scale-set/templates/manager_role.yaml
nikola-jokic: Added to the [charts/gha-runner-scale-set-controller/templates/manager_single_namespace_watch_role.yaml](https://github.com/actions/actions-runner-controller/pull/2433/files#diff-02543eb6ef7a43c83200a5e15d70f2d292661ab94e0561e69daaec42075f50e0)

charts/gha-runner-scale-set/templates/kube_mode_serviceaccount.yaml
TingluoHuang: the service account might be used by a pod, it feels safe to keep the service account until all the pod is cleaned up first.

nikola-jokic: Added finalizer to the no-permission service account, added to tests and tested manually.
Also, updated PR description :smile:

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -113,6 +116,16 @@ func (r *AutoscalingRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl
 			return ctrl.Result{}, err
 		}
 
+		requeue, err := r.cleanupPermissions(ctx, autoscalingRunnerSet, log)
```
TingluoHuang: maybe use `done, err` just like others in this method?

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -113,6 +116,16 @@ func (r *AutoscalingRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl
...
+		if err != nil {
+			log.Error(err, "Failed to clean up permissions")
+			return ctrl.Result{}, nil
+		}
+
+		if requeue {
+			return ctrl.Result{Requeue: true}, nil
```
TingluoHuang, line 126: add an extra trace for something `log.Info("Waiting for role/rolebinding/serviceaccount to be deleted")`?

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -467,8 +723,12 @@ func (r *AutoscalingRunnerSetReconciler) updateRunnerScaleSetName(ctx context.Co
 }
 
 func (r *AutoscalingRunnerSetReconciler) deleteRunnerScaleSet(ctx context.Context, autoscalingRunnerSet *v1alpha1.AutoscalingRunnerSet, logger logr.Logger) error {
+	scaleSetId, ok := autoscalingRunnerSet.Annotations[runnerScaleSetIdAnnotationKey]
+	if !ok {
+		return nil
```
TingluoHuang, line 520: you might want to double-check and make sure we follow the behavior comment on line 733

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -489,6 +749,13 @@ func (r *AutoscalingRunnerSetReconciler) deleteRunnerScaleSet(ctx context.Contex
 		return err
 	}
 
+	err = patch(ctx, r.Client, autoscalingRunnerSet, func(obj *v1alpha1.AutoscalingRunnerSet) {
+		delete(obj.Annotations, runnerScaleSetIdAnnotationKey)
+	})
+	if err != nil {
+		return err
```
TingluoHuang, line 549: let's add a trace to indicate what is erroring out.

controllers/actions.github.com/autoscalingrunnerset_controller.go
TingluoHuang:
```
// If the annotation is not set correctly, or if it does not exist, we are going to get stuck in a loop trying to parse the scale set id.
		// If the configuration is invalid (secret does not exist for example), we never get to the point to create runner set. But then, manual cleanup
		// would get stuck finalizing the resource trying to parse annotation indefinitely
```

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -305,6 +318,249 @@ func (r *AutoscalingRunnerSetReconciler) deleteEphemeralRunnerSets(ctx context.C
...
+func (r *AutoscalingRunnerSetReconciler) cleanupContainerModeKubernetesPermissions(ctx context.Context, autoscalingRunnerSet *v1alpha1.AutoscalingRunnerSet, logger logr.Logger) (requeue bool, err error) {
+	if roleBindingName, ok := autoscalingRunnerSet.Annotations[AnnotationKeyKubernetesModeRoleBindingName]; ok {
+		logger.Info("Cleaning up container mode kubernetes role binding")
+
+		roleBinding := new(rbacv1.RoleBinding)
+		err = r.Get(ctx, types.NamespacedName{Name: roleBindingName, Namespace: autoscalingRunnerSet.Namespace}, roleBinding)
+		switch {
+		case err == nil:
+			if roleBinding.DeletionTimestamp.IsZero() {
+				err := r.Delete(ctx, roleBinding)
```
TingluoHuang: do we need to delete it? or do we just need to remove the finalizer?

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -305,6 +318,249 @@ func (r *AutoscalingRunnerSetReconciler) deleteEphemeralRunnerSets(ctx context.C
...
+
+	if roleName, ok := autoscalingRunnerSet.Annotations[AnnotationKeyKubernetesModeRoleName]; ok {
+		logger.Info("Cleaning up container mode kubernetes role")
+
+		role := new(rbacv1.Role)
+		err = r.Get(ctx, types.NamespacedName{Name: roleName, Namespace: autoscalingRunnerSet.Namespace}, role)
+		switch {
+		case err == nil:
+			if role.DeletionTimestamp.IsZero() {
+				err := r.Delete(ctx, role)
```
TingluoHuang: same, do we need to delete them or just remove the finalizer?

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -305,6 +318,249 @@ func (r *AutoscalingRunnerSetReconciler) deleteEphemeralRunnerSets(ctx context.C
...
+
+	if serviceAccountName, ok := autoscalingRunnerSet.Annotations[AnnotationKeyKubernetesModeServiceAccountName]; ok {
+		logger.Info("Cleaning up container mode kubernetes service account")
+
+		serviceAccount := new(corev1.ServiceAccount)
+		err = r.Get(ctx, types.NamespacedName{Name: serviceAccountName, Namespace: autoscalingRunnerSet.Namespace}, serviceAccount)
+		switch {
+		case err == nil:
+			if serviceAccount.DeletionTimestamp.IsZero() {
+				err := r.Delete(ctx, serviceAccount)
```
TingluoHuang: same about deleting or removing the finalizer.

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -305,6 +318,249 @@ func (r *AutoscalingRunnerSetReconciler) deleteEphemeralRunnerSets(ctx context.C
...
+		}
+	default:
+		if serviceAccountName, ok := autoscalingRunnerSet.Annotations[AnnotationKeyNoPermissionServiceAccountName]; ok {
+			logger.Info("Cleaning up no permission service account")
+			serviceAccount := new(corev1.ServiceAccount)
+			err = r.Get(ctx, types.NamespacedName{Name: serviceAccountName, Namespace: autoscalingRunnerSet.Namespace}, serviceAccount)
+			switch {
+			case err == nil:
+				if serviceAccount.DeletionTimestamp.IsZero() {
+					err := r.Delete(ctx, serviceAccount)
```
TingluoHuang: delete or remove finalizer

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -305,6 +318,249 @@ func (r *AutoscalingRunnerSetReconciler) deleteEphemeralRunnerSets(ctx context.C
...
+
+	if githubSecretName, ok := autoscalingRunnerSet.Annotations[AnnotationKeyGitHubSecretName]; ok {
+		logger.Info("Cleaning up GitHub secret")
+
+		secret := new(corev1.Secret)
+		err = r.Get(ctx, types.NamespacedName{Name: githubSecretName, Namespace: autoscalingRunnerSet.Namespace}, secret)
+		switch {
+		case err == nil:
+			if secret.DeletionTimestamp.IsZero() {
+				err := r.Delete(ctx, secret)
```
TingluoHuang: same.

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -305,6 +318,249 @@ func (r *AutoscalingRunnerSetReconciler) deleteEphemeralRunnerSets(ctx context.C
...
+
+	if managerRoleName, ok := autoscalingRunnerSet.Annotations[AnnotationKeyManagerRoleName]; ok {
+		logger.Info("Cleaning up manager role")
+
+		role := new(rbacv1.Role)
+		err = r.Get(ctx, types.NamespacedName{Name: managerRoleName, Namespace: autoscalingRunnerSet.Namespace}, role)
+		switch {
+		case err == nil:
+			if role.DeletionTimestamp.IsZero() {
+				err := r.Delete(ctx, role)
```
TingluoHuang: same

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -305,6 +318,249 @@ func (r *AutoscalingRunnerSetReconciler) deleteEphemeralRunnerSets(ctx context.C
...
+
+	if managerRoleBindingName, ok := autoscalingRunnerSet.Annotations[AnnotationKeyManagerRoleBindingName]; ok {
+		logger.Info("Cleaning up manager role binding")
+
+		roleBinding := new(rbacv1.RoleBinding)
+		err = r.Get(ctx, types.NamespacedName{Name: managerRoleBindingName, Namespace: autoscalingRunnerSet.Namespace}, roleBinding)
+		switch {
+		case err == nil:
+			if roleBinding.DeletionTimestamp.IsZero() {
+				err := r.Delete(ctx, roleBinding)
```
TingluoHuang: same

charts/gha-runner-scale-set-controller/templates/manager_cluster_role.yaml
TingluoHuang: Should we flip the order to remove finalizer on the rolebinding first and then the role?
also, we might only need the `patch` permission on `role` then, I think?

charts/gha-runner-scale-set/templates/manager_role.yaml
```diff
@@ -22,6 +30,7 @@ rules:
   - ""
   resources:
   - secrets
+  - serviceaccounts
```
TingluoHuang, line 40: may be duplicate the section only for serviceaccounts?
like we have 2 sections for role and rolebinding.

controllers/actions.github.com/autoscalingrunnerset_controller.go
nikola-jokic: I thought to be more explicit that we need to requeue
Done is signalling if we should reconcile again, while requeue is signalling that we need to requeue the same request.

Since the behaviour is different, one would require explicit requeue while another will wait for update to trigger reconcile, I thought to separate it to be more explicit

controllers/actions.github.com/autoscalingrunnerset_controller.go
nikola-jokic: Since we are still operating on the resource while we have permissions, I thought it would be safer to catch the case deletion timestamp is zero.
This case should never occur while doing the `helm uninstall`, but if anyone chooses to use other method, we can still safely do it as long as finalizers are applied

nikola-jokic: But also, what should we do if we get to this point and the deletion timestamp is zero? Removing finalizer is of no use since the resource will still exist. And since we know that this `if` will never evaluate to true if used by supported helm charts, it does not hurt to have extra logic here. At least, if someone is using a different way of applying our resources, it can signal that the permission is missing to clean up.

TingluoHuang: My thought is whoever creates the resource is responsible for its deletion. 
Removing the finalizer basic means the AutoScalingRunnerSet gets deleted no longer taking a lock on those resources.

charts/gha-runner-scale-set/templates/autoscalingrunnerset.yaml
```diff
@@ -12,6 +12,24 @@ metadata:
   labels:
     app.kubernetes.io/component: "autoscaling-runner-set"
     {{- include "gha-runner-scale-set.labels" . | nindent 4 }}
+  annotations:
+    {{- $containerMode := .Values.containerMode }}
+    {{- if (and $containerMode $containerMode.type) }}
+    actions.github.com/mode: {{ $containerMode.type }}
```
TingluoHuang: maybe reduce `mode`?

controllers/actions.github.com/autoscalingrunnerset_controller.go
nikola-jokic: Oh I see, removing the delete and changing the name to explain the call correctly :relaxed:

controllers/actions.github.com/autoscalingrunnerset_controller.go
nikola-jokic: Updated the doc comment and added further explanation how it should work :relaxed:

charts/gha-runner-scale-set/templates/manager_role.yaml
```diff
@@ -3,6 +3,14 @@ kind: Role
 metadata:
   name: {{ include "gha-runner-scale-set.managerRoleName" . }}
   namespace: {{ .Release.Namespace }}
+  labels:
```
TingluoHuang, line 6: reuse/extend `{{- include "gha-runner-scale-set.labels" . | nindent 4 }}` if possible?

charts/gha-runner-scale-set/templates/manager_role_binding.yaml
```diff
@@ -1,13 +1,21 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
-  name: {{ include "gha-runner-scale-set.managerRoleBinding" . }}
+  name: {{ include "gha-runner-scale-set.managerRoleBindingName" . }}
   namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/part-of: gha-runner-scale-set-controller
```
TingluoHuang: same about using `{{- include "gha-runner-scale-set.labels" . | nindent 4 }}`?

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -42,11 +43,12 @@ import (
...
-	LabelKeyRunnerSpecHash            = "runner-spec-hash"
-	autoscalingRunnerSetFinalizerName = "autoscalingrunnerset.actions.github.com/finalizer"
-	runnerScaleSetIdAnnotationKey     = "runner-scale-set-id"
-	runnerScaleSetNameAnnotationKey   = "runner-scale-set-name"
+	autoscalingRunnerSetOwnerKey              = ".metadata.controller"
+	LabelKeyRunnerSpecHash                    = "runner-spec-hash"
+	autoscalingRunnerSetFinalizerName         = "autoscalingrunnerset.actions.github.com/finalizer"
+	runnerScaleSetIdAnnotationKey             = "runner-scale-set-id"
+	runnerScaleSetNameAnnotationKey           = "runner-scale-set-name"
+	autoscalingRunnerSetCleanupFinalizerLabel = "actions.github.com/cleanup-protection"
```
TingluoHuang: `autoscalingRunnerSetCleanupFinalizerLabel` or `autoscalingRunnerSetCleanupFinalizerName`

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -113,6 +115,17 @@ func (r *AutoscalingRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl
 			return ctrl.Result{}, err
 		}
 
+		requeue, err := r.removeFinalizersFromPermissionResources(ctx, autoscalingRunnerSet, log)
+		if err != nil {
+			log.Error(err, "Failed to clean up permissions")
```
TingluoHuang:
```suggestion
			log.Error(err, "Failed to remove finalizers on dependent resources")
```

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -113,6 +115,17 @@ func (r *AutoscalingRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl
 			return ctrl.Result{}, err
 		}
 
+		requeue, err := r.removeFinalizersFromPermissionResources(ctx, autoscalingRunnerSet, log)
```
TingluoHuang: `removeFinalizersFromPermissionResources`-> `removeFinalizersFromDependentResources`?

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -113,6 +115,17 @@ func (r *AutoscalingRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl
...
+		if err != nil {
+			log.Error(err, "Failed to clean up permissions")
+			return ctrl.Result{}, nil
+		}
+
+		if requeue {
+			log.Info("Waiting for role/rolebinding/serviceaccount to be deleted")
```
TingluoHuang:
```suggestion
			log.Info("Waiting for dependent resources to be deleted")
```

if you change the method name. 😄

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -305,6 +318,203 @@ func (r *AutoscalingRunnerSetReconciler) deleteEphemeralRunnerSets(ctx context.C
 	return nil
 }
 
+func (r *AutoscalingRunnerSetReconciler) removeFinalizersFromPermissionResources(ctx context.Context, autoscalingRunnerSet *v1alpha1.AutoscalingRunnerSet, logger logr.Logger) (requeue bool, err error) {
+	requeue, err = r.removeFinalizersContainerModeKubernetesPermissions(ctx, autoscalingRunnerSet, logger)
+	if err != nil {
+		return true, err
```
TingluoHuang: let's add some trace before each return, so we know which code path it runs during debug.

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -305,6 +318,203 @@ func (r *AutoscalingRunnerSetReconciler) deleteEphemeralRunnerSets(ctx context.C
...
+			}
+			err = patch(ctx, r.Client, serviceAccount, func(obj *corev1.ServiceAccount) {
+				controllerutil.RemoveFinalizer(obj, autoscalingRunnerSetCleanupFinalizerLabel)
+			})
+			return true, err
+		case err != nil && !kerrors.IsNotFound(err):
+			return true, err
+		}
+		logger.Info("Removed finalizer from no permission service account", "name", serviceAccountName)
+	} else {
```
TingluoHuang: should we refactor the code to remove `else`, if possible?

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -467,12 +677,28 @@ func (r *AutoscalingRunnerSetReconciler) updateRunnerScaleSetName(ctx context.Co
...
+	runnerScaleSetId, err := strconv.Atoi(scaleSetId)
 	if err != nil {
-		// If the annotation is not set correctly, or if it does not exist, we are going to get stuck in a loop trying to parse the scale set id.
-		// If the configuration is invalid (secret does not exist for example), we never get to the point to create runner set. But then, manual cleanup
-		// would get stuck finalizing the resource trying to parse annotation indefinitely
+		// If the annotation is not set correctly, we are going to get stuck in a loop trying to parse the scale set id.
+		// If the configuration is invalid (secret does not exist for example), we never got to the point to create runner set.
+		// But then, manual cleanup would get stuck finalizing the resource trying to parse annotation indefinitely
 		logger.Info("autoscaling runner set does not have annotation describing scale set id. Skip deletion", "err", err.Error())
 		return nil
```
TingluoHuang, line 529: do we want to return nil to err? since we fail to convert string to int for `scaleSetId`?

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -305,6 +318,203 @@ func (r *AutoscalingRunnerSetReconciler) deleteEphemeralRunnerSets(ctx context.C
 	return nil
 }
 
+func (r *AutoscalingRunnerSetReconciler) removeFinalizersFromPermissionResources(ctx context.Context, autoscalingRunnerSet *v1alpha1.AutoscalingRunnerSet, logger logr.Logger) (requeue bool, err error) {
+	requeue, err = r.removeFinalizersContainerModeKubernetesPermissions(ctx, autoscalingRunnerSet, logger)
```
TingluoHuang: it's a little weird that we abstract resources related to kube container mode into `removeFinalizersContainerModeKubernetesPermissions` and leave everything else as a flat list.

Is it possible to refactor out a common method that takes 3 parameters `remove FinalizerFromResource(resourceType sting,  resource object, annotationName string) (bool, err)`?
Then we can just use it for all resources we try to remove finalizers.

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -305,6 +318,203 @@ func (r *AutoscalingRunnerSetReconciler) deleteEphemeralRunnerSets(ctx context.C
...
+			return true, err
+		}
+		logger.Info("Removed finalizer from GitHub secret", "name", githubSecretName)
+	} else {
+		logger.Info(
+			"Skipping cleaning up GitHub secret",
+			"reason",
+			fmt.Sprintf("annotation key %q not present", AnnotationKeyGitHubSecretName),
+		)
+	}
```
TingluoHuang: it would nice to refactor this into some generic common method since we are using the same pattern to remove finalizers for all resources.

controllers/actions.github.com/autoscalingrunnerset_controller.go
nikola-jokic: Yes, here ideally we should return an error and log it as error
But from here we can't really recover unless someone manually fixes the scaleset id, so we will be stuck reconciling on error. That is the only thing why I'm not sure if we should return an error
Logging it as error yes, we should change level from Info

controllers/actions.github.com/autoscalingrunnerset_controller.go
nikola-jokic: Yes, my mistake :smile:

controllers/actions.github.com/autoscalingrunnerset_controller.go
nikola-jokic: With this new cleanup, I added error message that is descriptive so we can know where the failure is happening. Good call

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -113,6 +115,17 @@ func (r *AutoscalingRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl
 			return ctrl.Result{}, err
 		}
 
+		requeue, err := r.removeFinalizersFromDependentResources(ctx, autoscalingRunnerSet, log)
+		if err != nil {
+			log.Error(err, "Failed to remove finalizers on dependent resources")
+			return ctrl.Result{}, nil
```
TingluoHuang:
```suggestion
			return ctrl.Result{}, err
```

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -658,6 +718,307 @@ func (r *AutoscalingRunnerSetReconciler) SetupWithManager(mgr ctrl.Manager) erro
...
+			c.err = fmt.Errorf("failed to patch kubernetes mode role binding without finalizer: %w", err)
+			return
+		}
+		c.requeue = true
+		return
+	case err != nil && !kerrors.IsNotFound(err):
+		c.err = fmt.Errorf("failed to fetch kubernetes mode role binding: %w", err)
+		return
+	default:
+		c.logger.Info("Removed finalizer from container mode kubernetes role binding", "name", roleBindingName)
```
TingluoHuang: should the trace be `c.logger.Info("Container mode kubernetes role binding is already deleted", "name", roleBindingName)`?

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -658,6 +718,307 @@ func (r *AutoscalingRunnerSetReconciler) SetupWithManager(mgr ctrl.Manager) erro
...
+			c.err = fmt.Errorf("failed to patch kubernetes mode role without finalizer: %w", err)
+			return
+		}
+		c.requeue = true
+		return
+	case err != nil && !kerrors.IsNotFound(err):
+		c.err = fmt.Errorf("failed to fetch kubernetes mode role: %w", err)
+		return
+	default:
+		c.logger.Info("Removed finalizer from container mode kubernetes role")
```
TingluoHuang: same comment about the trace

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -658,6 +718,307 @@ func (r *AutoscalingRunnerSetReconciler) SetupWithManager(mgr ctrl.Manager) erro
...
+			c.err = fmt.Errorf("failed to patch kubernetes mode service account without finalizer: %w", err)
+			return
+		}
+		c.requeue = true
+		return
+	case err != nil && !kerrors.IsNotFound(err):
+		c.err = fmt.Errorf("failed to fetch kubernetes mode service account: %w", err)
+		return
+	default:
+		c.logger.Info("Removed finalizer from container mode kubernetes service account")
```
TingluoHuang: same comment about the trace

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -658,6 +718,307 @@ func (r *AutoscalingRunnerSetReconciler) SetupWithManager(mgr ctrl.Manager) erro
...
+			break
+		}
+		err = patch(ctx, c.client, roleBinding, func(obj *rbacv1.RoleBinding) {
+			controllerutil.RemoveFinalizer(obj, autoscalingRunnerSetCleanupFinalizerName)
+		})
+		if err != nil {
+			c.err = fmt.Errorf("failed to patch kubernetes mode role binding without finalizer: %w", err)
+			return
+		}
+		c.requeue = true
```
TingluoHuang, line 768:
```suggestion
                 c.logger.Info("Removed finalizer from container mode kubernetes role binding", "name", roleBindingName)
		c.requeue = true
```

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -658,6 +718,307 @@ func (r *AutoscalingRunnerSetReconciler) SetupWithManager(mgr ctrl.Manager) erro
...
+			break
+		}
+		err = patch(ctx, c.client, role, func(obj *rbacv1.Role) {
+			controllerutil.RemoveFinalizer(obj, autoscalingRunnerSetCleanupFinalizerName)
+		})
+		if err != nil {
+			c.err = fmt.Errorf("failed to patch kubernetes mode role without finalizer: %w", err)
+			return
+		}
+		c.requeue = true
```
TingluoHuang, line 811: same comment about the trace

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -658,6 +718,307 @@ func (r *AutoscalingRunnerSetReconciler) SetupWithManager(mgr ctrl.Manager) erro
...
+			break
+		}
+		err = patch(ctx, c.client, serviceAccount, func(obj *corev1.ServiceAccount) {
+			controllerutil.RemoveFinalizer(obj, autoscalingRunnerSetCleanupFinalizerName)
+		})
+		if err != nil {
+			c.err = fmt.Errorf("failed to patch kubernetes mode service account without finalizer: %w", err)
+			return
+		}
+		c.requeue = true
```
TingluoHuang, line 855: same comment about the trace

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -658,6 +718,307 @@ func (r *AutoscalingRunnerSetReconciler) SetupWithManager(mgr ctrl.Manager) erro
...
+			break
+		}
+		err = patch(ctx, c.client, serviceAccount, func(obj *corev1.ServiceAccount) {
+			controllerutil.RemoveFinalizer(obj, autoscalingRunnerSetCleanupFinalizerName)
+		})
+		if err != nil {
+			c.err = fmt.Errorf("failed to patch service account without finalizer: %w", err)
+			return
+		}
+		c.requeue = true
```
TingluoHuang, line 899: same comment about the trace

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -658,6 +718,307 @@ func (r *AutoscalingRunnerSetReconciler) SetupWithManager(mgr ctrl.Manager) erro
...
+		return
+	case err != nil && !kerrors.IsNotFound(err):
+		c.err = fmt.Errorf("failed to fetch service account: %w", err)
+		return
+	default:
+		c.logger.Info("Removed finalizer from no permission service account", "name", serviceAccountName)
```
TingluoHuang: same comment about the trace

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -658,6 +718,307 @@ func (r *AutoscalingRunnerSetReconciler) SetupWithManager(mgr ctrl.Manager) erro
...
+			break
+		}
+		err = patch(ctx, c.client, githubSecret, func(obj *corev1.Secret) {
+			controllerutil.RemoveFinalizer(obj, autoscalingRunnerSetCleanupFinalizerName)
+		})
+		if err != nil {
+			c.err = fmt.Errorf("failed to patch GitHub secret without finalizer: %w", err)
+			return
+		}
+		c.requeue = true
```
TingluoHuang, line 943: same comment about the trace

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -658,6 +718,307 @@ func (r *AutoscalingRunnerSetReconciler) SetupWithManager(mgr ctrl.Manager) erro
...
+		return
+	case err != nil && !kerrors.IsNotFound(err) && !kerrors.IsForbidden(err):
+		c.err = fmt.Errorf("failed to fetch GitHub secret: %w", err)
+		return
+	default:
+		c.logger.Info("Removed finalizer from GitHub secret", "name", githubSecretName)
```
TingluoHuang: same comment about the trace

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -658,6 +718,307 @@ func (r *AutoscalingRunnerSetReconciler) SetupWithManager(mgr ctrl.Manager) erro
...
+			break
+		}
+		err = patch(ctx, c.client, roleBinding, func(obj *rbacv1.RoleBinding) {
+			controllerutil.RemoveFinalizer(obj, autoscalingRunnerSetCleanupFinalizerName)
+		})
+		if err != nil {
+			c.err = fmt.Errorf("failed to patch manager role binding without finalizer: %w", err)
+			return
+		}
+		c.requeue = true
```
TingluoHuang, line 987: same comment about the trace

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -658,6 +718,307 @@ func (r *AutoscalingRunnerSetReconciler) SetupWithManager(mgr ctrl.Manager) erro
...
+		return
+	case err != nil && !kerrors.IsNotFound(err):
+		c.err = fmt.Errorf("failed to fetch manager role binding: %w", err)
+		return
+	default:
+		c.logger.Info("Removed finalizer from manager role binding", "name", managerRoleBindingName)
```
TingluoHuang: same comment about the trace

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -658,6 +718,307 @@ func (r *AutoscalingRunnerSetReconciler) SetupWithManager(mgr ctrl.Manager) erro
...
+			break
+		}
+		err = patch(ctx, c.client, role, func(obj *rbacv1.Role) {
+			controllerutil.RemoveFinalizer(obj, autoscalingRunnerSetCleanupFinalizerName)
+		})
+		if err != nil {
+			c.err = fmt.Errorf("failed to patch manager role without finalizer: %w", err)
+			return
+		}
+		c.requeue = true
```
TingluoHuang, line 1031: same comment about the trace

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -658,6 +718,307 @@ func (r *AutoscalingRunnerSetReconciler) SetupWithManager(mgr ctrl.Manager) erro
...
+		return
+	case err != nil && !kerrors.IsNotFound(err):
+		c.err = fmt.Errorf("failed to fetch manager role: %w", err)
+		return
+	default:
+		c.logger.Info("Removed finalizer from manager role", "name", managerRoleName)
```
TingluoHuang: same comment about the trace

TingluoHuang (approved):

PR #2435, opened 2023-03-23
Enhance quickstart guide with less noise and more troubleshooting guidelines

Link-:
### Context

The commit log is available already, no need to include in the quickstart guide, the change logs suffice. I've also included the recommendations to resolve the permissions mismatch problem we've been observing when running in kubernetes mode.

### DIFF
--- a/docs/preview/gha-runner-scale-set-controller/README.md
+++ b/docs/preview/gha-runner-scale-set-controller/README.md
@@ -187,7 +187,7 @@ You might see this error if you're using kubernetes mode with persistent volumes
 
 To fix this, you can either:
 
-1. Use a volume type that supports `securityContext.fsGroup`. Update the `fsGroup` of your runner pod to match the GID of the runner. You can do that by updating the `gha-runner-scale-set` helm chart values to include the following:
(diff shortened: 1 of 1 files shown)


nikola-jokic (dismissed):
Awesome! Thank you for documenting this! :rocket:

docs/preview/gha-runner-scale-set-controller/README.md
```diff
@@ -181,6 +181,42 @@ Error: INSTALLATION FAILED: execution error at (gha-runner-scale-set/templates/a
...
+You might see this error if you're using kubernetes mode with persistent volumes. This is because the runner container is running with a non-root user and is causing a permissions mismatch with the mounted volume.
+
+To fix this, you can either:
+
+1. Use a volume type that supports `securityContext.fsGroup`. Update the `fsGroup` of your runner pod to match the GID of the runner. You can do that by updating the `gha-runner-scale-set` helm chart values to include the following:
+
+    ```yaml
+    spec:
+        securityContext:
+            fsGroup: 123
```
TingluoHuang, line 195: will 1001 also work?

Link-: Yes it does!

Link-: Is there a reason we would want to map to the runner's UID instead of its GID?

TingluoHuang: not really, just asking, in case the customer doesn't like 123. 😄

nikola-jokic (approved):

PR #2437, opened 2023-03-23
Proposal: refactor controller tests

rentziass:
This is a proposal to make the (many) tests in `controllers/actions.github.com` more idiomatic and easier to maintain (and faster in some cases as a side effect). This currently feature only two tests, the rest have been commented out to make it easier to talk about this. If we buy into this I'll finish the conversion and open this PR, but wanted to discuss before because it's a _lot_ of code.

What's worth reviewing:
- [test setup](https://github.com/actions/actions-runner-controller/blob/rentziass/no-ginkgo/controllers/actions.github.com/controllers_test.go)
- [example tests](https://github.com/actions/actions-runner-controller/blob/rentziass/no-ginkgo/controllers/actions.github.com/autoscalingrunnerset_controller_test.go)

What's changed and what is proposed:
- Removal of ginkgo and gomega: this is a serious commitment for anyone approaching the project, it's nothing like standard go testing (and as such doesn't support code built to use standard go tools) and requires learning.
- switch to `require` and `assert` for assertions (the same we use in the rest of the project)
- use TestMain to setup and teardown k8s api server we use for tests; we could potentially set this up per test but it's rather costly to do so
- since we use different namespaces for all tests, we can now use `t.Parallel` 🎉 
- due to the asynchronous nature of k8s, we relied heavily on Gomega's Eventually. I could not find an idiomatic alternative to that (testify has its own `Eventually` but also not idiomatic) so in _mild frustration_ I've put [this](https://github.com/rentziass/eventually) together (featured in the examples linked above). Let me know what you think or if you have better alternatives!

### DIFF
--- a/controllers/actions.github.com/autoscalinglistener_controller_test.go
+++ b/controllers/actions.github.com/autoscalinglistener_controller_test.go
@@ -1,720 +1,720 @@
 package actionsgithubcom
 
-import (
-	"context"
-	"fmt"
-	"os"
-	"path/filepath"
-	"time"
-
-	corev1 "k8s.io/api/core/v1"
-	rbacv1 "k8s.io/api/rbac/v1"
-	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
-
-	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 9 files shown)


nikola-jokic:
+1 From me! Thank you for putting this together! :rocket: 

I'd love to remove ginkgo+gomega dependency for our tests

Link-:
We will come back to this effort when we're ready for it.

PR #2442, opened 2023-03-24
Remove containerMode from values

rentziass:
This removes the need to have `containerMode` defined in values when you don't actually need it.


Link- (approved):
Thank you for pushing this

PR #2450, opened 2023-03-29
Fix bug preventing env variables from being specified

Link-:
The chart was unable to properly parse the `env` list to add the variables to the controller pod. This fixes it.

### DIFF
--- a/charts/gha-runner-scale-set-controller/templates/deployment.yaml
+++ b/charts/gha-runner-scale-set-controller/templates/deployment.yaml
@@ -69,10 +69,10 @@ spec:
             fieldRef:
               fieldPath: metadata.namespace
         {{- with .Values.env }}
-          {{- if kindIs "slice" .Values.env }}
-        {{- toYaml .Values.env | nindent 8 }}
+          {{- if kindIs "slice" . }}
+            {{- toYaml . | nindent 8 }}
           {{- else }}
-            {{- range $key, $val := .Values.env }}
+            {{- range $key, $val := . }}
         - name: {{ $key }}
           value: {{ $val | quote }}
             {{- end }}
--- a/charts/gha-runner-scale-set-controller/templates/deployment.yaml
+++ b/charts/gha-runner-scale-set-controller/templates/deployment.yaml
@@ -71,11 +71,6 @@ spec:
         {{- with .Values.env }}
           {{- if kindIs "slice" . }}
             {{- toYaml . | nindent 8 }}
-          {{- else }}
-            {{- range $key, $val := . }}
-        - name: {{ $key }}
-          value: {{ $val | quote }}
-            {{- end }}
           {{- end }}
         {{- end }}
         {{- with .Values.resources }}
--- a/charts/gha-runner-scale-set-controller/tests/template_test.go
+++ b/charts/gha-runner-scale-set-controller/tests/template_test.go
@@ -390,6 +390,8 @@ func TestTemplate_ControllerDeployment_Customize(t *testing.T) {
 			"imagePullSecrets[0].name":     "dockerhub",
 			"nameOverride":                 "gha-runner-scale-set-controller-override",
 			"fullnameOverride":             "gha-runner-scale-set-controller-fullname-override",
+			"env[0].name":                  "ENV_VAR_NAME_1",
+			"env[0].value":                 "ENV_VAR_VALUE_1",
 			"serviceAccount.name":          "gha-runner-scale-set-controller-sa",
 			"podAnnotations.foo":           "bar",
 			"podSecurityContext.fsGroup":   "1000",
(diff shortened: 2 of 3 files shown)


charts/gha-runner-scale-set-controller/templates/deployment.yaml
```diff
@@ -69,10 +69,10 @@ spec:
             fieldRef:
               fieldPath: metadata.namespace
         {{- with .Values.env }}
-          {{- if kindIs "slice" .Values.env }}
-        {{- toYaml .Values.env | nindent 8 }}
+          {{- if kindIs "slice" . }}
+            {{- toYaml . | nindent 8 }}
```
TingluoHuang, line 73: do we want to kill the support for `slice`?

nikola-jokic (dismissed):
LGTM!

charts/gha-runner-scale-set-controller/templates/deployment.yaml
TingluoHuang: I would add an L0 test for this.

nikola-jokic: I would vote for yes, but I did not know if anyone uses it like that.
It would be a breaking change

TingluoHuang: it never work before. 😄

nikola-jokic: Then I'd vote for removing a slice :smile:

mumoshu: FYI, this is how to enable specifying env from secrets and external sources. We recently added this to our legacy chart at https://github.com/actions/actions-runner-controller/pull/2377/files 😄 
Not saying this is a must-have for the new chart right now but I thought it would be worth being considered because we eventually need it!

nikola-jokic: Now that I think about it, can we only support slices?

`pod.spec.containers.env` is a slice of objects, and if we go with the object, we can lose the `valueFrom` which might be useful. Maybe we should kill object support and only support slice?

Link-: I like the flexibility of retaining the slice support, less headaches for us in the future. We've already adopted this philosophy in exposing the pod spec, we might as well do it here at no cost to us

charts/gha-runner-scale-set-controller/tests/template_test.go
```diff
@@ -704,6 +712,45 @@ func TestTemplate_ControllerDeployment_WatchSingleNamespace(t *testing.T) {
...
+
+	var deployment appsv1.Deployment
+	helm.UnmarshalK8SYaml(t, output, &deployment)
+
+	assert.Equal(t, namespaceName, deployment.Namespace)
+	assert.Equal(t, "test-arc-gha-runner-scale-set-controller", deployment.Name)
+
+	assert.Len(t, deployment.Spec.Template.Spec.Containers[0].Env, 4)
+	assert.Equal(t, "ENV_VAR_NAME_1", deployment.Spec.Template.Spec.Containers[0].Env[2].Name)
+	assert.Equal(t, "ENV_VAR_VALUE_1", deployment.Spec.Template.Spec.Containers[0].Env[2].Value)
```
nikola-jokic, line 750: Maybe assert that ValueFrom is empty

charts/gha-runner-scale-set-controller/tests/template_test.go
```diff
@@ -704,6 +712,45 @@ func TestTemplate_ControllerDeployment_WatchSingleNamespace(t *testing.T) {
...
+	assert.Equal(t, "ENV_VAR_NAME_2", deployment.Spec.Template.Spec.Containers[0].Env[3].Name)
+	assert.Equal(t, "secret-name", deployment.Spec.Template.Spec.Containers[0].Env[3].ValueFrom.SecretKeyRef.Name)
```
nikola-jokic, line 752: Assert that Value is empty

charts/gha-runner-scale-set-controller/values.yaml
```diff
@@ -31,27 +42,27 @@ serviceAccount:
 podAnnotations: {}
 
 podSecurityContext: {}
-  # fsGroup: 2000
```
TingluoHuang, line 34: why we need to change this?

TingluoHuang: nvm...

charts/gha-runner-scale-set-controller/tests/template_test.go
Link-: Added below

charts/gha-runner-scale-set-controller/tests/template_test.go
Link-: Added below

TingluoHuang (approved):

nikola-jokic (approved):
:+1:

PR #2458, opened 2023-03-30
Add E2E test to assert self-signed CA support.

TingluoHuang:


charts/gha-runner-scale-set/values.yaml
```diff
@@ -65,7 +65,7 @@ githubConfigSecret:
 #   certificateFrom:
 #     configMapKeyRef:
 #       name: config-map-name
-#       key: ca.pem
+#       key: ca.crt
```
TingluoHuang, line 68: `update-ca-certificate` only checks files with `.crt` extension. 
https://manpages.ubuntu.com/manpages/xenial/man8/update-ca-certificates.8.html

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -565,3 +565,119 @@ jobs:
...
+
+      - name: Install gha-runner-scale-set
+        id: install_arc
+        run: |
+          docker run -d \
+            --rm \
+            --name mitmproxy \
+            --publish 8080:8080 \
+            -v ${{ github.workspace }}/mitmproxy:/home/mitmproxy/.mitmproxy \
+            mitmproxy/mitmproxy:latest \
```
TingluoHuang, line 623: `mitmproxy/mitmproxy:latest` can decrypt HTTP requests, I am using that as the self-signed CA test server.

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -565,3 +565,119 @@ jobs:
...
+          helm install "$ARC_NAME" \
+          --namespace "arc-runners" \
+          --create-namespace \
+          --set githubConfigUrl="https://github.com/${{ env.TARGET_ORG }}/${{env.TARGET_REPO}}" \
+          --set githubConfigSecret.github_token="${{ steps.setup.outputs.token }}" \
+          --set proxy.https.url="http://host.minikube.internal:8080" \
+          --set "proxy.noProxy[0]=10.96.0.1:443" \
+          --set "githubServerTLS.certificateFrom.configMapKeyRef.name=ca-cert" \
+          --set "githubServerTLS.certificateFrom.configMapKeyRef.key=mitmproxy-ca-cert.crt" \
+          --set "githubServerTLS.runnerMountPath=/usr/local/share/ca-certificates/" \
```
TingluoHuang, line 653: Feed in the cert

.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -565,3 +565,119 @@ jobs:
...
+              exit 1
+            fi
+            sleep 1
+          done
+          sudo cp ${{ github.workspace }}/mitmproxy/mitmproxy-ca-cert.pem ${{ github.workspace }}/mitmproxy/mitmproxy-ca-cert.crt
+          sudo chown runner ${{ github.workspace }}/mitmproxy/mitmproxy-ca-cert.crt
+          kubectl create namespace arc-runners
+          kubectl -n arc-runners create configmap ca-cert --from-file="${{ github.workspace }}/mitmproxy/mitmproxy-ca-cert.crt"
+          kubectl -n arc-runners get configmap ca-cert -o yaml
+          ARC_NAME=${{github.job}}-$(date +'%M%S')$((($RANDOM + 100) % 100 + 1))
```
nikola-jokic, line 643:
```suggestion
          ARC_NAME=${{github.job}}-$(date +'%M%S')$(($RANDOM % 100 + 1))
```
Do we need + 100? This should also work fine as far as I know

TingluoHuang: +100, so you always get 3 digit number. 😄 not required.

nikola-jokic: I know but I tested to see if it is okay to have for example:
```bash
echo $((5 % 100 + 1))
```
that would result in:
```
6
```
If you prefer this, ignore my suggestion :relaxed: I was just like, maybe it is cleaner for the reader to see what is going on :relaxed:  The result is exactly the same

TingluoHuang: I will keep this for now. if we want to change, we will change all other jobs to follow the same pattern in a different PR. 😄

nikola-jokic (approved):
LGTM!

Issue #2457, opened 2023-03-30
Running ARC with github_app_id, github_app_installation_id and github_app_private_key on repository or organization on Enterprise Server 3.6 fails to start runner pod configured with RunnerDeployment


mspasoje:
### Checks

- [X] I've already read https://github.com/actions/actions-runner-controller/blob/master/TROUBLESHOOTING.md and I'm sure my issue is not covered in the troubleshooting guide.
- [X] I'm not using a custom entrypoint in my runner image

### Controller Version

0.27.0

### Helm Chart Version

0.22.0

### CertManager Version

1.9.1

### Deployment Method

Helm

### cert-manager installation

I did not install cert-manager myself.

### Checks

- [X] This isn't a question or user support case (For Q&A and community support, go to [Discussions](https://github.com/actions/actions-runner-controller/discussions). It might also be a good idea to contract with any of contributors and maintainers if your business is so critical and therefore you need priority support
- [X] I've read [releasenotes](https://github.com/actions/actions-runner-controller/tree/master/docs/releasenotes) before submitting this issue and I'm sure it's not due to any recently-introduced backward-incompatible changes
- [X] My actions-runner-controller version (v0.x.y) does support the feature
- [X] I've already upgraded ARC (including the CRDs, see charts/actions-runner-controller/docs/UPGRADING.md for details) to the latest and it didn't fix the issue
- [X] I've migrated to the workflow job webhook event (if you using webhook driven scaling)

### Resource Definitions

```yaml
values.yaml for ARE setup:

githubURL: https://github.my_enterprise_name.com/api/v3
githubUploadURL: https://github.my_enterprise_name.com/api/uploads
runnerGithubURL: https://github.my_enterprise_name.com
scope:
  watchNamespace: my-actions-runner-ns
  singleNamespace: true
authSecret:
  create: true
  github_app_id: my_app_id
  github_app_installation_id: my_installation_id
  github_app_private_key: |
my_private_key

RunnerDeployment yaml:

apiVersion: actions.summerwind.dev/v1alpha1
kind: RunnerDeployment
metadata:
  name: my-deployment-name
  namespace: my-actions-runner-ns
spec:
  replicas: 1
  template:
    spec:
      repository: my-org/my-repo
      image: summerwind/actions-runner:ubuntu-20.04
```

### To Reproduce

```markdown
1. We use Github Enterprise Server 3.6
2. install ARC using values file above. ie
[...]

PR #2464, opened 2023-03-31
Proposed fix for ghes org and repo runners when authorized through github app

mspasoje:
Ref #2457

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -84,6 +84,8 @@ func (c *Config) NewClient() (*Client, error) {
 				return nil, fmt.Errorf("enterprise url incorrect: %v", err)
 			}
 			tr.BaseURL = githubAPIURL
+		} else if c.URL != "" && tr.BaseURL != c.URL {
+			tr.BaseURL = c.URL
 		}
 		transport = tr
 	}


github/github.go
```diff
@@ -84,6 +84,8 @@ func (c *Config) NewClient() (*Client, error) {
 				return nil, fmt.Errorf("enterprise url incorrect: %v", err)
 			}
 			tr.BaseURL = githubAPIURL
+		} else if c.URL != "" && tr.BaseURL != c.URL {
+			tr.BaseURL = c.URL
```
mumoshu, line 88: Hey @mspasoje! Could you clarify a bit more on how this works?

We do turn the URL into the BaseURL used by the gh client in L114-123 too. Perhaps yours is needed to let ghinstallation(not gh client) use the correct BaseURL too?

Some more information on how you tested this would be super helpful too! Thanks in advance for your cooperation.

mspasoje: Hi @mumoshu,
The intention was to make sure gh client uses correct BaseURL. However, I do not know why L122 did not do the job. Without a change on ghinstallation, githubclient was trying to get registration token using the default github endpoint (https://api.github.com) instead of the correct one (enterprise api url). Indicated in the below log:

_2023-03-30T18:05:08Z ERROR runner Failed to get new registration token {"runner": "xyz-test-dev-lwtc8-wzqb8", "error": "failed to create registration token: Post "[https://github.enterprise-name.com/api/v3/repos/my-org/my-repo/actions/runners/registration-token\](https://github.enterprise-name.com/api/v3/repos/my-org/my-repo/actions/runners/registration-token%5C)": could not refresh installation id XYZ's token: received non 2xx response status "404 Not Found" when fetching [https://api.github.com/app/installations/XYZ/access_tokens"}](https://api.github.com/app/installations/XYZ/access_tokens%22%7D)_

it seems that go-github client does not use the gh client base url and instead it uses the one specified in the ghinstallation in the second call as indicated by the GET on app installations access tokens in the log.

As for testing, I built the image with the changed code and deployed it in our production using GHES to make sure it works. 
To ensure it didn't break anything I ran makefile tests and also multiple acceptance with makefile deployed kind cluster locally on WSL running Ubuntu. With acceptance tests I had some problems running them out of the box so I had to do lots of manual interventions by deploying runners after running some of the acceptance steps and then running some acceptance tests steps after that. If needed I can explain in detail in another message what was done. In any case I thing I covered all permutations of runnerset/runnerdeployment with repo/org with PAT/GHI. They checked out.

If you need more details please let me know. 

Thanks

mumoshu: @mspasoje Thanks a lot for your detailed explanation and testing!

I realized that the current implementation, especially L81-L86, implies that you should have used `githubEnterpriseURL` instead of `githubURL` in values (and the corresponding envvar GITHUB_ENTERPRISE_URL).

You used this:

```
githubURL: https://github.my_enterprise_name.com/api/v3
```

but it would have worked if it was like:

```
githubEnterpriseURL: https://github.my_enterprise_name.com/api/v3
```

That said, your fix doesn't break anything so I'm going to merge this anyway, so that there will be less confusion because it would just work regardless of which value you specify, either `githubURL` or `githubEnterpriseURL`.

Does that make sense to you? Thanks in advance for your patience and support!

mumoshu (approved):
As I've written at https://github.com/actions/actions-runner-controller/pull/2464/files#r1178613216, I believe this is safe to be merged!
Thanks for your contribution @mspasoje!

github/github.go
mspasoje: Hi @mumoshu,
It makes sense. It is worth pointing out that I tried to deploy ARC and (org and/or repo) runner deployment with githubEnterpriseURL setting but I was not successful (regardless of PAT or githubapp authorization). My conclusion at the time was that githubEnterpriseURL is probably useful only when deploying enterprise level runners. I could not deploy enterprise level runners to verify due to company policy. That is why I have the setup only for repo/org runner deployments with this ARC configuration (values.yaml):

```
githubURL: https://github.my_enterprise_name.com/api/v3
githubUploadURL: https://github.my_enterprise_name.com/api/uploads
runnerGithubURL: https://github.my_enterprise_name.com

```

Do you expect the following configuration instead to work as well for repo/org runnerdeployments (values.yaml):
`githubEnterpriseURL: https://github.my_enterprise_name.com/api/v3`
?

I am asking because as far as I remember it does not work.

Anyways, if the expectation is that it should work, I can find some time and try it and collect logs in case of failure.
Please let me know.
Best Regards

PR #2144, opened 2023-01-12
Fix broken links in docs

SiaraMist:
When [restructuring the documentation](https://github.com/actions/actions-runner-controller/pull/2114), I deleted one file and renamed another. This caused some broken links in the existing documentation.

To resolve this, I ran a search for the two changed file names: `detailed-docs.md` and `Actions-Runner-Controller-Overview.md`. I updated any links to point to the new file names, and changed the link text where necessary.

There weren't too many broken links, so let me know if you think this strategy may have missed anything and I'll make those updates too. Thank you! 🚀

PR #2473, opened 2023-04-04
Fix more broken links in docs

rosesori:
After [the docs were recently restructured](https://github.com/actions/actions-runner-controller/pull/2114), and even after some broken links were fixed in #2144, I've found just a few more links that could be corrected.

I've made sure to click on these locally to make sure that they work. Hope this helps!


Link-:
Thank you @rosesori

Link- (approved):
LGTM

PR #2477, opened 2023-04-04
Add ImagePullPolicy to the AutoscalingListener, configurable through Manager env

nikola-jokic:
Fixes https://github.com/github/c2c-actions-runtime/issues/2416

Configuration is applied using manager's env variable `CONTROLLER_MANAGER_LISTENER_IMAGE_PULL_POLICY`.

The policy will be applied to all autoscaling listeners regardless of the `AutoscalingRunnerSet`.

### DIFF
--- a/charts/gha-runner-scale-set-controller/templates/deployment.yaml
+++ b/charts/gha-runner-scale-set-controller/templates/deployment.yaml
@@ -68,6 +68,8 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        - name: CONTROLLER_MANAGER_LISTENER_IMAGE_PULL_POLICY
+          value: "{{ .Values.image.pullPolicy | default "IfNotPresent" }}"
         {{- with .Values.env }}
           {{- if kindIs "slice" . }}
             {{- toYaml . | nindent 8 }}
--- a/main.go
+++ b/main.go
@@ -171,9 +171,10 @@ func main() {
 	}
 
 	listenerPullPolicy := os.Getenv("CONTROLLER_MANAGER_LISTENER_IMAGE_PULL_POLICY")
-	switch pullPolicy := corev1.PullPolicy(listenerPullPolicy); pullPolicy {
-	case corev1.PullAlways, corev1.PullNever, corev1.PullIfNotPresent:
-		actionsgithubcom.SetListenerImagePullPolicy(pullPolicy)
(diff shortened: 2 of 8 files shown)


charts/gha-runner-scale-set-controller/templates/deployment.yaml
```diff
@@ -68,6 +68,8 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        - name: CONTROLLER_MANAGER_LISTENER_IMAGE_PULL_POLICY
```
TingluoHuang, line 71: do you want to add an L0 test for the chart?

main.go
```diff
@@ -170,6 +170,12 @@ func main() {
 		}
 	}
 
+	listenerPullPolicy := os.Getenv("CONTROLLER_MANAGER_LISTENER_IMAGE_PULL_POLICY")
+	switch pullPolicy := corev1.PullPolicy(listenerPullPolicy); pullPolicy {
+	case corev1.PullAlways, corev1.PullNever, corev1.PullIfNotPresent:
+		actionsgithubcom.SetListenerImagePullPolicy(pullPolicy)
+	}
```
TingluoHuang: feels like a little bit of duplication on checking the `case` from both places?

charts/gha-runner-scale-set-controller/templates/deployment.yaml
nikola-jokic: Adding it, good call :smile: 
We have a mechanism to ignore value if it is incorrect or not set, but you are right!

main.go
nikola-jokic: Yes for sure
I left it here just in case we want to add logging maybe? Do you think we should trace?
If yes, we can add 
1. default case: "Using default image pull policy for AutoscalingListener"
1. On line 176: Setting AutoscalingListener image pull policy to: %q

TingluoHuang: i think we should trace it in `main.go` and remove the `switch/case` in resourceBuilder?

nikola-jokic: How do you feel about accepting a string and returning a bool from SetListenerImagePullPolicy?
So set can control what values are okay, and main can know if the configuration is actually applied or we are using defaults

nikola-jokic: I just wanted to have safety that `actionsgithubcom` package decides what values are correct and can be applied. It is already driving the logic for almost everything, so it feels to me like this logic belongs there

TingluoHuang (approved):

PR #2480, opened 2023-04-04
Treat `.ghe.com` domain as hosted environment

TingluoHuang:
Copy the logic from https://github.com/actions/runner/pull/2420 and make ARC consider the URL ends with  `.ghe.com` as the hosted environment.

https://github.com/github/c2c-actions/issues/6135

### DIFF
--- a/github/actions/config.go
+++ b/github/actions/config.go
@@ -75,21 +75,20 @@ func ParseGitHubConfigFromURL(in string) (*GitHubConfig, error) {
 func (c *GitHubConfig) GitHubAPIURL(path string) *url.URL {
 	result := &url.URL{
 		Scheme: c.ConfigURL.Scheme,
+		Host:   c.ConfigURL.Host, // default for Enterprise mode
+		Path:   "/api/v3",        // default for Enterprise mode
 	}
 
 	isHosted := isHostedGitHubURL(c.ConfigURL)
 
 	if isHosted {
+		result.Host = fmt.Sprintf("api.%s", c.ConfigURL.Host)
+		result.Path = ""
+
 		if strings.EqualFold("www.github.com", c.ConfigURL.Host) {
 			// re-routing www.github.com to api.github.com
 			result.Host = "api.github.com"
-		} else {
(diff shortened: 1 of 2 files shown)


github/actions/config.go
```diff
@@ -96,3 +99,16 @@ func (c *GitHubConfig) GitHubAPIURL(path string) *url.URL {
 
 	return result
 }
+
+func isHostedGitHubURL(u *url.URL) bool {
+	forceGhes := os.Getenv("GITHUB_ACTIONS_FORCE_GHES")
+
+	if len(forceGhes) > 0 {
```
nikola-jokic: If it is only important that the env is set, we can leverage `os.LookupEnv`, although this function returns true if env is set but value is empty

github/actions/config.go
```diff
@@ -78,16 +77,20 @@ func (c *GitHubConfig) GitHubAPIURL(path string) *url.URL {
...
+	if isHosted {
+		// Hosted
+		if strings.EqualFold("github.com", c.ConfigURL.Host) ||
+			strings.EqualFold("github.localhost", c.ConfigURL.Host) ||
+			strings.HasSuffix(c.ConfigURL.Host, ".ghe.com") {
+			result.Host = fmt.Sprintf("api.%s", c.ConfigURL.Host)
+		} else if strings.EqualFold("www.github.com", c.ConfigURL.Host) {
+			// re-routing www.github.com to api.github.com
+			result.Host = "api.github.com"
+		}
```
nikola-jokic:
```suggestion
		if strings.EqualFold("www.github.com", c.ConfigURL.Host) {
			// re-routing www.github.com to api.github.com
			result.Host = "api.github.com"
		} else {
		        result.Host = fmt.Sprintf("api.%s", c.ConfigURL.Host)
		}
```
Since we passed the filter for hosted, can we simplify it like this?

nikola-jokic (approved):
LGTM :rocket:

github/actions/config.go
TingluoHuang: i like this, i even refactor further to remove all the `else`

nikola-jokic: Your version is even better :smile:

PR #2488, opened 2023-04-06
Add description for "dockerRegistryMirror" key

kalined:
More detailed explanation added. It should help users to avoid issues with the docker container.

In case a user accidentally specifies a registry without **HTTP(S)** protocol it could cause an issue: 

**Runner container log:**

```
Cannot connect to the Docker daemon at tcp://localhost:2376. Is the docker daemon running?
```

**Docker container log:**

```
Status: invalid argument "private-registry.example.net" for "--registry-mirror" flag: invalid mirror: unsupported scheme "" in "private-registry.example.net"
See 'dockerd --help'., Code: 125
```


mumoshu (approved):
LGTM. Thank you very much for your contribution @EdgeSan!

PR #2489, opened 2023-04-06
Fix the publish chart workflow

Link-:
### Context

This is a follow-up fix to https://github.com/actions-runner-controller/actions-runner-controller.github.io/pull/2 and https://github.com/actions/actions-runner-controller/pull/2452

- A force publish input was added so that we can force publish the charts manually in case of issues
- The `Generate updated index.yaml` step was updated to push changes to the necessary branch

### DIFF
--- a/.github/workflows/publish-chart.yaml
+++ b/.github/workflows/publish-chart.yaml
@@ -5,14 +5,14 @@ name: Publish Helm Chart
 on:
   push:
     branches:
-      - master
+    - master
     paths:
-      - 'charts/**'
-      - '.github/workflows/publish-chart.yaml'
-      - '!charts/actions-runner-controller/docs/**'
-      - '!charts/gha-runner-scale-set-controller/**'
-      - '!charts/gha-runner-scale-set/**'
-      - '!**.md'
+    - 'charts/**'
+    - '.github/workflows/publish-chart.yaml'
+    - '!charts/actions-runner-controller/docs/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-chart.yaml
```diff
@@ -29,197 +35,173 @@ jobs:
...
+        wget https://github.com/zegl/kube-score/releases/download/v${{ env.KUBE_SCORE_VERSION }}/kube-score_${{ env.KUBE_SCORE_VERSION }}_linux_amd64 -O kube-score
+        chmod 755 kube-score
+
+    - name: Kube-score generated manifests
+      run: helm template  --values charts/.ci/values-kube-score.yaml charts/* | ./kube-score score - --ignore-test pod-networkpolicy --ignore-test deployment-has-poddisruptionbudget --ignore-test deployment-has-host-podantiaffinity --ignore-test container-security-context --ignore-test pod-probes --ignore-test container-image-tag --enable-optional-test container-security-context-privileged --enable-optional-test container-security-context-readonlyrootfilesystem
+
+    # python is a requirement for the chart-testing action below (supports yamllint among other tests)
+    - uses: actions/setup-python@v4
+      with:
+        python-version: '3.7'
```
nikola-jokic:
```suggestion
        python-version: '3.11'
```
Can we use latest version? It has some performance improvements :relaxed:

nikola-jokic (approved):

PR #2497, opened 2023-04-10
Fix GET validation for lowercase http methods

argokasper:
Some requests send method in lowercase (verified with curl and as a default for AWS ALB health check requests), but Go HTTP library constant MethodGet is in upper.


mumoshu (dismissed):
Hey @argokasper! This is absolutely a good catch... I was blindly thinking that go's HTTP server would automatically upper-case the method on composing a request object, but apparently not.

I've just approved the CI workflow run on this pull request so I'll be able to merge this once it passes.

Thank you for your patience and contribution!

controllers/actions.summerwind.net/horizontal_runner_autoscaler_webhook.go
```diff
@@ -115,7 +115,7 @@ func (autoscaler *HorizontalRunnerAutoscalerGitHubWebhook) Handle(w http.Respons
 	}()
 
 	// respond ok to GET / e.g. for health check
-	if r.Method == http.MethodGet {
+	if ToUpper(r.Method) == http.MethodGet {
```
mumoshu:
```suggestion
	if strings.ToUpper(r.Method) == http.MethodGet {
```

Something is wrong with this 😉

mumoshu (approved):

PR #2501, opened 2023-04-11
Reordering methods and constants so it is easier to look it up

nikola-jokic:
The purpose of this change is not to add any new functionality, but to make code easier to follow, remove some repetition and document the finalizer protecting resources.


controllers/actions.github.com/autoscalinglistener_controller.go
```diff
@@ -41,7 +41,6 @@ import (
 
 const (
 	autoscalingListenerContainerName = "autoscaler"
-	autoscalingListenerOwnerKey      = ".metadata.controller"
```
nikola-jokic, line 44: Extracted to constants since the owner key is a field selector

controllers/actions.github.com/autoscalinglistener_controller.go
```diff
@@ -246,65 +245,6 @@ func (r *AutoscalingListenerReconciler) Reconcile(ctx context.Context, req ctrl.
 	return ctrl.Result{}, nil
 }
 
-// SetupWithManager sets up the controller with the Manager.
```
nikola-jokic, line 249: Moved to be the last method on the reconciler, just like for rest of the reconcilers

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -42,13 +42,10 @@ import (
 )
 
 const (
-	// TODO: Replace with shared image.
-	autoscalingRunnerSetOwnerKey             = ".metadata.controller"
-	LabelKeyRunnerSpecHash                   = "runner-spec-hash"
```
nikola-jokic, line 47: Removed TODO and unexported `LabelKeyRunnerSpecHash` since it is only used in `autoscalingrunnerset_controller`

controllers/actions.github.com/constants.go
```diff
@@ -16,3 +18,47 @@ const (
 	EnvVarHTTPSProxy = "https_proxy"
 	EnvVarNoProxy    = "no_proxy"
 )
+
+// Labels applied to resources
+const (
```
nikola-jokic, line 23: Moved label constants from resource builder to constants where I think is more appropriate

controllers/actions.github.com/resourcebuilder.go
```diff
@@ -84,6 +48,62 @@ func SetListenerImagePullPolicy(pullPolicy string) bool {
 
 type resourceBuilder struct{}
 
+func (b *resourceBuilder) newAutoScalingListener(autoscalingRunnerSet *v1alpha1.AutoscalingRunnerSet, ephemeralRunnerSet *v1alpha1.EphemeralRunnerSet, namespace, image string, imagePullSecrets []corev1.LocalObjectReference) (*v1alpha1.AutoscalingListener, error) {
```
nikola-jokic, line 51: Ordered methods so that `newX` is close to its related methods.
It should follow the order of creation: listener, listener resources, ephemeral runner set, ephemeral runner, ephemeral runner resources...

controllers/actions.github.com/constants.go
```diff
@@ -16,3 +18,47 @@ const (
...
+	AnnotationKeyGitHubSecretName                 = "actions.github.com/cleanup-github-secret-name"
+	AnnotationKeyNoPermissionServiceAccountName   = "actions.github.com/cleanup-no-permission-service-account-name"
+)
+
+// DefaultScaleSetListenerImagePullPolicy is the default pull policy applied
+// to the listener when ImagePullPolicy is not specified
+const DefaultScaleSetListenerImagePullPolicy = corev1.PullIfNotPresent
+
+// ownerKey is field selector matching the owner name of a particular resource
+const resourceOwnerKey = ".metadata.controller"
```
TingluoHuang, line 64:
```suggestion
const ResourceOwnerKey = ".metadata.controller"
```

controllers/actions.github.com/resourcebuilder.go
TingluoHuang: I assume there is no real code change other than moving methods around? (maybe changes to use new const)

controllers/actions.github.com/constants.go
nikola-jokic: Do we want to export it? It is internal detail used only for controller and it only means something for the controller logic
No need to make it public on a package level

controllers/actions.github.com/resourcebuilder.go
nikola-jokic: No changes in the code logic, only renames in case I exported finalizer so it is documented in godoc if someone looks at the api there

I only moved things around and for 2 constants changed the visibility

TingluoHuang (approved):

Issue #2466, opened 2023-04-02
Jobs expire while on queue


Nuru:
### Checks

- [X] I've already read https://github.com/actions/actions-runner-controller/blob/master/TROUBLESHOOTING.md and I'm sure my issue is not covered in the troubleshooting guide.
- [X] I'm not using a custom entrypoint in my runner image

### Controller Version

0.27.1

### Helm Chart Version

0.23.0

### CertManager Version

1.10.2

### Deployment Method

Helm

### cert-manager installation

Yes, installed cert manager from official sources.

### Checks

- [X] This isn't a question or user support case (For Q&A and community support, go to [Discussions](https://github.com/actions/actions-runner-controller/discussions). It might also be a good idea to contract with any of contributors and maintainers if your business is so critical and therefore you need priority support
- [X] I've read [releasenotes](https://github.com/actions/actions-runner-controller/tree/master/docs/releasenotes) before submitting this issue and I'm sure it's not due to any recently-introduced backward-incompatible changes
- [X] My actions-runner-controller version (v0.x.y) does support the feature
- [X] I've already upgraded ARC (including the CRDs, see charts/actions-runner-controller/docs/UPGRADING.md for details) to the latest and it didn't fix the issue
- [X] I've migrated to the workflow job webhook event (if you using webhook driven scaling)

### Resource Definitions

```yaml
apiVersion: actions.summerwind.dev/v1alpha1
kind: HorizontalRunnerAutoscaler
metadata:
  annotations:
    meta.helm.sh/release-name: infra-runner-arm64
    meta.helm.sh/release-namespace: actions-runner-system

  labels:
    app.kubernetes.io/managed-by: Helm
  name: infra-runner-arm64
  namespace: actions-runner-system

spec:
  maxReplicas: 64
  minReplicas: 0
  scaleDownDelaySecondsAfterScaleOut: 300
  scaleTargetRef:
    name: infra-runner-arm64
  scaleUpTriggers:
    - amount: 1
      duration: 4m
      githubEvent:
        workflowJob: {}
---
apiVersion: actions.summerwind.dev/v1alpha1
kind: RunnerDeployment
metadata:
  name: infra-runner-arm64
  namespace: actions-runner-system

spec:
  template:
    spec:
      dockerdWithinRunnerContainer: true
      env:
        - name: RUNNER_GRACEFUL_STOP_TIMEOUT
          value: '90'
      group: armEnabled
      image: summerwind/actions-runner-dind
[...]

Issue #2306, opened 2023-02-21
Webhook ARC Scaling issue


uralsemih:
### Checks

- [X] I've already read https://github.com/actions/actions-runner-controller/blob/master/TROUBLESHOOTING.md and I'm sure my issue is not covered in the troubleshooting guide.
- [X] I'm not using a custom entrypoint in my runner image

### Controller Version

v0.27.0

### Helm Chart Version

0.22.0

### CertManager Version

1.5.3

### Deployment Method

Helm

### cert-manager installation

Yes

### Checks

- [X] This isn't a question or user support case (For Q&A and community support, go to [Discussions](https://github.com/actions/actions-runner-controller/discussions). It might also be a good idea to contract with any of contributors and maintainers if your business is so critical and therefore you need priority support
- [X] I've read [releasenotes](https://github.com/actions/actions-runner-controller/tree/master/docs/releasenotes) before submitting this issue and I'm sure it's not due to any recently-introduced backward-incompatible changes
- [X] My actions-runner-controller version (v0.x.y) does support the feature
- [X] I've already upgraded ARC (including the CRDs, see charts/actions-runner-controller/docs/UPGRADING.md for details) to the latest and it didn't fix the issue
- [X] I've migrated to the workflow job webhook event (if you using webhook driven scaling)

### Resource Definitions

```yaml
$ kubectl get runnerdeployment gha-staging-8c-32gb -n staging-8c-32gb -o yaml
apiVersion: actions.summerwind.dev/v1alpha1
kind: RunnerDeployment
metadata:
  annotations:
    meta.helm.sh/release-name: staging-8c-32gb
    meta.helm.sh/release-namespace: staging-8c-32gb
  creationTimestamp: "2022-10-26T17:57:43Z"
  generation: 368232
  labels:
    app.kubernetes.io/managed-by: Helm
  name: gha-staging-8c-32gb
  namespace: staging-8c-32gb
  resourceVersion: "94653290"
  uid: 4c1cd628-8c0a-489a-bb6b-4475cae150db
spec:
  effectiveTime: "2023-02-21T23:07:20Z"
  replicas: 5
  selector: null
  template:
    metadata: {}
    spec:
      dockerVolumeMounts:
      - mountPath: /var/lib/docker
        name: docker-extra
      dockerdContainerResources: {}
      dockerdWithinRunnerContainer: true
      env:
      - name: RUNNER_FEATURE_FLAG_EPHEMERAL
        value: "true"
[...]

Issue #2254, opened 2023-02-06
Ephemeral RunnerDeployments do not scale up for 10 minutes or until all jobs are complete when maxReplicas is set


Smirl:
### Checks

- [X] I've already read https://github.com/actions/actions-runner-controller/blob/master/TROUBLESHOOTING.md and I'm sure my issue is not covered in the troubleshooting guide.
- [X] I'm not using a custom entrypoint in my runner image

### Controller Version

0.27.0

### Helm Chart Version

0.22.0

### CertManager Version

v1.3.1

### Deployment Method

Helm

### cert-manager installation

Our cert manager is a little old in this cluster but I have no reason to believe it is the issue in this particular bug.

### Checks

- [X] This isn't a question or user support case (For Q&A and community support, go to [Discussions](https://github.com/actions/actions-runner-controller/discussions). It might also be a good idea to contract with any of contributors and maintainers if your business is so critical and therefore you need priority support
- [X] I've read [releasenotes](https://github.com/actions/actions-runner-controller/tree/master/docs/releasenotes) before submitting this issue and I'm sure it's not due to any recently-introduced backward-incompatible changes
- [X] My actions-runner-controller version (v0.x.y) does support the feature
- [X] I've already upgraded ARC (including the CRDs, see charts/actions-runner-controller/docs/UPGRADING.md for details) to the latest and it didn't fix the issue
- [X] I've migrated to the workflow job webhook event (if you using webhook driven scaling)

### Resource Definitions

```yaml
apiVersion: actions.summerwind.dev/v1alpha1
kind: RunnerDeployment
metadata:
  labels:
    argocd.argoproj.io/instance: gha-enterprise-runners-github-actions-1-eu-west-1-dev-1
  name: gha-enterprise-runner
  namespace: gha-enterprise
spec:
  effectiveTime: "2023-02-06T12:29:04Z"
  replicas: 0
  template:
    metadata:
      annotations:
        iam.amazonaws.com/role: arn:aws:iam::XXXXXXXXX:role/GHA-actions-test@Actions-runners-test
    spec:
      dockerdWithinRunnerContainer: true
      enterprise: enterprise
      env:
      - name: STARTUP_DELAY_IN_SECONDS
        value: "1"
      - name: DISABLE_RUNNER_UPDATE
        value: "true"
      ephemeral: true
[...]

PR #2502, opened 2023-04-12
Update unconsumed HRA capacity reservation's expiration more frequently and consistently

### LINKED ISSUE #2073: Scaling runners based on webhook is sometimes stuck
### Checks

- [X] I've already read https://github.com/actions-runner-controller/actions-runner-controller/blob/master/TROUBLESHOOTING.md and I'm sure my issue is not covered in the troubleshooting guide.

### Controller Version

0.26.0

### Helm Chart Version

0.21.1

### CertManager Version

v1.10.1

### Deployment Method

Helm

### cert-manager installation

source: https://charts.jetstack.io
Values:
```
installCRDs: true
podDnsPolicy: 'None'
podDnsConfig:
  nameservers:
    - '1.1.1.1'
    - '8.8.8.8'
```

Standard `helm upgrade --install`

### Checks

- [X] This isn't a question or user support case (For Q&A and community support, go to [Discussions](https://github.com/actions-runner-controller/actions-runner-controller/discussions). It might also be a good idea to contract with any of contributors and maintainers if your business is so critical and therefore you need priority support
- [X] I've read [releasenotes](https://github.com/actions-runner-controller/actions-runner-controller/tree/master/docs/releasenotes) before submitting this issue and I'm sure it's not due to any recently-introduced backward-incompatible changes

### Resource Definitions

```yaml
kind: RunnerDeployment
metadata:
  name: small-gha-runner
spec:
  template:
    spec:
      nodeSelector:
        kubernetes.azure.com/agentpool: "fourvcpueph"
      image: {{ .Values.image.repository }}/{{ .Values.image.name }}:{{ .Values.image.tag }}
      imagePullPolicy: {{ .Values.image.imagePullPolicy }}
      group: {{ .Values.github.runnersGroup }}
      organization: {{ .Values.github.organization }}
      labels:
        - small-gha-runner
        - ubuntu-latest-small
      resources:
        limits:
          memory: 5Gi
---
metadata:
  name: small-gha-runner-autoscaler
spec:
  scaleDownDelaySecondsAfterScaleOut: 30
  minReplicas: 1
  maxReplicas: 20
  scaleTargetRef:
    kind: RunnerDeployment
    name: small-gha-runner
  scaleUpTriggers:
    - githubEvent:
        workflowJob: {}
      duration: '10m'
```

### To Reproduce

```markdown
1. Define several github workflows with trigger set on Pull Requests
2. Ask developers to start working
3. Observe the situation when many workflows are triggered, new commits pushed. We also have:

concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true

set in workflows, but that doesn't seem to happen only when such cancel takes place for a particular branch.
```

### Describe the bug

Scaling is sometimes stuck for no visible reason. 
Editiing the horizontalrunnerautoscalers.actions.summerwind.dev and forcing `minReplicas` to be higher than current value also doesn't take place during that period.

No errors in any ACR component.

Jobs are pending in `Queued` state for up to 1hour during the day.

Runners are scaled up/down based on the number of queued jobs in Github workflow.

### Whole Controller Logs

```shell
https://gist.github.com/damyan90/7ffacb6f48ae10f13fd5cf168da142ac
```

### Whole Runner Pod Logs

```shell
Not relevant really, but here's an example one:
https://gist.github.com/damyan90/567979a84cbbd1210ad1ac423e7bac38
```

### Additional Context

Webhook delivery:
https://user-images.githubusercontent.com/24733538/205882331-086724a1-48b7-4e3f-ae77-23d35f959d02.png - 100% successful.

Runner's definition:
```
FROM summerwind/actions-runner:latest

RUN lsb_release -ra

ENV DIR=opt
COPY apt.sh /$DIR/
RUN sudo chmod +x /$DIR/apt.sh && sudo sh /$DIR/apt.sh 

COPY azure.sh /$DIR/
RUN sudo chmod +x /$DIR/azure.sh && sudo sh /$DIR/azure.sh

COPY software.sh software.json /$DIR/
RUN cd $DIR && sudo chmod +x /$DIR/software.sh && sudo sh /$DIR/software.sh

COPY cleanup.sh /$DIR/
RUN sudo chmod +x /$DIR/cleanup.sh && sudo sh /$DIR/cleanup.sh

RUN sudo apt-get update && sudo apt-get dist-upgrade -y

```

Nuru:
This is in response to #2466. This PR mainly updates `batchScale` and the documentation about how HRA webhook scaling works. 

This improves on the fix for #2254 and might fix #2073 and #2306. 

The existing code has a number of issues, particularly in the case where capacity reservations exceeded `maxReplicas`, that may have only been obvious when `minReplicas` was set to zero, HRA `scaleUpTriggers.duration` was close to the length of time jobs actually took to run, and more than `maxReplicas * 3` jobs were queued in a very short period of time. 

Of note is that with `foundIdx = 0` (the normal case) and `maxReplicas = 3` as posited in the comments, `i == 3` would have been skipped because `3 > 0+3` is `false`. 

Summary of changes:
- Capacity reservation expiration times of reservations on the backlog because of `maxReplicas` (see #2254)  happens once on every batch scale operation, rather than once on every scale down.
- Updates occur on all the reservations in the list that are blocked by `maxReplicas` before the list is pruned of expired reservations. This way, if reservations paired with active replicas expire, the replacement reservations will still have their expiration time updated.
- Because reservations could have been added by multiple triggers with different durations, duration is calculated from the existing expiration and effective times rather than by the duration of whatever scale-down trigger has fired.
- `CapacityReservation.EffectiveTime` is no longer marked optional.
- Because `batchScale` makes the implicit assumption that `len(hra.Spec.CapacityReservations)` is the number of replicas reserved, the scale amount dictates the number of capacity reservations added or removed, not the value of the `Replicas` field of the `CapacityReservation`.
- The logged `completed` number is now positive rather than negative. 
- The logged `expired` number now accurately reflects the number of capacity reservations removed because they have expired. Previously it included all scale-up and scale-down activity.


controllers/actions.summerwind.net/horizontal_runner_autoscaler_batch_scale.go
```diff
@@ -157,72 +200,40 @@ func (s *batchScaler) batchScale(ctx context.Context, batch batchScaleOperation)
...
-						//
-						// We address that, by resetting the expiration time for fourth(i=3 in the above example) and subsequent reservations when the first reservation gets cancelled.
-						r.ExpirationTime = metav1.Time{Time: now.Add(scale.trigger.Duration.Duration)}
-					}
-
-					reservations = append(reservations, r)
-				}
+			// Remove the requested number of reservations unless there are not that many left
+			if len(copy.Spec.CapacityReservations) > -amount {
+				copy.Spec.CapacityReservations = copy.Spec.CapacityReservations[-amount:]
```
mumoshu, line 257: You changed it from FIFO to FILO, right? If so, I guess that's incorrect- the earlier workflow jobs would correspond to earlier capacity reservations. Every scale down request(that corresponds to `completed` or `cancelled` workflow job webhook event) should thus result in removing the earliest capacity reservations that matches. As the `amount` is either 1 or -1, foundIdx should  be 0, resulting in removing the first capacity reservation on every scale down.

Nuru: No, it's still FIFO, just like before. New reservations are appended to the end of the slice, old reservations are removed from the beginning of the slice. When `amount == -1` this code simply removes the entry at index zero:

```go
copy.Spec.CapacityReservations = copy.Spec.CapacityReservations[1:]
```

The code in this PR has the added advantage of working properly when `amount` is a number other than +1 or -1.

controllers/actions.summerwind.net/horizontal_runner_autoscaler_batch_scale.go
```diff
@@ -143,8 +143,51 @@ func (s *batchScaler) batchScale(ctx context.Context, batch batchScaleOperation)
 	}
 
 	copy := hra.DeepCopy()
+	now := time.Now()
+
+	if hra.Spec.MaxReplicas != nil && len(copy.Spec.CapacityReservations) > *hra.Spec.MaxReplicas {
```
mumoshu: > The code in this PR has the added advantage of working properly when amount is a number other than +1 or -1.

@Nuru Hey! I was reviewing this once again and I think your statement and this line doesn't align with each other- shouldn't maxReplicas be compared against the sum of all the amount, instead of the length of cap reservations slice, if you're going to support other amounts?

controllers/actions.summerwind.net/horizontal_runner_autoscaler_batch_scale.go
mumoshu: > When amount == -1 this code simply removes the entry at index zero:

Got it! I somehow missed the preceding `-`.

> The code in this PR has the added advantage of working properly when amount is a number other than +1 or -1.

Why do we need this...? (Just unsure

If we are to support other amounts, it would be very likely we add "amount" also to the capacity reservation, so that we no longer need multiple capacity reservations to add multiple replicas- instead we need only one of those at a time.

If we don't, amount would never be other than 1 or -1- so we don't need support for that in code.

Some context: Currently, a scale trigger is a "queued" capacity reservation. So a scale trigger's "amount" corresponds to how many replicas are added/removed via a corresponding capacity reservation. The amount if currently either 1 or -1 because capacity reservation can only express a single replica(it has no `amount` field).

mumoshu: Well anyway, your goal for this pull request was two-fold, right?

- First, you wanted to let ARC update capacity reservation expiration on both scale up and scale down attempt, rather than only on scale down attempt that we currently do
- Second, you wanted to move the expiration update logic before the `getValidCapacityReservations`, so that (hopefully) capacity reservation can start to expire only after it was added a real replica.

Your theory is that these two enable us to (hopefully) fix the issue that a capacity reservation can expire while in the queue(due to it being above maxReplicas), right?

controllers/actions.summerwind.net/horizontal_runner_autoscaler_batch_scale.go
mumoshu: NVM. Could you check https://github.com/actions/actions-runner-controller/pull/2502#discussion_r1169376047?

controllers/actions.summerwind.net/horizontal_runner_autoscaler_batch_scale.go
mumoshu: I've updated the pull request title based on my understanding. WDYT?

Nuru: @mumoshu wrote
> If we are to support other amounts, it would be very likely we add "amount" also to the capacity reservation, so that we no longer need multiple capacity reservations to add multiple replicas- instead we need only one of those at a time.

Capacity reservations already have an "amount" field, it's just called `Replicas`. The problem is that if you have 1 capacity reservation, and it is reserving, for example, 3 replicas, how do you scale down by 1? You have to decrement the Replica field in the reservation rather than just remove it. Also, in this scenario, calculating the number of reserved replicas mean looping over all the reservations and summing their Replica fields, rather than just taking the length of the replica array. 

Supporting reservations reserving other than exactly 1 replica makes all of the code more complex and IMHO it is not worth it. This PR ensures that every capacity reservation reserves exactly 1 replica, no matter what the scale up amount is. 

> If we don't, amount would never be other than 1 or -1- so we don't need support for that in code.

Unfortunately, `scaleOps` have a `scale.trigger.Amount`, and the old code read that field and used it (somewhat incorrectly). If the field is there, someone is likely to use it, and this PR ensures that when they do, the scaling will still work.  I had a choice of either adding an assertion that the value was either 1 or -1 or fixing the code to handle other values, and I chose to fix it. 

> I've updated the pull request title based on my understanding. WDYT?
 
I don't like "Update unconsumed HRA capacity reservation's expiration more frequently" because that implies that the only problem this fixes is somehow related to the update frequency. One problem this PR fixes is that under certain conditions (some rare, some common), a reservation that was backlogged would not have its expiration time adjusted at all. 

I would prefer the title "Update all unconsumed HRA capacity reservation's expiration correctly".

mumoshu: @Nuru Hey! Thanks for your time.

> "Update all unconsumed HRA capacity reservation's expiration correctly".

Can we somehow rephrase "correctly" to a more concrete alternative?
What do you think you did to make it work "correctly"? My impression was that you tried to make it more correct by changing logic to let ARC update expirations "more frequently", but you have another opinion- right?

mumoshu: > Capacity reservations already have an "amount" field, it's just called Replicas. The problem is that if you have 1 capacity reservation, and it is reserving, for example, 3 replicas, how do you scale down by 1? You have to decrement the Replica field in the reservation rather than just remove it

Yeah good point! And that's why I made it only have two values, either 1 or -1. I thought you were willing to do that by saying "The code in this PR has the added advantage of working properly when amount is a number other than +1 or -1.", but apparently not.

> the old code read that field and used it (somewhat incorrectly). If the field is there, someone is likely to use it, and this PR ensures that when they do, the scaling will still work. I had a choice of either adding an assertion that the value was either 1 or -1 or fixing the code to handle other values, and I chose to fix it.

Okay! So- you did it just for consistency with existing code, not to actually add support for non-zero amount. Am I following you correctly?

mumoshu: > Your theory is that these two enable us to (hopefully) fix the issue that a capacity reservation can expire while in the queue(due to it being above maxReplicas), right?

@Nuru Did this part make sense to you? I'm still missing the exact goal of your seemingly awesome work. I do believe you put a lot of considerations and efforts into this work but that blurred the goal of this pull request to me.

Re:

> "Update all unconsumed HRA capacity reservation's expiration correctly".

I think this is good to go once we can agree on the definition of "correct" here.

My impression was it was working as expected and has been correct already. However, you thought it incorrect and therefore phrased "correctly". All in all, the term "correct" is more subjective. I'm more interested in exactly which part of the existing logic looked incorrect to you and how you fixed it!

And my understanding was https://github.com/actions/actions-runner-controller/pull/2502#discussion_r1169376047. Is that correct?

controllers/actions.summerwind.net/horizontal_runner_autoscaler_batch_scale.go
```diff
@@ -143,8 +143,51 @@ func (s *batchScaler) batchScale(ctx context.Context, batch batchScaleOperation)
...
+			//
+			// We address that, by resetting the expiration time for fourth(i=3 in the above example)
+			// and subsequent reservations whenever a batch is run (which is when expired reservations get deleted).
+
+			// There is no guarantee that all the reservations have the same duration, and even if there were,
+			// at this point we have lost the reference to the duration that was intended.
+			// However, we can compute the intended duration from the existing interval.
+			duration := copy.Spec.CapacityReservations[i].ExpirationTime.Time.Sub(copy.Spec.CapacityReservations[i].EffectiveTime.Time)
+			copy.Spec.CapacityReservations[i].EffectiveTime = metav1.Time{Time: now}
+			copy.Spec.CapacityReservations[i].ExpirationTime = metav1.Time{Time: now.Add(duration)}
```
mumoshu, line 213: @Nuru Can I take it as this line corresponds to the following description?

> Because reservations could have been added by multiple triggers with different durations, duration is calculated from the existing expiration and effective times rather than by the duration of whatever scale-down trigger has fired.

We've previously used the "latest capacity reservation duration" for the HRA contained in the scaleTrigger object. You made it reuse the capacity reservation duration when the capacity reservation was initially added, thinking this is more correct than the previous logic. Am I getting it correctly?

controllers/actions.summerwind.net/horizontal_runner_autoscaler_batch_scale.go
mumoshu: Maybe https://github.com/actions/actions-runner-controller/pull/2502/files#r1169468039 relates to the "correctness" you mentioned, too?

controllers/actions.summerwind.net/horizontal_runner_autoscaler_batch_scale.go
Nuru: Yes. The expiration duration is set when the scale-up event occurs and should not change. Plus, you could (in theory, anyway) have multiple triggers with different duration configurations acting at the same time and you would not want them to interfere with each other.

controllers/actions.summerwind.net/horizontal_runner_autoscaler_batch_scale.go
Nuru: The correctness I am referring to is explained in the PR description:

>The existing code has a number of issues, particularly in the case where capacity reservations exceeded maxReplicas, that may have only been obvious when minReplicas was set to zero, HRA scaleUpTriggers.duration was close to the length of time jobs actually took to run, and more than maxReplicas * 3 jobs were queued in a very short period of time.
>
>Of note is that with foundIdx = 0 (the normal case) and maxReplicas = 3 as posited in the comments, i == 3 would have been skipped because 3 > 0+3 is false.

In other words, in the normal case, the oldest backlogged reservation was not having its expiration time extended. Once it expired, the next in line would become the oldest backlogged reservation and its expiration time would top being extended. So the backlogged items were still being expired prematurely.

I also considered it an error that when running jobs expired, the replacements from the backlog were not having the expiration times adjusted before being paired with runners, but that would only be a few seconds of difference.

I also fixed the erroneous reporting of the number of replicas expired, but that is tangential.

mumoshu: Thanks a lot for your patience and effort @Nuru!

So maybe you're correct! I still think that part of the PR description contains only one thing fixed in this pull request. I'm still interested in the whole picture of this pull request.

Based on your explanation, I can definitely think that probably "Summary of changes:" in your PR description covers both "real" fixes that I can agree with and other tiny improvements for potential issues(have not been reported yet or are not too important).

Presuming you're out of assigned time now, I'll try to sort this out somehow- As you changed quite a lot of code here, probably I'd need to write tests for existing and new code, and group your change descriptions into seemingly-important fixes and nice-to-have fixes. Maybe it will take a week or two. Hope that works for you.

apis/actions.summerwind.net/v1alpha1/horizontalrunnerautoscaler_types.go
```diff
@@ -114,11 +114,9 @@ type PushSpec struct {
 // to the scale target until ExpirationTime.
 type CapacityReservation struct {
 	Name           string      `json:"name,omitempty"`
+	EffectiveTime  metav1.Time `json:"effectiveTime,omitempty"`
 	ExpirationTime metav1.Time `json:"expirationTime,omitempty"`
 	Replicas       int         `json:"replicas,omitempty"`
-
-	// +optional
-	EffectiveTime metav1.Time `json:"effectiveTime,omitempty"`
```
mumoshu: @Nuru May I ask why we changed this in this pull request? What exact issue is this change supposed to solve?

controllers/actions.summerwind.net/horizontal_runner_autoscaler_batch_scale.go
```diff
@@ -157,72 +200,40 @@ func (s *batchScaler) batchScale(ctx context.Context, batch batchScaleOperation)
...
+			if len(copy.Spec.CapacityReservations) > -amount {
+				copy.Spec.CapacityReservations = copy.Spec.CapacityReservations[-amount:]
+			} else {
+				copy.Spec.CapacityReservations = nil
 			}
-
-			copy.Spec.CapacityReservations = reservations
-
-			completed += amount
+			// amount is negative, make completed amount positive by negating it
```
mumoshu: Maybe we'd better explain "why" we need this, like

```suggestion
			// This "completed" represents the number of completed and therefore removed runners in this batch,
			// which is logged later.
			// As the amount is negative for a scale-down trigger, we make the "completed" amount positive by negating the amount.
			// That way, the user can see the number of removed runners(like 3), rather than the delta (like -3) in the number of runners.
```

apis/actions.summerwind.net/v1alpha1/horizontalrunnerautoscaler_types.go
Nuru: Removing "optional" to ensure it set, so that this code works: https://github.com/actions/actions-runner-controller/blob/0b5f9d9b930454e64610374143998c207fcfd977/controllers/actions.summerwind.net/horizontal_runner_autoscaler_batch_scale.go#L181-L182

Remove it if you want to.

controllers/actions.summerwind.net/horizontal_runner_autoscaler_batch_scale.go
Nuru: @mumoshu I would be grateful if you wrote tests. Thank you.

Please, in particular, test the case where capacity reservations exceed `maxReplicas`, `minReplicas` is set to zero, HRA `scaleUpTriggers.duration` are close to the length of time jobs actually take to run, and more than `maxReplicas * 3 jobs` are queued in a very short period of time. I expect that in that situation, reservations are expired that should not be, leading to the pool scaling to zero with jobs still on the queue. 

Yes, a lot of the PR falls under the category of "code cleanup", fixing trivial bugs or things that work now because of the limitations of how it is being used, but could break in the future with different usage (like scale amount other than 1). 

This can wait a couple of weeks for you to write tests. Please prioritize #2520, which is having a greater impact in production.

alexandrst88:
any ETA, when it will be merged? I'm still getting the same #2254, #2306, #2073. Thanks!

mumoshu:
Hi, @alexandrst88! I'm still struggling to figure out how to test this properly.
Out of curiosity, what were your exact issues and reproduction steps, and how did you correlate those to this pull request?

alexandrst88:
> Hi, @alexandrst88! I'm still struggling to figure out how to test this properly. Out of curiosity, what were your exact issues and reproduction steps, and how did you correlate those to this pull request? 

Sometimes, webhook has been delivered, but neither runner or job is running. I have set minReplicas to 0. Now, I set replica to 1, with schedule override, and monitor situation. I've read all issues that mention in the PR description and it's similar to mine. Thanks

mumoshu:
@alexandrst88 Thanks for your confirmation! Just to be extra sure, have you also seen workflow job webhook events whose `status` are `completed` before/after the `worfklow_job` webhook events that didn't trigger the expected scale up?
What was your "scale trigger duration" in your HRA spec? If it's too short, you will likely encounter an issue this PR can address.

alexandrst88:
> @alexandrst88 Thanks for your confirmation! Just to be extra sure, have you also seen workflow job webhook events whose `status` are `completed` before/after the `worfklow_job` webhook events that didn't trigger the expected scale up? What was your "scale trigger duration" in your HRA spec? If it's too short, you will likely encounter an issue this PR can address.

duration: 30m.

about `status` are `completed` before/after the `worfklow_job`, I didn't remember, but if it happens again, i will post a screenshot here.

apis/actions.summerwind.net/v1alpha1/horizontalrunnerautoscaler_types.go
mumoshu: Thanks for your confirmation. Let me revert this part then!

Whenever you upgrade CRDs in a backward-incompatible way, you should upgrade the controller code first so that it never omits EffectiveTime, so that the controller is forward-compatible!

That's because our current recommended upgrade process is to upgrade CRDs first, and followed by a controller upgrade. Without an extensive explanation of why this is safe, I'd consider any single pull request that modifies both controller and CRDs in a backward-incompatible way lacks consideration into a safe upgrade.

mumoshu: And it turned out this `// +optional` never produced an entry under the `required` field of the CRD. Not sure how this has ever worked!
The existence of `// +optional` did not make differences across `make manifests` runs. Odd.

controllers/actions.summerwind.net/horizontal_runner_autoscaler_batch_scale.go
mumoshu: > multiple triggers with different duration configurations acting at the same time

Acting at the same time against the same HorizontalRunnerAutoscaler, right? 🤔
I don't think that is supposed to work anyway, because a workflow job is routed to only one eligible HRA.
Anyway, making it to reuse the duration sounds good so I'm definitely going to keep this one!
Thanks for your confirmation.

mumoshu (approved):
I think this is good to go now! Let's merge this and get the ball rolling.
Thanks for your contribution @Nuru!

PR #2504, opened 2023-04-12
Add CR and CRB to the helm chart

totogtr:
In response to https://github.com/actions/actions-runner-controller/issues/2212 , the ARC helm chart is missing ClusterRoleBinding and ClusterRole for the ActionsMetricsServer resulting on missing permissions.

This also fix the labels of the ActionsMetricsServer Service as it is selected by the ServiceMonitor with those labels.

### DIFF
--- a/charts/actions-runner-controller/templates/actionsmetrics.role.yaml
+++ b/charts/actions-runner-controller/templates/actionsmetrics.role.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.githubWebhookServer.enabled }}
+{{- if .Values.actionsMetricsServer.enabled }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
--- a/charts/actions-runner-controller/templates/actionsmetrics.role.yaml
+++ b/charts/actions-runner-controller/templates/actionsmetrics.role.yaml
@@ -1,0 +1,90 @@
+{{- if .Values.githubWebhookServer.enabled }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  name: {{ include "actions-runner-controller-actions-metrics-server.roleName" . }}
+rules:
+- apiGroups: 
+  - actions.summerwind.dev
+  resources:
+  - horizontalrunnerautoscalers
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+  - watch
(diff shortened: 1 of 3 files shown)


charts/actions-runner-controller/templates/actionsmetrics.role.yaml
```diff
@@ -0,0 +1,90 @@
+{{- if .Values.githubWebhookServer.enabled }}
```
mumoshu:
```suggestion
{{- if .Values.actionsMetricsServer.enabled }}
```

Shouldn't this refer to the actionsMetricsServer.enabled as it's for actions-metrics-server, not for github-webhook-server?

charts/actions-runner-controller/templates/actionsmetrics.role_binding.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.githubWebhookServer.enabled }}
```
mumoshu:
```suggestion
{{- if .Values.actionsMetricsServer.enabled }}
```

charts/actions-runner-controller/templates/actionsmetrics.service.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   name: {{ include "actions-runner-controller-actions-metrics-server.fullname" . }}
   namespace: {{ .Release.Namespace }}
   labels:
-    {{- include "actions-runner-controller.labels" . | nindent 4 }}
+    {{- include "actions-runner-controller-actions-metrics-server.selectorLabels" . | nindent 4 }}
```
mumoshu, line 8: Wow, good catch!

mumoshu (commented):
Hey @totogtr! Thank you for your contribution! I just reviewed this and had two comments regarding the values used to enable the CR and CRB creations. Would you mind confirming when you have some time?

charts/actions-runner-controller/templates/actionsmetrics.role.yaml
totogtr: Of course you're right, my bad.

totogtr:
Hey @mumoshu, thanks a lot for reviewing. Indeed the enable conditions were wrong and fixed by your suggestion, I must have done some bad copy pasting here sorry !

mumoshu (approved):
LGTM. Thanks a lot for your contribution @totogtr!

PR #2508, opened 2023-04-12
Update ubuntu-22.04 Dockerfile to add python user script dir

cavila-evoliq:
I was running into an issue after upgrading our ubuntu image to build off of 22.04 instead of 20.04 where pipx was looking like it wasn't found even though pulling my image showed pipx was installed. I noticed this line missing from the 22.04 Dockerfile that is present in 20.04.

I figured I'd open a PR adding it to keep the behavior consistent for future upgraders.


mumoshu (approved):
Good catch! Thanks for your contribution @cavila-evoliq ☺️

PR #2515, opened 2023-04-17
Fix the path of the index.yaml in job summary

Link-:
Fixes the path of the `index.yaml` in the publish helm chart workflow's job summary. The default branch was incorrectly defined as main instead of master.

CleanShot 2023-04-17 at 15 06 07


johnsudol (approved):

mumoshu (approved):
LGTM! Thanks for fixing it!

PR #2517, opened 2023-04-17
Fix how hosted URLs are identified

Link-:
### Context

This change ignores anything that comes before the domain and tld in identifying hosted config URLs.

### DIFF
--- a/github/actions/config.go
+++ b/github/actions/config.go
@@ -102,8 +102,7 @@ func isHostedGitHubURL(u *url.URL) bool {
 		return false
 	}
 
-	return strings.EqualFold(u.Host, "github.com") ||
-		strings.EqualFold(u.Host, "www.github.com") ||
+	return strings.HasSuffix(u.Host, "github.com") ||
 		strings.EqualFold(u.Host, "github.localhost") ||
 		strings.HasSuffix(u.Host, ".ghe.com")
 }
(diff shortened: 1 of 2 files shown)


benwells (approved):

github/actions/config.go
```diff
@@ -102,8 +102,7 @@ func isHostedGitHubURL(u *url.URL) bool {
 		return false
 	}
 
-	return strings.EqualFold(u.Host, "github.com") ||
-		strings.EqualFold(u.Host, "www.github.com") ||
+	return strings.HasSuffix(u.Host, "github.com") ||
```
TingluoHuang, line 105: if we change this, does it means we also need to change the runner?

nikola-jokic: I think this should be:
```go
strings.EqualFold(u.Host, "github.com") || strings.HasSuffix(u.Host, ".github.com")
```
Domain "examplegithub.com" would pass this check if we just check for the suffix

Link-: Good catch @nikola-jokic!

Link-: @TingluoHuang I'm not familiar how this impacts the runner, can you or @nikola-jokic elaborate on the impact please?

TingluoHuang: the runner and arc should follow the same logic to identify whether a URL is hosted or GHES

Link-:
Closing this for being a low priority high effort fix.

PR #2521, opened 2023-04-19
Use build.Version to check if resource version is a mismatch

nikola-jokic:
If ARC deployment is removed in case of upgrade before rest of the resources are being cleaned up, use build.Version to check the version of the ARC and try to remove resources not matching this binary version

Fixes https://github.com/github/c2c-actions-runtime/issues/2426

### DIFF
--- a/.github/workflows/e2e-test-linux-vm.yaml
+++ b/.github/workflows/e2e-test-linux-vm.yaml
@@ -16,7 +16,7 @@ env:
   TARGET_ORG: actions-runner-controller
   TARGET_REPO: arc_e2e_test_dummy
   IMAGE_NAME: "arc-test-image"
-  IMAGE_VERSION: "dev"
+  IMAGE_VERSION: "0.4.0"
 
 jobs:
   default-setup:
--- a/controllers/actions.github.com/autoscalingrunnerset_controller.go
+++ b/controllers/actions.github.com/autoscalingrunnerset_controller.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 
 	"github.com/actions/actions-runner-controller/apis/actions.github.com/v1alpha1"
+	"github.com/actions/actions-runner-controller/build"
 	"github.com/actions/actions-runner-controller/github/actions"
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
@@ -133,6 +134,22 @@ func (r *AutoscalingRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl
 		}
 
 		log.Info("Successfully removed finalizer after cleanup")
+		return ctrl.Result{}, nil
+	}
+
+	if autoscalingRunnerSet.Labels[LabelKeyKubernetesVersion] != build.Version {
+		if err := r.Delete(ctx, autoscalingRunnerSet); err != nil {
+			log.Error(err, "Failed to delete autoscaling runner set on version mismatch",
+				"targetVersion", build.Version,
+				"observedVersion", autoscalingRunnerSet.Labels[LabelKeyKubernetesVersion],
+			)
+			return ctrl.Result{}, nil
+		}
+
+		log.Info("Successfully issued a delete for autoscaling runner set on version mismatch",
+			"targetVersion", build.Version,
+			"observedVersion", autoscalingRunnerSet.Labels[LabelKeyKubernetesVersion],
+		)
 		return ctrl.Result{}, nil
 	}
 
--- a/controllers/actions.github.com/autoscalingrunnerset_controller.go
+++ b/controllers/actions.github.com/autoscalingrunnerset_controller.go
@@ -141,14 +141,14 @@ func (r *AutoscalingRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl
 		if err := r.Delete(ctx, autoscalingRunnerSet); err != nil {
 			log.Error(err, "Failed to delete autoscaling runner set on version mismatch",
 				"targetVersion", build.Version,
-				"observedVersion", autoscalingRunnerSet.Labels[LabelKeyKubernetesVersion],
+				"actualVersion", autoscalingRunnerSet.Labels[LabelKeyKubernetesVersion],
 			)
 			return ctrl.Result{}, nil
 		}
 
-		log.Info("Successfully issued a delete for autoscaling runner set on version mismatch",
+		log.Info("Autoscaling runner set version doesn't match the build version. Deleting the resource.",
 			"targetVersion", build.Version,
-			"observedVersion", autoscalingRunnerSet.Labels[LabelKeyKubernetesVersion],
+			"actualVersion", autoscalingRunnerSet.Labels[LabelKeyKubernetesVersion],
 		)
 		return ctrl.Result{}, nil
 	}
--- a/controllers/actions.github.com/autoscalingrunnerset_controller_test.go
+++ b/controllers/actions.github.com/autoscalingrunnerset_controller_test.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/actions/actions-runner-controller/apis/actions.github.com/v1alpha1"
+	"github.com/actions/actions-runner-controller/build"
 	"github.com/actions/actions-runner-controller/github/actions"
 	"github.com/actions/actions-runner-controller/github/actions/fake"
 	"github.com/actions/actions-runner-controller/github/actions/testserver"
@@ -38,12 +39,24 @@ const (
 	autoscalingRunnerSetTestGitHubToken = "gh_token"
 )
 
-var _ = Describe("Test AutoScalingRunnerSet controller", func() {
+var _ = Describe("Test AutoScalingRunnerSet controller", Ordered, func() {
 	var ctx context.Context
 	var mgr ctrl.Manager
 	var autoscalingNS *corev1.Namespace
 	var autoscalingRunnerSet *v1alpha1.AutoscalingRunnerSet
 	var configSecret *corev1.Secret
+
+	var originalBuildVersion string
+	buildVersion := "0.1.0"
+
+	BeforeAll(func() {
+		originalBuildVersion = build.Version
+		build.Version = buildVersion
+	})
+
+	AfterAll(func() {
+		build.Version = originalBuildVersion
+	})
 
 	BeforeEach(func() {
 		ctx = context.Background()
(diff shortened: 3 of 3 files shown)


.github/workflows/e2e-test-linux-vm.yaml
```diff
@@ -16,7 +16,7 @@ env:
   TARGET_ORG: actions-runner-controller
   TARGET_REPO: arc_e2e_test_dummy
   IMAGE_NAME: "arc-test-image"
-  IMAGE_VERSION: "dev"
+  IMAGE_VERSION: "0.4.0"
```
Link-, line 19: Why are we pinning the image version here?

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -136,6 +137,22 @@ func (r *AutoscalingRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl
...
+			log.Error(err, "Failed to delete autoscaling runner set on version mismatch",
+				"targetVersion", build.Version,
+				"observedVersion", autoscalingRunnerSet.Labels[LabelKeyKubernetesVersion],
+			)
+			return ctrl.Result{}, nil
+		}
+
+		log.Info("Successfully issued a delete for autoscaling runner set on version mismatch",
+			"targetVersion", build.Version,
+			"observedVersion", autoscalingRunnerSet.Labels[LabelKeyKubernetesVersion],
```
Link-:
```suggestion
			"actualVersion", autoscalingRunnerSet.Labels[LabelKeyKubernetesVersion],
```
nit: Might this be clearer?

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -136,6 +137,22 @@ func (r *AutoscalingRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl
 		return ctrl.Result{}, nil
 	}
 
+	if autoscalingRunnerSet.Labels[LabelKeyKubernetesVersion] != build.Version {
+		if err := r.Delete(ctx, autoscalingRunnerSet); err != nil {
+			log.Error(err, "Failed to delete autoscaling runner set on version mismatch",
+				"targetVersion", build.Version,
+				"observedVersion", autoscalingRunnerSet.Labels[LabelKeyKubernetesVersion],
```
Link-:
```suggestion
				"actualVersion", autoscalingRunnerSet.Labels[LabelKeyKubernetesVersion],
```
nit: here too?

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -136,6 +137,22 @@ func (r *AutoscalingRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl
...
+			)
+			return ctrl.Result{}, nil
+		}
+
+		log.Info("Successfully issued a delete for autoscaling runner set on version mismatch",
```
Link-: So this message is correct, but it doesn't really explain why there's a version mismatch in the first place. Maybe we can something like:

> Autoscaling runner set version doesn't match the build version. Deleting the resource.

controllers/actions.github.com/autoscalingrunnerset_controller_test.go
```diff
@@ -707,7 +753,18 @@ var _ = Describe("Test AutoscalingController creation failures", func() {
 	})
 })
 
-var _ = Describe("Test Client optional configuration", func() {
+var _ = Describe("Test Client optional configuration", Ordered, func() {
```
Link-:
```suggestion
var _ = Describe("Test client optional configuration", Ordered, func() {
```

nit

controllers/actions.github.com/autoscalingrunnerset_controller.go
nikola-jokic: Awesome, I like it

controllers/actions.github.com/autoscalingrunnerset_controller_test.go
```diff
@@ -1465,3 +1551,76 @@ var _ = Describe("Test external permissions cleanup", func() {
 		).Should(BeTrue(), "Expected role to be cleaned up")
 	})
 })
+
+var _ = Describe("Version mismatch cleanup", func() {
```
Link-:
```suggestion
var _ = Describe("Test resource version and build version mismatch", func() {
```

.github/workflows/e2e-test-linux-vm.yaml
nikola-jokic: We need to do this because the build.Version will be replaced with "dev"
Chart.yaml uses "0.4.0" while build.Version uses dev, causing mismatch each time. So no e2e test can run because whenever we create something, we delete it on the first pass in the controller

controllers/actions.github.com/autoscalingrunnerset_controller_test.go
```diff
@@ -1465,3 +1551,76 @@ var _ = Describe("Test external permissions cleanup", func() {
...
+	It("Should delete autoscaling runner set on a version mismatch", func() {
```
Link-:
```suggestion
	It("Should delete and recreate the autoscaling runner set to match the build version", func() {
```

controllers/actions.github.com/autoscalingrunnerset_controller.go
nikola-jokic:
```suggestion
		log.Info("Autoscaling runner set version doesn't match the build version. Deleting the resource.",
```

controllers/actions.github.com/autoscalingrunnerset_controller_test.go
```diff
@@ -1465,3 +1551,76 @@ var _ = Describe("Test external permissions cleanup", func() {
...
+			Scheme:                             mgr.GetScheme(),
+			Log:                                logf.Log,
+			ControllerNamespace:                autoscalingNS.Name,
+			DefaultRunnerScaleSetListenerImage: "ghcr.io/actions/arc",
+			ActionsClient:                      fake.NewMultiClient(),
+		}
+		err := controller.SetupWithManager(mgr)
+		Expect(err).NotTo(HaveOccurred(), "failed to setup controller")
+
+		originalVersion := build.Version
```
Link-, line 1573: Why are we introducing a new pattern here while everywhere else we're using beforeAll / afterAll?

controllers/actions.github.com/autoscalingrunnerset_controller_test.go
```diff
@@ -1465,3 +1551,76 @@ var _ = Describe("Test external permissions cleanup", func() {
...
+		Expect(err).NotTo(HaveOccurred())
+
+		startManagers(GinkgoT(), mgr)
+
+		Eventually(func() bool {
+			ars := new(v1alpha1.AutoscalingRunnerSet)
+			err := k8sClient.Get(ctx, types.NamespacedName{Namespace: autoscalingRunnerSet.Namespace, Name: autoscalingRunnerSet.Name}, ars)
+			return errors.IsNotFound(err)
+		}).Should(BeTrue())
+	})
```
Link-, line 1625: I'm not really sure how we're testing that the deletion / creation actually happened, could you help me understand this part please?

controllers/actions.github.com/autoscalingrunnerset_controller_test.go
```diff
@@ -1465,3 +1551,76 @@ var _ = Describe("Test external permissions cleanup", func() {
...
+			ActionsClient:                      fake.NewMultiClient(),
+		}
+		err := controller.SetupWithManager(mgr)
+		Expect(err).NotTo(HaveOccurred(), "failed to setup controller")
+
+		originalVersion := build.Version
+		defer func() {
+			build.Version = originalVersion
+		}()
+		build.Version = "0.2.0"
```
Link-, line 1577: Why are we introducing a new pattern here while everywhere else we're using beforeAll / afterAll?

Link- (changes requested):
Awesome stuff as always! Thank you for taking care of this ❤️

.github/workflows/e2e-test-linux-vm.yaml
Link-: Clear, how are we going to maintain this down the line or it's inconsequential because we're not really pulling anything and we're building the image locally?

nikola-jokic: Each time we bump the chart version, e2e IMAGE_VERSION must align with it. That is it :relaxed:

controllers/actions.github.com/autoscalingrunnerset_controller_test.go
nikola-jokic: Sure, so once we issue a `k8sClient.Create`, it is going to create a resource in etcd.
Then our controller is triggered, we fetch the autoscaling runner set and the controller should check the version.
The resource is intentionally made with the version mismatch, so eventually, the controller needs to delete this resource as soon as it can.
Every other test have resource version matching the `build.Version`, so when we run tests, happy path is covered.
This test tries make sure that we are actually going to delete this resource because of the version mismatch. Eventually is there because we can't know the exact time this is going to be called, and with finalizer checks, it can take some time.
And since we are configuring the autoscaling runner set with correct parameters, but incorrect version, the deletion should happen without anyone (except the controller) issuing a `DELETE` request

controllers/actions.github.com/autoscalingrunnerset_controller_test.go
nikola-jokic: Because I wanted to scope the mismatch to this exact test.
I refactored it to use BeforeAll/AfterAll, but then it looked strange in a sense that this test sets the version `"0.1.0"`, which should be deleted because build.Version is "0.2.0".

When you read the test, you can see two constants and it draws attention to it.

On the rest of them, BeforeAll/AfterAll was appropriate (in my mind) because I am using the same version both for the label, and for the `build.Version`.

The intention is to draw attention here that we are actually changing the version and that is the most important part of the test

nikola-jokic: Maybe I shouldn't have used the separate describe for it... That would be an improvement, because this `Describe` block is only used to hold one `It` test, which is self-contained

Link-: I'll leave this to you to make the call on, the rest if fine with me

Link- (approved):
🚀

PR #2524, opened 2023-04-21
Check release tag version and chart versions during the release process

nikola-jokic:
Charts published along with the image should match the image version. This check aims to make sure that binary version and version of charts are the same

### DIFF
--- a/hack/check-gh-chart-versions.sh
+++ b/hack/check-gh-chart-versions.sh
@@ -14,7 +14,7 @@ target_version=$1
 if [[ $# -eq 0 ]]; then
     echo "Release version argument is required"
     echo
-    echo "Usage:  ${0} <release version>"
+    echo "Usage:  ${0} <VERSION>"
     exit 1
 fi
 
--- a/hack/check-gh-chart-versions.sh
+++ b/hack/check-gh-chart-versions.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-#
+set -eo pipefail
 TEXT_RED='\033[0;31m'
 TEXT_RESET='\033[0m'
 TEXT_GREEN='\033[0;32m'
--- a/hack/check-gh-chart-versions.sh
+++ b/hack/check-gh-chart-versions.sh
@@ -20,8 +20,12 @@ controller_app_version=$(cat $chart_dir/gha-runner-scale-set-controller/Chart.ya
 scaleset_version=$(cat $chart_dir/gha-runner-scale-set/Chart.yaml | yq .version)
 scaleset_app_version=$(cat $chart_dir/gha-runner-scale-set/Chart.yaml | yq .appVersion)
 
-if [[ $controller_version != $controller_app_version ]] || [[ $controller_version != $scaleset_version ]] || [[ $controller_version != $scaleset_app_version ]]; then
+if [[ $controller_version != $target_version ]] ||
+   [[ $controller_app_version != $target_version ]] ||
+   [[ $scaleset_version != $target_version ]] ||
+   [[ $scaleset_app_version != $target_version ]]; then
     echo -e "${TEXT_RED}Chart versions do not match${TEXT_RESET}"
+    echo "Target version:         $target_version"
     echo "Controller version:     $controller_version"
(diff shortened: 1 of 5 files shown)


hack/check-gh-chart-versions.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+#
```
Link-, line 3:
```suggestion
set -eo pipefail
```

Sets the shell options to exit immediately if any command exits with a non-zero status, and to exit with a non-zero status if any command in a pipeline fails.
- set -e: Exit immediately if a command exits with a non-zero status.
- set -o pipefail: Exit with a non-zero status if any command in a pipeline fails.

hack/check-gh-chart-versions.sh
```diff
@@ -0,0 +1,36 @@
...
+
+target_version=$1
+if [[ $# -eq 0 ]]; then
+    echo "Release version argument is required"
+    echo
+    echo "Usage:  ${0} <release version>"
+    exit 1
+fi
+
+chart_dir=$(pwd)/charts
```
Link-: Always double quote variables, including subshells. No naked $ signs. More on that here: http://mywiki.wooledge.org/Quotes

hack/check-gh-chart-versions.sh
```diff
@@ -0,0 +1,36 @@
...
+
+controller_version=$(cat $chart_dir/gha-runner-scale-set-controller/Chart.yaml | yq .version)
```
Link-:
```suggestion
controller_version="$(cat $chart_dir/gha-runner-scale-set-controller/Chart.yaml | grep version: | cut -d ' ' -f 2)"
```

Let's not introduce `yq` into the mix

hack/check-gh-chart-versions.sh
```diff
@@ -0,0 +1,36 @@
...
+controller_app_version=$(cat $chart_dir/gha-runner-scale-set-controller/Chart.yaml | yq .appVersion)
```
Link-: Same as above and let's not forget the double quotes here

hack/check-gh-chart-versions.sh
```diff
@@ -0,0 +1,36 @@
...
+
+scaleset_version=$(cat $chart_dir/gha-runner-scale-set/Chart.yaml | yq .version)
+scaleset_app_version=$(cat $chart_dir/gha-runner-scale-set/Chart.yaml | yq .appVersion)
```
Link-: Same as above + the double quotes

hack/check-gh-chart-versions.sh
```diff
@@ -0,0 +1,36 @@
...
+
+if [[ $controller_version != $target_version ]] ||
+   [[ $controller_app_version != $target_version ]] ||
+   [[ $scaleset_version != $target_version ]] ||
+   [[ $scaleset_app_version != $target_version ]]; then
```
Link-: Double quotes for all variables as well

hack/check-gh-chart-versions.sh
Link-: Also here, can we add some comments to describe what this script is / does and how it can be called?

```bash
# Checks the chart versions against an input version. Fails on mismatch.
#
# Usage:
#   check-gh-chart-versions.sh <VERSION>
```

Link-:
Can you have this pass `shellcheck` please? I think we should add it as a linter

Link- (changes requested):
Can you have this pass `shellcheck` please? I think we should add it as a linter

hack/check-gh-chart-versions.sh
nikola-jokic: yq is already a dependency https://github.com/actions/actions-runner-controller/blob/9859bbc7f2b1d9006fd137e6ed1db807cbfe86a0/Makefile#L344

And I think it would be nicer to use yq since yaml like this should return the same value:
```yaml
version: 0.1.0
appVerison: "0.1.0"
```
yq would evaluate correctly while grep will have a mismatch

Link- (approved):

PR #2535, opened 2023-04-25
Fix the default version of kube-rbac-proxy in the docs

sengaya:


mumoshu (approved):
Good catch...! Thanks a lot for fixing it @sengaya!

PR #2542, opened 2023-04-26
docs: minor correction for actions metrics server secret

object88:
Aligning docs with what the Helm chart produces


mumoshu (approved):
LGTM. Thanks for your contribution!

PR #2551, opened 2023-04-28
Update runner container hooks to 0.3.1

syslxg:
hooks 0.3.x fixed a few issues.

Authored-by: Gary Liu <garyliu@vmware.com>


reiniertimmer:
@syslxg I'm not completely sure I understand the workflow here, but it seems that the runner hooks version is actually passed from the [release-runners.yml](https://github.com/actions/actions-runner-controller/blob/master/.github/workflows/release-runners.yaml#L21) workflow (which triggers the release workflow on [actions-runner-controller/releases](https://github.com/actions-runner-controller/releases/blob/main/.github/workflows/release-runners.yaml)).

So in short, I guess the version needs to be updated in the [release-runners.yml](https://github.com/actions/actions-runner-controller/blob/master/.github/workflows/release-runners.yaml#L21) workflow file as well.

Link-:
I'll look into this tomorrow. Ideally we'd want this to be automated. This workflow: https://github.com/actions/actions-runner-controller/blob/master/.github/workflows/update-runners.yaml should also check for container-hooks releases and create a PR when there's a new version.

Link-:
Closing as this is being addressed in: https://github.com/actions/actions-runner-controller/pull/2576

Link-:
@reiniertimmer @syslxg - new runner images have been published with container hooks 0.3.1. Could you please verify and let me know?

reiniertimmer:
Thanks @Link- , will do!

PR #2553, opened 2023-04-30
Add alternative dind runner image with nvidia container runtime

yhtang:
Sometimes we may want to run Actions workflows that use containerized GPU software, e.g.:
```yaml
name: A workflow that uses a GPU container
...
jobs:
  unit-test:
    runs-on: [self-hosted, compute, V100]
    steps:
      ...
      - name: Run GPU code inside container
        shell: bash -x -e {0}
        run: |
          # not working out-of-the-box currently
          docker run --gpus all ghcr.io/nvidia/jax command ...
...
```

Currently, the GPUs on a worker node can be passed into a runner/pod via the deploy manifest file. However, to further allow the GPUs to be used in a docker container in a workflow, the runner must have NVIDIA container runtime installed.

This PR adds a new alternative runner image `actions-runner-dind-nvcrt` that augments the `dind` flavor with NVIDIA container runtime. An example image built from the new Dockerfile [can be found here](https://github.com/users/yhtang/packages/container/package/actions-runner-dind-nvcrt).

Question:

Given that the NVIDIA container runtime adds only ~20 MB to the runner image, how about directly baking it into the existing `dind` runner instead of creating a new runner image?
```
actions-runner-dind-nvcrt    ubuntu-20.04    1.23GB
actions-runner-dind          ubuntu-20.04    1.21GB
actions-runner-dind-nvcrt    ubuntu-22.04    965MB
actions-runner-dind          ubuntu-22.04    948MB
```

### DIFF
--- a/runner/Makefile
+++ b/runner/Makefile
@@ -2,6 +2,7 @@ DOCKER_USER ?= summerwind
 DOCKER ?= docker
 DEFAULT_RUNNER_NAME ?= ${DOCKER_USER}/actions-runner
 DIND_RUNNER_NAME ?= ${DOCKER_USER}/actions-runner-dind
+DIND_NVCRT_RUNNER_NAME ?= ${DOCKER_USER}/actions-runner-dind-nvcrt
 DIND_ROOTLESS_RUNNER_NAME ?= ${DOCKER_USER}/actions-runner-dind-rootless
 OS_IMAGE ?= ubuntu-22.04
 TARGETPLATFORM ?= $(shell arch)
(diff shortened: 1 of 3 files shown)


Link-:
@yhtang we highly appreciate your contribution to this project. However, moving forward we have no intentions to continue maintaining these runner images. The only runner image we will be supplying out of the box is [actions-runner](https://github.com/actions/runner/pkgs/container/actions-runner).

Entities that wish to customise their runners are encouraged to build and maintain their own runner images.

If you’d like to publish this image and maintain it independently, feel free to do so in a separate repository.

I will be closing this PR as we will not move forward with it.

PR #2554, opened 2023-05-01
docs: Update github docs links under `/managing-self-hosted-runners`

sh-cho:
Some links redirected well, but most links failed to redirect, so I updated all links with the self-hosted runner


wasup-yash (approved):

Link- (approved):
Thank you for these fixes. It's good to go.

johnsudol (approved):

PR #2559, opened 2023-05-03
Provide scale-set listener metrics

nikola-jokic:
### Context

Fixes https://github.com/github/c2c-actions/issues/6816

Metrics are being introduced to the new autoscaling runner scale set mode. These metrics are separate from those published by the legacy modes.

The following is the list of metrics the controller-manager and listener pods will be emitting.

| Owner              | Metric                         | Type      | Description                                                                                                          |
|--------------------|--------------------------------|-----------|----------------------------------------------------------------------------------------------------------------------|
| controller-manager | pending_ephemeral_runners      | gauge     | Number of ephemeral runners in a pending state.                                                                      |
| controller-manager | running_ephemeral_runners      | gauge     | Number of ephemeral runners in a running state.                                                                      |
| controller-manager | failed_ephemeral_runners       | gauge     | Number of ephemeral runners in a failed state.                                                                       |
| listener           | available_jobs                 | gauge     | Number of jobs with `runs-on` matching the runner scale set name. Jobs are not yet assigned to the runner scale set. |
| listener           | acquired_jobs                  | gauge     | Number of jobs acquired by the runner scale set.                                                                     |
| listener           | assigned_jobs                  | gauge     | Number of jobs assigned to the runner scale set.                                                                     |
| listener           | running_jobs                   | gauge     | Number of jobs running (or about to be run).                                                                         |
| listener           | registered_runners             | gauge     | Number of runners registered by the runner scale set.                                                                |
| listener           | busy_runners                   | gauge     | Number of registered runners currently running a job.                                                                |
| listener           | min_runners                    | gauge     | Minimum number of runners configured for the runner scale set.                                                       |
| listener           | max_runners                    | gauge     | Maximum number of runners configured for the runner scale set.                                                       |
| listener           | desired_runners                | gauge     | Number of runners desired (scale up / down target) by the runner scale set.                                          |
| listener           | idle_runners                   | gauge     | Number of registered runners not running a job.                                                                      |
| listener           | started_jobs_total             | counter   | Total number of jobs started since the listener became ready (will reset on pod restart).                            |
| listener           | completed_jobs_total           | counter   | Total number of jobs completed since the listener became ready (will reset on pod restart).                          |
| listener           | job_queue_duration_seconds     | histogram | Time spent waiting for workflow jobs to get assigned to the runner scale set after queueing (in seconds).            |
| listener           | job_startup_duration_seconds   | histogram | Time spent waiting for workflow job to get started on the runner owned by the runner scale set (in seconds).         |
| listener           | job_execution_duration_seconds | histogram | Time spent executing workflow jobs by the runner scale set (in seconds).                                             |


controllers/actions.github.com/resourcebuilder.go
```diff
@@ -200,6 +200,12 @@ func (b *resourceBuilder) newScaleSetListenerPod(autoscalingListener *v1alpha1.A
 				Command: []string{
 					"/github-runnerscaleset-listener",
 				},
+				Ports: []corev1.ContainerPort{
```
TingluoHuang: do we need to expose based on certain settings enabled?

cmd/githubrunnerscalesetlistener/autoScalerService.go
```diff
@@ -25,6 +26,43 @@ type Service struct {
 	kubeManager        KubernetesManager
 	settings           *ScaleSettings
 	currentRunnerCount int
+	metricsExporter    metricsExporter
+	errs               []error
+}
+
+func WithPrometheusMetrics(conf RunnerScaleSetListenerConfig) func(*Service) {
```
TingluoHuang, line 33: move these `With***` method down after the `NewService()`?

cmd/githubrunnerscalesetlistener/autoScalerService.go
```diff
@@ -25,6 +26,43 @@ type Service struct {
...
+		default:
+			owner = parsedURL.Organization
+		}
+
+		svc.metricsExporter.withBaseLabels(baseLabels{
+			scaleSetName:                  conf.EphemeralRunnerSetName,
+			scaleSetConfigURL:             conf.ConfigureUrl,
+			autoscalingRunnerSetName:      conf.EphemeralRunnerSetName,
+			autoscalingRunnerSetNamespace: conf.EphemeralRunnerSetNamespace,
+			repositoryName:                repo,
```
TingluoHuang: also include enterprise if exist?

nikola-jokic: Can I include like:
repo, organization, enterprise?
And allow empty label

controllers/actions.github.com/resourcebuilder.go
nikola-jokic: How should we enable settings?
There is a `metrics-addr` flag there. What I would do is, if `metrics-addr` is set, enable metrics on both the controller and the listener. Otherwise, don't expose metrics.
The only problem is that it takes the default value of `:8080`, so basically, to turn it off, you would have to say: `--metrics-addr=""` explicitly. 
Follow-up question: can we use a constant port for the listener, or should we set it based on the `metrics-addr`?

cmd/githubrunnerscalesetlistener/autoScalerService.go
nikola-jokic: I personally prefer having options above the `New` function. When I read the code top to bottom, I can see an Option type, skip through options or just see what they are and then start with the constructor function. I like to keep options above the constructor and methods below it personally

charts/gha-runner-scale-set/values.yaml
```diff
@@ -170,3 +170,7 @@ template:
 # controllerServiceAccount:
 #   namespace: arc-system
 #   name: test-arc-gha-runner-scale-set-controller
+
+metrics:
```
TingluoHuang: do we need to expose this controller per runner-scale-set?
or we just query the controller and create a service by default of the customer enables service monitor at the controller level.

nikola-jokic: I'll try to see the easiest way to set it up
I wanted to create this PR as a POC to have something working and later improve it since it is already a big one
But I'll convert this to draft, base the controller metrics on it to explore the easiest way to combine both of them and create a follow-up PR for controller metrics :relaxed:

charts/gha-runner-scale-set-controller/templates/manager_exported_service.yaml
TingluoHuang, line 1: change the file name form exported -> exporter?

cmd/githubrunnerscalesetlistener/metrics.go
```diff
@@ -0,0 +1,381 @@
...
+
+	"github.com/actions/actions-runner-controller/github/actions"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// label names
+const (
+	labelKeyRunnerScaleSetName            = "name"
+	labelKeyRunnerScaleSetConfigURL       = "config_url"
+	labelKeyAutoScalingRunnerSetName      = "auto_scaling_runner_set_name"
```
nikola-jokic: Should we use a shorter name here?

cmd/githubrunnerscalesetlistener/metrics.go
```diff
@@ -0,0 +1,381 @@
...
+	labelKeyAutoScalingRunnerSetNamespace = "auto_scaling_runner_set_namespace"
```
nikola-jokic: Should we use a shorter name here?

charts/gha-runner-scale-set-controller/values.yaml
```diff
@@ -83,3 +83,15 @@ flags:
...
+
+# If metrics object is not provided, following flags won't be applied:
+# --metrics-addr, --listener-metrics-addr, --listener-metrics-endpoint
+# Keep in mind, metrics are exposed by default (flags have default values)
+# specifying the default port and default endpoint.
+#
+# To turn off metrics, specify empty strings for controllerAddr and listenerAddr
+metrics:
+  controllerAddr: ":8080"
+  listenerAddr: ":8080"
```
TingluoHuang: Do we want to use the fully qualified name?
Ex:
- `controllerManagerAddr`
- `autoscalinglistenerAddr`

main.go
```diff
@@ -102,6 +107,8 @@ func main() {
 		os.Exit(1)
 	}
 
+	flag.StringVar(&listenerMetricsAddr, "listener-metrics-addr", "8080", "The address applied to AutoscalingListener metrics server")
```
TingluoHuang: `8080` or `:8080` the default value to `metrics-addr` is `:8080`.

controllers/actions.github.com/ephemeralrunnerset_controller.go
```diff
@@ -163,6 +166,34 @@ func (r *EphemeralRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl.R
 		"deleting", len(deletingEphemeralRunners),
 	)
 
+	if r.PublishMetrics {
+		githubConfigURL := ephemeralRunnerSet.Spec.EphemeralRunnerSpec.GitHubConfigUrl
+		parsedURL, err := actions.ParseGitHubConfigFromURL(githubConfigURL)
+		if err != nil {
+			log.Error(err, "Github Config URL is invalid", "URL", githubConfigURL)
+			// stop reconciling on this object
+			return ctrl.Result{}, nil
```
TingluoHuang, line 175: return error?

controllers/actions.github.com/ephemeralrunnerset_controller.go
```diff
@@ -163,6 +166,34 @@ func (r *EphemeralRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+		}
+
+		owner := parsedURL.Organization
+		if parsedURL.Enterprise != "" {
+			owner = parsedURL.Enterprise
```
TingluoHuang: we might not want to merge Org/Enterprise into Owner, the owner only means Org or User in GitHub system.

controllers/actions.github.com/metrics/metrics.go
```diff
@@ -0,0 +1,75 @@
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var githubScaleSetControllerSubsystem = "github_runner_scale_set_controller"
+
+var labels = []string{
```
TingluoHuang, line 10: i would just have `repo/owner/enterprise` all three.

charts/gha-runner-scale-set-controller/templates/deployment.yaml
```diff
@@ -59,8 +59,18 @@ spec:
...
+        {{- with .Values.metrics }}
+        - "--listener-metrics-addr={{ .listenerAddr }}"
+        - "--listener-metrics-endpoint={{ .listenerEndpoint }}"
+        - "--metrics-addr={{ .controllerAddr }}"
+        {{- end }}
         command:
         - "/manager"
+        {{- with .Values.metrics }}
+        ports:
+        - containerPort: {{regexReplaceAll ":([0-9]+)" .controllerAddr "${1}"}}
```
TingluoHuang: why do we need to have this regexReplace?

cmd/githubrunnerscalesetlistener/metrics.go
```diff
@@ -0,0 +1,381 @@
...
+import (
+	"strconv"
+
+	"github.com/actions/actions-runner-controller/github/actions"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// label names
+const (
+	labelKeyRunnerScaleSetName            = "name"
```
TingluoHuang: `labelKeyRunnerScaleSetName` and `labelKeyAutoScalingRunnerSetName` might confuse customers?

cmd/githubrunnerscalesetlistener/metrics.go
```diff
@@ -0,0 +1,381 @@
...
+			Help:      "Number of jobs running (or about to be run).",
+		},
+		scaleSetLabels,
+	)
+
+	registeredRunners = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: githubScaleSetSubsystem,
+			Name:      "registered_runners",
+			Help:      "Number of registered runners.",
```
TingluoHuang: what about `Number of runners registered by this scale set`?

cmd/githubrunnerscalesetlistener/metrics.go
```diff
@@ -0,0 +1,381 @@
...
+			Help:      "Number of registered runners running a job.",
+		},
+		scaleSetLabels,
+	)
+
+	minRunners = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: githubScaleSetSubsystem,
+			Name:      "min_runners",
+			Help:      "Number of runners desired bu the scale set.",
```
TingluoHuang: `bu`->`by`

What about `Minimal number of runners....`

cmd/githubrunnerscalesetlistener/metrics.go
```diff
@@ -0,0 +1,381 @@
...
+		},
+		scaleSetLabels,
+	)
+
+	maxRunners = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: githubScaleSetSubsystem,
+			Name:      "max_runners",
+			Help:      "Number of runners desired bu the scale set.",
```
TingluoHuang: same. bu->by

`Maximum number of runners....`

cmd/githubrunnerscalesetlistener/metrics.go
```diff
@@ -0,0 +1,381 @@
...
+		},
+		scaleSetLabels,
+	)
+
+	desiredRunners = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: githubScaleSetSubsystem,
+			Name:      "desired_runners",
+			Help:      "Number of runners desired bu the scale set.",
```
TingluoHuang: `bu` -> `by`

charts/gha-runner-scale-set-controller/templates/deployment.yaml
nikola-jokic: I needed to extract the port in case someone specifies `address:port`, or `:port`.

controllers/actions.github.com/ephemeralrunnerset_controller.go
nikola-jokic: It is going to trigger reconcile again, while if the URL is invalid, we can't really do anything about it, so we would be just wasting cycles

main.go
nikola-jokic: My mistake :relaxed: Thanks!

Link-:
Do we want to add E2E tests to this PR or shall we have another task just for that?

controllers/actions.github.com/metrics/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+	"name",
+	"namespace",
+	"repo",
```
TingluoHuang: `repo` -> `repository` ?

controllers/actions.github.com/metrics/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+	Organization string
+	Enterprise   string
+	ConfigURL    string
+}
+
+func (l *CommonLabels) labels() prometheus.Labels {
+	return prometheus.Labels{
+		"name":         l.Name,
+		"namespace":    l.Namespace,
+		"repo":         l.Repo,
```
TingluoHuang: `repo` -> `repository`?

controllers/actions.github.com/ephemeralrunnerset_controller.go
TingluoHuang: should we not return and keep the reconcile loop?
basically, the publish metrics is best effort?

cmd/githubrunnerscalesetlistener/main.go
```diff
@@ -67,17 +74,98 @@ func main() {
...
+	srv    *http.Server
+}
+
+func (s *metricsServer) shutdown() error {
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+	return s.srv.Shutdown(ctx)
+}
+
+func (s *metricsServer) run() error {
```
TingluoHuang: we have 2 `run()` functions in this file, is it possible to give them different name to make find/search accurate?

`func (s *metricsServer) startMetricsServer()`?

cmd/githubrunnerscalesetlistener/main.go
```diff
@@ -67,17 +74,98 @@ func main() {
...
+			g.Go(func() error {
+				<-ctx.Done()
+				return metricsServer.shutdown()
+			})
+			return metricsServer.run()
+		})
+	}
+
+	if err := g.Wait(); err != nil {
+		logger.Error(err, "Error encountered")
```
TingluoHuang, line 120: Will we get error during SIGINT?

cmd/githubrunnerscalesetlistener/metrics.go
```diff
@@ -0,0 +1,364 @@
...
+	jobStartupDurationLabels   = append(jobLabels, labelKeyRunnerID, labelKeyRunnerName)
+)
+
+// metrics
+var (
+	availableJobs = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: githubScaleSetSubsystem,
+			Name:      "available_jobs",
+			Help:      "Number of jobs with `runs-on` matching the runner scale set name. Jobs are not yet assigned but are acquired by the runner scale set.",
```
TingluoHuang:
```suggestion
			Help:      "Number of jobs with `runs-on` matching the runner scale set name. Jobs are not yet assigned to the runner scale set.",
```

cmd/githubrunnerscalesetlistener/metrics.go
```diff
@@ -0,0 +1,364 @@
...
+			Help:      "Number of registered runners not running a job.",
+		},
+		scaleSetLabels,
+	)
+
+	availableJobsTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: githubScaleSetSubsystem,
+			Name:      "available_jobs_total",
+			Help:      "Total number of jobs available for the scale set (runs-on matches and scale set passes all of the runner group permission checks).",
```
TingluoHuang:
```suggestion
			Help:      "Total number of jobs with `runs-on` matching the runner scale set name.",
```

cmd/githubrunnerscalesetlistener/main.go
nikola-jokic: No, unless I made a mistake :smile: 

The error will be returned in case g.Go callback returns an error. The SIGINT will cancel the context, and if we return `ctx.Err()`, then the g.Wait will return the error back causing `os.Exit(1)`. Otherwise, if callback stops on context cancellation and returns nil, this will exit with exit code 0.

controllers/actions.github.com/ephemeralrunnerset_controller.go
nikola-jokic: That is right, but the failure is due to the wrong URL. We should never reach this point, the listener would fail, the autoscaling runner set should fail before this. I left it as a precaution, but we should never reach this point. If we are failing to parse GitHub URL, we have a bigger problem somewhere and just in case, this would serve as an indicator. But I don't think we should ever reach this condition

charts/gha-runner-scale-set-controller/values.yaml
```diff
@@ -102,3 +102,15 @@ flags:
...
+# If metrics object is not provided, following flags won't be applied:
+# --metrics-addr, --listener-metrics-addr, --listener-metrics-endpoint
+# Keep in mind, metrics are exposed by default (flags have default values)
+# specifying the default port and default endpoint.
+#
+# To turn off metrics, specify empty strings for controllerAddr and listenerAddr
+metrics:
+  controllerManagerAddr: ":8080"
+  listenerAddr: ":8080"
+  listenerEndpoint: "/metrics"
```
Link-:
```suggestion
# metrics:
#  controllerManagerAddr: ":8080"
#  listenerAddr: ":8080"
#  listenerEndpoint: "/metrics"
```

I suggest disabling them by default. Otherwise customers will have this endpoint exposed and will not be aware of the change. Let it be opt-in  as opposed to opt-out

Link- (changes requested):

cmd/githubrunnerscalesetlistener/metrics.go
```diff
@@ -0,0 +1,330 @@
...
+			Help:      "Number of registered runners running a job.",
+		},
+		scaleSetLabels,
+	)
+
+	minRunners = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: githubScaleSetSubsystem,
+			Name:      "min_runners",
+			Help:      "Minimum number of runners.",
```
Link-, line 112:
```suggestion
			Help:      "Minimum number of runners configured for the runner scale set.",
```

cmd/githubrunnerscalesetlistener/metrics.go
```diff
@@ -0,0 +1,330 @@
...
+		},
+		scaleSetLabels,
+	)
+
+	maxRunners = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: githubScaleSetSubsystem,
+			Name:      "max_runners",
+			Help:      "Maximum number of runners.",
```
Link-, line 121:
```suggestion
			Help:      "Maximum number of runners configured for the runner scale set.",
```

cmd/githubrunnerscalesetlistener/metrics.go
```diff
@@ -0,0 +1,330 @@
...
+		},
+		scaleSetLabels,
+	)
+
+	desiredRunners = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: githubScaleSetSubsystem,
+			Name:      "desired_runners",
+			Help:      "Number of runners desired by the scale set.",
```
Link-, line 130:
```suggestion
			Help:      "Number of runners desired (scale up / down target) by the runner scale set.",
```

cmd/githubrunnerscalesetlistener/metrics.go
```diff
@@ -0,0 +1,330 @@
...
+			Help:      "Number of registered runners not running a job.",
+		},
+		scaleSetLabels,
+	)
+
+	startedJobsTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: githubScaleSetSubsystem,
+			Name:      "started_jobs_total",
+			Help:      "Total number of jobs started.",
```
Link-, line 148:
```suggestion
			Help:      "Total number of jobs started since the listener became ready (will reset on pod restart).",
```

cmd/githubrunnerscalesetlistener/metrics.go
```diff
@@ -0,0 +1,330 @@
...
+		},
+		startedJobsTotalLabels,
+	)
+
+	completedJobsTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name:      "completed_jobs_total",
+			Help:      "Total number of jobs completed.",
```
Link-, line 156:
```suggestion
			Help:      "Total number of jobs completed since the listener became ready (will reset on pod restart).",
```

charts/gha-runner-scale-set-controller/values.yaml
```diff
@@ -102,3 +102,14 @@ flags:
   ##   This can lead to a longer time to apply the change but it will ensure
   ##   that you don't have any overprovisioning of runners.
   updateStrategy: "immediate"
+
+# # If metrics object is not provided, following flags won't be applied:
+# # --metrics-addr, --listener-metrics-addr, --listener-metrics-endpoint
+# # Keep in mind, metrics are exposed by default (flags have default values)
+# # specifying the default port and default endpoint.
+# #
+# # To turn off metrics, specify empty strings for controllerAddr and listenerAddr
```
Link-: Maybe we should remove this line since they're disabled by default.

nikola-jokic: I have to update the comment above as well

Link-:
### TODOs:
- [ ] gha_registered_runners continues to report on the last value the listener receives
  - This requires a server side change 
- [x] Revisit the labels and make sure the common labels are applied 
- [x] Expose number of listeners gauge with all the common labels applied

Link- (approved):
Brilliant work @nikola-jokic this is an exciting ship!

chlunde:
@nikola-jokic any chance of getting job_queue_duration_seconds? This seems like a very good metric to monitor the health of our configuration/service quality?  Awesome work :heart:

nikola-jokic:
Hey @chlunde, thank you for your kind words! We are planning to introduce this metric in future releases. While testing we noticed a small issue when the scale set transitions from being active to idle, so we decided not to publish this metric until it is fixed :relaxed:

PR #2563, opened 2023-05-04
Fix typo in HorizontalRunnerAutoscaler

becheran:


mumoshu (approved):
@becheran Thank you for fixing it! I had to update the relevant Go code and generate another manifest to complete it. It is now good to go. Thanks again for your contribution!

PR #2566, opened 2023-05-08
chore: Service accounts in Kubernetes mode can now be annotated.

kahirokunn:
If annotations are added, IRSA and other settings can be configured.


kahirokunn:
I'd be happy to get a review.

nikola-jokic:
Hey @kahirokunn, sorry for the delay.

We are going to issue a new 0.5.0 release soon, and we will review this change for the next release. :relaxed:

kahirokunn:
Okay

kahirokunn:
@nikola-jokic
Congratulations on the release of 0.5.0!
How about now?

charts/gha-runner-scale-set/values.yaml
```diff
@@ -170,3 +170,7 @@ template:
 # controllerServiceAccount:
 #   namespace: arc-system
 #   name: test-arc-gha-runner-scale-set-controller
+
+kubernetesMode:
```
nikola-jokic: I think that the better place for this object would be under `containerMode` object, similar to the [`kubernetesModeWorkVolumeClaim`](https://github.com/actions/actions-runner-controller/blob/f1d7c52253b89f0beae60141f8465d9495cdc2cf/charts/gha-runner-scale-set/values.yaml#L80)
It can be named something like `kubernetesModeServiceAccount`
What do you think @Link- ?

kahirokunn: @nikola-jokic  I agree. Opinion reflected.

Link-:
Pending tests

charts/gha-runner-scale-set/templates/kube_mode_serviceaccount.yaml
```diff
@@ -5,6 +5,10 @@ kind: ServiceAccount
 metadata:
   name: {{ include "gha-runner-scale-set.kubeModeServiceAccountName" . }}
   namespace: {{ .Release.Namespace }}
+  {{- if .Values.containerMode.kubernetesModeServiceAccount.annotations }}
```
nikola-jokic: Can you also include a test?

kahirokunn: @nikola-jokic @Link- okay. I added a test.

nikola-jokic (approved):
LGTM

PR #2567, opened 2023-05-08
chore: remove 16 characters from -service-account

kahirokunn:
The service account must be less than 63 characters. Currently, a normal install will result in ${INSTALLATION_NAME}-gha-runner-scale-set-kube-mode-service-account and 47 characters are reserved.
This is too long, so we will remove 16 characters from -service-account


kahirokunn:
@Link- How about this?

bjw-s:
I'm also quite interested in seeing this being reviewed/implemented. The current limitation forces my ARC runner deployments to have super short names since otherwise the SA name will become too long. Alternatively, allowing us to set a `fullnameOverride` (as the default `helm create` template also implements) could help a lot with this in a more generic way.

nikola-jokic:
@Link- this is a good suggestion (in my opinion) that we might want to apply to all resources.
When you search for the role, you know that the kind is `Role`. We don't really need a suffix in the name.
If you are troubleshooting, what I would usually do is `kubectl get roles`. The `-role` or `-service-account` suffix does not really provide much value, but it limits the name used as a label

Link-:
@nikola-jokic alright, let's take this on. It's on our board.

kahirokunn:
I need this 👍

nikola-jokic:
Hey @kahirokunn,

The change seems good to me, but can you help me understand where is this causing you problems? I'm trying to reproduce an issue where the name of the service account is going to cause problems, so I registered the scale set with the name length of 45 characters. The no-permission service account did not cause any troubles there since we used its name as an annotation in cleanup (not having the 63 character limit), and the name of the resource has the 253 characters in the form of the DNS name

kahirokunn:
@nikola-jokic 
The problem lies in this link.
https://github.com/cilium/cilium/issues/16579#issuecomment-1352364004
cilium is something that many vendors, such as GKE, provide by default and share.
But this is not just an issue on the cilium side, it should be able to be adjusted to not exceed the dns rule of 63 characters.

nikola-jokic:
Oh, I see, thanks!

Then we might want to consider different naming conventions that will allow shorter names (@Link-)
I updated your branch to trigger chart tests but for now. I think that we should merge this change before a new release. But it seems like the naming limitation of 45 characters is not correct for situations like this one

nikola-jokic (approved):
LGTM

PR #2568, opened 2023-05-08
Scale Set Metrics ADR

nikola-jokic:
Propose a solution for exposing metrics by the `gha-runner-scale-set` and `gha-runner-scale-set-controller`

### DIFF
--- a/docs/adrs/2023-05-08-exposing-metrics.md
+++ b/docs/adrs/2023-05-08-exposing-metrics.md
@@ -7,36 +7,121 @@ Date: 2023-05-08
 ## Context
 
 Prometheus metrics are a common way to monitor the cluster. Providing metrics
-can be a helpful way to monitor scale sets and health of the ephemeral runners.
+can be a helpful way to monitor scale sets and the health of the ephemeral runners.
 
 ## Proposal
 
-There are two main components driving the behaviour of the scale set:
+Two main components are driving the behavior of the scale set:
 
-1. ARC controllers responsible for managing kubernetes resources.
+1. ARC controllers responsible for managing Kubernetes resources.
 2. The `AutoscalingListener`, driver of the autoscaling solution responsible for
    describing the desired state.
 
+We can approach publishing those metrics in 3 different ways
+
+### Expose Service Monitor, Services for listeners and a Service for the manager
+
 To expose metrics, we would need to create 3 additional resources:
 
-1. `ServiceMonitor` - resource used by prometheus to match namespaces and
-    services from where it needs to gather metrics
+1. `ServiceMonitor` - a resource used by Prometheus to match namespaces and
+   services from where it needs to gather metrics
 2. `Service` for the `gha-runner-scale-set-controller` - service that will
-    target ARC controller `Deployment`
-3. `Service` for the `gha-runner-scale-set` listeners - service that will target
-    all listeners for each `AutoscalingRunnerSet`
+   target ARC controller `Deployment`
+3. `Service` for each `gha-runner-scale-set` listener - service that will target
+   a single listener pod for each `AutoscalingRunnerSet`
 
-Metrics can be enabled or disabled on the controller level. If the flag
-`metrics-addr` is not empty, the controller and listeners are going to register
-the appropriate collectors.
+#### Pros
 
-Alternatively, if there is a need to control the metrics exposure per
-`AutoscalingRunnerSet`, we can control it by applying additional label on the
-listener.
+- Easy to control which scale set exposes metrics and which does not.
+- Easy to implement using helm charts in case they are enabled per chart
+  installation.
+
+#### Cons
+
+- With a cluster running many scale sets, we are going to create a lot of
+  resources.
+- In case metrics are enabled on the controller manager level, and they should
+  be applied across all `AutoscalingRunnerSets`, it is difficult to inherit this
+  configuration by applying helm charts.
+
+### Create an aggregator service
+
+To create an aggregator service, we can create a simple web application
+responsible for publishing and gathering metrics. All listeners would be
+responsible to communicate the metrics on each message, and controllers are
+responsible to communicate the metrics on each reconciliation.
+
+The application can be executed as a single pod, or as a side container next to
+the manager.
+
+Pros for running the aggregator next to the manager:
+
+- It exists side by side and is following the life cycle of the controller
+  manager
+- We don't need to introduce another controller managing the state of the pod
+
+Cons for running the aggregator next to the manager:
+
+- Crashes of the aggregator can influence the controller manager execution
+- The controller manager pod needs more resources to run
+
+Pros for running the aggregator as a stand-alone:
+
+- Does not influence the controller manager pod
+- The life cycle of the metric can be controlled by the controller manager (by
+  implementing another controller)
+
+Cons for running the aggregator as a stand-alone:
+
+- We need to implement the controller that can spin up the aggregator in case of
+  the crash.
+- If we choose not to implement the controller, the resource like `Deployment`
+  can be used to manage the aggregator, but we lose control over its life cycle.
+
+To expose metrics in this way, we need to:
+
+1. Create a web server with a single `/metrics` endpoint. The endpoint will have
+   `POST` and `GET` methods registered. The `GET` is used by Prometheus to
+   fetch the metrics, while the `POST` is going to be used by controllers and
+   listeners to publish their metrics.
+2. `ServiceMonitor` - to target the metrics aggregator service
+3. `Service` sitting in front of the web server.
+
+#### Pros
+
+- This implementation requires a few additional resources to be created
+  in a cluster.
+- Web server is easy to implement and easy to document - all metrics are aggregated in a
+  single package, and the web server only needs to apply them to its state on
+  `POST`. The `GET` handler is simple.
+- We can avoid Pushgateway from Prometheus.
+
+#### Cons
+
+- Another image that we need to publish on release.
+- Change in metric configuration (on manager update) would require re-creation
+  of all listeners. This is not a big problem but is something to point out.
+- Managing requests/limits can be tricky.
+
+### Use a Prometheus Pushgateway
+
+#### Pros
+
+- Using a supported way of pushing the metrics.
+- Easy to implement using their library.
+
+#### Cons
+
+- In the Prometheus docs, they specify that: "Usually, the only valid use case
+  for Pushgateway is for capturing the outcome of a service-level batch job".
+  The listener does not really fit this criteria.
+- Pushgateway is a single point of failure and potential bottleneck.
+- You lose Prometheus's automatic instance health monitoring via the up metric (generated on every scrape).
+- The Pushgateway never forgets series pushed to it and will expose them to Prometheus forever unless those series are manually deleted via the Pushgateway's API.
 
 ### Metrics exposed by the controller
 
-To get a better understanding about health and workings of the cluster
+To get a better understanding of health and workings of the cluster
 resources, we need to expose the following metrics:
 
 - `pending_ephemeral_runners` - Number of ephemeral runners in a pending state.
@@ -88,6 +173,7 @@ so the names are going to have `github_runner_scale_set_controller` prefix.
 
 ## Consequences
 
-Users can define alerts, monitor the behaviour of both the actions-based metrics
-(gathered from the listener) and the kubernetes resource-based metrics
+Users can define alerts, monitor the behavior of both the actions-based metrics
+(gathered from the listener) and the Kubernetes resource-based metrics
 (gathered from the controller manager).
+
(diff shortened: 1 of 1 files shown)


docs/adrs/2023-05-08-exposing-metrics.md
```diff
@@ -0,0 +1,93 @@
...
+`metrics-addr` is not empty, the controller and listeners are going to register
+the appropriate collectors.
+
+Alternatively, if there is a need to control the metrics exposure per
+`AutoscalingRunnerSet`, we can control it by applying additional label on the
+listener.
+
+### Metrics exposed by the controller
+
+To get a better understanding about health and workings of the cluster
```
TingluoHuang: i assume each metrics will be tagged with the RunnerScleSet id/name/url etc?

docs/adrs/2023-05-08-exposing-metrics.md
```diff
@@ -0,0 +1,93 @@
...
+- `running_ephemeral_runners` - Number of ephemeral runners currently running.
+  This information is helpful to see how many ephemeral runner pods are running
+  at any given time.
+- `failed_ephemeral_runners` - Number of ephemeral runners in a `Failed` state.
+  This information is helpful to catch the faulty image, or some underlying
+  problem. When the ephemeral runner controller is not able to start the
+  ephemeral runner pod after multiple retries, it will set the state of the
+  `EphemeralRunner` to failed. Since the controller can not recover from this
+  state, it can be useful to set Prometheus alerts to catch this issue quickly.
+
```
TingluoHuang, line 174: do we need metrics for EpheneralRunnerSet and AutoScalingRunnerSet level?

docs/adrs/2023-05-08-exposing-metrics.md
```diff
@@ -0,0 +1,93 @@
...
+### Metrics exposed by the `AutoscalingListener`
+
+Since the listener is responsible for communicating the state with the actions
+service, it can expose actions service related data through metrics. In
+particular:
+
+- `available_jobs` - Number of jobs with `runs-on` matching the runner scale set name. Jobs are not yet assigned but are acquired by the runner scale set.
```
TingluoHuang, line 181: `available_jobs` is for not acquired job.

docs/adrs/2023-05-08-exposing-metrics.md
```diff
@@ -0,0 +1,93 @@
...
+- `job_execution_duration_seconds` - Time spent executing workflow jobs by the scale set (in seconds).
+
+### Metric names
+
+Listener metrics belong to the `github_runner_scale_set` subsystem, so the names
+are going to have the `github_runner_scale_set_` prefix.
+
+Controller metrics belong to the `github_runner_scale_set_controller` subsystem,
+so the names are going to have `github_runner_scale_set_controller` prefix.
+
```
TingluoHuang, line 207: should we also include a section to doc all labels we will add to each metric?

nikola-jokic: Not sure, I kind of wanted to have them displayed here in case we need them. Maybe it should not even be in the ADR in case we want to extend metrics. I just included it for comments :relaxed:

docs/adrs/2023-05-08-exposing-metrics.md
nikola-jokic: I used labels that you assigned in POC which are great! We can always add or delete ones. That is the reason I said I'm not sure if we actually need to include them in the ADR. Changing labels or metrics would require an update on the ADR. But I wanted to display the basic structure and then trim it down if we need or extend it if you think we should

docs/adrs/2023-05-08-exposing-metrics.md
nikola-jokic: They are just controllers that are up. Not sure what added value will they bring but we can
EphemeralRunnerSet is exposing pending, failed and running ephemeral runners. I'd start from here and then extend them if there is a need for more metrics :relaxed:

docs/adrs/2023-05-08-exposing-metrics.md
nikola-jokic: Sorry, mistake in copying...

docs/adrs/2023-05-08-exposing-metrics.md
chrispat: It would be useful as we create documentation to have the set of labels we apply for each metric.

chrispat (dismissed):

docs/adrs/2023-05-08-exposing-metrics.md
```diff
@@ -0,0 +1,93 @@
...
+   describing the desired state.
+
+To expose metrics, we would need to create 3 additional resources:
+
+1. `ServiceMonitor` - resource used by prometheus to match namespaces and
+    services from where it needs to gather metrics
+2. `Service` for the `gha-runner-scale-set-controller` - service that will
+    target ARC controller `Deployment`
+3. `Service` for the `gha-runner-scale-set` listeners - service that will target
+    all listeners for each `AutoscalingRunnerSet`
```
Link-: I'm thinking that end users will not understand why we have 2 metrics servers for what is 1 controller in their mind. I feel like we're exposing our implementation here.

Is there no way to expose the data from the listener(s) and controller through the same service? Technically it might be challenging and/or not worth the effort, but did we at least consider what it would take?

docs/adrs/2023-05-08-exposing-metrics.md
```diff
@@ -0,0 +1,93 @@
...
+- `started_jobs_total` - Total number of jobs started.
+- `completed_jobs_total` - Total number of jobs completed.
+- `job_queue_duration_seconds` - Time spent waiting for workflow jobs to get assigned to the scale set after queueing (in seconds).
+- `job_startup_duration_seconds` - Time spent waiting for a workflow job to get started on the runner owned by the scale set (in seconds).
+- `job_execution_duration_seconds` - Time spent executing workflow jobs by the scale set (in seconds).
+
+### Metric names
+
+Listener metrics belong to the `github_runner_scale_set` subsystem, so the names
+are going to have the `github_runner_scale_set_` prefix.
```
Link-, line 203: Will we cross any label name limits with this prefix? Also why not `gh_runner_scale_set` and `gh_runner_scale_set_controller` to stay consistent with the chart names?

Link-: Also: https://prometheus.io/docs/practices/naming/#metric-names

docs/adrs/2023-05-08-exposing-metrics.md
```diff
@@ -0,0 +1,179 @@
...
+
+Two main components are driving the behavior of the scale set:
+
+1. ARC controllers responsible for managing Kubernetes resources.
+2. The `AutoscalingListener`, driver of the autoscaling solution responsible for
+   describing the desired state.
+
+We can approach publishing those metrics in 3 different ways
+
+### Expose Service Monitor, Services for listeners and a Service for the manager
```
Link-:
```suggestion
### Option 1: Expose a metrics endpoint for the controller-manager and every instance of the listener
```

docs/adrs/2023-05-08-exposing-metrics.md
```diff
@@ -0,0 +1,179 @@
...
+
+#### Cons
+
+- With a cluster running many scale sets, we are going to create a lot of
+  resources.
+- In case metrics are enabled on the controller manager level, and they should
+  be applied across all `AutoscalingRunnerSets`, it is difficult to inherit this
+  configuration by applying helm charts.
+
+### Create an aggregator service
```
Link-:
```suggestion
### Option 2: Create a single metrics aggregator service
```

docs/adrs/2023-05-08-exposing-metrics.md
```diff
@@ -0,0 +1,179 @@
...
+- We can avoid Pushgateway from Prometheus.
+
+#### Cons
+
+- Another image that we need to publish on release.
+- Change in metric configuration (on manager update) would require re-creation
+  of all listeners. This is not a big problem but is something to point out.
+- Managing requests/limits can be tricky.
+
+### Use a Prometheus Pushgateway
```
Link-:
```suggestion
### Option 3: Use a Prometheus Pushgateway
```

Link-:
```suggestion
### Option 3: Use a Prometheus Pushgateway
```

docs/adrs/2023-05-08-exposing-metrics.md
```diff
@@ -0,0 +1,179 @@
...
+Pros for running the aggregator next to the manager:
+
+- It exists side by side and is following the life cycle of the controller
+  manager
+- We don't need to introduce another controller managing the state of the pod
+
+Cons for running the aggregator next to the manager:
+
+- Crashes of the aggregator can influence the controller manager execution
+- The controller manager pod needs more resources to run
```
Link-:
```suggestion
#### Running the aggregator as a container in the controller-manager pod

**Pros:**
- It exists side by side and is following the life cycle of the controller
  manager
- We don't need to introduce another controller managing the state of the pod

**Cons**

- Crashes of the aggregator can influence the controller manager execution
- The controller manager pod needs more resources to run
```

docs/adrs/2023-05-08-exposing-metrics.md
```diff
@@ -0,0 +1,179 @@
...
+- Does not influence the controller manager pod
+- The life cycle of the metric can be controlled by the controller manager (by
+  implementing another controller)
+
+Cons for running the aggregator as a stand-alone:
+
+- We need to implement the controller that can spin up the aggregator in case of
+  the crash.
+- If we choose not to implement the controller, the resource like `Deployment`
+  can be used to manage the aggregator, but we lose control over its life cycle.
```
Link-:
```suggestion
#### Running the aggregator in a separate pod

**Pros**

- Does not influence the controller manager pod
- The life cycle of the metric can be controlled by the controller manager (by
  implementing another controller)

**Cons**

- We need to implement the controller that can spin up the aggregator in case of
  the crash.
- If we choose not to implement the controller, the resource like `Deployment`
  can be used to manage the aggregator, but we lose control over its life cycle.
```

docs/adrs/2023-05-08-exposing-metrics.md
```diff
@@ -0,0 +1,179 @@
...
+
+To expose metrics in this way, we need to:
```
Link-:
```suggestion
#### Metrics webserver requirements
```

docs/adrs/2023-05-08-exposing-metrics.md
```diff
@@ -0,0 +1,179 @@
...
+
+1. Create a web server with a single `/metrics` endpoint. The endpoint will have
+   `POST` and `GET` methods registered. The `GET` is used by Prometheus to
+   fetch the metrics, while the `POST` is going to be used by controllers and
+   listeners to publish their metrics.
+2. `ServiceMonitor` - to target the metrics aggregator service
+3. `Service` sitting in front of the web server.
+
+#### Pros
```
Link-:
```suggestion
**Pros**
```

docs/adrs/2023-05-08-exposing-metrics.md
```diff
@@ -0,0 +1,179 @@
...
+
+- This implementation requires a few additional resources to be created
+  in a cluster.
+- Web server is easy to implement and easy to document - all metrics are aggregated in a
+  single package, and the web server only needs to apply them to its state on
+  `POST`. The `GET` handler is simple.
+- We can avoid Pushgateway from Prometheus.
+
+#### Cons
```
Link-:
```suggestion
**Cons**
```

docs/adrs/2023-05-08-exposing-metrics.md
Link-: I am personally highly in favour of this approach, especially since this is the same way it was implemented in the legacy modes

docs/adrs/2023-05-08-exposing-metrics.md
mumoshu: In addition, I'd love it if we could have the following:

- An ability to configure which labels to be added to metrics via command-line flags, and
- An ability to add useful but can-be-source-of-prometheus-cardinality-issue labels (like runner id, job id, and any kind of "ids" that differ across instances of jobs = high cardinality).

Regardless of if we document every supported label in ADR, I do prefer it if the ADR clearly says that it is supposed to support dangerous-and-useful labels, and at the same time, it is going to have the ability to toggle which labels to be added. Those two would enable users to label metrics with various IDs for easier debugging and monitoring for relatively small-scale ARC deployments and to turn those ID labels off for large-scale deployments.

This old thread might provide more context about that https://github.com/actions/actions-runner-controller/issues/2176#issuecomment-1457183099

Link-:
We've done enough iterations on this, let's ship it 🚀 thank you for drafting a great doc.

Link- (approved):

PR #2569, opened 2023-05-08
Add DrainJobsMode

Link-:
### Context

Fixes https://github.com/github/c2c-actions-runtime/issues/2417

This PR adds `Drain Jobs Mode` flag (disabled by default) which will prevent the Listener and EphemeramRunnerSet from provisioning new runners whenever the gha-runner-scale-set has been patched while there are running jobs. This mode will force the controller manager to wait until all jobs have been completed before applying the changes.

Contrary to our conversations, this mode is disabled by default and will need to be explicitly enabled.

### Side quests

- ~This PR also fixes the `validate-gha-chart.yaml` workflow by using a released version of chart-tester `2.4.0` instead of master branch~
- ~Install kubebuilder-tools instead of a standalone kubebuilder in [go.yaml](https://github.com/actions/actions-runner-controller/blob/6e290fe10500e50878d512720ec308b761e8680c/.github/workflows/go.yaml)~
- These have been addressed in https://github.com/actions/actions-runner-controller/pull/2589

### Problem details

Reproduction steps:
- Setup ARC with runner image tag set to latest
- Run N number of jobs
- Change the runner image tag to a specific version
- Helm upgrade

### TODO

- [x] This PR needs to be merged as well: https://github.com/actions-runner-controller/arc_e2e_test_dummy/pull/3
- [x] This PR needs to be merged: https://github.com/actions-runner-controller/arc_e2e_test_dummy/pull/5


main.go
```diff
@@ -241,10 +248,12 @@ func main() {
...
 		}
+
 		if err = (&actionsgithubcom.AutoscalingListenerReconciler{
-			Client: mgr.GetClient(),
-			Log:    log.WithName("AutoscalingListener"),
-			Scheme: mgr.GetScheme(),
+			Client:        mgr.GetClient(),
+			Log:           log.WithName("AutoscalingListener"),
+			Scheme:        mgr.GetScheme(),
+			DrainJobsMode: drainJobsMode,
```
Link-: Remove this not needed

cmd/githubrunnerscalesetlistener/autoScalerService.go
```diff
@@ -12,10 +12,11 @@ import (
...
 type ScaleSettings struct {
-	Namespace    string
-	ResourceName string
-	MinRunners   int
-	MaxRunners   int
+	Namespace     string
+	ResourceName  string
+	MinRunners    int
+	MaxRunners    int
+	DrainJobsMode bool
```
Link-: Remove everything from the listener, not needed

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -218,35 +220,16 @@ func (r *AutoscalingRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl
...
-		if err != nil {
-			log.Error(err, "Failed to clean up old runner sets")
-			return ctrl.Result{}, err
-		}
-	}
-
 	// Make sure the AutoscalingListener is up and running in the controller namespace
 	listener := new(v1alpha1.AutoscalingListener)
 	if err := r.Get(ctx, client.ObjectKey{Namespace: r.ControllerNamespace, Name: scaleSetListenerName(autoscalingRunnerSet)}, listener); err != nil {
 		if kerrors.IsNotFound(err) {
```
nikola-jokic: If we fail to fetch listener, we should return with an error. It is likely that the API call failed, and we need to reconcile again

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -218,35 +220,16 @@ func (r *AutoscalingRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl
...
-			// We don't have a listener
-			log.Info("Creating a new AutoscalingListener for the runner set", "ephemeralRunnerSetName", latestRunnerSet.Name)
-			return r.createAutoScalingListenerForRunnerSet(ctx, autoscalingRunnerSet, latestRunnerSet, log)
+			log.Info("AutoscalingListener does not exist.")
```
nikola-jokic: We should keep the state of not found in a variable instead of checking with `reflect.DeepEqual`
```go
	listener := new(v1alpha1.AutoscalingListener)
        listenerFound := true
	if err := r.Get(ctx, client.ObjectKey{Namespace: r.ControllerNamespace, Name: scaleSetListenerName(autoscalingRunnerSet)}, listener); err != nil {
		if kerrors.IsNotFound(err) {
			log.Info("AutoscalingListener does not exist.")
                        listenerFound = false
		}
	}
```
Then it becomes:
```go
if listenerFound && listener.Labels[labelKeyRunnerSpecHash] != autoscalingRunnerSet.ListenerSpecHash() {
// delete listener
}
if listenerFound && desiredSpecHash != latestRunnerSet.Labels[labelKeyRunnerSpecHash] {
    if r.drainingJobs(&latestRunnerSet.Status) {
    // patch
    }
    // create ephemeral runner set
}
// old runner set checks

if !listenerFound  {
    if r.drainingJobs(&latestRunnerSet.Status) {
           // log and return
    }
    // create 
}
// update status
```

Link-: Smart! I like it much more efficient

Link-: Fixed

controllers/actions.github.com/autoscalingrunnerset_controller.go
Link-: Fixed

main.go
```diff
@@ -131,6 +132,7 @@ func main() {
 	flag.StringVar(&logLevel, "log-level", logging.LogLevelDebug, `The verbosity of the logging. Valid values are "debug", "info", "warn", "error". Defaults to "debug".`)
 	flag.StringVar(&logFormat, "log-format", "text", `The log format. Valid options are "text" and "json". Defaults to "text"`)
 	flag.BoolVar(&autoScalingRunnerSetOnly, "auto-scaling-runner-set-only", false, "Make controller only reconcile AutoRunnerScaleSet object.")
+	flag.BoolVar(&drainJobsMode, "drain-jobs-mode", false, "Wait for jobs to finish before mutating resources.")
```
nikola-jokic: Suggestion:

Should we use `update-strategy` as a flag and assign strategies to it?
To turn on draining: `--update-strategy=drain`
To use eventual replace: `--update-strategy=replace` 
Or something similar. We can always have a default value if no strategy is specified

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -218,35 +219,21 @@ func (r *AutoscalingRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl
...
 	listener := new(v1alpha1.AutoscalingListener)
+	listenerFound := true
 	if err := r.Get(ctx, client.ObjectKey{Namespace: r.ControllerNamespace, Name: scaleSetListenerName(autoscalingRunnerSet)}, listener); err != nil {
 		if kerrors.IsNotFound(err) {
-			// We don't have a listener
-			log.Info("Creating a new AutoscalingListener for the runner set", "ephemeralRunnerSetName", latestRunnerSet.Name)
-			return r.createAutoScalingListenerForRunnerSet(ctx, autoscalingRunnerSet, latestRunnerSet, log)
+			listenerFound = false
+			log.Info("AutoscalingListener does not exist.")
+		} else {
```
nikola-jokic: Since this branch is early return, how about:
```go
if !kerrors.IsNotFound(err) {
    log.Error(err, "Failed to get AutoscalingListener resource")
    return ctrl.Result{}, err
}
listenerFound = false
log.Info("AutoscalingListener does not exist.")
```

main.go
Link-:
```
  ## Defines how the controller should handle upgrades while having running jobs.
  ##
  ## The srategies available are:
  ## - "immediate": (default) The controller will immediately apply the change causing the
  ##   recreation of the listener and ephemeral runner set. This can lead to an 
  ##   overprovisioning of runners, if there are pending / running jobs. This should not
  ##   be a problem at a small scale, but it could lead to a significant increase of 
  ##   resources if you have a lot of jobs running concurrently.
  ## 
  ## - "eventual": The controller will remove the listener and ephemeral runner set 
  ##   immediately, but will not recreate them (to apply changes) until all 
  ##   pending / running jobs have completed.
  ##   This can lead to a longer time to apply the change but it will ensure
  ##   that you don't have any overprovisioning of runners.
  updateStrategy: "immediate"
```

nikola-jokic: This looks great! :+1:

controllers/actions.github.com/autoscalingrunnerset_controller.go
Link-: This makes it hard to read for the sake of removing an else.. 😄 but I'll bite

Link-: Double negatives ..

nikola-jokic: Oh yes, we have it in multiple places :joy: Maybe I'm just used to it :smile:

Link-: Fixed

main.go
Link-: Fixed

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -260,6 +247,44 @@ func (r *AutoscalingRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl
 		return ctrl.Result{}, nil
 	}
 
+	if desiredSpecHash != latestRunnerSet.Labels[labelKeyRunnerSpecHash] {
+		if r.drainingJobs(&latestRunnerSet.Status) {
```
mumoshu, line 269: Hey @Link-! I guess this might result in an upgrade never being completed when the old listener continuously received new workflow jobs... right?

If that's the case, can we instead do the following:

- Create a new listener immediately
- Delete the old listener (Can this result in the old ephemeralrunnerset to never drain? If that's the case this won't work...
- Wait for the latest ephemeralrunnerset to drain
- Delete the old ephemeralrunnerset

The idea is to prepare the new ephemeralrunnerset and the listener first so that the new workflow jobs immediately go to new runners, where workflow jobs that are already running on old runners are unaffected.

nikola-jokic: Hey @mumoshu 

The issue is that the listener would not be aware that the old ephemeral runner set is scaling down. 

Let's say that the old ephemeral runner set has 3 jobs and as soon as they are done, it will scale down. If we replace the listener and prepare a new runner set, the listener will patch the new runner set with the desired count. It should be up to the ephemeral runner set controller to create ephemeral runners based on the difference between the desired state, and the current state. But since the new ephemeral runner set is unaware of the old one, it will over provision the number of ephemeral runners (current state for the new ephemeral runner set is 0, desired state is 6, but we still have 3 ephemeral runners running from the old ephemeral runner set).

This was one of the approaches to drain first.

Thinking about your approach, I just realized another way we can implement it. We could extend the EphemeralRunnerSet CRD to contain the number of old ephemeral runners running in a different ephemeral runner set. That one could be controlled by the autoscaling runner set. On status update of the old ephemeral runner set, autoscaling runner set is going to be notified, and it can issue the patch to the newest ephemeral runner set and the replacement in that case would be faster.

The ephemeral runner set can calculate the number of runners needed based on `.spec.replicas`, current state and the previous state. What do you think (@Link-, @mumoshu)?

Link-: > Hey @Link-! I guess this might result in an upgrade never being completed when the old listener continuously received new workflow jobs... right?

@mumoshu - When the upgrade happens, the listener is removed so that it won't process any new jobs and the existing ephemeral runner set replicas is set to 0. This "should" in principle block new jobs from being process and they'll be queueing until a listener is back up!

Once the jobs are completed, the listener is recreated and a new ephemeral runner set spawns which will be able to pickup the jobs that have queued (if they're still available).

With that said, this was never tested at scale (for example when a new job is dispatched every second). I don't know how this will behave with that amount of load.

> The idea is to prepare the new ephemeralrunnerset and the listener first so that the new workflow jobs immediately go to new runners, where workflow jobs that are already running on old runners are unaffected.

That's actually what happens by default right now :) and it's what leads to the over-provisioning issue

Link-: > Thinking about your approach, I just realized another way we can implement it. We could extend the EphemeralRunnerSet CRD to contain the number of old ephemeral runners running in a different ephemeral runner set. That one could be controlled by the autoscaling runner set. On status update of the old ephemeral runner set, autoscaling runner set is going to be notified, and it can issue the patch to the newest ephemeral runner set and the replacement in that case would be faster.

@nikola-jokic - I'm afraid we'll be opening pandora's box here for so much unexpected behaviour. Let's keep it simple and stupid for now, see how it performs in the field and re-iterate based on that.

nikola-jokic (dismissed):
LGTM

.github/actions/execute-assert-arc-e2e/action.yaml
```diff
@@ -23,6 +23,14 @@ inputs:
...
     description: 'The namespace of the configured gha-runner-scale-set-controller'
     required: true
+  wait-to-finish:
+    description: 'Wait for the workflow run to finish'
+    required: true
+    default: "true"
+  wait-to-running:
+    description: 'Wait for the workflow run to start running'
+    required: true
+    default: "false"
```
Link-, line 33: These have been added to make the action a bit more generic. Sometimes, we just want to trigger the workflow run and not necessarily wait until it finishes successfully. These inputs allow for that.

.github/actions/execute-assert-arc-e2e/action.yaml
```diff
@@ -118,7 +126,36 @@ runs:
...
+              run_id: workflow_run_id
+            })
+            console.log(`${getRunResponse.data.html_url}: ${getRunResponse.data.status} (${getRunResponse.data.conclusion})`);
+            if (getRunResponse.data.status == 'in_progress') {
+              console.log(`Workflow run is in progress.`)
+              return
+            }
+          }
+          core.setFailed(`The triggered workflow run didn't start properly using ${{inputs.arc-name}}`)
+
```
Link-, line 156: This waits until the workflow run is `in_progress` as opposed to `finished`

main.go
```diff
@@ -131,6 +132,7 @@ func main() {
 	flag.StringVar(&logLevel, "log-level", logging.LogLevelDebug, `The verbosity of the logging. Valid values are "debug", "info", "warn", "error". Defaults to "debug".`)
 	flag.StringVar(&logFormat, "log-format", "text", `The log format. Valid options are "text" and "json". Defaults to "text"`)
 	flag.BoolVar(&autoScalingRunnerSetOnly, "auto-scaling-runner-set-only", false, "Make controller only reconcile AutoRunnerScaleSet object.")
+	flag.StringVar(&updateStrategy, "update-strategy", "immediate", "Immediately or eventually mutate resources on upgrade with running/pending jobs.")
```
nikola-jokic: Might be nice to provide more hints in flag docs :relaxed: 
```suggestion
	flag.StringVar(&updateStrategy, "update-strategy", "immediate", `Strategy for mutating resources on upgrade with running/pending jobs. (values: "eventual", "immediate"; default: "immediate")`
```

main.go
```diff
@@ -169,6 +171,10 @@ func main() {
 		if len(watchSingleNamespace) > 0 {
 			newCache = cache.MultiNamespacedCacheBuilder([]string{managerNamespace, watchSingleNamespace})
 		}
+
+		if len(updateStrategy) > 0 {
+			log.Info("update-strategy is set to: ", "updateStrategy", updateStrategy)
+		}
```
nikola-jokic:
```suggestion
		switch updateStrategy {
		case "eventual", "immediate":
		default:
			log.Info(`Update strategy not recognized. Defaulting to "immediately"`, "updateStrategy", updateStrategy)
			updateStrategy = "immediate"
		}
```

controllers/actions.github.com/autoscalingrunnerset_controller.go
```diff
@@ -57,6 +57,7 @@ type AutoscalingRunnerSetReconciler struct {
 	ControllerNamespace                           string
 	DefaultRunnerScaleSetListenerImage            string
 	DefaultRunnerScaleSetListenerImagePullSecrets []string
+	UpdateStrategy                                string
```
nikola-jokic: It would be awesome if we create a type for update strategy.
```go
type UpdateStrategy string
// Update strategies
const (
        // UpdateStrategyImmediate will not recreate resources until all
        // pending / running jobs have completed.
        // This can lead to a larger time to apply the change but it will ensure
        // that you don't have any overprovisioning of runners.
        UpdateStrategyImmediate = UpdateStrategy("immediate")
        // docs for update strategy
        UpdateStrategyEventual = UpdateStrategy("eventual")
)
```

And then the UpdateStrategy fields is of type `UpdateStrategy`. We can document the strategies on package level

main.go
Link-: Fixed, thank you

main.go
Link-: Fixed

controllers/actions.github.com/autoscalingrunnerset_controller.go
Link-: Fixed

controllers/actions.github.com/autoscalingrunnerset_controller_test.go
```diff
@@ -1617,10 +1722,14 @@ var _ = Describe("Test resource version and build version mismatch", func() {
...
-		}).Should(BeTrue())
+		Eventually(
+			func() bool {
+				ars := new(v1alpha1.AutoscalingRunnerSet)
+				err := k8sClient.Get(ctx, types.NamespacedName{Namespace: autoscalingRunnerSet.Namespace, Name: autoscalingRunnerSet.Name}, ars)
+				return errors.IsNotFound(err)
+			},
+			autoscalingRunnerSetTestTimeout,
+			autoscalingRunnerSetTestInterval,
+		).Should(BeTrue())
```
Link-, line 1733: @nikola-jokic I added these to this test you wrote. This test was timing out randomly: https://github.com/actions/actions-runner-controller/actions/runs/5047834860/jobs/9073153675

nikola-jokic (approved):
LGTM :rocket:

PR #2582, opened 2023-05-12
fix: labels on github webhook service template

vrivellino:
Similar to the fix contained in #2504 (specifically aefc024e35105d809a5a2bc2ff2a599cddb06eac)


mumoshu (approved):
@vrivellino Awesome! Thanks for fixing it ☺️

PR #2585, opened 2023-05-13
Make `github_app_installation_id` optional

### LINKED ISSUE #2584: Make `github_app_installation_id` optional

`github_app_installation_id` is currently required. This creates an overhead since k8s users who are allowed to create RunnerSet - most often are not allowed to create Secrets. Requiring `github_app_installation_id` also requires to create a separate secret per each organization.

Also, it doesn't seem like webhook-based scaling can be used at all right now in a multi-tenancy mode, unless one install apps and have multiple webhook deployments.

### Why is this needed?

To allow users to self-register their runners in my cluster securely. To support webhook based scaling.

### Additional context

https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app explains how to authenticate to GitHub "as app". The process is as simple as generating a local JWT and signing it with that same private key from `github_app_private_key`. The token has to be expiring within 15 minutes. Then you can list all installations of the app using a request like this:

```
curl --request GET \
--url "https://api.github.com/app/installations" \
--header "Accept: application/vnd.github+json" \
--header "Authorization: Bearer YOUR_JWT" \
--header "X-GitHub-Api-Version: 2022-11-28"
```

In the response you can find `github_app_installation_id` mapping to the respective organizations by their slug. Therefore the `github_app_id` and the `github_app_private_key` is sufficient enough information, and that one secret can be reused across as many organizations as needed as long as app is installed on them.

dee-kryvenko:
Fixes #2584

https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app explains how to authenticate to GitHub "as app". The process is as simple as generating a local JWT and signing it with that same private key from github_app_private_key. The token has to be expiring within 15 minutes. Then you can list all installations of the app using a request like this:

```
curl --request GET \
--url "https://api.github.com/app/installations" \
--header "Accept: application/vnd.github+json" \
--header "Authorization: Bearer YOUR_JWT" \
--header "X-GitHub-Api-Version: 2022-11-28"
```

In the response you can find github_app_installation_id mapping to the respective organizations by their slug. Therefore the github_app_id and the github_app_private_key is sufficient enough information, and that one secret can be reused across as many organizations as needed as long as app is installed on them.

### DIFF
--- a/github/actions/multi_client.go
+++ b/github/actions/multi_client.go
@@ -2,11 +2,16 @@ package actions
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"io"
+	"net/http"
 	"strconv"
 	"sync"
+	"time"
 
 	"github.com/go-logr/logr"
+	jwt "github.com/golang-jwt/jwt/v4"
 )
 
 type MultiClient interface {
@@ -40,6 +45,15 @@ type ActionsAuth struct {
 type ActionsClientKey struct {
 	Identifier string
 	Namespace  string
+}
+
+type AppInstallationAccount struct {
+	Login string `json:"login"`
+}
+
+type AppInstallation struct {
+	ID      int64                  `json:"id"`
+	Account AppInstallationAccount `json:"account"`
 }
 
 func NewMultiClient(userAgent string, logger logr.Logger) MultiClient {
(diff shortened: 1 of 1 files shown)


dee-kryvenko:
I have very little context, like how often is this function is called and is it efficient enough to list installations in that place, or does it require some sort of caching. Would appreciate any further guidance as to what needs to be done to make this ready to be merged.

dee-kryvenko:
Never mind, I see now it is not that easy as I hoped. There's at least 3 different github clients and after an all-nighter on this I give up.

PR #2590, opened 2023-05-15
Add RUNNER_NO_DEFAULT_LABELS flag for --no-default-labels

7Hazard:
As per https://github.com/actions/runner/pull/2443 addition, this gives the ability to pass RUNNER_NO_DEFAULT_LABELS to not include default labels to the runners such as `Linux` and `self-hosted`.


mumoshu:
@7Hazard Hey! Thanks for submitting this. First of all, may I ask for your motivation for doing this?

The current ARC's webhook-based autoscaler assumes the runner to have those labels and would misbehave (like autoscaling wrong RunnerDeployments) if there are no default labels present in runners. So, if you really need this, you would need to update the ARC's autoscaler code too to make it work in concert with runners without the default labels.

Could it be your goal?

On the other hand, the new ARC RunnerScaleSet controller creates runners without those default labels. If your goal is to have runners without the default labels, you could just give the new RunnerScaleSet a shot, without modifying anything. The new controller does not use `config.sh` to register runners so no need to modify the config_args you touched in this PR.

7Hazard:
If the new RunnerScaleSet allows runners without these default labels, then I will give that a try and keep this branch as a draft. If I see the need for the older way of working without default labels, I will attempt to adapt the autoscaler in this branch.

brconnell4:
By RunnerScaleSet do you mean the RunnerSet? 
https://github.com/actions/actions-runner-controller/blob/master/docs/deploying-arc-runners.md ?

mumoshu:
@brconnell4 RunnerScaleSet is the CRD supported by the new version of ARC! See https://github.com/actions/actions-runner-controller/discussions/2775

RunnerScaleSet is something entirely different from RunnerSet. RunnerSet is old ARC only stateful version of RunnerDeployment.

nkomm00:
when will this PR get merged? and ETA?

appgate-miguel-ampudia:
Why was this pr closed? The functionality no longer works or is there an alternative for this?

PR #2607, opened 2023-05-18
Discard logs on helm chart tests

nikola-jokic:
Logger pollutes the stdout, and it is hard to find the failing test. This change should discard the rendered chart in log and make it easy to find the failing test. 

Fixes https://github.com/github/c2c-actions-runtime/issues/2401


Link- (approved):
🚀

PR #2617, opened 2023-05-23
Adapt role name to prevent namespace collision

timmjd:
The current deployment is not working for a single namespace deployment of the `gha-runner-scale-set-controller`.

This is due to the `manager_single_namespace_controller_role` and the `manager_single_namespace_watch_role` would both result in the same role name: (`{{ include "gha-runner-scale-set-controller.managerSingleNamespaceRoleName" . }}`) - leading to the error: `roles.rbac.authorization.k8s.io "xxx-gha-runner-scale-set-controller-manager-single-namespace-role" already exists`

This fix does rename the `watcher` role. By this, a single namespace deployment should be possible.


nikola-jokic (approved):
LGTM

Link-:
Fixes https://github.com/actions/actions-runner-controller/issues/2668

PR #2637, opened 2023-05-31
fix: allow resource requests in the dind container

apr-1985:
Currently there is no way to set resource requests and limits on the DinD pod. 
It is possible for the runner pod so using the same mechanism to allow it for DinD so that it cannot take over nodes.

### DIFF
--- a/charts/gha-runner-scale-set/templates/_helpers.tpl
+++ b/charts/gha-runner-scale-set/templates/_helpers.tpl
@@ -88,6 +88,15 @@ volumeMounts:
 
 {{- define "gha-runner-scale-set.dind-container" -}}
 image: docker:dind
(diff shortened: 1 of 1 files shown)


apr-1985:
Would allow something like this to be added to the users values file
```
template:
  spec:
    containers:
      - name: runner
        resources:
          limits:
            cpu: "0.5"
            memory: "2Gi"
          requests:
            cpu: "0.25"
            memory: "1Gi"
      - name: dind
        resources:
          limits:
            cpu: "0.5"
            memory: "2Gi"
          requests:
            cpu: "0.25"
            memory: "1Gi"

```

kahirokunn:
Similar: https://github.com/actions/actions-runner-controller/pull/2586

nikola-jokic:
Closing this one as a duplicate

PR #2654, opened 2023-06-07
add interval and timeout configuration for the actions-runner-controler serviceMonitors

arielly-parussulo:
## What?
Allow the customization of the interval and the timeout of the ServiceMonitors scraping.

## Why?
I work in a company that has lots of workflows running and the controller generates lots of metrics in the /metrics path. When prometheus try to scrap the content of /metrics, it starts to face timeouts. 
The default values for this chart are the default values for the Prometheus rules.

## How?
By adding two new fields to the **serviceMonitor** map: **interval** and **timeout**. When the serviceMonitor is enabled (true), the serviceMonitor will be generated with these two properties.


eliasscosta (approved):

charts/actions-runner-controller/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.23.3
+version: 0.24.0
```
mumoshu:
```suggestion
version: 0.23.3
```

It would be great if you could let us batch some more changes before the next minor release!

mumoshu (approved):
LGTM. Thanks a lot for your contribution @arielly-parussulo! ☺️

PR #2657, opened 2023-06-09
Trim repo/org/enterprise to 63 characters in label values

nikola-jokic:


controllers/actions.github.com/resourcebuilder.go
```diff
@@ -545,14 +543,23 @@ func applyGitHubURLLabels(url string, labels map[string]string) error {
...
-		labels[LabelKeyGitHubRepository] = githubConfig.Repository
+		labels[LabelKeyGitHubRepository] = trimLabelValue(githubConfig.Repository)
 	}
 
 	return nil
 }
+
+const trimLabelVauleSuffix = "-trim"
+
+func trimLabelValue(val string) string {
```
TingluoHuang, line 560: do we have anything else need to trim? both in code and helm chart, might worth to double check all labels on an installed ARC.

TingluoHuang (approved):

controllers/actions.github.com/resourcebuilder.go
nikola-jokic: I'll test it :smile:

PR #2670, opened 2023-06-14
Create a new makefile for scaleset

nikola-jokic:
Fixes https://github.com/github/c2c-actions-runtime/issues/2425

Create a separate makefile that will help with building scalesets.

It also generates the role only required for `actions.github.com` since the controller is running using `actions.github.com` or `actions.summerwind.net` groups


gha.Makefile
```diff
@@ -0,0 +1,127 @@
+ifdef DOCKER_USER
```
Link-, line 1: I'm having some doubts about the name:

1. the editor will not recognise this as a Makefile and it's one or two extra steps to configure that
2. we might not need to go for this approach because we can namespace our targets, for example:

Instead of `fmt:` we could do `gha/fmt:`. And then in the original make file we can just include this file, for example:

```
-include ....
```

What do you think (I'm not convinced of this approach either, but let's have the conversation)?

gha.Makefile
```diff
@@ -0,0 +1,127 @@
...
+	DOCKER_IMAGE_NAME ?= ${DOCKER_USER}/actions-runner-controller
```
Link-, line 2: I really dislike the usage of `DOCKER_USER` in the original one, I think the more appropriate term would be: `CONTAINER_REGISTRY` and instead of `DOCKER_IMAGE_NAME` maybe we use `CONTROLLER_IMAGE_NAME`?

gha.Makefile
```diff
@@ -0,0 +1,127 @@
...
+ifeq (${IMG_RESULT}, load)
+	export PUSH_ARG="--load"
+	# if load is specified, image will be built only for the build machine architecture.
+	export PLATFORMS="local"
+else ifeq (${IMG_RESULT}, cache)
+	# if cache is specified, image will only be available in the build cache, it won't be pushed or loaded
+	# therefore no PUSH_ARG will be specified
+else
+	export PUSH_ARG="--push"
+endif
```
Link-, line 29: It's not very clear how this can be used 🤔 we either need to document the whole flow or use another approach

gha.Makefile
```diff
@@ -0,0 +1,127 @@
...
+
+.PHONY: help
+help: ## Display this help.
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+
+.PHONY: fmt
+fmt: ## Run go fmt.
```
Link-, line 36: The comment here doesn't say anything new that the command doesn't reflect already

gha.Makefile
```diff
@@ -0,0 +1,127 @@
...
+	go fmt ./...
+
+.PHONY: vet
+vet: ## Run go vet.
```
Link-, line 40: The comment here doesn't say anything new that the command doesn't reflect already

gha.Makefile
```diff
@@ -0,0 +1,127 @@
...
+	go vet ./...
+
+.PHONY: lint
+lint: # Run golangci-lint in docker
+	docker run --rm -v $(PWD):/app -w /app golangci/golangci-lint:v1.52.2 golangci-lint run
+
+.PHONY: test-ctrl
+test-ctrl: manifests generate fmt vet envtest ## Run test on actions.github.com controllers.
```
Link-, line 48:
```suggestion
test-controllers: manifests generate fmt vet envtest ## Run test on actions.github.com controllers.
```

gha.Makefile
```diff
@@ -0,0 +1,127 @@
...
+
+.PHONY: vet
+vet: ## Run go vet.
+	go vet ./...
+
+.PHONY: lint
+lint: # Run golangci-lint in docker
+	docker run --rm -v $(PWD):/app -w /app golangci/golangci-lint:v1.52.2 golangci-lint run
+
+.PHONY: test-ctrl
```
Link-, line 47:
```suggestion
.PHONY: test-controllers
```

gha.Makefile
```diff
@@ -0,0 +1,127 @@
...
+
+KUSTOMIZE ?= $(LOCALBIN)/kustomize
+CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
+ENVTEST ?= $(LOCALBIN)/setup-envtest
+
+## Tool Versions
+KUSTOMIZE_VERSION ?= v5.0.0
+CONTROLLER_TOOLS_VERSION ?= v0.11.3
+
+KUSTOMIZE_INSTALL_SCRIPT ?= "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"
```
Link-, line 107: nit: I feel like all of these URLs / versions / container image references, need to be bundled in one place (preferably) at the top so that we can easily revise them when upgrades are needed.

gha.Makefile
```diff
@@ -0,0 +1,127 @@
...
+	test -s $(LOCALBIN)/kustomize || { curl -Ss $(KUSTOMIZE_INSTALL_SCRIPT) --output install_kustomize.sh && bash install_kustomize.sh $(subst v,,$(KUSTOMIZE_VERSION)) $(LOCALBIN); rm install_kustomize.sh; }
+
+.PHONY: controller-gen
+controller-gen: $(CONTROLLER_GEN) ## Download controller-gen locally if necessary. If wrong version is installed, it will be overwritten.
+$(CONTROLLER_GEN): $(LOCALBIN)
+	test -s $(LOCALBIN)/controller-gen && $(LOCALBIN)/controller-gen --version | grep -q $(CONTROLLER_TOOLS_VERSION) || \
+	GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_TOOLS_VERSION)
+
+.PHONY: envtest
+envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
```
Link-, line 124: What do you think about adding a new target to install all the tools needed?

Link- (changes requested):
Thank you for this much needed cleanup!

config/crd/bases/actions.github.com_autoscalinglisteners.yaml
```diff
@@ -14,132 +15,136 @@ spec:
     singular: autoscalinglistener
   scope: Namespaced
   versions:
-    - additionalPrinterColumns:
```
Link-, line 17: Why do we need to change the CRDs in this PR?

gha.Makefile
nikola-jokic: It serves as a documentation when you just run make
Then it will tell you what does the `make fmt` do (You can possibly have different formatters (gofumpt etc)

config/crd/bases/actions.github.com_autoscalinglisteners.yaml
nikola-jokic: I don't, but I thought it would generate different ones on the next pr or on some completely unrelated one. So I thought, since we are changing controller runtime version for this makefile, it should be better to change charts now, than in an unrelated PR :smile:

Link-:
We'll come back for this when we're ready for it.

PR #2671, opened 2023-06-14
fix: remove callbacks resulting in scales due to incomplete response

### LINKED ISSUE #2659: Pull-based autoscaling causes unrelated scale-ups due to incomplete GitHub response
### Checks

### Controller Version

0.27.4

### Helm Chart Version

0.23.3

### CertManager Version

1.21.1

### Deployment Method

Kustomize

### cert-manager installation

- Yes
- Yes

### Checks

### Resource Definitions

```yaml
---
kind: RunnerDeployment
metadata:
  labels:
    app: gha-runner-controller
  name: aws-core-2-unstable
  namespace: gha-runner
spec:
  template:
    spec:
      env:
      - name: DISABLE_RUNNER_UPDATE
        value: 'true'
      ephemeral: true
      image: summerwind/actions-runner:v2.304.0-ubuntu-22.04
      labels:
      - aws-core-2-default
      nodeSelector:
        nodegroup: core-8-unstable
      organization: myorg
      resources:
        limits:
          cpu: '1.8'
          memory: 6Gi
        requests:
          cpu: '1.8'
          memory: 6Gi
      serviceAccountName: agent
      volumeMounts:
      - mountPath: /etc/arc/hooks/job-started.d/label-podself.sh
        name: actions-runner-agent-config
        subPath: label-podself.sh
      volumes:
      - configMap:
          defaultMode: 493
          name: actions-runner-agent-config
        name: actions-runner-agent-config
---
metadata:
  labels:
    app: gha-runner-controller
  name: aws-core-2-unstable
  namespace: gha-runner
spec:
  maxReplicas: 20
  metrics:
  - repositoryNames:
    - repo1
    - repo2
    - repo3
    type: TotalNumberOfQueuedAndInProgressWorkflowRuns
  minReplicas: 0
  scaleDownDelaySecondsAfterScaleOut: 60
  scaleTargetRef:
    name: aws-core-2-unstable
```

### To Reproduce

```markdown
1. Use RunnerDeployment with HorizontalRunnerAutoscaler. Use a label that not a single workflow is using since we want to showcase an unrelated trigger.
2. Only use pull-based scaling by using `TotalNumberOfQueuedAndInProgressWorkflowRuns` and defining some repositories. The more active those repositories the better. Active repositories = lots of GHA action is important since the controller is checking every single run whether it's of relevance and this introduced the unwanted behaviour.
3. Wait for an unrelated scale-up to happen. This is triggered by the GitHub API. Either it's an edge case in their API or eventual consistency resulting in them returning an empty Job Array for a run that actually has jobs.
```

### Describe the bug

GitHub's API is not always returning valid responses, while those are 200, they're empty.
The issue happens in the [listWorkflowJobs](https://github.com/actions/actions-runner-controller/blob/master/controllers/actions.summerwind.net/autoscaling.go#LL122C2-L122C18) function.
Two edge cases are caught by checking whether the `runID == 0` or whether the GH API response contains no jobs `len(allJobs) == 0`. In both cases, a callback is executed, which doesn't make any sense.
Neither case should trigger the callback of increasing the `queued` or `in_progress` counter, which ultimately determines the suggested desired amount of runners.
Ultimately, this results in unrelated RunnerDeployments being scaled since any workflow request can return an empty job array.

If the Job Array from the GitHub API response is empty, we should not blindly scale RunnerDeployments but rather ignore it and let the next pull scrape check again.

### Whole Controller Logs

```shell
I added my own debugging to figure out what was going on.
I changed the repo and job / run id. I checked them, they do contain jobs and they were some GitHub hosted label `ubuntu-`.


2023-06-08T12:59:12Z	INFO	horizontalrunnerautoscaler	Checking workflows for queued
2023-06-08T12:59:12Z	INFO	horizontalrunnerautoscaler	Found the following runnerLabels	{"runner_labels": ["aws-core-2-default"], "runner_name": "aws-core-2-unstable", "repository": "repo"}
2023-06-08T12:59:12Z	INFO	horizontalrunnerautoscaler	Triggered CallBack - AllJobs == 0	{"run_id": 111111111, "repository": "repo"}
2023-06-08T12:59:12Z	INFO	horizontalrunnerautoscaler	Checking workflows for queued
2023-06-08T12:59:12Z	INFO	horizontalrunnerautoscaler	Found the following runnerLabels	{"runner_labels": ["aws-core-2-default"], "runner_name": "aws-core-2-unstable", "repository": "repo"}
2023-06-08T12:59:12Z	INFO	horizontalrunnerautoscaler	Detected a job	{"labels": ["ubuntu-22.04"], "run_id": 111111111, "job_id": 111111111, "repository": "repo"}
2023-06-08T12:59:12Z	INFO	horizontalrunnerautoscaler	Checking workflows for queued
2023-06-08T12:59:12Z	INFO	horizontalrunnerautoscaler	Found the following runnerLabels	{"runner_labels": ["aws-core-2-default"], "runner_name": "aws-core-2-unstable", "repository": "repo"}
2023-06-08T12:59:13Z	INFO	horizontalrunnerautoscaler	Detected a job	{"labels": ["ubuntu-22.04"], "run_id": 111111111, "job_id": 2222222222, "repository": "repo"}
2023-06-08T12:59:13Z	INFO	horizontalrunnerautoscaler	Checking workflows for queued
2023-06-08T12:59:13Z	INFO	horizontalrunnerautoscaler	Found the following runnerLabels	{"runner_labels": ["aws-core-2-default"], "runner_name": "aws-core-2-unstable", "repository": "repo"}
2023-06-08T12:59:13Z	INFO	horizontalrunnerautoscaler	Triggered CallBack - AllJobs == 0	{"run_id": 111111111, "repository": "repo"}
2023-06-08T12:59:16Z	DEBUG	horizontalrunnerautoscaler	Suggested desired replicas of 2 by TotalNumberOfQueuedAndInProgressWorkflowRuns	{"workflow_runs_completed": 0, "workflow_runs_in_progress": 0, "workflow_runs_queued": 2, "workflow_runs_unknown": 0, "namespace": "gha-runner-camunda", "kind": "runnerdeployment", "name": "aws-core-2-unstable", "horizontal_runner_autoscaler": "aws-core-2-unstable"}
2023-06-08T12:59:16Z	DEBUG	horizontalrunnerautoscaler	Calculated desired replicas of 2	{"horizontalrunnerautoscaler": "gha-runner-camunda/aws-core-2-unstable", "suggested": 2, "reserved": 0, "min": 0, "max": 20}
```
```


### Whole Runner Pod Logs

```shell
Runner Pod Logs are irrelevant since it's about the controller.
```

### Additional Context

I found this [discussion](https://github.com/actions/actions-runner-controller/commit/a63860029a85ac35c2839c19d3bcf3b5294377e5#r67617091) about the callback stuff.
I checked the code and I don't see a reason for keeping the callback behaviour and would suggest for both edge cases of `allJobs == 0` and `runID == 0` to simply ignore them rather than scaling up RunnerDeployments.

Langleu:
Fixes https://github.com/actions/actions-runner-controller/issues/2659

It removes the callbacks in the `listWorkflowJobs` function, which is only used by 2 edge cases, which shouldn't scale the RunnerDeployment.
GitHub requests can return an empty job array for a run even though it contains jobs. This results in unpredicted scale-ups of unrelated RunnerDeployments since no labels are checked but just blindly scaled.

I can also rework it into a feature flag to maintain the same functionality, but I currently don't see the use case.
The logging may be helpful for users to figure out stuck jobs, e.g. when GitHub Actions has an internal error, workflows can get stuck and it may not be noticeable to users. Therefore, helps to cancel those jobs once and for all that would otherwise trigger the previous edge case.

### DIFF
--- a/controllers/actions.summerwind.net/autoscaling_test.go
+++ b/controllers/actions.summerwind.net/autoscaling_test.go
@@ -465,7 +465,7 @@ func TestDetermineDesiredReplicas_OrganizationalRunner(t *testing.T) {
 		err          string
 	}{
 		// case_0
-		// 2 demanded, max at 3
+		// 0 demanded due to zero runID, min at 2
 		{
 			org:                      "test",
 			repos:                    []string{"valid"},
--- a/controllers/actions.summerwind.net/autoscaling_test.go
+++ b/controllers/actions.summerwind.net/autoscaling_test.go
@@ -76,7 +76,7 @@ func TestDetermineDesiredReplicas_RepositoryRunner(t *testing.T) {
 		// case_1
 		// Explicitly specified the default `self-hosted` label which is ignored by the simulator,
 		// as we assume that GitHub Actions automatically associates the `self-hosted` label to every self-hosted runner.
-		// 2 demanded, max at 3
+		// 3 demanded, max at 3
 		{
 			repo:                     "test/valid",
 			labels:                   []string{"self-hosted"},
--- a/controllers/actions.summerwind.net/autoscaling_test.go
+++ b/controllers/actions.summerwind.net/autoscaling_test.go
@@ -63,7 +63,7 @@ func TestDetermineDesiredReplicas_RepositoryRunner(t *testing.T) {
 	}{
 		// case_0
 		// Legacy functionality
-		// 2 demanded, max at 3
+		// 0 demanded due to zero runID, min at 2
 		{
 			repo:                     "test/valid",
 			min:                      intPtr(2),
--- a/controllers/actions.summerwind.net/autoscaling_test.go
+++ b/controllers/actions.summerwind.net/autoscaling_test.go
@@ -61,8 +61,9 @@ func TestDetermineDesiredReplicas_RepositoryRunner(t *testing.T) {
 		want         int
 		err          string
 	}{
+		// case_0
 		// Legacy functionality
-		// 3 demanded, max at 3
+		// 2 demanded, max at 3
 		{
 			repo:                     "test/valid",
 			min:                      intPtr(2),
@@ -70,21 +71,28 @@ func TestDetermineDesiredReplicas_RepositoryRunner(t *testing.T) {
 			workflowRuns:             `{"total_count": 4, "workflow_runs":[{"status":"queued"}, {"status":"in_progress"}, {"status":"in_progress"}, {"status":"completed"}]}"`,
 			workflowRuns_queued:      `{"total_count": 1, "workflow_runs":[{"status":"queued"}]}"`,
 			workflowRuns_in_progress: `{"total_count": 2, "workflow_runs":[{"status":"in_progress"}, {"status":"in_progress"}]}"`,
-			want:                     3,
+			want:                     2,
 		},
-		// Explicitly speified the default `self-hosted` label which is ignored by the simulator,
+		// case_1
+		// Explicitly specified the default `self-hosted` label which is ignored by the simulator,
 		// as we assume that GitHub Actions automatically associates the `self-hosted` label to every self-hosted runner.
-		// 3 demanded, max at 3
+		// 2 demanded, max at 3
 		{
 			repo:                     "test/valid",
 			labels:                   []string{"self-hosted"},
 			min:                      intPtr(2),
 			max:                      intPtr(3),
-			workflowRuns:             `{"total_count": 4, "workflow_runs":[{"status":"queued"}, {"status":"in_progress"}, {"status":"in_progress"}, {"status":"completed"}]}"`,
-			workflowRuns_queued:      `{"total_count": 1, "workflow_runs":[{"status":"queued"}]}"`,
-			workflowRuns_in_progress: `{"total_count": 2, "workflow_runs":[{"status":"in_progress"}, {"status":"in_progress"}]}"`,
-			want:                     3,
+			workflowRuns:             `{"total_count": 4, "workflow_runs":[{"id": 1, "status":"queued"}, {"id": 2, "status":"in_progress"}, {"id": 3, "status":"in_progress"}, {"status":"completed"}]}"`,
+			workflowRuns_queued:      `{"total_count": 1, "workflow_runs":[{"id": 1, "status":"queued"}]}"`,
+			workflowRuns_in_progress: `{"total_count": 2, "workflow_runs":[{"id": 2, "status":"in_progress"}, {"id": 3, "status":"in_progress"}]}"`,
+			workflowJobs: map[int]string{
+				1: `{"jobs": [{"status": "queued", "labels":["self-hosted"]}]}`,
+				2: `{"jobs": [{"status": "in_progress", "labels":["self-hosted"]}]}`,
+				3: `{"jobs": [{"status": "in_progress", "labels":["self-hosted"]}]}`,
+			},
+			want: 3,
 		},
+		// case_2
 		// 2 demanded, max at 3, currently 3, delay scaling down due to grace period
 		{
 			repo:                     "test/valid",
@@ -97,6 +105,7 @@ func TestDetermineDesiredReplicas_RepositoryRunner(t *testing.T) {
 			workflowRuns_in_progress: `{"total_count": 1, "workflow_runs":[{"status":"in_progress"}]}"`,
 			want:                     3,
 		},
+		// case_3
 		// 3 demanded, max at 2
 		{
 			repo:                     "test/valid",
@@ -107,6 +116,7 @@ func TestDetermineDesiredReplicas_RepositoryRunner(t *testing.T) {
 			workflowRuns_in_progress: `{"total_count": 2, "workflow_runs":[{"status":"in_progress"}, {"status":"in_progress"}]}"`,
 			want:                     2,
 		},
+		// case_4
 		// 2 demanded, min at 2
 		{
 			repo:                     "test/valid",
@@ -117,6 +127,7 @@ func TestDetermineDesiredReplicas_RepositoryRunner(t *testing.T) {
 			workflowRuns_in_progress: `{"total_count": 1, "workflow_runs":[{"status":"in_progress"}]}"`,
 			want:                     2,
 		},
+		// case_5
 		// 1 demanded, min at 2
 		{
 			repo:                     "test/valid",
@@ -127,6 +138,7 @@ func TestDetermineDesiredReplicas_RepositoryRunner(t *testing.T) {
 			workflowRuns_in_progress: `{"total_count": 0, "workflow_runs":[]}"`,
 			want:                     2,
 		},
+		// case_6
 		// 1 demanded, min at 2
 		{
 			repo:                     "test/valid",
@@ -137,6 +149,7 @@ func TestDetermineDesiredReplicas_RepositoryRunner(t *testing.T) {
 			workflowRuns_in_progress: `{"total_count": 1, "workflow_runs":[{"status":"in_progress"}]}"`,
 			want:                     2,
 		},
+		// case_7
 		// 1 demanded, min at 1
 		{
 			repo:                     "test/valid",
@@ -147,6 +160,7 @@ func TestDetermineDesiredReplicas_RepositoryRunner(t *testing.T) {
 			workflowRuns_in_progress: `{"total_count": 0, "workflow_runs":[]}"`,
 			want:                     1,
 		},
+		// case_8
 		// 1 demanded, min at 1
 		{
 			repo:                     "test/valid",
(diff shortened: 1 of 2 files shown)


sergeykranga:
after digging through the code base trying to understand why we see random scale ups of runners that actually don't do anything, we came to the same issue described here (thanks for opening the pull request too)!

this is affecting us as well, causing extra cost for ec2 instances. it would be great to get some attention from maintainers here

Langleu:
@mumoshu, @toast-gear, @nikola-jokic anyone up or a review?
Would be nice to get it fixed since it can unknowingly result in tremendous bills.

mumoshu (dismissed):
Hi @Langleu! First of all, awesome job! I can admit that the removed logic was helpful only when there were a few jobs with different labels- that could be cases 2 or 3 years ago. Today, it might be worse than nothing as many folks started using self-hosted runners for relatively more complex use-cases.

LGTM. Let me merge this, and I will try to cut a new ARC patch version this week!

Thank you for your contribution!

Langleu:
Hey @mumoshu, thanks for having a look at it!

Any good suggestions on how to fix the tests?
From what I can see the tests are completely built around `runID == 0` callback, which in reality should never happen.

I played around a bit with it and we could give each single workflow run an id and mock via `workflowJobs` the response.
This only works if one passes a label like `self-hosted` since otherwise they're ignored. Which makes sense honestly, as the GHA controller shouldn't trigger on jobs that don't contain the `self-hosted` label.

Langleu:
@mumoshu added in https://github.com/actions/actions-runner-controller/pull/2671/commits/45c975bbbc7e0d0ce85af5ec89ea06d296eadad6 how I imagine the tests to be fixed atm.
Some were unconsciously relying on the `runID == 0` callback, which is the default in tests but not the reality.

Additionally added 2 small test cases to make sure that neither a hosted runner nor an empty Jobs array is scaling the runner.

Looking forward to getting it merged and having a new release! 🚀

controllers/actions.summerwind.net/autoscaling_test.go
```diff
@@ -534,8 +584,9 @@ func TestDetermineDesiredReplicas_OrganizationalRunner(t *testing.T) {
 			workflowRuns:             `{"total_count": 4, "workflow_runs":[{"status":"in_progress"}, {"status":"in_progress"}, {"status":"in_progress"}, {"status":"completed"}]}"`,
 			workflowRuns_queued:      `{"total_count": 0, "workflow_runs":[]}"`,
 			workflowRuns_in_progress: `{"total_count": 3, "workflow_runs":[{"status":"in_progress"},{"status":"in_progress"},{"status":"in_progress"}]}"`,
-			want:                     3,
+			want:                     1,
```
Langleu, line 587: this one is supposedly fixed at 3.
Maybe the test framework is broken but I didn't see the `fixed` value having any effect.
It only worked prior due to the `callback` secretly bumping the `in_progress` value

mumoshu: Good catch! However, I think this makes this test case useless- How about adding `"id":123` or whatever to every `workflow_runs` items so that it still falls into the desired code path, allowing this `want` value to remain `3`?

Langleu: @mumoshu, looked a bit through the code but I don't think it makes any difference.
The fixed value influences the replicas of the `RunnerDeployment` and that ultimately is the `scaleTarget` but to calculate the desired amount of replicas needed, none of that information is used.
The workflows don't have a `self-hosted` label, which would result in the requirement of 3. Otherwise they're treated as 3 hosted runners and that means the desired replicas == minimum defined as they're not of importance.

All autoscaler tests are running against the `AutoscalingMetricTypeTotalNumberOfQueuedAndInProgressWorkflowRuns` metric and none against `AutoscalingMetricTypePercentageRunnersBusy`. For the later, I could imagine the `fixed`, and therefore `scaleTarget` to have an influence but since we only check for `inQueue / inProgress` the `fixed` value and replicas of the RunnerDeployment have 0 influence on the `desired` value.

Langleu: @mumoshu, I'd propose removing the `fixed` tests, since they're irrelevant for `AutoscalingMetricTypeTotalNumberOfQueuedAndInProgressWorkflowRuns` tests.
Are you in accord with the change?

Langleu: @mumoshu , just another ping since this is stalling a bit ...

maxdanilov: @mumoshu pinging as well, would be amazing to have this PR merged for the stability of the pull-based approach.

alexanderkranga: @mumoshu another ping, we would really love this to get merged and released. This change is very important as it will have a positive impact on the cost of running multiple clusters

alexanderkranga: @nikola-jokic @toast-gear hey guys, any chance you can provide some feedback here too and help to get this merged? this is a very important change and we will all benefit from it

nikola-jokic: Hey, sorry for the delay,

I can't help much here since I am responsible for the `actions.github.com` API group. However, I asked @mumoshu to do another round of reviews and he is the best one to talk to related to the `actions.summerwind.net` group :relaxed: 

While waiting for the review, @Langleu do you mind updating the branch? That can fix rest of the tests. I can do it myself if you agree but it would be better if you do it

Langleu: @nikola-jokic, done.

nikola-jokic: Thank you!

Langleu: lint looks like a timeout, works locally.
Same for the RunnerSet related tests since this change doesn't touch that logic and runs into timeout as well.

nikola-jokic: No worries, I'll re-run the tests. I know it is not affecting that part of the code

controllers/actions.summerwind.net/autoscaling_test.go
```diff
@@ -61,30 +61,38 @@ func TestDetermineDesiredReplicas_RepositoryRunner(t *testing.T) {
 		want         int
 		err          string
 	}{
+		// case_0
 		// Legacy functionality
-		// 3 demanded, max at 3
+		// 2 demanded, max at 3
```
mumoshu:
```suggestion
		// 0 demanded due to zero runID, min at 2
```

Perhaps this is better to say!
`want` becomes 2 due to the min replicas set to 2, not due to the workflow runs statuses demanded 2 replicas.

controllers/actions.summerwind.net/autoscaling_test.go
```diff
@@ -422,7 +464,8 @@ func TestDetermineDesiredReplicas_OrganizationalRunner(t *testing.T) {
 		want         int
 		err          string
 	}{
-		// 3 demanded, max at 3
+		// case_0
+		// 2 demanded, max at 3
```
mumoshu:
```suggestion
		// 0 demanded due to zero runID, min at 2
```

controllers/actions.summerwind.net/autoscaling_test.go
```diff
@@ -61,30 +61,38 @@ func TestDetermineDesiredReplicas_RepositoryRunner(t *testing.T) {
...
 			workflowRuns_in_progress: `{"total_count": 2, "workflow_runs":[{"status":"in_progress"}, {"status":"in_progress"}]}"`,
-			want:                     3,
+			want:                     2,
 		},
-		// Explicitly speified the default `self-hosted` label which is ignored by the simulator,
+		// case_1
+		// Explicitly specified the default `self-hosted` label which is ignored by the simulator,
 		// as we assume that GitHub Actions automatically associates the `self-hosted` label to every self-hosted runner.
-		// 3 demanded, max at 3
+		// 2 demanded, max at 3
```
mumoshu:
```suggestion
		// 3 demanded, max at 3
```

You actually demanded 3 replicas via those workflow jobs.

mumoshu (approved):
Sorry for the delay! Apparently, I got confused only due to how tests are commented.
I've corrected some comments, and this looks good overall now.
Thanks for your patience and contribution @Langleu!
// I've been a bit out of bandwidth due to a lack of sponsorships.

controllers/actions.summerwind.net/autoscaling_test.go
mumoshu: Thanks for your assistance @nikola-jokic!

PR #2682, opened 2023-06-18
Fix serviceMonitor labels in actionsMetrics

marcin-motyl:
`serviceMonitorLabels` are defined under `actionsMetrics`, not `actionsMetricsServer`.

See: https://github.com/actions/actions-runner-controller/blob/master/charts/actions-runner-controller/values.yaml#L308


mumoshu (approved):
Good catch! Thanks a lot for fixing it @marcin-motyl ☺️

PR #2683, opened 2023-06-18
Fix deployment & service values in actionsMetrics

marcin-motyl:
In some places Actions Metrics Server uses values from `metrics`, but in other from `actionsMetrics`.

For example, see:
- https://github.com/actions/actions-runner-controller/blob/master/charts/actions-runner-controller/templates/actionsmetrics.deployment.yaml#L125
- https://github.com/actions/actions-runner-controller/blob/master/charts/actions-runner-controller/templates/actionsmetrics.servicemonitor.yaml.yml#L16

My understanding is that `metrics` section is for Webhook Server, while Actions Metrics Server should use `actionsMetrics` values.

### DIFF
--- a/charts/actions-runner-controller/templates/actionsmetrics.deployment.yaml
+++ b/charts/actions-runner-controller/templates/actionsmetrics.deployment.yaml
@@ -36,8 +36,8 @@ spec:
       {{- end }}
       containers:
       - args:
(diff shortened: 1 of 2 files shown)


mumoshu (approved):
Awesome! Thanks for the fixes @marcin-motyl! ☺️

PR #2686, opened 2023-06-19
Add configurable log format to values.yaml and propagate it to listener

nikola-jokic:
### Context

Expose a `log-format` parameter through the helm charts that will allow end users to specify the log format for the autoscaling runner set controller manager and listener.

This `log-format` will not propagate to the runners because they have a distinct logging structure.

### DIFF
--- a/charts/gha-runner-scale-set-controller/values.yaml
+++ b/charts/gha-runner-scale-set-controller/values.yaml
@@ -79,6 +79,8 @@ flags:
   ## Log level can be set here with one of the following values: "debug", "info", "warn", "error".
   ## Defaults to "debug".
   logLevel: "debug"
+  ## Log format can be set with one of the following values: "text", "json"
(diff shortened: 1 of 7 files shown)


charts/gha-runner-scale-set-controller/values.yaml
```diff
@@ -79,6 +79,7 @@ flags:
   ## Log level can be set here with one of the following values: "debug", "info", "warn", "error".
   ## Defaults to "debug".
   logLevel: "debug"
+  logFormat: "text"
```
Link-, line 84: Can we please add a comment with the accepted log formats?

Link- (changes requested):
Thank you for the quick turnaround for this! It looks great 🙏

Link- (approved):
LGTM! Let's ship it

PR #2699, opened 2023-06-23
Add status check before deserializing runner-registration response

nikola-jokic:
The purpose of this PR is to ensure the listener does not panic if the status code returned by the actions service is not in 2xx range.

Previously, we would try to unmarshal JSON which would leave nil values for actions service admin token. When later used, the listener would panic.

Added check similar to one in the runner making sure the status code is success, and log the error message if returned from the actions service

https://github.com/actions/runner/blob/9c57205c1c1cd7ada73cb08b556293ec0411f8d4/src/Runner.Listener/Configuration/ConfigurationManager.cs#L777C50-L777C50

### DIFF
--- a/github/actions/client.go
+++ b/github/actions/client.go
@@ -922,7 +922,13 @@ func (c *Client) getActionsServiceAdminConnection(ctx context.Context, rt *regis
 	defer resp.Body.Close()
 
 	if resp.StatusCode < 200 || resp.StatusCode > 299 {
-		return nil, fmt.Errorf("unexpected response from Actions service during registration call: %v", resp.StatusCode)
+		registrationErr := fmt.Errorf("unexpected response from Actions service during registration call: %v", resp.StatusCode)
+
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return nil, fmt.Errorf("%v - %v", registrationErr, err)
+		}
+		return nil, fmt.Errorf("%v - %v", registrationErr, string(body))
 	}
 
 	var actionsServiceAdminConnection *ActionsServiceAdminConnection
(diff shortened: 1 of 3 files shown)


github/actions/testserver/server.go
```diff
@@ -73,10 +70,45 @@ func WithActionsToken(token string) actionsServerOption {
...
+
+	if s.actionRegistrationTokenHandler == nil {
+		s.actionRegistrationTokenHandler = func(w http.ResponseWriter, r *http.Request) {
+			if s.token == "" {
+				s.token = DefaultActionsToken(t)
+			}
+
+			w.WriteHeader(http.StatusCreated)
+			w.Write([]byte(`{"url":"` + s.URL + `/tenant/123/","token":"` + s.token + `"}`))
+			return
```
nikola-jokic:
```suggestion

```

github/actions/client.go
```diff
@@ -921,6 +921,10 @@ func (c *Client) getActionsServiceAdminConnection(ctx context.Context, rt *regis
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		return nil, fmt.Errorf("unexpected response from Actions service during registration call: %v", resp.StatusCode)
```
Link-: Are we dumping that response body somewhere else? We might need to see what the service is returning in order to troubleshoot the root cause of this error

Link- (changes requested):
Can we please add more context to the PR description so that we use it as documentation for why this change was introduced? 

1. Describe the problem
2. How this change addresses that problem

Link- (approved):
This looks good, thank you

PR #2701, opened 2023-06-23
feat: allow more dockerd options

twildber:
- Added a description how to mount a configmap to configure multiple docker daemon options
- Moved the directory structure creation for the rootless-docker runner to the image.
  That way it will not break permissions while mounting the configmap.

provides feature: https://github.com/actions/actions-runner-controller/issues/2700


mumoshu (approved):
LGTM. Thank you for your patience and awesome contribution @twildber!

PR #2705, opened 2023-06-28
fix: template test of service account

Langleu:
test regression of https://github.com/actions/actions-runner-controller/pull/2567

Adjusts the tests as they fail on master and any other branch. See [Go test](https://github.com/actions/actions-runner-controller/actions/runs/5389106723/jobs/9782621900) on master.
`-service-account` was removed but the tests to check that weren't adjusted.


Langleu:
@nikola-jokic, please have a look since you've also reviewed the previous PR.
I ran it locally with `make test-with-deps` and the errors were resolved.

nikola-jokic (approved):
Thanks for fixing this :relaxed:

PR #2708, opened 2023-06-28
add opportunity write dnsPolicy for controller by helm values

sobolevskaya-k:
hello, i read that if using hostNetwork: true then need dnsPolicy: ClusterFirstWithHostNet. from this [doc](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy). can you add possibility for defining this parameter in helm values?


mumoshu (approved):
Good finding! Thanks a lot for sharing your awesome work! @sobolevskaya-k ☺️

PR #2712, opened 2023-07-01
Clarify multiple metrics in docs

julietrb1:
I hadn't seen any examples or documentation about how combining multiple HorizontalRunnerAutoscalers behaves in close proximity to the explanation of the metric types in the documentation, so I didn't know if it was even allowed.

This commit moves the explanation from *Autoscaling to/from 0* to *Pull Driven Scaling*, where it clarifies usage closer to the metric definitions.


mumoshu (approved):
LGTM. Thanks for your attention and fixing it @julietrb1!

PR #2713, opened 2023-07-01
chore: Set build version on make-runscaleset

mumoshu:
This enables us to locally test various runner versions.

For example, you can now run `VERSION=0.4.0 make run-scaleset` to let the local runnerscaleset controller to manage runners deployed using the runner-scale-set chart v0.4.0.


nikola-jokic (approved):
Great catch!

Link- (approved):

Link-:
🚢 it!

PR #2718, opened 2023-07-05
fix(startup): only show the "using latest tag" warning on relevant images

TPXP:
We'll set the SHOW_LATEST_TAG_WARN env var for images build with the latest tag only (see PR on the releases repo).

Since the warning is now only shown in relevant cases, bump its severity.

See corresponding PR for the build pipeline https://github.com/actions-runner-controller/releases/pull/11


TPXP:
@Link- Hello, thanks for your work on this repo! This PR has been open for a while, is there anything I can do to help getting it merged?

Link-:
@TPXP - these runner images including the startup script are maintained by @mumoshu. My 2 cents are that this type of warnings is not good to ship. We expect a certain level of experience when adopting a solution like ARC and we don't want to be throwing warnings for every trivial matter.

It's valid to use the latest tag, albeit not advisable, and I expect end-users to be aware what they're getting into here.

TPXP:
@Link- Thanks for your feedback, I just wish to clarify that the warning is currently raised unconditionally in the current version, which adds confusion over unneeded noise in the logs. If you prefer not to have any warning raised even when using the latest tag, could we just drop the warning from the script? https://github.com/actions/actions-runner-controller/blob/c8216e13968cab03475b90f8abdc8dbabbbccc90/runner/startup.sh#L176-L177

Link-:
I'll let @mumoshu weigh in on this ☝️

TPXP:
@mumoshu  Hello there, any updates?

mumoshu:
@TPXP Sorry for the delay! Yeah, let's drop it and update our `defaultRunnerImage` default to not point to the `latest` image.
That's the way to go and it's just that I didn't have much time to do it myself!

PR #2725, opened 2023-07-06
fixed indent in a README example

prein:
`ephemeral` is in `.spec.template.spec` path not `.spec.template`. Took us some time to figure out after c&p from the README

Apologies for not following the contributing guide, but this is such a minor change...

BTW, I should probably open an issue for that but sharing `/tmp` in my setup leads to
```
Failed to create CoreCLR, HRESULT: 0x80004005
```
preventing the runner from starting. Why share `/tmp` at all?

### DIFF
--- a/docs/using-custom-volumes.md
+++ b/docs/using-custom-volumes.md
@@ -78,7 +78,7 @@ spec:
       - hostPath:
           path: /mnt/disks/ssd0
         name: tmp
-    ephemeral: true # VERY important. otherwise data inside the workdir and /tmp is not cleared between builds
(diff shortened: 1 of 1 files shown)


mrsensey (approved):

mumoshu (approved):
Good catch!! Thanks a lot for your contribution @prein ☺️ 
Re `tmp` sharing, I'm not entirely sure.. You only need to mount it only when you want the tmp directory to persists between ephemeral runner pods.
If you don't need persitency, you can just omit the volume mount(as you might have already done).

PR #2758, opened 2023-07-20
Extend and generate crds allowing listener pod spec change

### LINKED ISSUE #2434: Add support for Affinity and Tolerations on runner scale set

Configure the actions runner helm chart to allow runner scale set inherits its affinity and tolerations from the runner scale set controller

### Why is this needed?

We do have dedicated kubernetes node groups for our action runner and use tolerations and affinity to allow runners to run on the correct group of nodes. Currenty only the runner scale set controller runs on a dedicated node because we can set affinity and toleration in the values.yaml file. 

The runner scale set on the other hand do not have  affinity and toleration settings causing them to be deployed on non dedicated nodes.

### LINKED ISSUE #2652: Listeners are sometimes scheduled on Windows nodes
### Checks

### Controller Version

0.4.0

### Helm Chart Version

0.4.0

### CertManager Version

N/a

### Deployment Method

Helm

### cert-manager installation

N/a

### Checks

### Resource Definitions

```yaml
# install arc
helm install arc `
    --namespace my-arc-namespace `
    --create-namespace `
    --set githubConfigSecret.github_token=$myPat `
    --set nodeSelector."kubernetes\.io/os"=linux `
    oci://ghcr.io/actions/actions-runner-controller-charts/gha-runner-scale-set-controller `
    --version 0.4.0

# linux-values.yaml contents (for linux based jobs):
template:
  spec:
    nodeSelector:
      kubernetes.io/os: linux

# windows-values.yaml contents (for windows based jobs):
template:
  spec:
    nodeSelector:
      kubernetes.io/os: windows
    containers:
    - name: runner
      image: location-of-my-windows-runner-container/my-windows-runner:latest

# scale sets are installed using the helm chart and one of the overrides above, e.g. for windows:
helm upgrade --install my-windows-runner-set `
    --namespace my-runners-namespace `
    --create-namespace `
    --set githubConfigUrl=$myGitHubUrl `
    --set githubConfigSecret.github_token=$myPat `
    --values windows-values.yaml `
    oci://ghcr.io/actions/actions-runner-controller-charts/gha-runner-scale-set `
    --version 0.4.0
```

### To Reproduce

```markdown
Using AKS:

# option 1:

1. Create an aks cluster which will have a linux based node pool for the control plane called "nodepool1"
2. Add a Windows node pool called "wpool" with a single Windows node
3. Install the ARC controller via the helm chart as shown above
4. Install a Windows runner scale set via helm chart as shown above
5. The scale set listener is scheduled on the Windows node created in step 2 and not on a linux node in "nodepool1" (despite its name - see below)

# option 2: As for #1 - but add a linux based user nodepool called "zzzpool" - same issue

# option 3: As for #1 - but add a linux based user nodepool called "aaapool" - this works, listeners are scheduled on this and Windows runners are correctly scheduled on the windows node pool

# option 4: As for #1 - but create the Windows node pool with zero nodes and auto scaling enabled - this also works but isn't a feasible workaround (and is obviously also slower when a Windows runner is required)

Note - if a listener pod needs to be recreated for any reason this same issue can occur.
```

### Describe the bug

Described in the steps to reproduce above.

Listener pods are always scheduled on linux nodes regardless of the nodepool name.

### Whole Controller Logs

```shell
-
```

### Whole Runner Pod Logs

```shell
-
```

### Additional Context

_No response_

### LINKED ISSUE #2881: Support listener pod annotations
It would be nice if we can set a listener pod spec such as `.metadata.annotations`.

### Why is this needed?
We'd like to monitor the metrics of listener pods by Datadog. It requires the pod annotation like:

```yaml
apiVersion: v1
kind: Pod
metdata:
  annotations:
    ad.datadoghq.com/autoscaler.checks: |
      {
        "openmetrics": {
          "init_config": {},
          "instances": [
            {
              "openmetrics_endpoint": "http://%%host%%:8080/metrics",
              "namespace": "arc",
              "metrics": ["gha_*"]
            }
          ]
        }
      }
```

Currently, there is no way to set annotations to a listener pod.

### Additional context
I notice the change of https://github.com/actions/actions-runner-controller/pull/2758. I hope this issue will be also resolved.

nikola-jokic:
This change aims to extend the CRD, allowing listener pod customization.

Warning: Previously, there was a way to set the default image pull policy for listener pods, that will be respected by every autoscaling runner set. With this change, we are removing this behaviour and instead of it, allowing customization per autoscaling runner set, instead of the global one.

Fixes https://github.com/github/actions-runners/issues/31
Fixes #2652
Fixes #2434
Fixes #2881


charts/actions-runner-controller/crds/actions.summerwind.dev_horizontalrunnerautoscalers.yaml
```diff
@@ -1,8 +1,9 @@
+---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.7.0
+    controller-gen.kubebuilder.io/version: v0.11.3
```
Link-, line 6: Do we have to do this upgrade in this PR?

charts/gha-runner-scale-set/templates/autoscalingrunnerset.yaml
```diff
@@ -88,6 +88,11 @@ spec:
   minRunners: {{ .Values.minRunners | int }}
   {{- end }}
 
+  {{- with .Values.listenerTemplate}}
+  listenerTemplate:
+    {{- toYaml . | nindent 4}}
+  {{- end }}
```
Link-, line 94: How is this different than `template` below it? And is there a valid reason why we aren't restrictive in the configuration we allow?

charts/gha-runner-scale-set-controller/templates/NOTES.txt
```diff
@@ -1,3 +1,5 @@
 Thank you for installing {{ .Chart.Name }}.
 
 Your release is named {{ .Release.Name }}.
+
+WARNING: value specified under `image.pullPolicy` will no longer be applied to the listener pod spec in the next release. Please use the `listenerTemplate` in the gha-runner-scale-set chart to control the image pull policy of the listener.
```
Link-: Awesome, I would also suggest that we create an announcement: https://github.com/actions/actions-runner-controller/discussions/categories/announcements with:
- what's being deprecated and why
- timelines
- apply the [deprecation](https://github.com/actions/actions-runner-controller/labels/deprecation) and [breaking](https://github.com/actions/actions-runner-controller/labels/breaking) labels on it

Link- (changes requested):
Thank you for putting this together! I think the proposal is solid but let's go through the details a little bit further

charts/actions-runner-controller/crds/actions.summerwind.dev_horizontalrunnerautoscalers.yaml
nikola-jokic: I guess not, this was generated from my machine when I generated CRDs
 Would you like me to revert it back? I guess it won't affect anything

charts/gha-runner-scale-set/templates/autoscalingrunnerset.yaml
nikola-jokic: I mean it is completely different field on the AutoscalingRunnerSet. The `Template` field is ephemeral runner template while the `ListenerField` is pod template spec for the listener
We don't have restrictions on the helm level the same way we don't have any restrictions for the ephemeral runner :relaxed:

charts/actions-runner-controller/crds/actions.summerwind.dev_horizontalrunnerautoscalers.yaml
Link-: Doesn't it require an upgrade of codegen or something?

nikola-jokic: I guess not really. It just used this version to generate CRD that will be fed to the API server
As long as the API server understands the CRD and serves those objects to the controller, we should be okay
And it is just an annotation, so I guess it just provides more information of how this CRD is generated. It shouldn't influence the execution

charts/gha-runner-scale-set/values.yaml
```diff
@@ -163,9 +163,19 @@ template:
...
+
+## template is the PodSpec for each listener Pod
+# listenerTemplate:
+#   spec:
+#     containers:
+#     - name: listener # merge extra container configuration to the default configuration for the listener container.
+#       securityContext:
+#         runAsUser: 1000
+#     - name: side-car # container used as a side-car container. Spec for this container will be applied without any modifications.
+#       image: example-sidecar
```
Link-: nit: I think this will be better if we move it up before the runner pod template, wdyt?

charts/gha-runner-scale-set/values.yaml
```diff
@@ -163,9 +163,19 @@ template:
...
-      command: ["/home/runner/run.sh"]
+      - name: runner
+        image: ghcr.io/actions/actions-runner:latest
+        command: ["/home/runner/run.sh"]
+
+## template is the PodSpec for each listener Pod
+# listenerTemplate:
+#   spec:
+#     containers:
+#     - name: listener # merge extra container configuration to the default configuration for the listener container.
```
Link-: What happens when end-users change this value?

nikola-jokic: It will be a side-container then

charts/gha-runner-scale-set/values.yaml
```diff
@@ -163,9 +164,25 @@ template:
...
+#     # Use this section to append additional container configuration listener container.
+#     # If you change the name of the container, the configuration will not be applied to the listener,
+#     # and it will be treated as a side-car container.
+#     - name: listener
+#       securityContext:
+#         runAsUser: 1000
+#     # Use this section to add the configuration of a side-car container. 
+#     # Spec for this container will be applied as is without any modifications.
+#     - name: side-car
+#       image: example-sidecar
```
Link-: @nikola-jokic - I took the liberty and added some clarifying comments directly

charts/gha-runner-scale-set-controller/templates/NOTES.txt
Link-:
```suggestion
WARNING: value specified under image.pullPolicy will be ignored and no longer be applied to the listener pod spec as of gha-runner-scale-set-0.7.0. Please use the listenerTemplate in the gha-runner-scale-set chart to control the image pull policy of the listener.
```

Let's be more specific

Link-:
Ran the tests, this is great!

### TODO before merging:
- [ ] Create an announcement: https://github.com/actions/actions-runner-controller/discussions/categories/announcements with:
    - [ ] what's being deprecated and why
    - [ ] timelines
    - [ ] apply the [deprecation](https://github.com/actions/actions-runner-controller/labels/deprecation) and [breaking](https://github.com/actions/actions-runner-controller/labels/breaking) labels on it
- [ ] Update the e2e tests to include a test for `listenerTemplate` changes

charts/gha-runner-scale-set/values.yaml
nikola-jokic: Oh sure! Thanks!

Link-:
> DRAFT deprecation announcement

# Listener `imagePullPolicy` will no longer be inherited from the controller manager as of gha-runner-scale-set-0.7.0

In gha-runner-scale-set-0.6.0 we introduced the ability to configure the listener pod spec via the `listenerTemplate` in `charts/gha-runner-scale-set`. Before this release, the `imagePullPolicy` was inherited from the configuration of the controller manager pod and used for every autoscaling listener.

With the ability to update the spec of any autoscaling listener pod, we are removing the dependency of the listener pods on the `imagePullPolicy` of the controller manager.

## How does this change affect me?

If you upgraded to `gha-runner-scale-set-0.6.0` and configured the `imagePullPolicy` in the `listenerTemplate` object for your listener, **this value will be ignored and we will use the `imagePullPolicy` field, if it has been specified in the controller chart.** This is to maintain temporary backward compatibility until everyone has had the chance to upgrade.

Otherwise, we will fall back to the controller manager `imagePullPolicy` value.

Cluster administrators will get a deprecation warning through helm.

## What happens after gha-runner-scale-set-0.7.0?

The `imagePullPolicy` will only be applied to the listener pod if the `listenerTemplate` specifies it.

Otherwise, it will be left unspecified, and will apply the default Kubernetes behaviour. The `imagePullPolicy` field is specified in the controller chart will not be applied to the listener pod.

If you depend on this field, make sure you update your `listenerTemplate` object to explicitly specify the `imagePullPolicy` for the listener.

Link- (approved):
Let's 🚢 it

PR #2780, opened 2023-07-26
Logs - Add missing formatting

luka5:
Follow up to https://github.com/actions/actions-runner-controller/pull/2320

Right now the logs look like:
```
INFO    -watch-namespace is %q. Only HorizontalRunnerAutoscalers in %q are watched, cached, and considered as scale targets.    {"my-namespace": "my-namespace"}
```

But I expect it to look like:
```
INFO    -watch-namespace is "my-namespace". Only HorizontalRunnerAutoscalers in "my-namespace" are watched, cached, and considered as scale targets.
```


mumoshu (approved):
LGTM. Thank you for your contribution @luka5!

PR #2785, opened 2023-07-29
:memo: fix docs in `deploy-arc-runners.md`

Shuliyey:
`dockerdWithinRunnerContainer` should be under `spec.template.spec.dockerdWithinRunnerContainer` in `RunnerSet`


Shuliyey:
after testing this is not right, issue was something else seems `spec.dockerdWithinRunnerContainer` and `spec.template.spec.containers[1]` with docker should be used together

PR #2797, opened 2023-08-04
Add runner_name as new metric label

zanehala-wk:
A common task we've been finding ourselves in as maintainers of CICD systems is improving the cost efficiency of our workflows/jobs.

While we have metrics on the resource usage of our individual runner pods, it's quite tricky to link the pods to a specific job or workflow since there are no common metrics between the two. This PR is an attempt to reconcile that. 

This PR simply adds the runner name as a new label on the already exposed `github_workflow_job_*` histograms. This will increase the cardinality of the metrics already reported however, so I will leave the decision to accept this up to the maintainers. 

I built the dockerfile and deployed to our dev environment and was able to get the metric working as expected. Below is a screenshot of it reporting the metrics (I blurred some things out but you should be able to get the gist) of it working. 
Screenshot 2023-08-04 at 4 07 08 PM

### DIFF
--- a/pkg/actionsmetrics/event_reader.go
+++ b/pkg/actionsmetrics/event_reader.go
@@ -112,6 +112,13 @@ func (reader *EventReader) ProcessWorkflowJobEvent(ctx context.Context, event in
 	labels["workflow_name"] = wn
 	labels["head_branch"] = hb
 
+	var rn string
+	if n := e.WorkflowJob.RunnerName; n != nil {
+		rn = *n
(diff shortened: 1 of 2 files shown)


neilharris1:
+1 on this. This is exactly what I am looking also.

daveeickelberg-wf:
@mumoshu @toast-gear @nikola-jokic Any update on this one? It would be a big win for us.

chrisgustavsen-wf:
Seems like a good feature to have 😄

PR #2706, opened 2023-06-28
Trim gha-runner-scale-set to gha-rs in names and remove role type suffixes

nikola-jokic:
Fixes #2697

This PR tries to mitigate the problem when service account names are used as labels. The suffix being added by supported charts can be too long (the `-gha-runner-scale-set` suffix takes 21 characters).

# Summary:

- All suffixes with `gha-runner-scale-set` are replaced with `gha-rs`
- Resource types are removed from the name suffix (for example `-role`)

# Change

By trimming down the suffix, we can allow longer names for clusters with such limitations.

## gha-runner-scale-set-controller
| Templates | Resource Kind | Name | Reserved Length | Description | Notes |
|-------|---------------|------|-----------------|-------------|-------|
| deployment.yaml | Deployment | [installation name]-gha-rs-controller | 18  | The resource running controller-manager | The pods created by this resource will have replica set suffix and the pod suffix |
| serviceaccount.yaml | ServiceAccount | [installation name]-gha-rs-controller | 18 | Created if serviceAccount.create in values.yaml is set to true | The name can be customized in values.yaml |
| manager_cluster_role.yaml | ClusterRole | [installation name]-gha-rs-controller | 18 | ClusterRole for the controller manager | Created if the value of flags.watchSingleNamespace is empty |
| manager_cluster_role_binding.yaml | ClusterRoleBinding | [installation name]-gha-rs-controller | 18 | ClusterRoleBinding for the controller manager | Created if the value of flags.watchSingleNamespace is empty |
| manager_single_namespace_controller_role.yaml | Role | [installation name]-gha-rs-controller-single-namespace | 35 | Role for the controller manager | Created if the value of flags.watchSingleNamespace is set |
| manager_single_namespace_controller_role_binding.yaml | RoleBinding | [installation name]-gha-rs-controller-single-namespace | 35 | RoleBinding for the controller manager | Created if the value of flags.watchSingleNamespace is set |
| manager_single_namespace_watch_role.yaml | Role | [installation name]-gha-rs-controller-single-namespace-watch | 41 | Role for the controller manager for the namespace configured | Created if the value of flags.watchSingleNamespace is set |
| manager_single_namespace_watch_role_binding.yaml | RoleBinding | [installation name]-gha-rs-controller-single-namespace-watch | 41 | RoleBinding for the controller manager for the namespace configured | Created if the value of flags.watchSingleNamespace is set |
[...]

PR #2824, opened 2023-08-21
Revert chart name and use helper constant to trim the name base

nikola-jokic:
Addendum to #2706 - this PR will revert back the helm chart renaming.

### DIFF
--- a/charts/gha-runner-scale-set-controller/Chart.yaml
+++ b/charts/gha-runner-scale-set-controller/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-name: gha-rs-controller
+name: gha-runner-scale-set-controller
 description: A Helm chart for install actions-runner-controller CRD
 
 # A chart can be either an 'application' or a 'library' chart.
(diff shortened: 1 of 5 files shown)


Link- (dismissed):

Link- (approved):

brendanlafond:
This change doesn't allow for helm upgrade resulting in the below error. Trying to uninstall and install also returns the same error. It looks like there may have been a missed item in the reverting of the chart name as the error implies that it's still looking for gha-rs-controller in the manager_role_binding.yaml.

```shell
Error: UPGRADE FAILED: execution error at (gha-runner-scale-set/templates/manager_role_binding.yaml:17:11): No gha-rs-controller deployment found using label (app.kubernetes.io/part-of=gha-rs-controller). Consider setting controllerServiceAccount.name in values.yaml to be explicit if you think the discovery is wrong.
``` 

Any thoughts on where this is being set? I tried looking at the broader change to see if there may have been something missed, but can't spot it without diving much deeper.

cc @nikola-jokic

PR #2828, opened 2023-08-22
Fix - Actually Enable Sets in addition to Slices in `env`

luka5:
As mentioned in the comments and as implemented in [githubwebhook.deployment.yaml](https://github.com/actions/actions-runner-controller/blob/4536707/charts/actions-runner-controller/templates/githubwebhook.deployment.yaml#L120-L127) and [deployment.yaml](https://github.com/actions/actions-runner-controller/blob/4536707af666b3f3b1b548d2fbe961d32805eebf/charts/actions-runner-controller/templates/deployment.yaml#L124-L131) one can provide a set or slice in `env`.

This works, when one uses the helm chart source code and modifies the `values.yaml`. It does work - but prints warnings, when using the released chart directly. Hereby, when I set a slice, I get the following warning, since it it already defined as a set.

```
cannot overwrite table with non table for 
```

I'd recommend to remove the default `{}` value. Then users can actually use sets or slices.

*Edit:* Previously I mentioned, that it is an error, it is not. It is "only" a warning and works.

### DIFF
--- a/charts/actions-runner-controller/values.yaml
+++ b/charts/actions-runner-controller/values.yaml
@@ -302,7 +302,7 @@ githubWebhookServer:
   #       key: GITHUB_WEBHOOK_SECRET_TOKEN
   #       name: prod-gha-controller-webhook-token
   #       optional: true
-  env: {}
+  # env:
 
 actionsMetrics:
   serviceAnnotations: {}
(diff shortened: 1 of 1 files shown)


mumoshu (approved):
LGTM. Thank you very much for your contribution @luka5! ☺️

luka5:
Thanks, it resolves my issue :)

PR #2833, opened 2023-08-23
Fix #2809 : replace TCP docker daemon connection with unix socket

### LINKED ISSUE #2809: Working non-trivial example for gha-runner-scale-set

I would really like there to be a working example that can do more than just hello world.

### Why is this needed?

Because nothing really works beyond hello world

### Additional context

Example Docker:

```yaml
name: demo
on:
  workflow_dispatch:
jobs:
  demo:
    runs-on: demo
    steps:
    - name: Set up Docker Buildx
      uses: docker/setup-buildx-action@v2
```

*dind*: fails because of "ERROR: could not create a builder instance with TLS data loaded from environment."

*kubernetes*: fails because of "/usr/bin/tar: ../../../.docker: Cannot mkdir: Permission denied"

Example Cache volume:

```yaml
name: demo
on:
  workflow_dispatch:
jobs:
  first-run:
    runs-on: demo
    steps:
    - run: echo 'demo' > $RUNNER_TOOL_CACHE/demo

  second-run:
    runs-on: demo
    needs: [first-run]
    steps:
    - run: cat $RUNNER_TOOL_CACHE/demo
```

*default*: doesn't output the expected value

Add pvc resource + extra config:
```yaml
  template:
    spec:
      securityContext:
        fsGroup: 123
      containers:
      - name: runner
        image: 'ghcr.io/actions/actions-runner:latest'
        volumeMounts:
        - name: cache
          mountPath: /home/runner/_work/_tool
      nodeSelector:
        kubernetes.io/os: linux
      volumes:
      - name: cache
        persistentVolumeClaim:
          claimName: github-cache
```

and yes, you need  the image and command also, or things break

*dind*: Works

*kubernetes*: $RUNNER_TOOL_CACHE is different path in running container. Fails with mountpath set to either the above or the changed path

Example windows:

*default*:  Doesn't work with the windows dockerfile example: Cannot start and Fails with `AutoscalingRunnerSet.actions.github.com "gha-windows" is invalid: [spec.template.spec.containers[0].volumeMounts: Invalid value: "null": `

*dind*: Doesn't work with the windows dockerfile example. Cannot start and Fails to run the dind things in linux, or have a windows image for them

*kubernetes*: Doesn't work with the windows dockerfile example. Can start but fails on the "initialize containers" step with "
Container operations are only supported on Linux runners"

*Other issues*

* Everything is marked with deletion protection, even though everything is ephemeral.
* Helm-chart uses lookup, so requires work-around for ArgoCD

dm3ch:
TCP docker connection causes the following problem:
```
dind: fails because of "ERROR: could not create a builder instance with TLS data loaded from environment."
kubernetes: fails because of "/usr/bin/tar: ../../../.docker: Cannot mkdir: Permission denied"
```

This issue have been workaround on legacy work mode by using unix socket for interaction with docker daemon ( https://github.com/actions/actions-runner-controller/pull/2324 )

This PR reimplements same approach for RunnerSets.

P.S. I preferred this approach cause in this case RunnerSets can work with workflows wrote for public action runners without any changes in workflow to make it work (without adding steps to create buildx context properly)

/fix #2809


nikola-jokic (dismissed):
LGTM

dm3ch:
Do I need to fix something or merge fresh master?

charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -93,19 +93,26 @@ volumeMounts:
 
 {{- define "gha-runner-scale-set.dind-container" -}}
 image: docker:dind
+args:
+  - dockerd
+  - --host=unix:///run/docker/docker.sock
+  - --group=$(DOCKER_GROUP_GID)
+env:
+  - name: DOCKER_GROUP_GID
+    value: "123"
```
Link-, line 102: This is could potentially cause issues especially since we don't provide a way to configure it. It's true we set `123` in our runner image even: https://github.com/actions/runner/blob/2908d82845c018193655e68f3c20a5ad03bc0efd/images/Dockerfile#L43 but the idea from the runner image is that users can override and build an image that works best for them.

I'm not sure how to address this without exposing the `dind` configuration through the helm charts which we intentionally left out.

Link- (changes requested):
Need to address the concern of group id before we merge this

charts/gha-runner-scale-set/values.yaml
```diff
@@ -103,16 +103,12 @@ template:
   ##       image: ghcr.io/actions/actions-runner:latest
   ##       env:
   ##         - name: DOCKER_HOST
-  ##           value: tcp://localhost:2376
-  ##         - name: DOCKER_TLS_VERIFY
-  ##           value: "1"
-  ##         - name: DOCKER_CERT_PATH
-  ##           value: /certs/client
+  ##           value: unix:///run/docker/docker.sock
```
Link-, line 125: The changes introduced on lines 96 - 102 of `charts/gha-runner-scale-set/templates/_helpers.tpl` need to be included here in the comments to stay consistent

Link-:
As we plan to upgrade the Docker engine of the [runner image](https://github.com/actions/runner/blob/c1a5dc71a5f33e1dd61f9bb76566e8f59cb0ad32/images/Dockerfile#L8) to `24.0.6` this fix is going to be a necessity for DinD mode to work.

It resolves the error from running `docker/setup-buildx-action@v3`:
```
 ERROR: could not create a builder instance with TLS data loaded from environment. Please use `docker context create <context-name>` to create a context for current environment and then create a builder instance with `docker buildx create <context-name>`
```

Other workarounds are discussed here: https://github.com/actions/actions-runner-controller/issues/893

charts/gha-runner-scale-set/values.yaml
dm3ch: Added missing part of dind container the configuration.
Please re check

charts/gha-runner-scale-set/templates/_helpers.tpl
dm3ch: I also have no thoughts how to better workaround this fact right now.

dm3ch: @Link- Do you have any ideas how to better deal with this?

Link-: I'll keep you posted

Link-:
### TODO before releasing

- [ ] Create an announcement to make it clear for folks what changed and the consequences of that. The announcement should clarify that this is an opinionated approach and that folks who want to do things different are always welcome to create their own chart
- [x] Pass E2E tests: https://github.com/actions/actions-runner-controller/actions/runs/6266116102

charts/gha-runner-scale-set/templates/_helpers.tpl
Link-: I don't see an alternative. Let's move forward with it.

Link- (approved):

alpozcan:
There is no more `docker.sock` on AWS EKS cluster, as they have been switched over to containerd. We now have`/var/run/containerd/containerd.sock` in its place, which is not the same thing interface-wise.

My understanding is that we would have to deploy [nerdctl](https://github.com/containerd/nerdctl) or similar into our runner images if we want to do this seamlessly, but have to drop docker & docker-compose client tools in favour of this, at a minimum.

Are there any plans to support and/or documented that setup?

shreelola:
Running containers inside containers is not working with this change. 
Do you have any workaround for this?

` docker: Cannot connect to the Docker daemon at unix:///run/docker.sock. Is the docker daemon running?.`

nikola-jokic:
Hey @shreelola,

Did you configure your autoscaling runner set with dind mode?

shreelola:
Hi @nikola-jokic 

Yes. 
Standard docker pulls and run works with this change inside the runner pod, but when we try to run the container inside 
the container, it fails. 

Edit:

I delved further and discovered that the issue is not related to this change; it appears that the problem stems from the controller modification.

The old controller version was using TCP here: https://github.com/actions/actions-runner-controller/blob/v0.26.0/controllers/runner_controller.go#L1011
with the new runner version, it is using unix socket: https://github.com/actions/actions-runner-controller/blob/v0.27.5/controllers/actions.summerwind.net/runner_controller.go#L1072. 

We started experiencing this issue after upgrading the runner version. Any insights would be highly appreciated. Thank you for your time.

nikola-jokic:
The latest version uses unix sockets instead of tcp sockets. [This issue](https://github.com/actions/actions-runner-controller/issues/2967) references the same problem. As a workaround, may I suggest commenting out the entire `containerMode` object and providing [the dind spec](https://github.com/actions/actions-runner-controller/blob/abc0b678d323b9016af467bdedd2c4fea85a7769/charts/gha-runner-scale-set/values.yaml#L92) as it was described in prior releases?

PR #2837, opened 2023-08-25
Fix parsing AcquireJob MessageQueueTokenExpiredError

nikola-jokic:


github/actions/client.go
```diff
@@ -634,7 +634,18 @@ func (c *Client) AcquireJobs(ctx context.Context, runnerScaleSetId int, messageQ
 	}
 
 	if resp.StatusCode != http.StatusOK {
-		return nil, ParseActionsErrorFromResponse(resp)
+		if resp.StatusCode != http.StatusUnauthorized {
```
TingluoHuang, line 637: do we have test for these clients? 😄

nikola-jokic: I just added one :relaxed:

TingluoHuang (approved):

PR #2838, opened 2023-08-25
Switch git-lfs source to packagecloud

jb-2020:
`git-lfs` from the [Ubuntu source](https://launchpad.net/ubuntu/+source/git-lfs) is using golang `github.com/golang/go:1.13.5` for Ubuntu 20 and `github.com/golang/go:1.18.1` for Ubuntu 22.  

Both of these versions of Go have known vulnerabilities causing our internal security scanning to fail. 

Update to the recommended install instructions for [git-lfs](https://github.com/git-lfs/git-lfs/blob/main/INSTALLING.md) (packagecloud source). 

We're now pulling in: `git-lfs/3.3.0 (GitHub; linux arm64; go 1.19.3)`

### DIFF
--- a/runner/actions-runner-dind-rootless.ubuntu-20.04.dockerfile
+++ b/runner/actions-runner-dind-rootless.ubuntu-20.04.dockerfile
@@ -27,7 +27,6 @@ RUN apt-get update -y \
     dnsutils \
     ftp \
     git \
-    git-lfs \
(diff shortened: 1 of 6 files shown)


mumoshu (approved):
LGTM. Thank you so much for your contribution @jb-2020 ☺️

PR #2843, opened 2023-08-28
Add docs for setting the RUNNER_GRACEFUL_STOP_TIMEOUT env var on docker container

nsheaps:
While investigating some instability with our long running builds using ARC, I found that when termination signals were sent, the runner container would terminate gracefully according to the docs, but the docker daemon would not. The cause was that this env var was not set on the docker container. After looking at the Pod and it's preStop hook, the `docker` container is already setup to handle graceful termination, but the documentation was not clear that this env var needed to be set on that container as well.

This PR clarifies the docs and explicitly says where to add the variable

### DIFF
--- a/docs/automatically-scaling-runners.md
+++ b/docs/automatically-scaling-runners.md
@@ -554,7 +554,7 @@ This can be problematic in two scenarios:
 
 > RunnerDeployment is not affected by the Scenario 1 as RunnerDeployment-managed runners are already tolerable to unlimitedly long in-progress running job while being replaced, as it's graceful termination process is handled outside of the entrypoint and the Kubernetes' pod termination process.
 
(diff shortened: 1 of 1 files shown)


mumoshu (approved):
Good catch! Thank you so much for finding and fixing it @nsheaps ☺️

nsheaps:
No problem @mumoshu ! Please merge this when you feel ready, as I don't have permission to merge it.

Also I think your first-interaction action might be missing some permissions for the token, but I see it's not required.

mumoshu:
@nsheaps Thanks for your confirmation! I was just waiting for the CI build to pass and it looks alright now!
Re: the first-interaction workflow- yeah it seems not working but I think it isn't strictly needed.
Let me merge this now!
Thanks again for your contribution ☺️

PR #2866, opened 2023-09-04
Update about-arc.md

bery:
fix missing slash


Link- (approved):
Thank you for the fix!

PR #2870, opened 2023-09-05
Add missing command

timm-drevensek:
Extends the existing examples with the needed `command` to startup the runner. This is mentioned a few line later - but actually copy & paste of the example is not working and lead to debugging. So having the required line it would have made my day less debugging stuff 😉


nikola-jokic (approved):
LGTM

Link-:
Fixed in #2854

PR #2892, opened 2023-09-14
Extend the user agent and fix the build version for the listener app

nikola-jokic:
Fixes https://github.com/github/actions-runners/issues/75

### Context

ARC's http client user agent needs to be changed to:

`actions-runner-controller/<build version> CommitSHA/<sha> RunnerScaleSetId/<id>`

For calls outside of the listener, we can skip the RunnerScaleSetId.


Link- (approved):
Awesome! Looks good to me let's 🚢 it!

PR #2897, opened 2023-09-14
Add relabeling

stefanandres:
* Add relabelings and metricRelabelings attributes to ServiceMonitors
* Fix template name being lower case for all ServiceMonitor templates
* Fix file extension for template files


gplessis:
This would be super useful to keep the cardinality of some metrics under control 👍 

This needs a rebase though 🙏

stefanandres:
It's rebased now :)

stefanandres:
Closing in favor of using the new scaleset chart. (which doesn't support a podMonitor at all currently, so we have it deployed manually)

Issue #2873, opened 2023-09-06
Runner-scale-set metric cardinality


hamishforbes:
The listener histograms `gha_job_execution_duration_seconds` and `gha_job_startup_duration_seconds` have extremely high cardinality. 
After less than a week and ~800 jobs executed they are the top 2 highest cardinality metrics in my prometheus deployment, with more series than the next 4 metrics combined.

---

Specifically the `runner_id` and `runner_name` labels effectively results in a new unique histogram metric for every single job the scale set runs.

This would be a problem even if the metrics were just counters, but because they're histograms you actually get
(46 bucket metrics + sum + count) * 2 (execution+startup) + 2 (gha_started_jobs_total / gha_completed_jobs_total) = 96 new unique metrics for every executed job.
Only 1 of which is useful (sum) because actually you're only recording a single number, how many seconds it took for that unique job to run.

---

The `job_workflow_ref` is also problematic especially when running jobs on the PR event as the PR numeric ID is included and will create a unique `job_workflow_ref` label.

---

Finally `job_name` is also a potential issue, it is effectively 'user-controlled' and jobs using matrices can and will blow this up massively.

---

As these histograms also persist per-listener pod the number of samples scraped (and time/cpu required) also grows forever.
If your listener pod has been running for a week you'll still be scraping the ingesting the unique metric from the first job run on that scale set a week ago.

After about a week I was ingesting 80k samples per-scrape in around 0.5s.
Restarting all the listeners dropped that to 400 samples and around 10ms.

---

These labels should at minimum *not* be enabled by default as they will eventually break any prometheus deployment that doesn't take extra action to clean them up.

This graph shows the effect on cardinality of removing the various labels
![Screenshot 2023-09-06 at 14 59 38](https://github.com/actions/actions-runner-controller/assets/1282135/5190be09-12b4-452a-9764-7ba7f6149dec)
Note this is just for `gha_job_execution_duration_seconds_count`, you have to multiply by 96 for the actual total number of unique series.

Partial workaround is to add a metric_relabeling config to your prometheus setup like:

```
  - source_labels: [__name__]
    regex: gha_job_(execution|startup)_duration_seconds_bucket
    action: drop
```

This will prevent scraping of the bucket and while it won't stop the sum and count metrics it should keep things under control, you'll probably want to periodically restart the listener pods as well.

MPV:
Agree. 👍 

[...]

PR #2903, opened 2023-09-16
Add the way to configure bucket size for actions-metric-server

### LINKED ISSUE #2739: Disable high-cardinality tags in the metrics-server

A way to stop the metrics-server from tagging metrics with tags of high cardinality.

### Why is this needed?

The metrics-server currently tags metrics with high-cardinality tags like `job_name`, `workflow_name`, `repository`, `head_branch` (see https://github.com/actions/actions-runner-controller/blob/e0a7e142e0fcd446c58e7875d4d44a7eea6e72f2/pkg/actionsmetrics/metrics.go#L79). On top of that it publishes two distribution metrics which have about 45 buckets each (see https://github.com/actions/actions-runner-controller/blob/e0a7e142e0fcd446c58e7875d4d44a7eea6e72f2/pkg/actionsmetrics/metrics.go#L24). Since a separate value is publish for each combination of metric and tags, the number of published values grows very fast.

Example: If you have 10 different jobs names and 5 workflow names across 5 repositories with 5 branches each, you get `10*5*5*5=1250` different values. Since two metrics are distributions with 45 buckets, we get `2*45*1250=112500` values just from those two metrics.

### Additional context

By default when you use Datadog to collect the metrics, it stops at the first 2000 values. The limit can be increased, but the data easily grows to several tens of MBs large per scrape.

If you let me know how you would like this configuration to work (e.g., a toggle per tag?) I'd be happy to open a PR for it.

kirillbilchenko:
Add the way to configure bucket size for actions-metric-server histogram metrics
To set the different bucket need to set the env variable in the values file under `actionsMetricsServer.env`
`PROMETHEUS_RUN_BUCKET_INTERVALS: 1,5,10,300` etc. 
`PROMETHEUS_QUEUE_BUCKET_INTERVALS: 0.5,1,3,5,10,30` etc

resolve https://github.com/actions/actions-runner-controller/issues/2739


MPV:
If you haven't seen this one already it might be worth looking into as well:
- https://github.com/actions/actions-runner-controller/issues/2873

kirillbilchenko:
> If you haven't seen this one already it might be worth looking into as well:
> 
> * [Runner-scale-set metric cardinality #2873](https://github.com/actions/actions-runner-controller/issues/2873)

@MPV Hey thanks for pointing but in my case metrics relabelling not solving the issue, to relabel prometheus first need to scrape it, in my case the payload produced by metric server with default amount of bucket is 30+ megabyte, so prometheus simply fail to do it. By controlling this on application level I can reduce the amount of produced data and better control the buckets, for example for scheduling I'm not sure if we need to have 3000+ second bucket or even 1800 and 600, because this is already a bad indication that something going wrong with queue.

kirillbilchenko:
@Link- @mumoshu do we need this or we can drop it?

Link-:
As this relates to the community modes, I'll leave it to @mumoshu to take action on this

PR #2904, opened 2023-09-18
Helm-Template-Variable created for dind image

kpinarci:
During the implementation of the Actions-runner-controller, I encountered several challenges. A primary issue was my inability to download images within the `dind container` due to persistent timeout issues. Upon closer examination, I realized that the `dind` container wasn't recognizing my specific configurations, including certificates, proxy settings, and `registry-mirror`. Despite numerous attempts, I couldn't correctly relay this information to the `dind` container. While I was able to execute Docker commands within the container, downloading images remained an obstacle.

To circumvent this issue, I created a custom `docker:dind` image that incorporated all the necessary configurations. I then uploaded this image to my own `registry-mirror` and replaced the default hard-coded `dind` image with my tailored version. This approach also allowed me to bypass the rate-limiting challenges I faced when downloading images within the container.

Although this method resolved the issues related to certificates, proxy, and `registry-mirror`, it introduced a downside: I had to make continuous adjustments to the Helm-Chart. Ideally, the chart should offer the flexibility to use a user-defined `docker:dind` image instead of relying on a hard-coded one.

Considering this, I've extended the chart to provide this flexibility. I hope these modifications prove useful to others, and I welcome any feedback or suggestions for improvement.

Issue #211, opened 2020-11-25
is it possible to enable docker registry mirror for the docker:dind container?


lasse-aagren:
We have a local docker registry mirror to avoid hitting docker hub ratelimits. Is there a way of adding something like:
```
$ cat /etc/docker/daemon.json
{
    "registry-mirrors": [
      "https://my-docker-mirror.example.com"
    ]
}
```
Or otherwise pass mirror settings to dockerd in the dind container?

lasse-aagren:
Answering my self:

One solution is to create my own docker:dind image with that modification, and use the `--docker-image=myawesomedocker:dind` parameter for the manager

PR #2910, opened 2023-09-19
Allow adding registry mirror for the DinD mode

timmjd:
Will allow to adapt the Docker in Docker container by the following options:
* Modify the `image` to be use (optional)
* Modify the `command` of the dind container (optional)
* Modify the `args` passed to the `dockerd-entrypoint.sh` of the dind container (optional)
  * See [dockerd](https://docs.docker.com/engine/reference/commandline/dockerd/) options

Background is to be capable to add a __mirror for the DinD__ to come arround rate-limit issues:
> Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit

Example how to add a mirror to be used:
```yaml
template:
  spec:
    containers:
    - name: dind
      args:
      - --registry-mirror=https://<my-docker-mirror-host>
```

Modify the image to be used:
```yaml
template:
  spec:
    containers:
    - name: dind
      image: docker:dind-rootless
```


timmjd:
* Implementation is partly covered by #2904 - but in a more general way
* Solves #211 without the need of a custom image

timmjd:
Hey @mumoshu / @nikola-jokic any chance to get this reviewed?

Link-:
@timmjd - you can inject your own configuration for DinD. Disable `containerMode` and then define your own DinD configuration and add whatever you like. We will not be making customisations to the template as there is no end for the combinations of configurations that can be used.

timmjd:
@Link- But by this statement, I can skip the whole `gha-runner-scale-set` chart - it will become worthless or reduced to a PoC - only the helm operator deployment would stay of interest.

Link-:
> But by this statement, I can skip the whole gha-runner-scale-set chart - it will become worthless or reduced to a PoC - only the helm operator deployment would stay of interest.

@timmjd - I'm not entirely sure what your expectations of the helm charts are here. Our helm charts are a highly opinionated and a fast way for you to break into ARC. As we stand today, they're not a fit for all purposes and most likely won't be. 

We've added the [dind configuration as comments to the values.yaml](https://github.com/actions/actions-runner-controller/blob/2117fd18923aa0c37369426713f0169bb6b2ccf6/charts/gha-runner-scale-set/values.yaml#L110) file for this exact purpose, for whoever needs to customise it further, they can copy / paste and tweak whatever they want.

timmjd:
> We've added the [dind configuration as comments to the values.yaml](https://github.com/actions/actions-runner-controller/blob/2117fd18923aa0c37369426713f0169bb6b2ccf6/charts/gha-runner-scale-set/values.yaml#L110) file for this exact purpose, for whoever needs to customise it further, they can copy / paste and tweak whatever they want.

@Link- But this is exactly what is currently __not working__ - and was was intended with this change!

# Usecase

If you want to enforce the Docker engine of the DinD container to make use of a mirror. This requires you to start the DinD container with a `registry-mirror` parameter.

# The Code

So it must be pasted to the `dind` container as startup parameter. Looking at the helm, this is done [here](https://github.com/actions/actions-runner-controller/blob/2117fd18923aa0c37369426713f0169bb6b2ccf6/charts/gha-runner-scale-set/templates/autoscalingrunnerset.yaml#L137-L138). Expanding this [function](https://github.com/actions/actions-runner-controller/blob/2117fd18923aa0c37369426713f0169bb6b2ccf6/charts/gha-runner-scale-set/templates/_helpers.tpl#L98-L116) you will notice, that there is nothing you can paste to this function.

Therefore, referencing to the [extension point](https://github.com/actions/actions-runner-controller/blob/2117fd18923aa0c37369426713f0169bb6b2ccf6/charts/gha-runner-scale-set/values.yaml#L134-L151) of the template you mentioned: Any option a user would add to this `template` section related to the `dind` container currently will have absolute no effect. This is also why #2904 is also touching the same line within the `_helpers.tpl`.

```yaml
template:
  spec:
    containers:
    - name: dind
      args:
      - --registry-mirror=https://<my-docker-mirror-host>
``` 

This is simply due to nothing gets forwarded trough helm into the k8s yaml.

# Problem

The current implementation you can not configure anything regarding the `dind` container. Therefore, this change is providing an implementation is reflecting how the code is currently written for other parts like the [runner container](https://github.com/actions/actions-runner-controller/blob/2117fd18923aa0c37369426713f0169bb6b2ccf6/charts/gha-runner-scale-set/templates/_helpers.tpl#L189) or even the [init container](https://github.com/actions/actions-runner-controller/blob/2117fd18923aa0c37369426713f0169bb6b2ccf6/charts/gha-runner-scale-set/templates/_helpers.tpl#L85-L96).

But I also agree with you: Adding there more and more options will become a mess sooner or later. Maybe another approach would be possible:
* If nothing according to the `dind` container was specified, use the default config provided via the `_helpers.tpl`
* If there is a `dind` section within the `template`: 100% use it - and don't do a mixin.

Link-:
@timmjd - if you leave the `containerMode` section commend out, you can **specify whatever you want** under `template`.

timmjd:
> @timmjd - if you leave the `containerMode` section commend out, you can **specify whatever you want** under `template`.

Wou, ok - now I got you. Let me try this!

dmitrijrub:
@timmjd can you share example of dockerfile how you build docker:dind-rootless ?

cmergenthaler:
@Link- I see your point but honestly, isn't this what the helm chart is there for? I mean you also allow customizing the runner container, so why not allow customizing the dind container? If we want to add one more argument to the docker daemon or using a custom dind image, it results in adding additional 30 lines of code to the helm-release that come with configuring dind by ourselves (init-container, using shared volume, etc.).

Would really appreciate more flexibility to configure the dind container.

mthemis-provenir:
I agree with @cmergenthaler and others in this thread. While it may indeed be an opinionated helm chart, this is such low hanging fruit and an easy win that it makes no sense _not_ to implement it. Hell, there are pull requests outstanding _that already did the work for you_. I would also argue writing half the manifest yourself in a helm chart values is an anti-pattern.

In our case, we were getting throttled by docker hub while pulling the docker:dind image, which ended up causing [this](https://github.com/actions/actions-runner-controller/issues/2721), which _also_ hasn't been fixed :roll_eyes:.

PR #2915, opened 2023-09-21
Fix name override labels when runnerScaleSetName value is set

nikola-jokic:
Fixing labels applied based on the runnerScaleSetName label.

Fixing missed modification thanks to [this comment](https://github.com/actions/actions-runner-controller/pull/2803#issuecomment-1726830304)


Link- (approved):
Let's ship 🚀

PR #2917, opened 2023-09-21
set pod labels

### LINKED ISSUE #2753: Be able to set pod labels

When installing **gha-runner-scale-set-controller** with the helm chart, have a `podlabels` field in the **values.yaml**, that can be used to set labels for the pods. In a similar way as `podannotations`.

### Why is this needed?

We have policies in place that require some specific pod labels in our clusters, so it would be nice to have this option. This template value is also available in the "old"  actions-runner-controller which we are migrating from.

### Additional context

Something like this https://github.com/actions/actions-runner-controller/blob/5271f316e6e068f0daf4b202d7c8b9f21301fbd1/charts/actions-runner-controller/values.yaml#L88

robwhitby:
allows adding optional `labels` to the controller pod.

fixes #2753


nikola-jokic:
Hey @robwhitby,

Closing this PR since it has been implemented [here](https://github.com/actions/actions-runner-controller/pull/2952).

PR #2919, opened 2023-09-21
Add an option to use rootless DinD

0xiso:
This pull request introduces a new option (`containerMode.rootless`) that allows the use of rootless DinD instead of rootful DinD. This feature provides security enhancements by reducing the risk of container breakout (https://github.com/actions/actions-runner-controller/issues/1288).

- To enable rootless DinD, set the following options:
  - `containerMode.type=dind`
  - `containerMode.rootless=true`
- To continue using rootful DinD, set the following options:
  - `containerMode.type=dind`
  - Omit `containerMode.rootless` or set it to `false`.

Existing setups will not be affected by this pull request, as the behavior remains the same if `containerMode.rootless` is not set.

There is an existing pull request (https://github.com/actions/actions-runner-controller/pull/2910) that allows users to specify the image for the `dind` container. However, supporting rootless DinD as an official and dedicated option is an important security improvement. It may also be worth considering making rootless DinD the default option in future releases.


kevholmes:
Isn’t this already supported?

https://github.com/actions/actions-runner-controller/blob/cf24ab584d01d0df2c03aff1b3805bd22a176e5f/docs/releasenotes/0.26.md?plain=1#L19

0xiso:
@kevholmes Thank you for your comment. That is for the legacy actions-runner-controller and not for the new Autoscaling Runner Scale Sets mode.

In addition to the above, compared to the legacy rootless DinD, my pull request has some benefits.

In a legacy rootless DinD setup, the user's workflow and Docker daemon run in the same container. This means that a user can execute any command in a privileged container, which I believe is not desirable from a security perspective. (Although you can't do a container breakout easily as `sudo` isn't allowed in that image, there's still a risk.)

On the other hand, my pull request has the following advantages.

- The runner and Docker daemon operate in different containers, so users can't execute arbitrary commands directly in a privileged container.
- The Docker daemon running in a privileged container runs as a non-root user, so the risk of a container breakout is _relatively_ low.
- By using the official Docker image (`docker:dind-rootless`), we don't need to upgrade a Docker daemon by ourselves.

kevholmes:
> @kevholmes Thank you for your comment. That is for the legacy actions-runner-controller and not for the new Autoscaling Runner Scale Sets mode.

It shows you how out of the loop I am here 😆 thank you for clarifying! I have not had a chance to try out the new Runner Scale Sets model.

> In a legacy rootless DinD setup, the user's workflow and Docker daemon run in the same container. This means that a user can execute any command in a privileged container, which is certainly undesirable from a security perspective. (Although you can't do a container breakout easily as `sudo` isn't allowed in that image, there's still a risk.)

Yes, it is absolutely a risk. In the past, I've tried to ensure daily runner image updates/rebuilds in CI/CD environments to avoid local privilege escalation concerns if a 0-day came out and the runner images were lagging on patches. It's nice to limit that concern even further to only the docker-dind container.

> On the other hand, my pull request has the following advantages.
> 
>  The runner and Docker daemon operate in different containers, so users can't execute arbitrary commands directly in a privileged container.
>  The Docker daemon running in a privileged container runs as a non-root user, so the risk of a container breakout is _relatively_ low.

Just checking the new Runner Scale Set template out now, and I see what you're saying - the runner container executing workflows/Actions itself is unprivileged, while the rootless-dind container gets escalated privileges. A good step in the right direction for sure. The only real risk still is what instructions are fed into the rootless-dind container from the runner container ... but it runs as a non-root user as you've mentioned, so the breakout concern is low but not entirely perfect as we'll still need to watch for 0-days impacting Docker or related dependencies within the docker-dind image.

>  By using the official Docker image (`docker:dind-rootless`), we don't need to upgrade a Docker daemon by ourselves.

Def a nice upgrade, also!

Thank you for the clarification @0xiso! I'll certainly be trying this out when I get a chance to upgrade ARC.

0xiso:
Due to the changes made in https://github.com/actions/actions-runner-controller/pull/2833, I have to update this PR. Please wait a few days as I will update it soon.

0xiso:
I've updated the pull request to support the UNIX socket dockerd connection that was introduced by https://github.com/actions/actions-runner-controller/pull/2833.

In rootless DinD mode, a runner container and a dind container run with different UID, and leading to numerous permission issues. (The runner container runs with the runner user (UID: 1001), and the rootless dind container runs with the rootless user (UID: 1000).) To solve those issues, we need to run the rootless dind with UID: 1001 in some way.

The most straightforward way would be to rebuild the rootless-dind image with a runner user added. However, this method introduces the necessity of maintaining the new container image. So, this pull request proposes using an initContainer to add a runner user into a dind container.

wilmardo:
Have you tested the container with just the securityContext set? If I remember correctly Kubernetes doesn't need the entries to exist in the `/etc/passwd`

0xiso:
@wilmardo You're correct. Kubernetes doesn't require an entry in `/etc/passwd` to run a process with a specific UID/GID. 

However, to run a rootless dockerd, the user's home directory must be properly set and writable. The user's home directory can be specified in an environment variable (`$HOME`), but I thought it more natural to specify in `/etc/passwd`.

wilmardo:
> However, to run a rootless dockerd, the user's home directory must be properly set and writable. The user's home directory can be specified in an environment variable ($HOME), but I thought it more natural to specify in /etc/passwd.

Ah explains, would setting $HOME to `/dev/shm` then be sufficient? That would remove the init container.
`/dev/shm` is by default provided on Linux and is a memory backed TMPFS and has `777` as default permission (even is writable in a readonly container).

0xiso:
@wilmardo
> Ah explains, would setting $HOME to /dev/shm then be sufficient

Rootless dockerd will extract container images to a directory in the user's home directory, and those won't fit in `/dev/shm`. We can use emptyDir for that purpose, which is proposed by this PR.

> That would remove the init container.

Modifying `/etc/subuid` and `/etc/subgid` is [necessary for rootless dockerd](https://docs.docker.com/engine/security/rootless/#prerequisites). The init container also serves this purpose. Although other alternatives exist for this purpose, such as the config map, I think the init container is the simplest solution.

ajschmidt8:
> I've updated the pull request to support the UNIX socket dockerd connection that was introduced by #2833.
> 
> In rootless DinD mode, a runner container and a dind container run with different UID, and leading to numerous permission issues. (The runner container runs with the runner user (UID: 1001), and the rootless dind container runs with the rootless user (UID: 1000).) To solve those issues, we need to run the rootless dind with UID: 1001 in some way.
> 
> The most straightforward way would be to rebuild the rootless-dind image with a runner user added. However, this method introduces the necessity of maintaining the new container image. So, this pull request proposes using an initContainer to add a runner user into a dind container.

I just stumbled upon this PR. I'm exploring the changes now. I posted a similar discussion below for reference:

- https://github.com/actions/actions-runner-controller/discussions/2930#discussioncomment-7123779

dnewhook:
I tested your proposal with the 0.6.0 release of the charts by un-setting containerMode and using the following template.  It seems to be working great.
```
template:
  spec:
    initContainers:
    - name: init-dind-externals
      image: private.registry/arc/actions-runner:2.309.0
      command: ["cp", "-r", "-v", "/home/runner/externals/.", "/home/runner/tmpDir/"]
      volumeMounts:
        - name: dind-externals
    - name: dind-rootless-init-container
      image: private.registry/arc/docker:24.0.6-dind-rootless
      command:
        - sh
        - -c
        - |
          set -x
          cp -a /etc/. /dind-etc/
          echo 'runner:x:1001:1001:runner:/home/runner:/bin/ash' >> /dind-etc/passwd
          echo 'runner:x:1001:' >> /dind-etc/group
          echo 'runner:100000:65536' >> /dind-etc/subgid
          echo 'runner:100000:65536' >>  /dind-etc/subuid
          chmod 755 /dind-etc;
          chmod u=rwx,g=rx+s,o=rx /dind-home
          chown 1001:1001 /dind-home
      securityContext:
        runAsUser: 0
      volumeMounts:
        - mountPath: /dind-etc
          name: dind-etc
        - mountPath: /dind-home
          name: dind-home          
    containers:
    - name: runner
      image: private.registry/arc/actions-runner:2.309.0
      env:
        - name: DOCKER_HOST
          value: unix:///run/docker/docker.sock
      volumeMounts:
        - name: work
          mountPath: /home/runner/_work
        - name: dind-sock
          mountPath: /run/docker
          readOnly: true
    - name: dind
      image: private.registry/arc/docker:24.0.6-dind-rootless
      args:
        - dockerd
        - --host=unix:///run/docker/docker.sock
      securityContext:
        privileged: true
        runAsUser: 1001
        runAsGroup: 1001        
      volumeMounts:
        - name: work
          mountPath: /home/runner/_work
        - name: dind-sock
          mountPath: /run/docker          
        - name: dind-externals
          mountPath: /home/runner/externals
        - name: dind-etc
          mountPath: /etc
        - name: dind-home
          mountPath: /home/runner
    volumes:
    - name: work
      emptyDir: {}
    - name: dind-externals
      emptyDir: {}
    - name: dind-sock
      emptyDir: {}
    - name: dind-etc
      emptyDir: {}
    - name: dind-home
      emptyDir: {}
    imagePullSecrets:
    - name: regcred
```

0xiso:
Hi @Link-, I'm eager to hear any feedback you might have. Let me know if there's anything I can do to help get it merged.

Link-:
@0xiso thank you for putting this together. The approach that @dnewhook highlighted is what we designed for. At this point in time we have not made the decision to add a new container mode for dind-rootless as it will bring with it a new set of complexities we're not able to deal with as things stand today.

I see no technical blocker for you to comment out `containerMode:` and just inject this configuration under `template` and it should work just fine!

Link-:
I wrote a detailed answer for this here: https://github.com/actions/actions-runner-controller/discussions/3117

I'm going to close this PR for now, we'll reopen it if the team decides this is an option we'll support

0xiso:
Hi @Link-, thank you for the detailed answer.

This PR was a feature addition but also a security-fix proposal. As you know, rootful DinD allows access to host devices and, in many cases, allows container breakout.

My colleague shared with me an article about the security risk of shared CI/CD runners. This article is about GitLab, but it also applies to this project. https://pulsesecurity.co.nz/articles/OMGCICD-gitlab

I respect your decision, but I think this risk should at least be clearly stated in the documentation. (It is likely that not a few organizations are using ARC without recognizing this risk.) And ideally, rootless DinD should be the default, and rootful DinD should be optional or obsolete in the near future.

Link-:
@0xiso - thank you, I think indeed we should document this risk. I'll work with the team on adding it to our documentation here: https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/deploying-runner-scale-sets-with-actions-runner-controller#using-docker-in-docker-mode

PR #2920, opened 2023-09-22
Ignore summerwind resources in go test until controller version is the same

nikola-jokic:


config/crd/bases/actions.summerwind.dev_runnerdeployments.yaml
```diff
@@ -102,6 +103,7 @@ spec:
                       description: matchLabels is a map of {key,value} pairs. A single {key,value} in the matchLabels map is equivalent to an element of matchExpressions, whose key field is "key", the operator is "In", and the values array contains only "value". The requirements are ANDed.
                       type: object
                   type: object
+                  x-kubernetes-map-type: atomic
```
Link-, line 106: What are these changes?

nikola-jokic: It [extends the definition for an object](https://github.com/kubernetes/kubernetes/pull/84113) since 1.17 release
Updated controller-gen generates it by default and it is how we generated listener CRD, so I think it is safe :relaxed:

Link- (approved):

PR #2937, opened 2023-09-27
docs: how to extend the PATH environment variable

### LINKED ISSUE #1135: How to Customize Runner Environment?
We are extending the runner base image to provision additional software similar to what GitHub does in their virtual environments. These provisioners also need to set up additional environment variables dynamically since they install software with dynamic versions. The natural choice to do so is by adding those environment variables to `/etc/environment`. However, this file is never loaded by the runners.

Hence, we set out to find out what is going on, and it seems that there is no reliable way to actually do this, other than manipulating files of the base image. This is obviously brittle, since those files can change at any time and are not standardized like the files of the operating system (e.g. `/etc/environment`). Let me explain what we found.

The entrypoint is set to `/usr/local/bin/dumb-init` and the initial command is set to `/entrypoint.sh`. The `/entrypoint.sh` script copies all non-dot files from `/runnertmp` to `/runner`.[^cpr] Later it replaces the standard `runsvc.sh` with a pachted `runsvc.sh`.[^patched] This file contains the following lines:

https://github.com/actions-runner-controller/actions-runner-controller/blob/0b9bef2c086a54b0731fa7aa0a6bd7cb856b48ee/runner/patched/runsvc.sh#L7-L13

The `.path` file can never exist (unless it would be created by some of the commands that are being called in `/entrypoint.sh`) since dot files are not copied over to `/runner`. The line that states that this is the place to customize the environment is for sure well intended, but assumes that editing alien scripts is done by a human, which is not the case for us. Customizing it with automation is – as stated before – a bad idea since it may break a lot of things at any point in time.

Imho it would be best to simply read `/etc/environment` since that is what it is intended for, but maybe there are good reasons not to do so?

[^cpr]: https://github.com/actions-runner-controller/actions-runner-controller/blob/0b9bef2c086a54b0731fa7aa0a6bd7cb856b48ee/runner/entrypoint.sh#L79-L84
[^patched]: https://github.com/actions-runner-controller/actions-runner-controller/blob/0b9bef2c086a54b0731fa7aa0a6bd7cb856b48ee/runner/entrypoint.sh#L158-L162

### LINKED ISSUE #2273: Document extending the runner environment

I would like the preferred solution for extending the runner environment/image and runner `PATH` to be documented in https://github.com/actions/actions-runner-controller/blob/master/docs/about-arc.md#software-installed-in-the-runner-image

I am currently setting `PATH` via `/runnertmp/.env` after reading https://github.com/actions/actions-runner-controller/issues/1135. I tried the `.path` solution mentioned and it did not work.

I would make a PR but am not sure if the `.env` solution is the preferred/correct solution for adding binaries to the runners path and would like an official stance

### Why is this needed?

Users who need custom paths in their runners do not have a clear path to integrating them into the runners image at this time.

I needed to add `gcloud` to the runner image to enable docker `gcr` authentication for the job `container` option and the override of `PATH` in my image was being ignored.

Previously, I used `summerwinds/action-runner:v2.285.1-ubuntu-20.04` as my base image, added the `gcloud` binaries and updated my `PATH`. When I upgraded to newer runner versions, my changes to the `PATH` were being ignored. 

### Additional context

I spent significant time reading the docs and could not find a solution. Reading through https://github.com/actions/actions-runner-controller/issues/1135 required me to try a variety of solutions and did not explicitly state which worked/was preferred

morremeyer:
This adds documentation for the extension of the `PATH` environment variable.

Resolves #1135.
Resolves #2273.

### DIFF
--- a/docs/about-arc.md
+++ b/docs/about-arc.md
@@ -6,7 +6,7 @@ This document provides a high-level overview of Actions Runner Controller (ARC).
 With this overview, you can get a foundation of basic scenarios and be capable of reviewing other advanced topics.
 
 ## GitHub Actions
-[GitHub Actions](https://github.com/features/actions) is a continuous integration and continuous delivery (CI/CD) platform to automate your build, test, and deployment pipeline. 
+[GitHub Actions](https://github.com/features/actions) is a continuous integration and continuous delivery (CI/CD) platform to automate your build, test, and deployment pipeline.
 
 You can create workflows that build and test every pull request to your repository, or deploy merged pull requests to production. Your workflow contains one or more jobs which can run in sequential order or in parallel. Each job will run inside its own runner and has one or more steps that either run a script that you define or run an action, which is a reusable extension that can simplify your workflow. To learn more about Actions - see "[Learn Github Actions](https://docs.github.com/en/actions/learn-github-actions)".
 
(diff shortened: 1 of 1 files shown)


rakeshramakrishnan-hbk:
Tried the following to persist the PATH changes on the new RunnerSets - `RUN echo "PATH=$PATH" >> /runnertmp/.env`
But ^ threw an error that the `/runnertmp` path does not exist.

Does this doc apply for the new RunnerSets, or for the legacy runners?

morremeyer:
We migrated to RunnerSets, where this does not apply anymore. Therefore, we also cleaned up our fork.

PR #2952, opened 2023-10-02
Allow custom labels to be specified for controller pods

### LINKED ISSUE #2951: gha-runner-scale-set-controller: supporting the custom labels on the pod template level

I would like to add the possibility to customise the labels on the pod template level similarly to how it's done on the annotations.

### Why is this needed?

Our organisation requires to set some labels on the pod level in order to apply some network policies. Without them the gatekeeper rejects the deployments.

The rationale behind is very simple: we can modify the annotations on the pod level as well as the labels on the deployment level, but these on the pod level cannot be changed without modifying the chart. It should be allowed IMHO. 

### Additional context

**Possible implementation**

`charts/gha-runner-scale-set-controller/templates/deployment.yaml`
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: {{ include "gha-runner-scale-set-controller.fullname" . }}
  labels:
    {{- include "gha-runner-scale-set-controller.labels" . | nindent 4 }}
    actions.github.com/controller-service-account-namespace: {{ .Release.Namespace }}
    actions.github.com/controller-service-account-name: {{ include "gha-runner-scale-set-controller.serviceAccountName" . }}
    {{- if .Values.flags.watchSingleNamespace }}
    actions.github.com/controller-watch-single-namespace: {{ .Values.flags.watchSingleNamespace }}
    {{- end }}
spec:
  replicas: {{ default 1 .Values.replicaCount }}
  selector:
    matchLabels:
      {{- include "gha-runner-scale-set-controller.selectorLabels" . | nindent 6 }}
  template:
    metadata:
      annotations:
        kubectl.kubernetes.io/default-container: "manager"
      {{- with .Values.podAnnotations }}
        {{- toYaml . | nindent 8 }}
      {{- end }}
      labels:
        app.kubernetes.io/part-of: gha-rs-controller
        app.kubernetes.io/component: controller-manager
        app.kubernetes.io/version: {{ .Chart.Version }}
        {{- include "gha-runner-scale-set-controller.selectorLabels" . | nindent 8 }}
        {{- with .Values.podTemplateLabels }}
          {{- toYaml . | nindent 8 }}
        {{- end }}
...
```
`charts/gha-runner-scale-set-controller/values.yaml`
```
# Default values for gha-runner-scale-set-controller.
# This is a YAML-formatted file.
# Declare variables to be passed into your templates.
labels: {}

podTemplateLabels: {}
...
```

wherka-ama:
Add ability to add custom labels to the controller pods.

Fixes #2951

### DIFF
--- a/charts/gha-runner-scale-set-controller/values.yaml
+++ b/charts/gha-runner-scale-set-controller/values.yaml
@@ -3,6 +3,8 @@
 # Declare variables to be passed into your templates.
 labels: {}
 
+podTemplateLabels: {}
+
 # leaderElection will be enabled when replicaCount>1,
 # So, only one replica will in charge of reconciliation at a given time
 # leaderElectionId will be set to {{ define gha-runner-scale-set-controller.fullname }}.
--- a/charts/gha-runner-scale-set-controller/values.yaml
+++ b/charts/gha-runner-scale-set-controller/values.yaml
@@ -3,8 +3,6 @@
 # Declare variables to be passed into your templates.
 labels: {}
 
-podTemplateLabels: {}
-
 # leaderElection will be enabled when replicaCount>1,
 # So, only one replica will in charge of reconciliation at a given time
 # leaderElectionId will be set to {{ define gha-runner-scale-set-controller.fullname }}.
@@ -42,6 +40,8 @@ serviceAccount:
   name: ""
 
 podAnnotations: {}
+
+podLabels: {}
 
 podSecurityContext: {}
 # fsGroup: 2000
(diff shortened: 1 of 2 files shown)


wherka-ama:
/fixes https://github.com/actions/actions-runner-controller/issues/2951

charts/gha-runner-scale-set-controller/values.yaml
```diff
@@ -3,6 +3,8 @@
 # Declare variables to be passed into your templates.
 labels: {}
 
+podTemplateLabels: {}
```
nikola-jokic: Could you move this below the podAnnotations and also name it `podLabels`?

wherka-ama: Done. 
Thanks a lot @nikola-jokic !

nikola-jokic: I'll run some tests just to make sure nothing breaks, but looks good to me! Thank you for your contribution!

nikola-jokic (approved):
LGTM! Thanks!

PR #2965, opened 2023-10-06
Configure listener pod with the secret instead of env

nikola-jokic:
This PR changes the way the listener pod is configured. With environment variables, there can be a possibility where a large certificate injected as env crashes the pod. Trimming down the certificate chain is an option, but switching to secrets allows us to extend the configuration without affecting other configuration variables or potentially breaking the same configuration from one version to another.

It can potentially also increase security since users are more likely to have access to inspect the pod where environment variables are being passed, than to inspect secrets.


cmd/githubrunnerscalesetlistener/config/config_test.go
```diff
@@ -0,0 +1,92 @@
+package config
```
Link-, line 1: Do we need to cover the rest of the scenarios? (LogLevel, LogFormat, Metrics, Certificate etc..)

controllers/actions.github.com/resourcebuilder.go
```diff
@@ -144,133 +132,101 @@ type listenerMetricsServerConfig struct {
...
-					},
-					Key: "github_app_private_key",
-				},
-			},
-		})
+	config := listenerconfig.Config{
+		ConfigureUrl:                autoscalingListener.Spec.GitHubConfigUrl,
+		AppID:                       appID,
+		AppInstallationID:           appInstallationID,
+		AppPrivateKey:               string(secret.Data["github_app_private_key"]),
```
Link-, line 183: Is there a reason why we need to have these secrets in the config as opposed to passing a secret reference that the listener can use to fetch the data? Is it because we want to allow the listener to maintain isolation from the controller-manager namespace?

This is iffy IMHO.. Maybe the best course of action is for the controller-manager to create the necessary secrets in the appropriate namespace that the listener can read.

All I'm saying is, we might not want these written to a configuration file.

controllers/actions.github.com/autoscalinglistener_controller_test.go
```diff
@@ -965,31 +1002,26 @@ var _ = Describe("Test GitHub Server TLS configuration", func() {
...
-					var env *corev1.EnvVar
-					for _, e := range pod.Spec.Containers[0].Env {
-						if e.Name == "GITHUB_SERVER_ROOT_CA" {
-							env = &e
-							break
-						}
-					}
-					g.Expect(env).NotTo(BeNil(), "pod should have an env variable named GITHUB_SERVER_ROOT_CA_PATH")
+
+					g.Expect(config.Data["config.json"]).ToNot(BeEmpty(), "config.json should not be empty")
```
Link-:
```suggestion
					g.Expect(config.Data["config.json"]).ToNot(BeEmpty(), "listener configuration file should not be empty")
```

controllers/actions.github.com/resourcebuilder.go
nikola-jokic: I personally wouldn't like to have many secrets wired up with the listener. The reason is that reading config from many secrets is less clear in the code and also, I would like to have a simple entrypoint with the configuration. If we end up using many secrets, then we would need to read them from many places. We would have to mount those secrets at different directories. If we do it like this, there is a single place from which we read the configuration. It is somewhat similar to the env. We know what env keys are, we try to read them, but essentially, environment is a single thing exposed to the process.

Having many secrets would unnecessarily complicate a simple operation. We just need the data that we can use to start the process. It also simplifies debugging.

controllers/actions.github.com/autoscalinglistener_controller_test.go
```diff
@@ -965,31 +1002,26 @@ var _ = Describe("Test GitHub Server TLS configuration", func() {
...
+
+					var listenerConfig listenerconfig.Config
+					err = json.Unmarshal(config.Data["config.json"], &listenerConfig)
+					g.Expect(err).NotTo(HaveOccurred(), "failed to unmarshal config.json")
```
Link-:
```suggestion
					g.Expect(err).NotTo(HaveOccurred(), "failed to parse listener configuration file")
```

controllers/actions.github.com/autoscalinglistener_controller.go
```diff
@@ -403,7 +418,35 @@ func (r *AutoscalingListenerReconciler) createListenerPod(ctx context.Context, a
...
+		if !kerrors.IsNotFound(err) {
+			logger.Error(err, "Unable to get listener config secret", "namespace", autoscalingListener.Namespace, "name", scaleSetListenerConfigName(autoscalingListener))
+			return ctrl.Result{Requeue: true}, err
+		}
+
+		logger.Info("Creating listener config secret")
+
+		podConfig, err := r.resourceBuilder.newScaleSetListenerConfig(autoscalingListener, secret, metricsConfig, cert)
+		if err != nil {
+			logger.Error(err, "Failed to build listener config secret")
```
Link-, line 464: Are we doing uppercase or lowercase for error messages?

nikola-jokic: When logging, uppercase first is preferred. When creating an error, lowercase is :relaxed: We don't always follow this, but I think it should be applied

controllers/actions.github.com/resourcebuilder.go
Link-: @nikola-jokic - I don't mean having multiple secrets per se. If we step back a minute and revisit this question: why is the listener not able to pull the secret from the API, and store / use it from memory? Why do we need to mount it?

Link- (approved):
Let's move this forward, thank you Nikola!

PR #2970, opened 2023-10-09
Fix role and rolebinding cleanup for the listener controller

### LINKED ISSUE #2781: helm uninstall leaves the autoscalingrunnerset and other resources behind(missing rbac?)
### Checks

### Controller Version

0.23.3

### Helm Chart Version

0.4.0

### CertManager Version

_No response_

### Deployment Method

Helm

### cert-manager installation

This is for ARC v2, so cert-manager may not be relevant here.

### Checks

### Resource Definitions

```yaml
Arc v2 - these are not relevant here.

<values.yaml>
githubConfigSecret: runners-secret

## maxRunners is the max number of runners the auto scaling runner set will scale up to.
maxRunners: 10

## minRunners is the min number of runners the auto scaling runner set will scale down to.
minRunners: 0

runnerGroup: "runner_group"

## template is the PodSpec for each runner Pod
template:
  spec:
    containers:
    - name: runner
      image: <internal_repo>/runner-image-infra:x.x.x
      command: ["./entrypoint-dind-rootless.sh"]
      securityContext:
        privileged: true
        runAsUser: 1001
        runAsGroup: 0
      env:
        - name: DOCKER_CLIENT_ADDRESS
          value: unix:///run/user/1001/docker.sock
        - name: ENTRYPOINT_SCRIPT
          value: /home/runner/run.sh
        - name: DISABLE_RUNNER_UPDATE
          value: "false"
      volumeMounts:
        - name: work
          mountPath: /home/runner/_work
      resources:
        requests:
          cpu: 100m
          memory: 100Mi
        limits:
          cpu: 1
          memory: 4Gi
    volumes:
    - name: work
      emptyDir: {}

## Optional controller service account that needs to have required Role and RoleBinding
## to operate this gha-runner-scale-set installation.
## The helm chart will try to find the controller deployment and its service account at installation time.
## In case the helm chart can't find the right service account, you can explicitly pass in the following value
## to help it finish RoleBinding with the right service account.
## Note: if your controller is installed to only watch a single namespace, you have to pass these values explicitly.
controllerServiceAccount:
  namespace: actions-runner-controller
  name: actions-runner-controller
```

### To Reproduce

```markdown
1. install chart release
helm upgrade --install --namespace actions-runner-controller --skip-crds --debug --values=values.yaml infra oci://ghcr.io/actions/actions-runner-controller-charts/gha-runner-scale-set
2. uninstall the release
helm uninstall infra
3. list resources (autoscalingrunnerset, role, rolebinding)
```

### Describe the bug

Helm uninstall does not seem to work in a consistent way. Sometimes it leaves autoscalingrunnerset, roles, rolebindings. However, it's not super consistent.
Sometimes these objects prevent a subsequent re-installations on the the release name as well.

Helm uninstall should enforce the clean up of all the resources managed by a specific release. 

### Whole Controller Logs

```shell
N/A
```

### Whole Runner Pod Logs

```shell
N/A
```

### Additional Context

The situation seems to be improved when the following patch is applied(it's not covering 100% of the cases though):

```
diff --git a/charts/gha-runner-scale-set/templates/manager_role.yaml b/charts/gha-runner-scale-set/templates/manager_role.yaml
index f6a1e49..df7a335 100644
--- a/charts/gha-runner-scale-set/templates/manager_role.yaml
+++ b/charts/gha-runner-scale-set/templates/manager_role.yaml
@@ -23,6 +23,15 @@ rules:
   - pods/status
   verbs:
   - get
+- apiGroups:
+  - ""
+  resources:
+  - pods/finalizers
+  verbs:
+  - create
+  - delete
+  - get
+  - list
 - apiGroups:
   - ""
   resources:
   ```

### PR DESCRIPTION
**Pull request by nikola-jokic**
After investigation, it seems like we forgot to delete `Role` and `RoleBinding` resources when doing the listener cleanup. This PR aims to fix it

Fixes #2781

### DIFF
--- a/controllers/actions.github.com/autoscalinglistener_controller.go
+++ b/controllers/actions.github.com/autoscalinglistener_controller.go
@@ -306,6 +306,38 @@ func (r *AutoscalingListenerReconciler) cleanupResources(ctx context.Context, au
 		logger.Info("Listener proxy secret is deleted")
 	}
 
+	listenerRoleBinding := new(rbacv1.RoleBinding)
+	err = r.Get(ctx, types.NamespacedName{Namespace: autoscalingListener.Spec.AutoscalingRunnerSetNamespace, Name: scaleSetListenerRoleName(autoscalingListener)}, listenerRoleBinding)
+	switch {
+	case err == nil:
+		if listenerRoleBinding.ObjectMeta.DeletionTimestamp.IsZero() {
+			logger.Info("Deleting the listener role binding")
+			if err := r.Delete(ctx, listenerRoleBinding); err != nil {
+				return false, fmt.Errorf("failed to delete listener role binding: %v", err)
+			}
+		}
+		return false, nil
+	case err != nil && !kerrors.IsNotFound(err):
+		return false, fmt.Errorf("failed to get listener role binding: %v", err)
+	}
+	logger.Info("Listener role binding is deleted")
+
+	listenerRole := new(rbacv1.Role)
+	err = r.Get(ctx, types.NamespacedName{Namespace: autoscalingListener.Spec.AutoscalingRunnerSetNamespace, Name: scaleSetListenerRoleName(autoscalingListener)}, listenerRole)
+	switch {
+	case err == nil:
+		if listenerRole.ObjectMeta.DeletionTimestamp.IsZero() {
+			logger.Info("Deleting the listener role")
+			if err := r.Delete(ctx, listenerRole); err != nil {
+				return false, fmt.Errorf("failed to delete listener role: %v", err)
+			}
+		}
+		return false, nil
+	case err != nil && !kerrors.IsNotFound(err):
+		return false, fmt.Errorf("failed to get listener role: %v", err)
+	}
+	logger.Info("Listener role is deleted")
+
 	logger.Info("Cleaning up the listener service account")
 	listenerSa := new(corev1.ServiceAccount)
 	err = r.Get(ctx, types.NamespacedName{Name: scaleSetListenerServiceAccountName(autoscalingListener), Namespace: autoscalingListener.Namespace}, listenerSa)
--- a/controllers/actions.github.com/autoscalinglistener_controller_test.go
+++ b/controllers/actions.github.com/autoscalinglistener_controller_test.go
@@ -190,7 +190,7 @@ var _ = Describe("Test AutoScalingListener controller", func() {
 	})
 
 	Context("When deleting a new AutoScalingListener", func() {
-		It("It should cleanup all resources for a deleting AutoScalingListener before removing it", func() {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: controllers/actions.github.com/autoscalinglistener_controller.go [hunk 1]
```diff
@@ -306,6 +306,38 @@ func (r *AutoscalingListenerReconciler) cleanupResources(ctx context.Context, au
 		logger.Info("Listener proxy secret is deleted")
 	}
 
+	listenerRoleBinding := new(rbacv1.RoleBinding)
+	err = r.Get(ctx, types.NamespacedName{Namespace: autoscalingListener.Spec.AutoscalingRunnerSetNamespace, Name: scaleSetListenerRoleName(autoscalingListener)}, listenerRoleBinding)
```
Link-, line 310: How do we guarantee that we're only deleting the roles we created? What if the end user installed the resources through some other mechanism?

nikola-jokic: These roles are created by the controller here: https://github.com/actions/actions-runner-controller/blob/ba91c183b505dfc7a619e0e56db2bf268ff3d569/controllers/actions.github.com/autoscalinglistener_controller.go#L179

We just never did a cleanup on them

Link- (approved):

controllers/actions.github.com/autoscalinglistener_controller.go
```diff
@@ -306,6 +306,38 @@ func (r *AutoscalingListenerReconciler) cleanupResources(ctx context.Context, au
...
+			}
+		}
+		return false, nil
+	case err != nil && !kerrors.IsNotFound(err):
+		return false, fmt.Errorf("failed to get listener role binding: %v", err)
+	}
+	logger.Info("Listener role binding is deleted")
+
+	listenerRole := new(rbacv1.Role)
+	err = r.Get(ctx, types.NamespacedName{Namespace: autoscalingListener.Spec.AutoscalingRunnerSetNamespace, Name: scaleSetListenerRoleName(autoscalingListener)}, listenerRole)
```
Link-, line 326: Do we need to create / delete the role in the controller? Why not create the role in the charts and the binding is applied by the controller?

Link- (approved):

controllers/actions.github.com/autoscalinglistener_controller.go
nikola-jokic: After some digging, I figured why :smile: 

We can't because the role is assigned to a specific ephemeral runner set by the name. So if AutoscalingRunnerSet is updated, the EphemeralRunnerSet name will change and at that point, we need to update the role: https://github.com/actions/actions-runner-controller/blob/ea2443a410fce4629488b1fde14ed4244d327d03/controllers/actions.github.com/autoscalinglistener_controller.go#L192

PR #3003, opened 2023-10-19
Trim down metrics cardinality

### LINKED ISSUE #2873: Runner-scale-set metric cardinality

nikola-jokic:
Runner ID increases metrics cardinality, since all runners are ephemeral.

To address this problem:
- Exclude `runner_id` and `runner_name` from all metrics labels
- Exclude `job_workflow_ref` and the `job_name` from the histogram

In the future releases, it should be considered adding configurable labels. However, removing the labels mentioned above makes sense since they increase cardinality substantially. If there is a need to return them back, we can do it in the same time when configurable metrics are introduced.

Fixes #2873

### DIFF
--- a/cmd/githubrunnerscalesetlistener/metrics.go
+++ b/cmd/githubrunnerscalesetlistener/metrics.go
@@ -1,8 +1,6 @@
 package main
 
 import (
-	"strconv"
-
 	"github.com/actions/actions-runner-controller/github/actions"
 	"github.com/prometheus/client_golang/prometheus"
 )
@@ -18,8 +16,6 @@ const (
 	labelKeyJobWorkflowRef          = "job_workflow_ref"
 	labelKeyEventName               = "event_name"
 	labelKeyJobResult               = "job_result"
-	labelKeyRunnerID                = "runner_id"
-	labelKeyRunnerName              = "runner_name"
 )
 
 const githubScaleSetSubsystem = "gha"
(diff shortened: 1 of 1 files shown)


Link- (approved):

MPV:
Did this PR remove these labels from both:
- `gha_job_execution_duration_seconds`
- `gha_job_startup_duration_seconds`

We are seeing lowered cardinality on: 🙏🏻
- `gha_job_startup_duration_seconds`

But still high cardinality on: 😅
- `gha_job_execution_duration_seconds` (especially the `_bucket`)

PR #3009, opened 2023-10-20
Remove inheritance of imagePullPolicy from manager to listeners

nikola-jokic:
Implementation of announcement: https://github.com/actions/actions-runner-controller/discussions/2902


.github/workflows/arc-publish-chart.yaml
```diff
@@ -63,7 +63,7 @@ jobs:
         python-version: '3.11'
 
     - name: Set up chart-testing
-      uses: helm/chart-testing-action@v2.3.1
+      uses: helm/chart-testing-action@v2.6
```
fhammerl, line 66: Unrelated change; the bump is only required because a the dependency of this action (sigstore/cosign) has moved and cannot be downloaded with 2.3.1 anymore.

fhammerl (approved):
LGTM
Fixes Helm chart validation workflow errors too!

Issue #2445, opened 2023-03-25
Multiple label support for gha-runner-scale-set


tiithansen:
### What would you like added?

Currently it is possible to set only one label gha-runner-scale-set through `runnerScaleSetName`. It would be nice it there would be `runnerSetLabels` which would take in array of labels.

Example:
```yaml
runnerSetName: 'generic'
runnerSetLabels:
- self-hosted
- arm64
```

### Why is this needed?

In previous versions of ARC it was possible to set multiple labels and self-hosted was sort of default label. Having this feature would enable easier switch from ARC to gha-runner-scale-set because it wouldn't be necessary to modify `runs-on: [...]` in workflows which already have these labels in place.

### Additional context

It seems to possible to pass in multiple labels for the runnerScaleSet [code](https://github.com/actions/actions-runner-controller/blob/64cedff2b42790afc00f7ede5a9df897a1afb1ac/controllers/actions.github.com/autoscalingrunnerset_controller.go#L347-L352).

jrhemstad:
Agreed, this is important for our use case.

For example, I only have enough Windows test machines to do nightly windows testing. However, during the day, those machines may sit idle and could be repurposed as build machines to improve build job throughput.

Ideally, I could label both my windows build runners with both "windows" and "windows-build" labels, and my test runners as "windows" and "windows-test".

Then I could just specify `runs-on: windows` for my build jobs.

dm3ch:
@Link- Could you please point me where I can find information on why it has been decided not to allow support of labels for runnersets?

And is there any chance that it will be added in the future?

```
Labels are not supported and will not be supported for runner scale sets. Please read the docs:

https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/about-actions-runner-controller#using-arc-runners-in-a-workflow
```

shakefu:
We also have this use case where we have existing runners labeled `[self-hosted, general-ubuntu]` and we want the scale sets to be able to pick up load from our EC2-based runners using those labels. But since the scale sets are restricted to the single (unique) name as their label, it seems they cannot share load.

From my limited understanding of the code, this seems to be due to the design choice to use k8s labels/annotations on the CRDs as a way to pass the name (label) information to the controller. 

It seems like it would be possible to extend the `AutoscalingListenerSpec` to include a `runnerLabels []string` field which could then be added into the controller reconciliation?

nethershaw:
[...]

PR #3019, opened 2023-10-24
Decouple runner scale-set-name from k8s deployment name

timmjd:
# Issue
The current implementation does not allow to create multiple runners with the same __GitHub Runner Name__ from within the same __Kubernetes Namespace__. As an example, it's not possible to have multiple `ubuntu-latest` on different GitHub orgs that are all deployed within a single namespace `foo`.

# Reason
This is due to the `metadata.name` of the `AutoscalingRunnerSet` is hard wired to the `spec.runnerScaleSetName` - so two runners with the same name would cause a namespace collision inside kubernetes.

# Fix
This fix will allow to decoupled the kubernetes __deployment name__ from the GitHub __runner name__. By this it will be possible to have a `ubuntu-latest-foo` and a `ubuntu-latest-bar` deployment inside the same k8s `namespace` that would both create a runner set named `ubuntu-latest` - both in different GitHub orgs like `foo` and `bar`.

Deployment `foo`
```yaml
runnerScaleSetName: ubuntu-latest
deploymentName: ubuntu-latest-foo
```

Deployment `bar`
```yaml
runnerScaleSetName: ubuntu-latest
deploymentName: ubuntu-latest-bar
```


timmjd:
This would probable provide a workaround for the absence of #2445 & #2921 due to it will reduce the complexity of infrastructure setup.

shapirus:
> This would probable provide a workaround for the absence of #2445 & #2921 due to it will reduce the complexity of infrastructure setup.

I don't see how it would provide a workaround, can you please elaborate? Those issues are related to being unable to tell GHA that a job can be run on **any of** the runners from a list. That's what is normally achieved by using labels: e.g., you have arch-specific runners ubuntu-arm64, ubuntu-amd64 and assign a more general "ubuntu" label to them which will pick whatever of the runners having this label is available.

Issue #2707, opened 2023-06-28
gha-runner-scale-set - pre-emptively scale runners based on new "min idle runners" setting


mikeclayton:
### What would you like added?

The new scaleset runners already support a ```minRunners``` and ```maxRunners``` setting in the gha-runner-scale-set helm charts:

https://github.com/actions/actions-runner-controller/blob/d134dee14b5ecd76ecbd2b3b2e1bc79fe3f4a95d/charts/gha-runner-scale-set/values.yaml#L39-L43

but it would be great if the runners also had a setting called something like ```minIdleRunners``` as well so that the runner group would scale up pre-emptively to ensure there were idle runners already available when a new workflow run arrives in the queue.

### Why is this needed?

Sometimes, scaling up can be a slow operation and waiting to do it on-demand when a new workflow is triggered could mean the workflow is delayed by several minutes while the scale-up event is processed - for example if a new node needs to be dynamically provisioned and added to the kubernetes cluster for the runner pod to be scheduled onto.

Scaling up pre-emptively to ensure there are always a predefined number of idle runners would mean workflows could start sooner.

For example, if the *current* gha-runner-scale-set ```values.yaml``` is configured as follows:

```
maxRunners: 10
minRunners: 2
```

... then the first two workflows that arrive will be immediately assigned to the two idle runners (lets assume they both take about 10-15 minutes to complete) . There are now 2 runners but they're both busy, so if a third workflow is triggered 5 minutes later it will need to wait for another runner pod to be provisioned and *that* could also require a new node to be dynamically provisioned and added to the kubernetes cluster, which might take several minutes.

If there were a ```minIdleRunners``` setting, the scale-up event would be triggered as soon as the two initial workflows were allocated to runners and became busy, so that the scale up to create 2 more *idle* runners would have already competed by the time the third workflow is triggered and a runner would be available to start the third workflow immediately.

### Additional context

**Note:** The new setting could potentially conflict with the existing ```minRunners``` and ```maxRunners``` in some cases - e.g.

```
maxRunners: 10
minRunners: 2
minIdleRunners: 100
```
or
```
maxRunners: 10
minRunners: 2
minIdleRunners: 0
```
or if there are, say, already 9 "busy" runners there could only be one additional idle runner provisioned.

The ```minIdleRunners``` setting would be "advisory" and would yield to the ```minRunners```and ```maxRunners``` settings when a conflict arises.

samchoraria:
[...]

PR #3040, opened 2023-11-02
ADR: Changing semantics of min runners to be min idle runners

nikola-jokic:
Proposing design for having two ways of handling `minRunners` field

Addresses #2707


docs/adrs/2023-11-02-min-runners-strategy.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Decision
+
+The decision is to include an additional `minRunnerStrategy` field per `AutoscalingRunnerSet`. This field will allow the scaling strategy to be picked based on the scale set. There will be two strategies: "lazy" and "eager".
+
+With the "lazy" strategy, the current behavior of `minRunners` will be preserved. The `minRunners` field will specify the minimum number of runners running in a cluster regardless of their state ("running" or "idle").
+
+With the "eager" strategy, `minRunners` will be treated as `minIdleRunners`. This strategy will calculate the number of runners based on the number of workflows acquired plus the number of minRunners. If the `maxRunners` field is specified, it will be respected, so the number of idle runners can be less than the number of idle `minRunners` when the number of acquired jobs plus the number of min runners is greater than `maxRunners`.
```
TingluoHuang: since we don't stop acquire jobs after we reaching maxRunners, would this cause we creating too many runners?
also since 2 runner scale sets can acquire the same job, but only one of them will get run, will we cause too much wasting resource?

docs/adrs/2023-11-02-min-runners-strategy.md
```diff
@@ -0,0 +1,23 @@
+# Introducing min runner strategies
+
+**Status**: Proposed
+
+## Context
+
+Current implementation treats the `minRunners` field as the number of runners that should be running on your cluster. They can be busy running the job, or they can be idle. This ensures faster cold startup time when workflows are acquired as well as trying to use the minimum amount of runners needed to fulfill the scaling requirement.
```
TingluoHuang: i think runner pod kind of having 3 stage:
- Running jobs
- Idle, runner online to the service
- Runner pod starting, runner is trying to be online to the service.

docs/adrs/2023-11-02-min-runners-strategy.md
nikola-jokic: 1. It should not because the patch from the listener should only be up to `maxRunners` regardless of how many are acquired
2. Here we can over-provision them if other scale set "steals" the job, but eventually we will scale down on the next message. The unlucky scenario here can be that no events are created after the job is stolen, so we may not scale down to the desired number of min runners. Good point...

docs/adrs/2023-11-02-min-runners-strategy.md
nikola-jokic: Good point! I'll add it (I merged this case with the idle one as it is not busy, but I should have been more precise)

docs/adrs/2023-11-02-min-runners-strategy.md
TingluoHuang: We might also want to fix the scale down process if we start doing this since it might make the extra scale up/down even worse.

nikola-jokic:
Hey everyone,

The ADR has been changed not to introduce multiple scaling strategies, but rather to change the semantics of the field.

Link- (approved):
Let's ship it

PR #3045, opened 2023-11-04
Allow dind image to be specified in Helm charts

kohbis:
Now, `docker:dind` image can only download from Docker Hub.
There can be problems like the following: 

- Rate limit
- Download speed when geographically distant
- Ease of customization

For that reason, it is preferable for users to be able to flexibly choose a mirror registry.

```yaml
template:
  spec:
    containers:
      - name: runner
      - name: dind
        # e.g. image: public.ecr.aws/docker/library/docker:dind
        image: {{ mirror-registory }}/docker:dind
```


timmjd:
Hey @kohbis , I did exactly the same PR just some weeks ago: #2910

PR #3061, opened 2023-11-13
feat: GHA controller Helm Chart quoted labels

tijmenstor:
The GHA controller Helm Chart currently applies the labels it gets as-is. If you have labels with spaces or special characters, the objects fail to be applied with the labels as they cannot be recognized as a string. By always quoting the values of the labels, anything can be passed along as a label.

In our case, we simply need to be able to set labels on the objects with spaces in the values for monitoring purposes.

### DIFF
--- a/charts/gha-runner-scale-set-controller/tests/template_test.go
+++ b/charts/gha-runner-scale-set-controller/tests/template_test.go
@@ -404,6 +404,8 @@ func TestTemplate_ControllerDeployment_Customize(t *testing.T) {
 		SetValues: map[string]string{
 			"labels.foo":                   "bar",
 			"labels.github":                "actions",
+			"labels.team":                  "GitHub Team",
+			"labels.teamMail":              "team@github.com",
 			"replicaCount":                 "1",
 			"image.pullPolicy":             "Always",
 			"image.tag":                    "dev",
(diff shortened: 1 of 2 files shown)


tijmenstor:
Pinging @TingluoHuang as well, as you have done a lot of work on the Chart as well.

TingluoHuang:
might be worth to add a test case in https://github.com/actions/actions-runner-controller/blob/05f1a6ec92f5996971e0e0faaa8bf6251e0f6a3b/charts/gha-runner-scale-set-controller/tests/template_test.go#L404

tijmenstor:
> might be worth to add a test case in
> 
> https://github.com/actions/actions-runner-controller/blob/05f1a6ec92f5996971e0e0faaa8bf6251e0f6a3b/charts/gha-runner-scale-set-controller/tests/template_test.go#L404

Added and ran successfully.

charts/gha-runner-scale-set-controller/tests/template_test.go
```diff
@@ -404,6 +404,8 @@ func TestTemplate_ControllerDeployment_Customize(t *testing.T) {
 		SetValues: map[string]string{
 			"labels.foo":                   "bar",
 			"labels.github":                "actions",
+			"labels.team":                  "GitHub Team",
+			"labels.teammail":              "team@github.com",
```
TingluoHuang: you might want to also assert these on line 449, similar to `assert.Equal(t, "actions", deployment.Labels["github"])`

tijmenstor: Aw man, I did not read further than I should have. Sorry, added assert statements for both cases now.

nikola-jokic (approved):
LGTM, Thanks!

tijmenstor:
Cheers guys, lighting fast approval and merge <3

PR #3096, opened 2023-11-21
Refactoring listener app with configurable fallback

nikola-jokic:
Listener app has been working well, but navigating through it is can be hard sometimes.

This PR aims to split the listener functionality into multiple testable packages:
- `app`: Actual listener app responsible for wiring listener, worker and metrics server together. Reads configuration and instantiates components.
- `config`: Shared package between the controller and the listener. 
- `listener`: Package responsible for communicating with the actions service and calling worker methods when needed
- `worker`: Actual k8s worker doing patches on resources
- `metrics`: Package exposing metrics server and metrics publishers (interface used to translate actions service message to an actual metric)

This change does not include any new features. It is just a refactor of the original app. In case of an issue, users can set `LISTENER_ENTRYPOINT` to `/github-runnerscaleset-listener` in order to use older listener app.

NOTE: Even with the ability to fall back to the previous implementation, please submit issues related to the newer implementation so we can fix them and later have only one listener app.


cmd/ghalistener/worker/worker.go
```diff
@@ -0,0 +1,214 @@
...
+			return fmt.Errorf("NewLogger failed: %w", err)
+		}
+		logger = logger.WithName(workerName)
+		w.logger = &logger
+	}
+
+	return nil
+}
+
+func (w *Worker) HandleJobStarted(ctx context.Context, jobInfo *actions.JobStarted) error {
```
Link-, line 96: I think it'll be helpful to add doc comments for these functions. We can use Copilot for that

cmd/ghalistener/listener/listener.go
```diff
@@ -0,0 +1,381 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("context cancelled: %w", ctx.Err())
+		default:
+		}
+
+		msg, err := l.getMessage(ctx)
+		if err != nil {
+			return fmt.Errorf("getMessage failed: %w", err)
```
Link-: maybe we should be more consistent with the other error messages below?

Link- (changes requested):
To be continued

main.go
```diff
@@ -203,6 +203,8 @@ func main() {
 		log.Info("Using default AutoscalingListener logging parameters", "LogLevel", actionsgithubcom.DefaultScaleSetListenerLogLevel, "LogFormat", actionsgithubcom.DefaultScaleSetListenerLogFormat)
 	}
 
+	actionsgithubcom.SetListenerEntrypoint(os.Getenv("LISTENER_ENTRYPOINT"))
```
Link-, line 206: Since we're feature flagging this, let's pass which listener we are using in the UserAgent so that we can troubleshoot problems pre-emptively and more effectively

nikola-jokic: After we merge [this PR](https://github.com/actions/actions-runner-controller/pull/3116), I will update the client :relaxed:

cmd/ghalistener/config/config.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	if hasToken && hasPrivateKeyConfig {
+		return fmt.Errorf("only one GitHub auth method supported at a time. Have both PAT and App auth: token length: '%d', appId: '%d', installationId: '%d', private key length: '%d", len(c.Token), c.AppID, c.AppInstallationID, len(c.AppPrivateKey))
+	}
+
+	return nil
+}
+
+func (c *Config) Logger() (logr.Logger, error) {
```
Link-, line 84: At some point in the near future, we'd want to adopt OTel standards for logging. This could be our next evolution from here.

cmd/ghalistener/config/config.go
```diff
@@ -0,0 +1,147 @@
...
+			return nil, fmt.Errorf("failed to load system cert pool: %w", err)
+		}
+		pool := systemPool.Clone()
+		ok := pool.AppendCertsFromPEM([]byte(c.ServerRootCA))
+		if !ok {
+			return nil, fmt.Errorf("failed to parse root certificate")
+		}
+
+		options = append(options, actions.WithRootCAs(pool))
+	}
```
Link-, line 133: Did you manage to test this? I remember you had some worries about it? If not, I'll run some tests

nikola-jokic: e2e test is covering this use-case. I can't test it without using a client, but that is already covered in the e2e :relaxed:

main.go
Link-: I approved the other PR, let's also log which listener we're using

Dockerfile
```diff
@@ -38,6 +38,7 @@ RUN --mount=target=. \
   export GOOS=${TARGETOS} GOARCH=${TARGETARCH} GOARM=${TARGETVARIANT#v} && \
   go build -trimpath -ldflags="-s -w -X 'github.com/actions/actions-runner-controller/build.Version=${VERSION}' -X 'github.com/actions/actions-runner-controller/build.CommitSHA=${COMMIT_SHA}'" -o /out/manager main.go && \
   go build -trimpath -ldflags="-s -w -X 'github.com/actions/actions-runner-controller/build.Version=${VERSION}' -X 'github.com/actions/actions-runner-controller/build.CommitSHA=${COMMIT_SHA}'" -o /out/github-runnerscaleset-listener ./cmd/githubrunnerscalesetlistener && \
+  go build -trimpath -ldflags="-s -w -X 'github.com/actions/actions-runner-controller/build.Version=${VERSION}' -X 'github.com/actions/actions-runner-controller/build.CommitSHA=${COMMIT_SHA}'" -o /out/ghalistener ./cmd/ghalistener && \
```
Link-, line 41: Aren't we locking the use of the new listener this way?

nikola-jokic: I am not sure that I understand, this is just a binary build. If you look 1 line up, we are also building `githubrunnerscalesetlistener` so both of them are included in the binary

Link-: Ah yes of course :facepalm:

Link- (approved):
Awesome work @nikola-jokic <3 let's get this rolling! 🚀

PR #3098, opened 2023-11-22
Propagate runner scale set name annotation to EphemeralRunner

ordovicia:
With this PR, AutoscalingRunnerSets' annotation that describes its runner scale set name will be propagated to EphemeralRunners.
This change would make it easy to inspect EphemeralRunners for troubleshooting.

This PR changes `runner-scale-set-name` annotation key to `actions.github.com/runner-scale-set-name` to align it with `actions.github.com/runner-group-name`.
The annotation is used only inside the controllers, but I am not sure if this change is acceptable.

### DIFF
--- a/controllers/actions.github.com/autoscalingrunnerset_controller.go
+++ b/controllers/actions.github.com/autoscalingrunnerset_controller.go
@@ -45,7 +45,6 @@ const (
 	labelKeyRunnerSpecHash            = "runner-spec-hash"
 	autoscalingRunnerSetFinalizerName = "autoscalingrunnerset.actions.github.com/finalizer"
 	runnerScaleSetIdAnnotationKey     = "runner-scale-set-id"
-	runnerScaleSetNameAnnotationKey   = "runner-scale-set-name"
 )
 
 type UpdateStrategy string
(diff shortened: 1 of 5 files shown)


nikola-jokic:
Hey @ordovicia,

Can you please help me understand what problem does this PR solve? We already have a label that specifies the scale set name.

ordovicia:
Yes, EphemeralRunner resources already have `actions.github.com/scale-set-name` label.
But its value is the name of the parent AutoscalingRunnerSet resource's name, not a GitHub runner scale set's name.

https://github.com/actions/actions-runner-controller/blob/7793e1974a0fc6c4dd346b2492ab82a1531593c3/controllers/actions.github.com/resourcebuilder.go#L494

Runner scale set name can be different from AutoscalingRunnerSet name when the user specifies `spec.runnerScaleSetName` field of the AutoscalingRunnerSet.

https://github.com/actions/actions-runner-controller/blob/7793e1974a0fc6c4dd346b2492ab82a1531593c3/controllers/actions.github.com/autoscalingrunnerset_controller.go#L445-L446

So, to resolve the runner scale set name for an EphemeralRunner, users need to first get its parent AutoscalingRunnerSet, and then read the AutoscalingRunnerSet's `spec.runnerScaleSetName` field.
This PR allows users to obtain the runner scale set name directly from an EphemeralRunner.

ordovicia:
Hi @nikola-jokic could you please take a look at my comment above? Thank you!

nikola-jokic:
Hey @ordovicia,

Sorry for the delay and thank you for the explanation. I don't think adding an annotation is the right approach here. If the purpose of the field is to query it, it should be a label. Also, we use annotations mostly to include some metadata, so they should be completely scoped to the cluster resources. 
Now, what I would propose is to install two more labels, like `actions.github.com/runner-scale-set-name` and `actions.github.com/runner-scale-set-group-name` and include them only in a resource builder. Let me reach out to the team for a second opinion and I will get back to you :relaxed:.

ordovicia:
Thank you @nikola-jokic for the reply!
I agree that using labels is better for this purpose.
I look forward to hearing the team’s opinion.

nikola-jokic (approved):
LGTM, Thank you and sorry for the delay!

Issue #3082, opened 2023-11-16
Upgrading to 0.7.0 from 0.6.1 broke listeners and couldn't revert back to 0.6.1


mattpopa:
### Checks

- [X] I've already read https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/troubleshooting-actions-runner-controller-errors and I'm sure my issue is not covered in the troubleshooting guide.
- [X] I am using charts that are officially provided

### Controller Version

0.7.0

### Deployment Method

Helm

### Checks

- [X] This isn't a question or user support case (For Q&A and community support, go to [Discussions](https://github.com/actions/actions-runner-controller/discussions)).
- [X] I've read the [Changelog](https://github.com/actions/actions-runner-controller/blob/master/docs/gha-runner-scale-set-controller/README.md#changelog) before submitting this issue and I'm sure it's not due to any recently-introduced backward-incompatible changes

### To Reproduce

```markdown
1. have 0.6.1 controller and scale-set running, auth with gh app
2. update version to 0.7.0 to controller and scale-set
3. listeners do not come up anymore, no errors in the controller
4. rolled back doesn't help, now listeners don't start on 0.6.1

managed to restore by
5. delete cdrs
6. delete/redeploy controller on 0.6.1
7. delete/redeploy sets on 0.6.1
```

### Describe the bug

Did the update to 0.7.0 mainly for the cardinality fix, and 0.7.0 broke the scale-sets, listeners gone, no error logs, not much to investigate/figure out why this is happening. Using github app for auth.

The most annoying part was that restore on 0.6.1 didn't help, and I had to remove the CRDs and reinstall the controller + sets one by one.

### Describe the expected behavior

1. Update to 0.7.0 from 0.6.1
2. Controller starts up successfully
3. Listeners start up and connect to gh using the same gh app config
4. business as usual

### Additional Context

```yaml
This is a dind setup, but the problem is related to CRDs and listeners. Deployment's via helm chart

Using terraform with helm provider to deploy

# controller vars
image:
  repository: "ghcr.io/actions/gha-runner-scale-set-controller"
  pullPolicy: IfNotPresent
  tag: ${arc_version}

resources:
  limits:
    cpu: 100m
    memory: 256Mi
  requests:
    cpu: 100m
    memory: 256Mi

metrics:
  controllerManagerAddr: ":8080"
  listenerAddr: ":8080"
  listenerEndpoint: "/metrics"

[...]

PR #3106, opened 2023-11-27
Remove deletion of Autoscaling Runner Set when versions dont match

larhauga:
Deleting the autoscalingrunnerset from the controller results in errors during upgrade procedures. The controller should enable the runnerset to live with a version skew, and leave the upgrade procedure to other tooling like flux, argo or helm.

This is one way of implementing https://github.com/actions/actions-runner-controller/discussions/3090 which aims to discuss issue #3082. 

I think that it should be possible to update the controller ahead of upgrading the scalesets, and this is a quick fix to mitigate the current behavior.
And an error in annotation of the version number should not result in the deletion of the scalesets.
I think that there could be more work to be done to ensure that the scale set controller adheres to the design of kubebuilder and controller manager. However, I have been unable to find the design of the choises done in this controller at the moment, and as such there could be technical reasons for this implementation. Can anyone elaborate on this?

Thank you

### DIFF
--- a/controllers/actions.github.com/autoscalingrunnerset_controller.go
+++ b/controllers/actions.github.com/autoscalingrunnerset_controller.go
@@ -156,19 +156,10 @@ func (r *AutoscalingRunnerSetReconciler) Reconcile(ctx context.Context, req ctrl
 	}
 
 	if autoscalingRunnerSet.Labels[LabelKeyKubernetesVersion] != build.Version {
-		if err := r.Delete(ctx, autoscalingRunnerSet); err != nil {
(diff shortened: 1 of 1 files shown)


nikola-jokic:
Hey @larhauga,

The reason we have this functionality is because CRDs are not stable. To conform with the kubernetes way, every CRD update would require the controller to correctly reconcile all stable versions. Having that in mind, our API is v1alpha1 to reflect that the API may change and we cannot promise backward compatibility. This mechanism ensures that the controller reconciles the tested version for that particular release.

PR #3116, opened 2023-11-30
Modify user agent format with subsystem and is proxy configured information

nikola-jokic:
Reformat the user agent string and include the component sending the data.

Include information specifying if the proxy is configured


Link- (approved):

Link- (approved):
This is good, let's merge it

PR #3139, opened 2023-12-08
Change minRunners behavior and fix the new listener min runners

### LINKED ISSUE #2707: gha-runner-scale-set - pre-emptively scale runners based on new "min idle runners" setting

nikola-jokic:
This PR implements the [proposed ADR](https://github.com/actions/actions-runner-controller/blob/master/docs/adrs/2023-11-02-min-runners-strategy.md) and changes its status.

Target runner count is now calculated as `min(assigned + minRunners, maxRunners)` where maxRunners is set to the maximum number that can fit in int32 if not configured.

Also, this PR fixes a bug in the new listener that is not yet published, introduced by #3096

Fixes #2707

### DIFF
--- a/.github/workflows/gha-e2e-tests.yaml
+++ b/.github/workflows/gha-e2e-tests.yaml
@@ -880,3 +880,98 @@ jobs:
           helm uninstall "${{ steps.install_arc.outputs.ARC_NAME }}" --namespace "arc-runners" --debug
           kubectl wait --timeout=10s --for=delete AutoScalingRunnerSet -n "${{ steps.install_arc.outputs.ARC_NAME }}" -l app.kubernetes.io/instance="${{ steps.install_arc.outputs.ARC_NAME }}"
           kubectl logs deployment/arc-gha-rs-controller -n "arc-systems"
+
+  init-with-min-runners:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.id == github.repository_id
+    env:
+      WORKFLOW_FILE: arc-test-workflow.yaml
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.head_ref }}
+
+      - uses: ./.github/actions/setup-arc-e2e
+        id: setup
+        with:
+          app-id: ${{secrets.E2E_TESTS_ACCESS_APP_ID}}
+          app-pk: ${{secrets.E2E_TESTS_ACCESS_PK}}
+          image-name: ${{env.IMAGE_NAME}}
+          image-tag: ${{env.IMAGE_VERSION}}
+          target-org: ${{env.TARGET_ORG}}
+
+      - name: Install gha-runner-scale-set-controller
+        id: install_arc_controller
+        run: |
+          helm install arc \
+          --namespace "arc-systems" \
+          --create-namespace \
+          --set image.repository=${{ env.IMAGE_NAME }} \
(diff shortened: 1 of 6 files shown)


.github/workflows/gha-e2e-tests.yaml
```diff
@@ -880,3 +880,98 @@ jobs:
...
+            if [ "$count" -ge 60 ]; then
+              echo "Timeout waiting for listener pod with label actions.github.com/scale-set-name=$ARC_NAME"
+              exit 1
+            fi
+            sleep 1
+            count=$((count+1))
+          done
+          kubectl wait --timeout=30s --for=condition=ready pod -n arc-systems -l actions.github.com/scale-set-name=$ARC_NAME
+          kubectl get pod -n arc-systems
+      - name: Ensure 5 runners are up
```
Link-, line 962: I think this test is not sufficient, what I think the test should do is also to trigger one or more workflow runs and make sure we have 5 runners all the time

Link- (approved):
I don't want to delay the release for an e2e test. I'm fine with this moving forward as long as we address the e2e right after

remover:
@nikola-jokic will this also apply to legacy ARC? i'm guessing not but would it be easy enough to apply there also?

Link-:
@remover this does not apply to legacy ARC. Legacy ARC is maintained by the community led by @mumoshu and we do not backport changes.

remover:
@mumoshu do you think this would be an easy enough change in legacy ARC? happy to help if so... :)

PR #3146, opened 2023-12-13
Fix assertion test in wait for delete

nikola-jokic:
Fixing typo in arc e2e assertion where we check for the event in a namespace that is scale sets name, not scale sets namespace.


Link- (approved):

Link- (changes requested):
I'll re-approve when the tests pass

Link- (approved):

PR #3156, opened 2023-12-14
add available and acquire jobs metrics

### LINKED ISSUE #3154: Metrics from acquired and available missing
### Checks

### Controller Version

0.7.0

### Deployment Method

Helm

### Checks

### To Reproduce

```markdown
1. enable listener metrics
2. try to check acquired or available metrics from the metrics path
```

### Describe the bug

https://github.com/actions/actions-runner-controller/blob/master/docs/adrs/2023-05-08-exposing-metrics.md?plain=1#L175 says that there should be `available_jobs` & `acquired_jobs` metrics exposed. However, for some reason these are disabled in code? https://github.com/actions/actions-runner-controller/blob/3e4201ac5f6c6d172a19b580154eaf5abf24a2ca/cmd/githubrunnerscalesetlistener/metrics.go#L54-L71

expected behaviour is that I could see values of these metrics in my prometheus.

Now I can see these only in pod logs, which makes it pretty difficult to make alerts based on metrics.

### Additional Context

```yaml
replicaCount: 2

flags:
  logLevel: "debug"

metrics:
  listenerAddr: ":8080"
  listenerEndpoint: "/metrics"
```

### Controller Logs

```shell
not relevant
```

### Runner Pod Logs

```shell
not relevant
```

zetaab:
fixes #3154

cc @nikola-jokic do you have idea why these two metrics were disabled? I added these back, compiled new version and trying it now and at least for me this works fine. Just wondering is there some another case that it does not work for instance?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -214,7 +214,7 @@ The process would look like the below:
 
 - Clone this repository locally
 - Checkout the branch. If you use the `gh` command, run `gh pr checkout $PR_NUMBER`
-- Run `NAME=$DOCKER_USER/actions-runner-controller VERSION=canary make docker-build docker-push` for a custom container image build
+- Run `NAME=$DOCKER_USER/actions-runner-controller VERSION=canary make docker-buildx docker-push` for a custom container image build
 - Update your actions-runner-controller's controller-manager deployment to use the new image, `$DOCKER_USER/actions-runner-controller:canary`
 
 Please also note that you need to replace `$DOCKER_USER` with your own DockerHub account name.
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM --platform=$BUILDPLATFORM golang:1.21.3 as builder
+FROM --platform=$BUILDPLATFORM golang:1.21.5 as builder
 
 WORKDIR /workspace
 
--- a/cmd/githubrunnerscalesetlistener/main.go
+++ b/cmd/githubrunnerscalesetlistener/main.go
@@ -118,8 +118,8 @@ func (s *metricsServer) shutdown() error {
 func (s *metricsServer) listenAndServe() error {
 	reg := prometheus.NewRegistry()
 	reg.MustRegister(
-		// availableJobs,
-		// acquiredJobs,
+		availableJobs,
+		acquiredJobs,
 		assignedJobs,
 		runningJobs,
 		registeredRunners,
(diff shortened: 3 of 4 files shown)


nikola-jokic:
Hey @zetaab,

You are absolutely right, they do work fine, but we are encountering an issue in a situation when scale set becomes idle. We are missing the event that sets acquired and available jobs back to 0 when there are no jobs coming in. :relaxed:

zetaab:
@nikola-jokic would it still make sense to merge this with known issue in it? We are currently facing issue that listeners for some reason are not making the progress without restarting of whole listener. There is no way to get alert about that if there are no metrics available (now we just get messages from developers that "is there some issues because my build is not starting"). So I would like to make prometheus rule to `count(available_jobs) > 10` longer than n minutes.

Example (in middle of the screen I restarted the listener, so for some reason it stopped fetching the jobs from github side?):

image

And after the listener were restarted the `available_jobs` count went quite fast back to 0 and jobs got assigned

btw. There is Github support ticket open about this. I wanted to contribute and add these metrics to get possibility to add alerts.

MPV:
Anything that can be done to help out here? I can to begin with chime in and say this does seem useful. 🙏

nikola-jokic:
Hey @zetaab, @MPV,

Quick update: we are reevaluating and investigating overall need for these calls, and we may remove these calls in the future. We would prefer to keep these metrics out, since they are likely to be removed. Not having to acquire jobs may increase scaling performance. Let's keep this PR open, and I will update you on the progress.

zetaab:
@nikola-jokic at least we are interested of queue times. We have SLOs to have 99% percentile of queue times less than 5 minutes. Currently we are doing this with own rest api, which receives webhook from github. Github does not provide this statistics currently.

MPV:
@nikola-jokic thanks for the update. ❤️ 

In related news, this turned out to be quite promising and helpful:
https://github.blog/changelog/2024-03-28-actions-usage-metrics-public-beta/

Only piece missing now is the queue/waiting times (so we can help developers get quicker CI/CD). 🙏 

In terms of [working with Service-Level Objectives for ARC](https://github.com/actions/actions-runner-controller/discussions/3177), here are some rough ideas we've thought one could implement if/when this becomes a reality:

>In an effort to validate that ARC delivers on its promise to handle its jobs:
>- 99.5% of all "available" jobs should also be "aquired"
>- 99.5% of all "available" jobs should also be "started"
>- 99.5% of all "available" jobs should also be "completed"
>
>The above could likely also include a timeframe on when this should have happened.

PR #3157, opened 2023-12-15
propogate arbitrary labels from runnersets to all created resources

Lazyshot:
Within our organization we have policies in place that require all resources to have specific labels for security and cost attribution purposes (e.g. team, business unit, etc.).

In order to support such a requirement, this change would have any arbitrary labels that are set at the 
AutoscalingRunnerSet level also be assigned to any subsequent resources the controller will create.

Similar behavior could be valuable for annotations as well, but I personally do not have any need for it.

### DIFF
--- a/controllers/actions.github.com/resourcebuilder.go
+++ b/controllers/actions.github.com/resourcebuilder.go
@@ -85,13 +85,13 @@ func (b *resourceBuilder) newAutoScalingListener(autoscalingRunnerSet *v1alpha1.
 		effectiveMinRunners = *autoscalingRunnerSet.Spec.MinRunners
 	}
 
-	labels := map[string]string{
+	labels := mergeLabels(autoscalingRunnerSet.Labels, map[string]string{
 		LabelKeyGitHubScaleSetNamespace: autoscalingRunnerSet.Namespace,
 		LabelKeyGitHubScaleSetName:      autoscalingRunnerSet.Name,
 		LabelKeyKubernetesPartOf:        labelValueKubernetesPartOf,
 		LabelKeyKubernetesComponent:     "runner-scale-set-listener",
 		LabelKeyKubernetesVersion:       autoscalingRunnerSet.Labels[LabelKeyKubernetesVersion],
-	}
+	})
 
 	annotations := map[string]string{
 		annotationKeyRunnerSpecHash: autoscalingRunnerSet.ListenerSpecHash(),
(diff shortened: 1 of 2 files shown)


MartinLesterSynamedia:
This is a really important feature for us.
We have many VMs added as runners right now and we want to migrate them to use this system.
There are 100s of jobs that reference our enterprise runners using the same tag so we want this tag on all available runners.
Further we want to be able to create variations e.g. `Enterprise, small` and `Enterprise, large`
Being restricted to 1 label, that is also the name of the scale set seems really limiting.

Please approve this.

controllers/actions.github.com/resourcebuilder.go
```diff
@@ -741,3 +737,15 @@ func trimLabelValue(val string) string {
 	}
 	return val
 }
+
+func mergeLabels(labels ...map[string]string) map[string]string {
```
nikola-jokic: Can we do something like:
```
func mergeLabels(base, overwrites map[string]string) map[string]string
```
It would be much easier to understand by reading the function definition.

nikola-jokic (changes requested):
Added a suggestion. Please fix merge conflict :relaxed:

Lazyshot:
> This is a really important feature for us. We have many VMs added as runners right now and we want to migrate them to use this system. There are 100s of jobs that reference our enterprise runners using the same tag so we want this tag on all available runners. Further we want to be able to create variations e.g. `Enterprise, small` and `Enterprise, large` Being restricted to 1 label, that is also the name of the scale set seems really limiting.
> 
> Please approve this.

To be clear this is **not** intended to touch runner labels, but instead only the kubernetes resource labels.

MartinLesterSynamedia:
> > This is a really important feature for us. We have many VMs added as runners ...
>
> To be clear this is **not** intended to touch runner labels, but instead only the kubernetes resource labels.

Dang. Am I missing something?
Currently our runners have `runs-on` lines like `[ Enterprise, ubuntu, docker, docker-compose ]` so their job gets allocated to a VM with the correct capabilities. As far as I can tell there is no way to get an ephemeral runner to pick up a job with a `runs-on` with more than 1 label. The best I could do is create a new scale set for each possible label set e.g. Enterprise-Ubuntu, Enterprise-Ubuntu-Docker, etc. I couldn't even find a way to add labels to a scale set or it's runners in the UI.

Sadly this requires all our engineers update their jobs to be able to use the new ephemeral runners and I wont be able silently replace our VMs with ephemeral

nikola-jokic (approved):
LGTM

PR #3161, opened 2023-12-17
Fix override listener pod spec (#3139)

sergelogvinov:
Hello, this small fix.

Thank you @nikola-jokic for you changes. I've checked it, and got small error.
So it will fix it.

### DIFF
--- a/controllers/actions.github.com/resourcebuilder.go
+++ b/controllers/actions.github.com/resourcebuilder.go
@@ -309,7 +309,7 @@ func mergeListenerPodWithTemplate(pod *corev1.Pod, tmpl *corev1.PodTemplateSpec)
 		c := &tmpl.Spec.Containers[i]
 
 		switch c.Name {
-		case "listener":
+		case autoscalingListenerContainerName:
 			mergeListenerContainer(listenerContainer, c)
 		default:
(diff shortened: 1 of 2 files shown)


controllers/actions.github.com/resourcebuilder.go
```diff
@@ -311,6 +311,8 @@ func mergeListenerPodWithTemplate(pod *corev1.Pod, tmpl *corev1.PodTemplateSpec)
 		switch c.Name {
 		case "listener":
```
nikola-jokic: Great catch! Can you remove this case because I mistakenly wrote that the container name for the listener is the `listener`? We don't really need this case anymore

nikola-jokic (changes requested):
Can you please do the following:
1. Modify the case to replace `listener` with `autoscalingListenerContainerName`
2. Fix [this](https://github.com/actions/actions-runner-controller/blob/3e4201ac5f6c6d172a19b580154eaf5abf24a2ca/controllers/actions.github.com/autoscalinglistener_controller_test.go#L857C6-L857C6) test also to replace `"listener"` with the appropriate variable

sergelogvinov:
@nikola-jokic Done 👍

nikola-jokic (approved):
LGTM, Thank you!

PR #3195, opened 2023-12-28
Add annotation with values hash to re-create listener

### LINKED ISSUE #3182: Runner Scale Set listener-config secret is static
### Checks

### Controller Version

0.7.0

### Deployment Method

Helm

### Checks

### To Reproduce

```markdown
1. Update the data in the `githubConfigSecret`
2. Force reconciliation (restart controller pod)
```

### Describe the bug

After attempting to rotate the credentials used by the runner scale sets it was observed the data in the secret: `*-listener-config` is not automatically updated like the base secret: `*-listener`.   This will cause issues if the old credentials are no longer valid.  

Current work around is to manually delete the `*-listener-config` and force reconciliation

After the 'base' secret  `*-listener`  is updated the `*-listener-config` secret should also be updated.

### Additional Context

```yaml
N/A - it is worth noting that the static nature of this config secret also cause issues if the runner groups are updated or recreated.  If the associated runner group's ID is modified but the name is preserved you will also run into issues.  

This issue becomes quite painful if the old credentials are no longer valid, all listeners would then start throwing 401s.
```

### Controller Logs

```shell
https://gist.github.com/jb-2020/bc063bcdead1e72ce44d4a0f0ceab32e
```

### Runner Pod Logs

```shell
N/A
```

nikola-jokic:
Using spec hash of the autoscaling runner set is not sufficient to indicate re-creation of the listener. As it is described in issue #3182, changing values.yaml to update secret does not re-create the listener.

Since we have avoided using cluster role to watch secrets, controller does not always have permission to list secrets (for more information, you can [read this ADR](https://github.com/actions/actions-runner-controller/blob/master/docs/adrs/2023-04-11-limit-manager-role-permission.md)).

Since all upgrades to the autoscaling runner set, initiated from helm, should re-create the listener, we can use the values json hash to create a fingerprint of the applied configuration. Instead of autoscaling runner set's spec hash, we can match the label against the installation to figure out if the listener should be re-created.

Fixes #3182


apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
```diff
@@ -250,12 +250,6 @@ type AutoscalingRunnerSetStatus struct {
 	FailedEphemeralRunners int `json:"failedEphemeralRunners"`
 }
 
-func (ars *AutoscalingRunnerSet) ListenerSpecHash() string {
```
nikola-jokic: Note: This does not take into account live patches to the listener itself. If we need to calculate the hash spec live as well, I propose using shorter hash of the values.yaml spec and later add the suffix to the annotation based on the listener spec hash

rentziass (dismissed):

rentziass (approved):

PR #3199, opened 2024-01-03
Add documentation warning and links to new docs

kenmuse:
The existing documentation (`/docs`) is for the legacy releases of ARC. These documents are still used to support older installations. These docs frequently show up in Google search results, which can be confusing for anyone trying to troubleshoot or install the newer version. It can also lead to users inadvertently configuring the wrong version of ARC.

These changes add a warning to the existing docs to indicate that they are intended for the legacy version of ARC. The warning also includes a link to the discussion announcing the new autoscaling runner scale sets version (to educate readers on the two versions) and a link to the Quickstart Docs for the newer version of ARC.

### DIFF
--- a/docs/about-arc.md
+++ b/docs/about-arc.md
@@ -1,7 +1,7 @@
 # About ARC
 
 > [!WARNING]
-> This documentation covers the legacy mode of ARC (resources in the `actions.summerwind.net` namespace). If you're looking for documentation on the newer [autoscaling runner scale sets](https://github.com/actions/actions-runner-controller/discussions/2775), it is available in [GitHub Docs](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/quickstart-for-actions-runner-controller).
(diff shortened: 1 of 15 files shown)


docs/about-arc.md
```diff
@@ -1,5 +1,8 @@
 # About ARC
 
+> [!WARNING]
+> This documentation covers the legacy mode of ARC (resources in the `actions.summerwind.net` namespace). If you're looking for documentation on the newer [autoscaling runner scale sets](https://github.com/actions/actions-runner-controller/discussions/2775), it is available in [GitHub Docs](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/quickstart-for-actions-runner-controller).
```
Link-: Let's reference this discussion to provide context on why these are legacy modes https://github.com/actions/actions-runner-controller/discussions/2775

Link- (changes requested):
Thank you for putting this together @kenmuse it'll help reduce the confusion around the different modes for sure

docs/about-arc.md
kenmuse: @Link- - are you wanting me to update the phrasing to be more explicit about the modes? Perhaps something like this:

```suggestion
> This documentation covers the legacy mode of ARC (resources in the `actions.summerwind.net` namespace). If you're looking for documentation on the newer autoscaling runner scale sets, it is available in [GitHub Docs](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/quickstart-for-actions-runner-controller). To understand why these resources are considered legacy (and the benefits of using the newer autoscaling runner scale sets), read [this discussion (#2775)](https://github.com/actions/actions-runner-controller/discussions/2775).
```

Link-: LGTM, let's go with it! @nebuk89 any thoughts?

kenmuse: I've updated the warning in the docs to match this thread.

Link- (approved):
Thank you @kenmuse let's ship this! 🚢

PR #3212, opened 2024-01-09
fix: gha controller needs further permissions

brett-au:
# Description

The current RBAC permissions cause several errors in the controller

https://github.com/actions/actions-runner-controller/discussions/3160
https://github.com/actions/actions-runner-controller/discussions/3197

These permissions allow my actions to run properly

# Testing

 Currently using this in our environment with kustomize to override the Cluster Role

### DIFF
--- a/charts/gha-runner-scale-set-controller/templates/manager_cluster_role.yaml
+++ b/charts/gha-runner-scale-set-controller/templates/manager_cluster_role.yaml
@@ -109,36 +109,65 @@ rules:
   resources:
   - ephemeralrunners/status
   verbs:
+  - create
+  - delete
   - get
+  - list
+  - watch
   - patch
-  - update
 - apiGroups:
   - ""
   resources:
   - pods
   verbs:
+  - create
+  - delete
+  - get
   - list
   - watch
+  - patch
 - apiGroups:
   - ""
   resources:
   - serviceaccounts
   verbs:
+  - create
+  - delete
+  - get
   - list
   - watch
+  - patch
 - apiGroups:
   - rbac.authorization.k8s.io
   resources:
   - rolebindings
   verbs:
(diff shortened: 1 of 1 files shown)


estolietnii (approved):

gr8jen:
I guess I have the same problem. please look at the solution so that I can go further without hacking/overriding

eugene100 (approved):

nikola-jokic:
Hey @brett-au,

Sorry, but these permissions are exactly the reason why we created this [ADR](https://github.com/actions/actions-runner-controller/blob/master/docs/adrs/2023-04-11-limit-manager-role-permission.md). The permissions you added are applied during scale set installation, so we can avoid having an overly permissive cluster role.

You can still use your own chart, but we won't be able to support it. I will close this PR since it reverts the security hardening that we are actually trying to have.

gr8jen:
Found the issue, the name of the serviceaccount wat not what I expected. It works without the modification :)

nikola-jokic:
I'm glad you resolved it!

brett-au:
> Hey @brett-au,
> 
> Sorry, but these permissions are exactly the reason why we created this [ADR](https://github.com/actions/actions-runner-controller/blob/master/docs/adrs/2023-04-11-limit-manager-role-permission.md). The permissions you added are applied during scale set installation, so we can avoid having an overly permissive cluster role.
> 
> You can still use your own chart, but we won't be able to support it. I will close this PR since it reverts the security hardening that we are actually trying to have.

Thank you. This document helps me understand the reasoning much better!

noamgreen:
@nikola-jokic HI , the change you made in ADR as you share a link have some issues as i add in the #3672 

main issues is to give runners access to AWS resources

PR #3216, opened 2024-01-10
feat(chart): make dind container template configurable

norman-zon:
In addition to the pre-populated values for the dind container template parse spec from Values.template.spec.containers["dind"] to enable users to add config

### DIFF
--- a/charts/gha-runner-scale-set/templates/_helpers.tpl
+++ b/charts/gha-runner-scale-set/templates/_helpers.tpl
@@ -96,6 +96,15 @@ volumeMounts:
 {{- end }}
 
 {{- define "gha-runner-scale-set.dind-container" -}}
+{{- range $i, $container := .Values.template.spec.containers }}
+  {{- if eq $container.name "dind" }}
+    {{- range $key, $val := $container }}
+      {{- if and (ne $key "env") (ne $key "volumeMounts") (ne $key "name") (ne $key "image") (ne $key "command") (ne $key "args") }}
+{{ $key }}: {{ $val | toYaml | nindent 2 }}
+      {{- end }}
+    {{- end }}
+  {{- end }}
+{{- end }}
 image: docker:dind
 args:
   - dockerd
--- a/charts/gha-runner-scale-set/tests/values_dind_merge_spec.yaml
+++ b/charts/gha-runner-scale-set/tests/values_dind_merge_spec.yaml
@@ -22,6 +22,17 @@ template:
         limits:
           memory: "64Mi"
           cpu: "250m"
+        requests:
+          cpu: "32Mi"
+          memory: "100Mi"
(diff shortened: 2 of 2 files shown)


norman-zon:
@nikola-jokic, can we get this merged? It's only a small helmchart change, but would make life much easier for us.

norman-zon:
I also would like to specify additional `args` to each container. Would it be preferred to make `args` completely overridable, or  rather have special values for each separate container like `{dind|runner|init}ContainerExtraArgs`, that are appended to the default `args`?
Just give me opinions and I'll add it.

nikola-jokic:
Hey @norman-zon,

We decided not to allow customizations in the default template. The reason behind this decision is that customization should either work for all fields, or no fields. So at this time, we decided to proceed with the following:
1. containerMode applies the default, out-of-box specification only
2. If any customization is needed, we documented the expanded spec, so you can uncomment the spec that would be used if containerMode is applied, and customize it however you'd like

norman-zon:
Hi @nikola-jokic, 
thanks for explaining the design decision.

First of all I overlooked the [line](https://github.com/actions/actions-runner-controller/blob/9e191cdd21621f4e43023e0bdbbd2ff9b139c8a6/charts/gha-runner-scale-set/values.yaml#L76)

> If any customization is required for dind or kubernetes mode, containerMode should remain empty, and configuration should be applied to the template

and thought it would not be possible to override the template completely. 

Regarding the everything-or-nothing policy, this does not seem to be consistent, because you already [allow setting certain customisations](https://github.com/actions/actions-runner-controller/blob/9e191cdd21621f4e43023e0bdbbd2ff9b139c8a6/charts/gha-runner-scale-set/templates/_helpers.tpl#L281) for the `runner` container.

Setting the template as a whole might be the straight-forward approach to customisation, but requires me to check the helmrelease for changed default templates on every update, to avoid breaking things, or missing important changes (like the recent change of the docker socket for example). Also, isn't partial templating the whole point of using helm?

I would suggest to switch to using `extra...` values to customise the default template like I've seen in many helmcharts and what seem to be canon in the helm world.


So we could end up with something like:
```yaml
runnerPod:
    extraVolumes: []
    runnerContainer:
        extraArgs: []
        extraEnv: []
        exraVolumeMounts: []
        ...
    dindContainer:
        extraArgs: []
    ...
...
```

I'd be happy to write a PR for that. What do you think?

AmitBenAmi:
Is there a way to specify a different image instead of being hardcoded?
`docker:dind` is coming from Dockerhub, for which we do not have the token for rate limiting purposes, but we do use the public ECR one.
However, given the current template, it is impossible to override the `image` from the values, so I need to unset `containermode`, and basically copy/paste the expected outcome as a configuration in the values

PR #3224, opened 2024-01-15
feat(helm): Include value for annotations in controller (deployment)

jcortejoso:
Hi! Small PR to include as values custom annotations for the controller deployment.

Rationale for this is that our github credentials are automatically rotated. ARC controller secrets are read when it starts, and it cannot reload the credentials if they are updated (causing error during runners registration). We want to add an annotation to the controller deployment to use [Reloader](https://github.com/stakater/Reloader) together with [external-secrets](https://github.com/external-secrets/external-secrets), so the controller is restarted every time the secret is rotated.


drew-viles:
This has been a long time needed. Any chance of this being rebased and merged any time soon? I don't want to have to fork for such a simple change if I can avoid it.

PR #3236, opened 2024-01-20
Revert docker socket path to `/var/run/docker.sock`

Link-:
This pull request primarily includes changes to the `charts/gha-runner-scale-set/templates/_helpers.tpl` file, where the Docker socket's path has been updated from `/run/docker/docker.sock` to `/var/run/docker.sock`. 

The default path for the docker daemon's unix socket is `/var/run/docker.sock` as per [this reference](https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-socket-option). There's no reason to use any other path.

### DIFF
--- a/charts/gha-runner-scale-set/templates/_helpers.tpl
+++ b/charts/gha-runner-scale-set/templates/_helpers.tpl
@@ -99,7 +99,7 @@ volumeMounts:
 image: docker:dind
 args:
   - dockerd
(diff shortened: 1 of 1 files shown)


Link-:
Useless PR, the `docker:dind` image has a symlink from `/var/run` to `/run`:

```
lrwxrwxrwx    1 root     root             4 Dec  7 09:43 run -> /run
```

PR #3251, opened 2024-01-27
Update _helpers.tpl

thedjpetersen:
There are cases where we need to extend the docker daemon and pass in additional arguments this provides a way to do so through the values.yaml file. My specific use case is having a way to set an insecure registry.


nikola-jokic:
Hey @thedjpetersen,

[We documented customizing container modes](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/deploying-runner-scale-sets-with-actions-runner-controller#customizing-container-modes). For now, we decided not to allow customizations on particular configuration. This PR allows customization on a single property, but that would mean we should allow it for all properties.

If you want to customize the `dind` mode, you can simply leave container mode object commented out, uncomment spec that we would expand and modify it to fit your case.

PR #3260, opened 2024-02-02
Expose volumeMounts and volumes in gha-runner-scale-set-controller

taliastocks:
My team needs to be able to mount our cert chain into these containers so they can trust our GHE deployment certs.


charts/gha-runner-scale-set-controller/templates/deployment.yaml
```diff
@@ -110,10 +110,16 @@ spec:
         volumeMounts:
         - mountPath: /tmp
           name: tmp
+        {{- range .Values.volumeMounts }}
+        - {{ toJson . }}
```
nikola-jokic: Can you please use toYaml? When someone renders a chart, it would be much easier to read

nikola-jokic (changes requested):
Can you please add tests?

charts/gha-runner-scale-set-controller/templates/deployment.yaml
taliastocks: Yes, I can do that. Thanks for the review.

taliastocks:
> Can you please add tests?

You probably want me to add to `TestTemplate_ControllerDeployment_Customize` right?

Do you have instructions somewhere on running these tests? It doesn't look like there's a Github Action enabled (somewhat ironically).

nikola-jokic:
Yeah, so to run tests, you do `cd charts/gha-runner-scale-set-controller/tests && go test`. The action is enabled only for maintainers :relaxed:

taliastocks:
> Yeah, so to run tests, you do `cd charts/gha-runner-scale-set-controller/tests && go test`. The action is enabled only for maintainers ☺️

Thank you! That is fair, security is important.

taliastocks:
Tests are updated and passing.

nikola-jokic (approved):
LGTM

PR #3280, opened 2024-02-12
Deprecation warning of older listener for 0.9.0 release

nikola-jokic:
This pull request includes a change to the `charts/gha-runner-scale-set-controller/templates/NOTES.txt` file. A warning message has been added to alert users that an older version of the listener, `githubrunnerscalesetlistener`, is deprecated and will be removed in the future `gha-runner-scale-set-0.10.0` release. Users are advised to remove the environment variable override and use the new listener, `ghalistener`, instead.

### DIFF
--- a/charts/gha-runner-scale-set-controller/templates/NOTES.txt
+++ b/charts/gha-runner-scale-set-controller/templates/NOTES.txt
@@ -2,3 +2,4 @@ Thank you for installing {{ .Chart.Name }}.
 
 Your release is named {{ .Release.Name }}.
 
+WARNING: Older version of the listener (githubrunnerscalesetlistener) is deprecated and will be removed in the future gha-runner-scale-set-0.10.0 release. If you are using environment variable override to force the old listener, please remove the environment variable and use the new listener (ghalistener) instead.


rentziass (approved):

EricWF:
Does this warning appear even if you're not using the old listener? 

I took it to mean "You _ARE_ using the old listener", and spent an hour trying to figure out how I was & how to fix it.
Because I am using environment variables to pick the chart versions, but they're all the same version.

Grepping for the strings `githubrunnerscalesetlistener` and `ghalistener` within the project files didn't yield any results, further confusing me.

Perhaps this warning could be rephrased to be more clear.

WyriHaximus:
> Does this warning appear even if you're not using the old listener?

Yes, given that there is no conditional this also showed on a fully fresh install I just did.

> Grepping for the strings `githubrunnerscalesetlistener` and `ghalistener` within the project files didn't yield any results, further confusing me.
> 
> Perhaps this warning could be rephrased to be more clear.

And maybe add links to documentation about them as well.

WyriHaximus:
@EricWF Been digging into the Docker image and I bet that it's about those binaries in the roof other filesystem:
image

PR #3296, opened 2024-02-21
Refer to the correct variable in discovery error message

ivarmedi:
`name` -> `namespace` in error message. 

The section in the contributing guidelines specifically mentions code in regards to creating an issue, and since the change is tiny and only touches text for human consumption, I opted not to. Let me know if I should :)


nikola-jokic (approved):
LGTM! Thanks!

PR #3310, opened 2024-02-28
Update metrics to include repository on job-based label

### LINKED ISSUE #3185: Scale Set Metrics - Missing Repo Name and Workflow Name
### Checks

### Controller Version

0.7.0

### Deployment Method

Helm

### Checks

### To Reproduce

```markdown
1. Install both ARC and the ScaleSet helm charts to a kubernetes cluster.
2. Run a job with a self hosted runner.
```

### Describe the bug

With the old self-hosted runners before the Scale Set, prometheus metrics like `github_workflow_job_failures_total` had repository/repository_full_name/workflow_name/job_name

Example
```
github_workflow_job_failures_total{container="actions-metrics-server", endpoint="metrics-port", exit_code="1", failed_step="10", instance="10.26.153.77:8080", job="actions-runner-controller-actions-metrics-server", job_name="client-codeql / client-codeql", namespace="github-actions", owner="orgname", pod="actions-runner-controller-actions-metrics-server-598bb5fb99lhl2", repository="reponame", repository_full_name="orgname/reponame", runs_on="self-hosted", service="actions-runner-controller-actions-metrics-server", workflow_name="build"}
```

I found this very helpful because I could monitor for when jobs began to fail or repos where the developers were running very long/costly ci.

Now with the new metrics we only have job_workflow_ref and job_name. job_workflow_ref only points to the org/repo/relative_path of the workflow, this means with "reuseable workflows" it only shows where the reusable workflow is (in our case a shared CI repo), not the dev's repo which called it.

Example:
```
gha_completed_jobs_total{container="autoscaler", endpoint="http-metrics", event_name="pull_request", instance="10.26.139.124:8080", job="self-hosted-listener", job_name=" / Build Test Publish", job_result="canceled", job_workflow_ref="myorg/github-actions-shared/.github/workflows/python-library.yaml@refs/heads/v1", namespace="github-actions", organization="myorg", pod="self-hosted-56cc585c-listener", service="self-hosted-listener"}
```

I expect in the prometheus metrics for the new Scale Set Listener for there to be the labels workflow_name and repository in addition to job_workflow_ref. As those labels existed in the prior self hosted runner paradigm.

I know that this will increase metric cardinality but it did not seem to be a problem when we ran the old self hosted runners.

I know we have a very unique case where we take advantage of the reusable workflows feature. Otherwise it makes it almost impossible to track what repos are initiating the workflows/jobs.

Thanks in advance 🙏 

### Additional Context

```yaml
Not Relevant
```

### Controller Logs

```shell
Not Relevant
```

### Runner Pod Logs

```shell
Not Relevant
```

nikola-jokic:
When creating labels used for the job-related metrics, apply repository name based on the job message, instead of `githubConfigUrl`.

Also, fix the comment in listener_test.

Partially addresses #3185

### DIFF
--- a/cmd/ghalistener/metrics/metrics.go
+++ b/cmd/ghalistener/metrics/metrics.go
@@ -223,7 +223,7 @@ type baseLabels struct {
 func (b *baseLabels) jobLabels(jobBase *actions.JobMessageBase) prometheus.Labels {
 	return prometheus.Labels{
 		labelKeyEnterprise:     b.enterprise,
-		labelKeyOrganization:   b.organization,
+		labelKeyOrganization:   jobBase.OwnerName,
 		labelKeyRepository:     jobBase.RepositoryName,
 		labelKeyJobName:        jobBase.JobDisplayName,
 		labelKeyJobWorkflowRef: jobBase.JobWorkflowRef,
--- a/cmd/ghalistener/metrics/metrics.go
+++ b/cmd/ghalistener/metrics/metrics.go
@@ -224,7 +224,7 @@ func (b *baseLabels) jobLabels(jobBase *actions.JobMessageBase) prometheus.Label
 	return prometheus.Labels{
 		labelKeyEnterprise:     b.enterprise,
 		labelKeyOrganization:   b.organization,
-		labelKeyRepository:     b.repository,
+		labelKeyRepository:     jobBase.RepositoryName,
 		labelKeyJobName:        jobBase.JobDisplayName,
 		labelKeyJobWorkflowRef: jobBase.JobWorkflowRef,
 		labelKeyEventName:      jobBase.EventName,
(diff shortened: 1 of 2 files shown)


cmd/ghalistener/metrics/metrics.go
```diff
@@ -224,7 +224,7 @@ func (b *baseLabels) jobLabels(jobBase *actions.JobMessageBase) prometheus.Label
 	return prometheus.Labels{
 		labelKeyEnterprise:     b.enterprise,
 		labelKeyOrganization:   b.organization,
```
github-samuel-rats: Hello! 

Should we do the same here? 
`b.organization` will be empty for runners assigned to an enterprise. 
```suggestion
		labelKeyOrganization:   jobBase.OwnerName,
```

nikola-jokic: Thanks!

rentziass (approved):
LGTM!

PR #3329, opened 2024-03-06
fix: set image,env for dind container from spec

verbotenj:


charts/gha-runner-scale-set/templates/_helpers.tpl
```diff
@@ -96,14 +96,20 @@ volumeMounts:
 {{- end }}
 
 {{- define "gha-runner-scale-set.dind-container" -}}
-image: docker:dind
+{{- range $i, $container := .Values.template.spec.containers }}
+  {{- if eq $container.name "dind" }}
```
miked-qc, line 100: How does this default the values if the "dind" container name is not found?

verbotenj: Nice catch, update please re-review

miked-qc (approved):
I was about to get a member of my team to look into this, thanks for proposing the improvement!

nikola-jokic:
Hey @verbotenj,

Thank you for suggesting this improvement, but as documented in [Customizing container modes](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/deploying-runner-scale-sets-with-actions-runner-controller#customizing-container-modes) section:
> To customize the spec, comment out or remove containerMode, and append the configuration you want in the template section.

I will close this PR now :relaxed:

miked-qc:
@nikola-jokic I'm not seeing a way to set where `docker:dind` comes from? We have the use case that we don't want to pull from docker hub but rather from our internal, scanned registry. Is there another way to do this other than the fix proposed?

PR #3335, opened 2024-03-08
Fix the deletion of ScaleSet

arkadiiemelianov-brt:
Using terraform `helm_release` resource with helm chart of version `0.8.3`: after successful deployment, when trying to remove the ScaleSet, controller is failing with missing permission, which fails the terraform run.

```
User \"system:serviceaccount:my-service-account:my-controller-rs-controller\" cannot delete resource \"rolebindings\" in API group \"rbac.authorization.k8s.io\" in the namespace \"my-namespace\""
```


arkadiiemelianov-brt:
Sorry, wrong issue.

PR #3336, opened 2024-03-08
fix(metrics): Make sure repository/organization is set for job metrics

github-samuel-rats:
These fields are empty when the runner is assigned to an enterprise/organization.

Jobs metrics are somewhat useless if not linked to their associated job.

NOT TESTED YET, will try do to that ASAP.


github-samuel-rats:
Looks like https://github.com/actions/actions-runner-controller/pull/3310 will do the trick.

PR #3371, opened 2024-03-20
Fix overscaling when the controller is much faster then the listener

nikola-jokic:
When the ephemeral runner finishes, the listener issues a scale-down patch to the ephemeral runner set. But if ephemeral runner set reacts quicker than the listener, it may re-create the ephemeral runner, and quickly remove it.
To avoid this problem,  this PR introduces the following changes:
1. Ephemeral runner cleans up its resources as soon as it is done
2. Ephemeral runner set keeps finished ephemeral runners until the next patch request is received.
3. The target count is calculated only on patch ID difference (or patchID == 0), i.e. when the listener patches the ephemeral runner set.
4. The listener issues a patch request when there is a change in the state, or in case it gets notified that the job is done.

This change also requires a change in the CRD.


apis/actions.github.com/v1alpha1/ephemeralrunnerset_types.go
```diff
@@ -24,6 +24,8 @@ import (
 type EphemeralRunnerSetSpec struct {
 	// Replicas is the number of desired EphemeralRunner resources in the k8s namespace.
 	Replicas int `json:"replicas,omitempty"`
+	// PatchID is unique identifier for the patch issued by the listener app
```
rentziass:
```suggestion
	// PatchID is the unique identifier for the patch issued by the listener app
```

charts/gha-runner-scale-set-controller/crds/actions.github.com_ephemeralrunnersets.yaml
```diff
@@ -6957,9 +6957,14 @@ spec:
                         - containers
                       type: object
                   type: object
+                patchID:
+                  description: PatchID is unique identifier for the patch issued by the listener app
```
rentziass:
```suggestion
                  description: PatchID is the unique identifier for the patch issued by the listener app
```

config/crd/bases/actions.github.com_ephemeralrunnersets.yaml
```diff
@@ -6957,9 +6957,14 @@ spec:
                         - containers
                       type: object
                   type: object
+                patchID:
+                  description: PatchID is unique identifier for the patch issued by the listener app
```
rentziass:
```suggestion
                  description: PatchID is the unique identifier for the patch issued by the listener app
```

rentziass (dismissed):
This looks good to me, I just left a bunch of suggestions over a type!

charts/gha-runner-scale-set-controller/crds/actions.github.com_ephemeralrunnersets.yaml
nikola-jokic: Generated with `make manifests` :relaxed:

rentziass (approved):

PR #3377, opened 2024-03-25
Add retry on 401 and 403 for runner-registration

nikola-jokic:


github/actions/client.go
```diff
@@ -975,28 +975,41 @@ func (c *Client) getActionsServiceAdminConnection(ctx context.Context, rt *regis
...
+		if resp.StatusCode != http.StatusUnauthorized && resp.StatusCode != http.StatusForbidden {
+			return nil, err
+		}
 
-	return actionsServiceAdminConnection, nil
+		retry++
+		if retry > 3 {
+			return nil, fmt.Errorf("unable to register runner after 3 retries: %v", err)
+		}
+		time.Sleep(1 * time.Second)
```
rentziass: Should we use a growing interval here? e.g. `250 * time.Millisecond * (retries + 1)` or something like that?

github/actions/client.go
```diff
@@ -975,28 +975,41 @@ func (c *Client) getActionsServiceAdminConnection(ctx context.Context, rt *regis
...
 
-	if resp.StatusCode < 200 || resp.StatusCode > 299 {
-		registrationErr := fmt.Errorf("unexpected response from Actions service during registration call: %v", resp.StatusCode)
+		if resp.StatusCode >= 200 && resp.StatusCode <= 299 {
+			var actionsServiceAdminConnection *ActionsServiceAdminConnection
+			if err := json.NewDecoder(resp.Body).Decode(&actionsServiceAdminConnection); err != nil {
+				return nil, err
+			}
 
+			return actionsServiceAdminConnection, nil
```
rentziass: Can we move handling the happy path out of the loop? I think the goal of the loop should be to handle retryable errors?

github/actions/client.go
```diff
@@ -975,28 +975,41 @@ func (c *Client) getActionsServiceAdminConnection(ctx context.Context, rt *regis
...
+		}
+
+		errStr := fmt.Sprintf("unexpected response from Actions service during registration call: %v", resp.StatusCode)
 		body, err := io.ReadAll(resp.Body)
 		if err != nil {
-			return nil, fmt.Errorf("%v - %v", registrationErr, err)
+			err = fmt.Errorf("%s - %v", errStr, err)
```
rentziass:
```suggestion
			err = fmt.Errorf("%s - %w", errStr, err)
```

github/actions/client.go
```diff
@@ -975,28 +975,41 @@ func (c *Client) getActionsServiceAdminConnection(ctx context.Context, rt *regis
 
 	c.logger.Info("getting Actions tenant URL and JWT", "registrationURL", req.URL.String())
 
-	resp, err := c.Do(req)
-	if err != nil {
-		return nil, err
-	}
-	defer resp.Body.Close()
+	retry := 0
+	for {
```
rentziass, line 980: This might be painful because this is unexported, but can we add a test for this?

nikola-jokic: Added, good call!

rentziass (approved):

PR #3394, opened 2024-03-31
Allow configuring dind image and service account

shaul75:
The PR addresses 2 of our issues
1. a service account with a specific annotation is needed to grant the runner pod IAM permissions in EKS. https://docs.aws.amazon.com/eks/latest/userguide/associate-service-account-role.html
2. Changing the dind image to pull from another repository
We have a base `values.yaml` with common options for all runner sets. this change allows to override just the dind image as a base value, without copy-pasting the entire block.

Related Discussion: https://github.com/actions/actions-runner-controller/discussions/3256


yardenasadosa:
Any updates? We have been waiting for 2.5 years for that.

PR #3405, opened 2024-04-04
Add topologySpreadConstraint to gha-runner-scale-set-controller chart

DaazKu:
Adds `topologySpreadConstraint` to `gha-runner-scale-set-controller` so that we can spread replicas across AZs like so:

```
topologySpreadConstraints:
  - labelSelector:
      matchLabels:
        app.kubernetes.io/name: gha-rs-controller # Or "nameOverride" if defined
    maxSkew: 1
    topologyKey: topology.kubernetes.io/zone
    whenUnsatisfiable: DoNotSchedule
```

Reference: https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/


charts/gha-runner-scale-set-controller/values.yaml
```diff
@@ -72,6 +72,8 @@ tolerations: []
 
 affinity: {}
 
+topologySpreadConstraints: {}
```
nikola-jokic:
```suggestion
topologySpreadConstraints: []
```

nikola-jokic (changes requested):
Please run go fmt :relaxed:

nikola-jokic (approved):
LGTM, thanks!

PR #3436, opened 2024-04-15
Fix doc comment for listenerTemplate

nasa9084:
The comment is for listenerTemplate but said `template is...` maybe it's copied from the next configuration, which is actually for `template`?


nikola-jokic (approved):
LGTM, thanks!

PR #3472, opened 2024-04-25
Crds custom annotations

GSZoominfo:
This PR will allow us to add custom annotations to the crds

use case:

- adding argo cd annotations such as: `argocd.argoproj.io/sync-options: ServerSideApply=true`


GSZoominfo:
@mumoshu @toast-gear @rentziass @nikola-jokic 
hey all - can someone please check this PR
it's a very small one and it could solve a lot of issues when using argo cd.

asafhm:
Hi @GSZoominfo ,
I was wondering, won't setting ServerSideApply option in the ArgoCD `Application` level be enough? 
If the main reason is to solve the infamous "Too long must have at most 262144 bytes" error, then setting this option in the application `SyncOptions` helped me with this very error in the `Prometheus` CRD

thomaspetit:
> Hi @GSZoominfo , I was wondering, won't setting ServerSideApply option in the ArgoCD `Application` level be enough? If the main reason is to solve the infamous "Too long must have at most 262144 bytes" error, then setting this option in the application `SyncOptions` helped me with this very error in the `Prometheus` CRD

You don't want to do this as setting it on Application level might break other functionalities like sync waves/hooks.
Reference to actual (open) issue: https://github.com/argoproj/argo-cd/issues/13320

The approach chosen in this PR is a good one. We see similar implementations on prometheus, kyverno, argocd:
- Prometheus:
  - https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-operator-crds/values.yaml#L4
  - https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-operator-crds/charts/crds/templates/crd-alertmanagerconfigs.yaml#L7
- Kyverno:
  - https://github.com/kyverno/kyverno/blob/main/charts/kyverno/charts/crds/values.yaml#L35
- ArgoCD:
  - https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/crds/crd-project.yaml#L9
  - https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/values.yaml#L37

PR #3506, opened 2024-05-08
feat(version): update runner version to 2.316.1

maljolani:


avsprashant:
same as https://github.com/actions/actions-runner-controller/pull/3496

PR #3528, opened 2024-05-17
Fix problem with ephemeralRunner Succeeded state before build executed

### LINKED ISSUE #3527: Ephemeral runner gets stuck in Successful state
### Checks

### Controller Version

0.9.0

### Deployment Method

Helm

### Checks

### To Reproduce

```markdown
1. I have ARC installed in GKE in one namespace: controller and runnersets in the same one
2. All works as expected in most of the cases
3. From time to time (randomly) we observe such behavior:
- ephemeral runnerset gets patch , e.g. from 0 desired to 1 replica
- ephemeral runner is created, but immediately its status is changed to Succeeded and nothing happens - the workload is "stuck" on waiting for runner
```

### Describe the bug

In controller logs I see that it already "Found the runner with the same name" - it looks like the controller is performing reconcile twice for the same ephemeralrunner in almost the same time, the second run "removes" runner and makes it hung.

The runner is eventually not created, and the ephemeral runner gets to stage Succeeded and stuck until workflow is cancelled. 

We started to observe this behavior when we moved to faster cluster. 

The controller should create runner always on ephemeral runner creation. 

### Additional Context

```yaml
N/A
```

### Controller Logs

```shell
https://gist.github.com/katarzynainit/ceccccde10d5454aa104d0f5a98f9b0d
```

### Runner Pod Logs

```shell
N/A
```

kr-sabre:
Fixes #3527

Seems that changing order of actions fixes the issues:
- first call GH services and update ephemeral runner status
- apply finalizers on CR

### DIFF
--- a/controllers/actions.github.com/ephemeralrunner_controller.go
+++ b/controllers/actions.github.com/ephemeralrunner_controller.go
@@ -149,6 +149,19 @@ func (r *EphemeralRunnerReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, nil
 	}
 
+	if !controllerutil.ContainsFinalizer(ephemeralRunner, ephemeralRunnerFinalizerName) {
+		log.Info("Adding finalizer")
+		if err := patch(ctx, r.Client, ephemeralRunner, func(obj *v1alpha1.EphemeralRunner) {
+			controllerutil.AddFinalizer(obj, ephemeralRunnerFinalizerName)
+		}); err != nil {
+			log.Error(err, "Failed to update with finalizer set")
+			return ctrl.Result{}, err
+		}
+
+		log.Info("Successfully added finalizer")
+		return ctrl.Result{}, nil
+	}
+
 	if !controllerutil.ContainsFinalizer(ephemeralRunner, ephemeralRunnerActionsFinalizerName) {
 		log.Info("Adding runner registration finalizer")
 		err := patch(ctx, r.Client, ephemeralRunner, func(obj *v1alpha1.EphemeralRunner) {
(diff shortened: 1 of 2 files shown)


kr-sabre:
But I'm wondering if returning result after each finalizer apply isn't more correct approach...
Please advise

controllers/actions.github.com/ephemeralrunner_controller.go
```diff
@@ -149,6 +149,11 @@ func (r *EphemeralRunnerReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, nil
 	}
 
+	if ephemeralRunner.Status.RunnerId == 0 {
```
nikola-jokic: Hey, I don't think that changing the registration and then patching the finalizers should be the right way to go. However, I noticed that the order in which we applied finalizers is incorrect.
We first need to apply the `ephemeralRunnerFinalizerName`, then the `ephemeralRunnerActionsFinalizerName`.

Would you be interested in trying out that approach?

kr-sabre: Sure, will do the tests, but shouldn't we return Result also after ephemeralrunner.actions.github.com/runner-registration-finalizer apply?

nikola-jokic: Good catch, we should!

kr-sabre: Updated the code, please check now

I don't see such issues with the updated code, let me check a little longer - will be back with update on Monday

nikola-jokic: Looks good to me. I'll wait for your feedback on Monday before merging it.

nikola-jokic: Can you please also fix the test :relaxed: The order of finalizers is different now

kr-sabre: > Looks good to me. I'll wait for your feedback on Monday before merging it.

We didn't observe the issue since Wednesday :)

nikola-jokic (approved):
LGTM! Thank you!

PR #3568, opened 2024-06-03
Check status code of fetch access token for github app

### LINKED ISSUE #3565: Misleading logs when blocked by IP allow list, does not show failure when requesting access token
### Checks

### Controller Version

0.9.2

### Deployment Method

ArgoCD

### Checks

### To Reproduce

```markdown
This problem happened with our Enterprise which has Enterprise Managed Users. Judging from the code, I would expect the same error to occur without EMU, but cannot test this.

1. Configure IP allow list so that it blocks ARC on the enterprise
2. Restart the controller
3. Observe the log
```

### Describe the bug

The logs indicate that obtaining an access token worked when in truth this fails with HTTP 403 due to the missing IP allow list entry.

```
2024-05-27T09:32:03Z	INFO	actions-clients	getting access token for GitHub App auth	{"accessTokenURL": "https://api.github.com/app/installations/<id>/access_tokens"}
```

GitHub Enterprise support has confirmed that this request actually failed with 403 due to the wrong IP allow list, but there is no log line indicating the error.

The next request then fails with 401, and this is the first error visible in the logs, when in truth this is a consequence of the failure to obtain an access token.

```
2024-05-27T09:32:03Z	INFO	actions-clients	getting runner registration token	{"registrationTokenURL": "https://api.github.com/orgs/<org>/actions/runners/registration-token"}
2024-05-27T09:32:03Z	ERROR	AutoscalingRunnerSet	Failed to get runner group by name	{"version": "0.9.2", "autoscalingrunnerset": {"name":"arc-test-runner","namespace":"cicd"}, "runnerGroup": "operator-runner-group", "error": "failed to get runner registration token on refresh: github api error: StatusCode 401, RequestID \"<omitted>\": {\"message\":\"Bad credentials\",\"documentation_url\":\"https://docs.github.com/rest\"}"}
```

I believe the error is due to not checking the status code on the HTTP response before parsing the JSON response.

Should insert a status check before this line (and fail and log an error, rather than continue):

https://github.com/actions/actions-runner-controller/blob/master/github/actions/client.go#L1058

The log should indicate the failure to obtain an access token.

### Additional Context

```yaml
not relevant
```

### Controller Logs

```shell
https://gist.github.com/StephanHCB/2028ecfe6ae94cc12a8bd298b1596346
```

### Runner Pod Logs

```shell
(no runner pod logs because controller fails to work, so no runners get started)
```

nikola-jokic:
Previously, the status code was not checked on `fetchAccessToken`.

This PR aims to add a check for the endpoint described [here](https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#create-an-installation-access-token-for-an-app). 

Fixes #3565

### DIFF
--- a/github/actions/client.go
+++ b/github/actions/client.go
@@ -1054,6 +1054,14 @@ func (c *Client) fetchAccessToken(ctx context.Context, gitHubConfigURL string, c
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode != http.StatusCreated {
+		return nil, &GitHubAPIError{
+			StatusCode: resp.StatusCode,
+			RequestID:  resp.Header.Get(HeaderGitHubRequestID),
+			Err:        fmt.Errorf("failed to get access token for GitHub App auth: %v", resp.Status),
+		}
+	}
+
 	// Format: https://docs.github.com/en/rest/apps/apps#create-an-installation-access-token-for-an-app
 	var accessToken *accessToken
 	if err = json.NewDecoder(resp.Body).Decode(&accessToken); err != nil {


kahirokunn:
LGTM

Link- (approved):

PR #3601, opened 2024-06-17
[JK] ARSS DinD Graceful termination - Wait for Runner

jaswanthikolla:
This is a decent workaround for this Issue https://github.com/actions/actions-runner-controller/issues/3600 . It basically waits until the runner worker is completed.

### DIFF
--- a/charts/gha-runner-scale-set/templates/_helpers.tpl
+++ b/charts/gha-runner-scale-set/templates/_helpers.tpl
@@ -113,6 +113,17 @@ volumeMounts:
     mountPath: /var/run
   - name: dind-externals
     mountPath: /home/runner/externals
+lifecycle:
+  preStop:
+    exec:
+      command:
+      - /bin/sh
+      - -c
+      - >
+        while pgrep "Runner.Worker" > /dev/null; do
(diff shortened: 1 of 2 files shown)


nicolaswebb (approved):

jaswanthikolla:
Karpenter support is essential to save significant amount of across all companies.  We save easily $200k+ per year, Scaling that across 1000's of tech companies, Karpenter support can easily save $50 million+ and CO2 Emissions. 

Can we be more responsible users, and prioritize reviewing and merging this PR.

velkovb:
Just saw this PR. I had the same issue but solved it in a different way - https://github.com/actions/actions-runner-controller/pull/3842. Moving it to a sidecar seems to do the job without additional hacks. Maybe combination of both for pre 1.29 clusters would be the ideal solution.

jaswanthikolla:
I did look into sidecar while working on this, But they need 1.29 which is still not not popular production version as earlier versions are not forced by EOL upgrade.

jaswanthikolla:
Now, that `1.29`+  is rolled out across the community, I am going to withdraw this PR.

PR #3607, opened 2024-06-20
Exclude label prefix propagation

### LINKED ISSUE #3533: Issue running 0.9.2. Listener crash restart loop when running inside ArgoCD
### Checks

### Controller Version

0.9.2

### Deployment Method

ArgoCD

### Checks

### To Reproduce

```markdown
Installed the arc controller and runner set via Helm onto ArgoCD. Once everything came up, the controller would start up a listener that would just get stuck in a crash loop attempting to call out to github and failing GETs.
```

### Describe the bug

When attempting to use the newest 0.9.2 version of the arc controller / runner sets, the listener that was spinning up kept having issues during startup. The screenshot below shows the issue that I was encountering. When downgrading to 0.9.1, the issue went away. 

image

### Additional Context

```yaml
N/A
```

### Controller Logs

```shell
The controller logs had nothing of use in them. They would spin up the listener, see that it died, and tried to spin up a new one. Over and over, forever.
```

### Runner Pod Logs

```shell
N/A
```

nikola-jokic:
Propagating labels through internal resources managed by ARC can cause issues when ARC is used in combination with other tools, such as ArgoCD. This change introduces new flag on the controller that accepts the list of label prefixes that shouldn't be propagated to internal resources.

The reason we choose to opt out of labels instead of opt in to the label propagation behavior is due to several reasons:
1. If cluster policies are configured to ensure every resource has a particular label, then propagating labels is a requirement
2. If you are applying specific labels in order to easily manage resources, you should be able to inspect sub-resources based on that label.
3. If you have cluster policies to propagate labels, and you want to use systems similar to Argo, that use resource tracking and act on it based on label, you should have a way to opt out of propagating specific labels.

Turning on/off label propagation would not solve the issue if you have cluster policies and you are managing your cluster with Argo. So this more granular approach ensures both scenarios are covered. 

The `resource builder` component is instantiated with the ruleset, and will ignore labels that match the prefix configured.

Fixes #3533

### DIFF
--- a/charts/gha-runner-scale-set-controller/templates/deployment.yaml
+++ b/charts/gha-runner-scale-set-controller/templates/deployment.yaml
@@ -79,6 +79,9 @@ spec:
         - "--listener-metrics-endpoint="
         - "--metrics-addr=0"
         {{- end }}
+        {{- range .Values.flags.excludeLabelPropagationPrefixes }}
+        - "--exclude-label-propagation-prefix={{ . }}"
+        {{- end }}
         command:
         - "/manager"
         {{- with .Values.metrics }}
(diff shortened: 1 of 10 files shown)


Link- (approved):
Pending updating the PR description to include the reasoning for why we adopted this approach.

PR #3638, opened 2024-07-03
ISSUE-3637: Do not copy verbosely in init container

schniedergers:
Do not use the `verbose` flag (`-v`) during copy of files in the init container.
This greatly reduces logs if the image contains additional package like node that have thousands of files.


nikola-jokic:
Closing in favor of #3805

PR #3642, opened 2024-07-04
feat(dind): allow to adjust log level and image name

PiotrPalkaSpotOn:
Allow users to specify docker log level and dind image


norman-zon:
would be included in #3645

bmatei-visma:
This would be appreciated in the context of docker 29.0.0 breaking change that causes issues like https://github.com/spring-projects/spring-boot/issues/48050

Until then you'd have to override the whole helper function with image `docker:28-dind`

PR #3697, opened 2024-08-06
Add PodMonitor to gha-runner-scale-set-controller chart

fabiiw05:


nikola-jokic:
Closing this one since the intention is to allow users to choose how to gather metrics on their cluster. Thank you for your effort!

PR #3710, opened 2024-08-16
Update README.md

SpksM69:
Don't allow 
```go

```


SpksM69:
[7808.pdf](https://github.com/user-attachments/files/16637208/7808.pdf)

PR #3714, opened 2024-08-19
Allow reading AutoscalingRunnerSet githubConfigSecret from controller namespace

lacarvalho91:
There isn't an open issue for this because the issue templates state that features for `actions.github.com` controllers should be raised elsewhere, which I have done [here](https://github.com/orgs/community/discussions/132518) but I haven't had a response yet. I also raised a discussion on this repo [here](https://github.com/actions/actions-runner-controller/discussions/3693).

The reason I am raising this PR is because in my setup the `AutoscalingRunnerSet`s we provide are specific to users, so users can mount their own config maps and secrets. This is a problem because it means they can see the secret referenced in `githubConfigSecret` since it must currently be in the same namespace as the runner itself. We are [configuring ARC to register at the organisation level](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/authenticating-to-the-github-api#authenticating-arc-with-a-github-app) so this means that users will have access to secrets for a GitHub App with "Self-hosted runners: Read and write" org level permissions, which we do not want.

To resolve this problem, this PR introduces a flag for the controller that, when enabled, tells the controller to read the secret referenced by `githubConfigSecret` from the same namespace as the controller. This is disabled by default.

I spent some time looking into the code and I couldn't actually see a technical reason for the runner secret to be in the same namespace as the runner, because the secret that gets read by the runner pod is a completely different secret - that one is the jitconfig secret that is created on the fly by the controller.

I have run this on a local kind cluster and verified that it works, from what I can see the e2e tests are only for the legacy controllers so I haven't added any coverage there. If there are e2e tests for the new controllers that I should add to please let me know.

### DIFF
--- a/controllers/actions.github.com/autoscalingrunnerset_controller_test.go
+++ b/controllers/actions.github.com/autoscalingrunnerset_controller_test.go
@@ -44,6 +44,7 @@ var _ = Describe("Test AutoScalingRunnerSet controller", Ordered, func() {
 	var mgr ctrl.Manager
 	var controller *AutoscalingRunnerSetReconciler
 	var autoscalingNS *corev1.Namespace
+	var controllerNS *corev1.Namespace
 	var autoscalingRunnerSet *v1alpha1.AutoscalingRunnerSet
 	var configSecret *corev1.Secret
 
@@ -61,7 +62,11 @@ var _ = Describe("Test AutoScalingRunnerSet controller", Ordered, func() {
 
 	BeforeEach(func() {
 		ctx = context.Background()
-		autoscalingNS, mgr = createNamespace(GinkgoT(), k8sClient)
+		var namespaces []*corev1.Namespace
+		namespaces, mgr = createNamespaces(GinkgoT(), 2)
+		autoscalingNS = namespaces[0]
+		controllerNS = namespaces[1]
+
 		configSecret = createDefaultSecret(GinkgoT(), k8sClient, autoscalingNS.Name)
 
 		controller = &AutoscalingRunnerSetReconciler{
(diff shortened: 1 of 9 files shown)


controllers/actions.github.com/autoscalingrunnerset_controller_test.go
```diff
@@ -197,6 +177,46 @@ var _ = Describe("Test AutoScalingRunnerSet controller", Ordered, func() {
 			Expect(err).NotTo(HaveOccurred(), "failed to list EphemeralRunnerSet")
 			Expect(len(runnerSetList.Items)).To(BeEquivalentTo(1), "Only one EphemeralRunnerSet should be created")
 		})
+
+		It("It should read GitHub config secret from controller namespace if ReadGitHubConfigSecretFromControllerNamespace is enabled", func() {
```
lacarvalho91, line 181: i've only added this test on the autoscalingrunnerset controller so far because I wanted to get some feedback before writing similar tests on the other controllers. I struggled to find a better way to write this test, it doesn't seem ideal that I'm setting up a second controller in this test but one is already setup in the `BeforeEach` but it isn't configurable. Any thoughts on this would be appreciated.

controllers/actions.github.com/autoscalingrunnerset_controller_test.go
```diff
@@ -197,6 +177,46 @@ var _ = Describe("Test AutoScalingRunnerSet controller", Ordered, func() {
...
+					if _, ok := created.Annotations[AnnotationKeyGitHubRunnerGroupName]; !ok {
+						return "", nil
+					}
+
+					return fmt.Sprintf("%s_%s", created.Annotations[runnerScaleSetIdAnnotationKey], created.Annotations[AnnotationKeyGitHubRunnerGroupName]), nil
+				},
+				autoscalingRunnerSetTestTimeout,
+				autoscalingRunnerSetTestInterval).Should(BeEquivalentTo("1_testgroup"), "RunnerScaleSet should be created/fetched and update the AutoScalingRunnerSet's annotation")
+
+		})
```
lacarvalho91, line 219: i've only added this test on the autoscalingrunnerset controller so far because I wanted to get some feedback before writing similar tests on the other controllers. I struggled to find a better way to write this test, it doesn't seem ideal that I'm setting up a second controller in this test but one is already setup in the `BeforeEach` but it isn't configurable. Any thoughts on this would be appreciated.

MichaelHudgins:
+1 to adding something like this.  I have a similar usecase and not having the secret in the runner's namespace would be a good win towards securing the api secret.

lacarvalho91:
@mumoshu @toast-gear @rentziass 

hey all, just wondering if theres anything that can be done to progress this 🙂

PR #3718, opened 2024-08-21
Don't copy with -v to reduce noise in log output

JohanLindvall:
Copying with `cp -r -v` logs several thousand lines per invocation.


nikola-jokic:
Closing as a duplicate of #3805

PR #3724, opened 2024-08-27
Add exponential backoff when generating runner reg tokens

Link-:
Improve the retry mechanism and add exponential backoff with jitter when refreshing the session token.

### DIFF
--- a/github/actions/client.go
+++ b/github/actions/client.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math/rand"
 	"net/http"
 	"net/url"
 	"strconv"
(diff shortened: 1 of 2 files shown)


rentziass (approved):

github/actions/client.go
```diff
@@ -1139,15 +1140,30 @@ func (c *Client) getActionsServiceAdminConnection(ctx context.Context, rt *regis
 		}
 
 		retry++
-		if retry > 3 {
+		if retry > 5 {
 			return nil, fmt.Errorf("unable to register runner after 3 retries: %w", &GitHubAPIError{
```
lucasfcnunes, line 1144: forgot to change `after 3` to `after 5`

PR #3736, opened 2024-09-11
docs: end markdown code block correctly

imnotjames:
the `automatically-scaling-runners.md` has a code block which isn't properly ended and that makes it harder to read (only a little)


nikola-jokic (approved):
Thanks!

PR #3782, opened 2024-10-22
Allow Actions Github Client HTTP timeout to be configurable

jb-2020:
We're on GHES and believe the 5 minute timeout on the listener to be too high for our network flow

This PR allows us to configure the HTTP timeout via the environment variable: `ACTIONS_HTTP_CLIENT_TIMEOUT_SECONDS`

If this variable is not provided then we will default to the current 5 minute (300 second) timeout.


rfpludwick:
Can we get some eyes on this PR please? Reducing the timeout will greatly help us reduce the number of hung runner pods. Thank you.

PR #3797, opened 2024-11-06
feat: allow namespace overrides

### LINKED ISSUE #3798: Cannot control target namespace

Add the ability to override the destination namespace.

### Why is this needed?

Sometimes it's handy to deploy both controller and scaler-set charts under the same Helm umbrella chart.
In such cases, since `.Release.Namespace` is the one inherited from the umbrella chart, we cannot follow the best practices of having separate namespaces for both controller and scalerset.

### Additional context

n/a

fernandezcuesta:
Add the ability to control the destination namespace for each chart, useful when deploying under umbrella charts.

Fix:  #3798


charts/gha-runner-scale-set/values.yaml
```diff
@@ -216,3 +216,6 @@ template:
 # controllerServiceAccount:
 #   namespace: arc-system
 #   name: test-arc-gha-runner-scale-set-controller
+
+# Add the option to deploy in another namespace rather than .Release.Namespace.
```
nikola-jokic:
```suggestion
# Overrides the default `.Release.Namespace` for all resources in this chart.
```

charts/gha-runner-scale-set-controller/values.yaml
```diff
@@ -136,6 +136,9 @@ flags:
   # excludeLabelPropagationPrefixes:
   #   - "argocd.argoproj.io/instance"
 
-  ## Defines the K8s client rate limiter parameters.
+# Add the option to deploy in another namespace rather than .Release.Namespace.
```
nikola-jokic:
```suggestion
# Overrides the default `.Release.Namespace` for all resources in this chart.
```

nikola-jokic (commented):
Can you please add tests?

fernandezcuesta:
@nikola-jokic done

nikola-jokic:
Hey @fernandezcuesta,

Can you also resolve conflicts. Sorry, I can do it for you if you want.

charts/gha-runner-scale-set/templates/kube_mode_role.yaml
```diff
@@ -6,8 +6,23 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: {{ include "gha-runner-scale-set.kubeModeRoleName" . }}
-  namespace: {{ include "gha-runner-scale-set.namespace" . }}
-  finalizers:
+    {{- toYaml . | nindent 4 }}
```
nikola-jokic, line 12: You are missing labels and namespace here and the finalizers

charts/gha-runner-scale-set/templates/kube_mode_serviceaccount.yaml
```diff
@@ -5,9 +5,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: {{ include "gha-runner-scale-set.kubeModeServiceAccountName" . }}
-  namespace: {{ include "gha-runner-scale-set.namespace" . }}
-  {{- if .Values.containerMode.kubernetesModeServiceAccount }}
-  {{- with .Values.containerMode.kubernetesModeServiceAccount.annotations }}
+  {{- if or .Values.annotations $hasCustomResourceMeta }}
```
nikola-jokic, line 9: Missing the namespace

nikola-jokic (changes requested):

charts/gha-runner-scale-set/templates/kube_mode_role.yaml
fernandezcuesta: yeah sorry, I figured it out right after pushing and was fixing that

charts/gha-runner-scale-set/templates/kube_mode_serviceaccount.yaml
fernandezcuesta: should be fixed

charts/gha-runner-scale-set/tests/template_test.go
```diff
@@ -2315,3 +2316,43 @@ func TestCustomAnnotations(t *testing.T) {
...
+	require.NoError(t, err)
+
+	for _, f := range templateFiles {
+		if filepath.Ext(f.Name()) != ".yaml" && filepath.Ext(f.Name()) != ".yml" {
+			continue
+		}
+		templateFile := filepath.Join("templates", f.Name())
+		output, err := helm.RenderTemplateE(t, options, chartPath, releaseName, []string{templateFile})
+
+		if err != nil {
```
nikola-jokic: I'm sorry for insisting, but I don't think this is the right way to test it. 

If there is an error in yaml format, render would fail, and we would skip it. The better way to probably do it is to have `map[string]*helm.Options`. I understand it is time consuming, but we really need to properly cover these test cases so we don't introduce errors in future releases.

fernandezcuesta: What I saw is that this fails if:

- there template is conditionally rendered (e.g. wrapped in an `{{- if .Values.xxxx }}`)
- the template invokes a helper which relies on a separate resource, but here we're deploying a single resource

I thought about rendering the whole chart as is (not template by template), then check every single resource's `metadata.name` and assert it matches `namespaceOverride` but didn't find an easy way to do it.

~Sorry but I don't fully understand your suggestion about using Helm options, do you refer to inject the values that may potentially hide certain resources (as explained above, the reason to ignore them)?~

I will follow your suggestion, makes sense now after thinking it twice. Thanks!

nikola-jokic: Thank you for responding quickly and applying fixes!

fernandezcuesta: let me know wdyt now

charts/gha-runner-scale-set/tests/template_test.go
```diff
@@ -2315,3 +2316,143 @@ func TestCustomAnnotations(t *testing.T) {
...
+		}
+
+		var renderedObject map[string]interface{}
+		helm.UnmarshalK8SYaml(t, output, &renderedObject)
+
+		if renderedObject["metadata"].(map[string]interface{})["namespace"] != nil {
+			assert.Equal(t, namespaceOverride, renderedObject["metadata"].(map[string]interface{})["namespace"], fmt.Sprintf("template %s from chart %s should have namespace %s", f.Name(), chartPath, namespaceOverride))
+		}
+	}
+}
```
nikola-jokic:
```suggestion
func TestNamespaceOverride(t *testing.T) {
	t.Parallel()

	chartPath := "../../gha-runner-scale-set"

	releaseName := "test"
	releaseNamespace := "test-" + strings.ToLower(random.UniqueId())
	namespaceOverride := "test-" + strings.ToLower(random.UniqueId())

	tt := map[string]struct {
		file    string
		options *helm.Options
	}{
		"manager_role": {
			file: "manager_role.yaml",
			options: &helm.Options{
				Logger: logger.Discard,
				SetValues: map[string]string{
					"namespaceOverride":                  namespaceOverride,
					"controllerServiceAccount.name":      "foo",
					"controllerServiceAccount.namespace": "bar",
					"githubConfigSecret.github_token":    "gh_token12345",
					"githubConfigUrl":                    "https://github.com",
				},
				KubectlOptions: k8s.NewKubectlOptions("", "", releaseNamespace),
			},
		},
		"manager_role_binding": {
			file: "manager_role_binding.yaml",
			options: &helm.Options{
				Logger: logger.Discard,
				SetValues: map[string]string{
					"namespaceOverride":                  namespaceOverride,
					"controllerServiceAccount.name":      "foo",
					"controllerServiceAccount.namespace": "bar",
					"githubConfigSecret.github_token":    "gh_token12345",
					"githubConfigUrl":                    "https://github.com",
				},
				KubectlOptions: k8s.NewKubectlOptions("", "", releaseNamespace),
			},
		},
		"no_permission_serviceaccount": {
			file: "no_permission_serviceaccount.yaml",
			options: &helm.Options{
				Logger: logger.Discard,
				SetValues: map[string]string{
					"namespaceOverride":                  namespaceOverride,
					"controllerServiceAccount.name":      "foo",
					"controllerServiceAccount.namespace": "bar",
					"githubConfigSecret.github_token":    "gh_token12345",
					"githubConfigUrl":                    "https://github.com",
				},
				KubectlOptions: k8s.NewKubectlOptions("", "", releaseNamespace),
			},
		},
		"autoscalingrunnerset": {
			file: "autoscalingrunnerset.yaml",
			options: &helm.Options{
				Logger: logger.Discard,
				SetValues: map[string]string{
					"namespaceOverride":                  namespaceOverride,
					"controllerServiceAccount.name":      "foo",
					"controllerServiceAccount.namespace": "bar",
					"githubConfigSecret.github_token":    "gh_token12345",
					"githubConfigUrl":                    "https://github.com",
				},
				KubectlOptions: k8s.NewKubectlOptions("", "", releaseNamespace),
			},
		},
		"githubsecret": {
			file: "githubsecret.yaml",
			options: &helm.Options{
				Logger: logger.Discard,
				SetValues: map[string]string{
					"namespaceOverride":                  namespaceOverride,
					"controllerServiceAccount.name":      "foo",
					"controllerServiceAccount.namespace": "bar",
					"githubConfigSecret.github_token":    "gh_token12345",
					"githubConfigUrl":                    "https://github.com",
				},
				KubectlOptions: k8s.NewKubectlOptions("", "", releaseNamespace),
			},
		},
		"kube_mode_role": {
			file: "kube_mode_role.yaml",
			options: &helm.Options{
				Logger: logger.Discard,
				SetValues: map[string]string{
					"namespaceOverride":                  namespaceOverride,
					"containerMode.type":                 "kubernetes",
					"controllerServiceAccount.name":      "foo",
					"controllerServiceAccount.namespace": "bar",
					"githubConfigSecret.github_token":    "gh_token12345",
					"githubConfigUrl":                    "https://github.com",
				},
				KubectlOptions: k8s.NewKubectlOptions("", "", releaseNamespace),
			},
		},
		"kube_mode_role_binding": {
			file: "kube_mode_role_binding.yaml",
			options: &helm.Options{
				Logger: logger.Discard,
				SetValues: map[string]string{
					"namespaceOverride":                  namespaceOverride,
					"containerMode.type":                 "kubernetes",
					"controllerServiceAccount.name":      "foo",
					"controllerServiceAccount.namespace": "bar",
					"githubConfigSecret.github_token":    "gh_token12345",
					"githubConfigUrl":                    "https://github.com",
				},
				KubectlOptions: k8s.NewKubectlOptions("", "", releaseNamespace),
			},
		},
		"kube_mode_serviceaccount": {
			file: "kube_mode_serviceaccount.yaml",
			options: &helm.Options{
				Logger: logger.Discard,
				SetValues: map[string]string{
					"namespaceOverride":                  namespaceOverride,
					"containerMode.type":                 "kubernetes",
					"controllerServiceAccount.name":      "foo",
					"controllerServiceAccount.namespace": "bar",
					"githubConfigSecret.github_token":    "gh_token12345",
					"githubConfigUrl":                    "https://github.com",
				},
				KubectlOptions: k8s.NewKubectlOptions("", "", releaseNamespace),
			},
		},
	}

	for name, tc := range tt {
		t.Run(name, func(t *testing.T) {
			t.Parallel()
			templateFile := filepath.Join("templates", tc.file)

			output, err := helm.RenderTemplateE(t, tc.options, chartPath, releaseName, []string{templateFile})
			if err != nil {
				t.Errorf("Error rendering template %s from chart %s: %s", tc.file, chartPath, err)
			}

			type object struct {
				Metadata metav1.ObjectMeta
			}
			var renderedObject object
			helm.UnmarshalK8SYaml(t, output, &renderedObject)
			assert.Equal(t, namespaceOverride, renderedObject.Metadata.Namespace)
		})
	}
}
```

charts/gha-runner-scale-set-controller/tests/template_test.go
```diff
@@ -1078,3 +1078,114 @@ func TestDeployment_excludeLabelPropagationPrefixes(t *testing.T) {
...
+				}
+			}
+			if mustHaveNamespaceOverriden {
+				assert.Equal(t, namespaceOverride, renderedObject["metadata"].(map[string]interface{})["namespace"], fmt.Sprintf("template %s from chart %s should have namespace %s", f.Name(), chartPath, namespaceOverride))
+			} else {
+				assert.NotEqual(t, namespaceOverride, renderedObject["metadata"].(map[string]interface{})["namespace"], fmt.Sprintf("template %s from chart %s should not have namespace %s", f.Name(), chartPath, namespaceOverride))
+			}
+		}
+	}
+}
```
nikola-jokic:
```suggestion
func TestNamespaceOverride(t *testing.T) {
	t.Parallel()

	chartPath := "../../gha-runner-scale-set-controller"

	releaseName := "test"
	releaseNamespace := "test-" + strings.ToLower(random.UniqueId())
	namespaceOverride := "test-" + strings.ToLower(random.UniqueId())

	tt := map[string]struct {
		file          string
		options       *helm.Options
		wantNamespace string
	}{
		"leader_election_role": {
			file: "leader_election_role.yaml",
			options: &helm.Options{
				Logger: logger.Discard,
				SetValues: map[string]string{
					"namespaceOverride": namespaceOverride,
					"replicaCount":      "2",
				},
				KubectlOptions: k8s.NewKubectlOptions("", "", releaseNamespace),
			},
			wantNamespace: namespaceOverride,
		},
		"leader_election_role_binding": {
			file: "leader_election_role_binding.yaml",
			options: &helm.Options{
				Logger: logger.Discard,
				SetValues: map[string]string{
					"namespaceOverride": namespaceOverride,
					"replicaCount":      "2",
				},
				KubectlOptions: k8s.NewKubectlOptions("", "", releaseNamespace),
			},
			wantNamespace: namespaceOverride,
		},
		"manager_single_namespace_controller_role": {
			file: "manager_single_namespace_controller_role.yaml",
			options: &helm.Options{
				Logger: logger.Discard,
				SetValues: map[string]string{
					"namespaceOverride":          namespaceOverride,
					"flags.watchSingleNamespace": "true",
				},
				KubectlOptions: k8s.NewKubectlOptions("", "", releaseNamespace),
			},
			wantNamespace: namespaceOverride,
		},
		"manager_single_namespace_controller_role_binding": {
			file: "manager_single_namespace_controller_role_binding.yaml",
			options: &helm.Options{
				Logger: logger.Discard,
				SetValues: map[string]string{
					"namespaceOverride":          namespaceOverride,
					"flags.watchSingleNamespace": "true",
				},
				KubectlOptions: k8s.NewKubectlOptions("", "", releaseNamespace),
			},
			wantNamespace: namespaceOverride,
		},
		"manager_single_namespace_watch_role": {
			file: "manager_single_namespace_watch_role.yaml",
			options: &helm.Options{
				Logger: logger.Discard,
				SetValues: map[string]string{
					"namespaceOverride":          namespaceOverride,
					"flags.watchSingleNamespace": "target-ns",
				},
				KubectlOptions: k8s.NewKubectlOptions("", "", releaseNamespace),
			},
			wantNamespace: "target-ns",
		},
		"manager_single_namespace_watch_role_binding": {
			file: "manager_single_namespace_watch_role_binding.yaml",
			options: &helm.Options{
				Logger: logger.Discard,
				SetValues: map[string]string{
					"namespaceOverride":          namespaceOverride,
					"flags.watchSingleNamespace": "target-ns",
				},
				KubectlOptions: k8s.NewKubectlOptions("", "", releaseNamespace),
			},
			wantNamespace: "target-ns",
		},
	}

	for name, tc := range tt {
		t.Run(name, func(t *testing.T) {
			t.Parallel()
			templateFile := filepath.Join("templates", tc.file)

			output, err := helm.RenderTemplateE(t, tc.options, chartPath, releaseName, []string{templateFile})
			if err != nil {
				t.Errorf("Error rendering template %s from chart %s: %s", tc.file, chartPath, err)
			}

			type object struct {
				Metadata metav1.ObjectMeta
			}
			var renderedObject object
			helm.UnmarshalK8SYaml(t, output, &renderedObject)
			assert.Equal(t, tc.wantNamespace, renderedObject.Metadata.Namespace)
		})
	}
}
```

nikola-jokic (commented):
Suggested few updates for clarity.
You need to add import for these:
```go
import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```

charts/gha-runner-scale-set-controller/tests/template_test.go
fernandezcuesta: :thinking:  in this case I guess we're not covering all the resources, for example the deployment is out of the scope. My point here was to add some fallback Values, and customize only those with conditions. In the case of `gha-runner-scale-set` chart, your suggestion fully makes sense and covers all resources (didn't realize that all of them had to have an entry), but in this case there're more resources which are OK with just setting `nameOverride` to validate it.

nikola-jokic: Yeah, but then you move the logic into a test, and the test is basically a documentation of what is expected and how. If the test is too complicated, it is hard to see what is expected.
With the approach I suggested, you can clearly see the expected value when you override the namespace.
Other tests already cover no namespace override, so you cannot introduce a bug that will not be caught by the test suite. 

Since I forgot to add the deployment, that is just one test case more. Each file is written separately, so each file has a test case.

fernandezcuesta: OK done, thanks for the help!

nikola-jokic (approved):
LGTM Thanks!

PR #3812, opened 2024-11-18
Clarify syntax for `githubConfigSecret`

jww3:
We've received several reports from customers that the embedded documentation on `githubConfigSecret` is confusing.  This PR attempts to add clarity there.


yacaovsnc (approved):

charts/gha-runner-scale-set/values.yaml
```diff
@@ -2,25 +2,40 @@
 ## ex: https://github.com/myorg/myrepo or https://github.com/myorg
```
corycalahan, line 2: Are there any specific configuration setting(s) that could also be supplied as examples when [Authenticating ARC with a personal access token (classic)](https://docs.github.com/en/enterprise-cloud@latest/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/authenticating-to-the-github-api#authenticating-arc-with-a-personal-access-token-classic) that could be possible clarifying text for "`githubConfigUrl: ""`"?

```yaml
## githubConfigUrl is the GitHub url for where you want to configure runners
## ex: https://github.com/myorg/myrepo or https://github.com/myorg
githubConfigUrl: ""
```

e.g. if yes, what is the string format?

```
https://github.com/enterprises/myenterprise
```
?

PR #3825, opened 2024-11-27
Enable k8s client cache for Secrets and ConfigMaps in the single namespace mode

tfujiwar:
## WHAT

- Allow the controller to watch Secrets / ConfigMaps in the single namespace mode.
- Enable the k8s API client cache for Secrets / ConfigMaps in the single namespace mode.

## WHY

To reduce latency of runner pod creation (https://github.com/actions/actions-runner-controller/issues/3276).

The EphemeralRunnerReconciler [retrieves](https://github.com/actions/actions-runner-controller/blob/8b36ea90ebe81710fcdcb4f96424b43203d24f1e/controllers/actions.github.com/ephemeralrunner_controller.go#L184-L193) a JITConfig Secret for every reconciliation once the Secret is created. It can be a performance bottleneck because the k8s API client cache is [disabled](https://github.com/actions/actions-runner-controller/blob/8b36ea90ebe81710fcdcb4f96424b43203d24f1e/main.go#L229-L236) for Secrets.

The cache has been disabled for Secrets because it requires cluster-wide list/watch permissions in the default mode. But in the single namespace mode, we can narrow down the permissions only to the single namespace and the controller namespace, which would be acceptable.

This change is aligned with [ADR 2023-04-11: Limit Permissions for Service Accounts in Actions-Runner-Controller](https://github.com/actions/actions-runner-controller/blob/master/docs/adrs/2023-04-11-limit-manager-role-permission.md#user-content-fn-1-e815e9ff5813b4c89ea144e793e80598).

> In this mode, you will end up with a manager `Role` that has all Get/List/Create/Delete/Update/Patch/Watch permissions on resources we need, and a `RoleBinding` to bind the `Role` with the controller `ServiceAccount` in the watched single namespace and the controller namespace

## TESTING

We have tested the change with a controller that manages 300-400 EphemeralRunners at peak times. We confirmed that the duration of getting k8s Secrets from the EphemeralRunnerReconciler has decreased due to the caching.

Screenshot 2024-11-27 at 12 06 56

We also confirmed that the number of Pending EphemeralRunners has decreased because they are reconciled faster.

Before

Screenshot 2024-11-27 at 12 23 39

After

Screenshot 2024-11-27 at 12 22 42


Okabe-Junya:
@nikola-jokic @rentziass 

Friendly reminder, can you take a look when you have time?
I can take over this PR. Thanks in advance!

PR #3832, opened 2024-11-30
Make EphemeralRunnerController MaxConcurrentReconciles configurable

mumoshu:
Improve the runner pod startup times by using more goroutines(almost "more cpus", assuming you are NOT K8s API or network bounded) for ephemeralrunner reconciliation.

Today, `cotroller-runtime` and hence ARC reconciles EphemeralRunnerPod one by one, which should be limiting the runner pod creation throughput to almost `unit time / latency of the GHA jit config API`.

If you are not bounded by CPU, K8s, network, or GHA-API yet, setting N MaxConcurrentReconciles should improve the throughput by N.

This is the PR for the second commit mentioned in https://github.com/actions/actions-runner-controller/issues/3276#issuecomment-2496636965.

### DIFF
--- a/controllers/actions.github.com/options.go
+++ b/controllers/actions.github.com/options.go
@@ -1,0 +1,90 @@
+package actionsgithubcom
+
+import (
+	"fmt"
+	"os"
+	"strconv"
+
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+)
+
+// Options is the optional configuration for the controllers, which can be
+// set via command-line flags or environment variables.
+type Options struct {
+	// RunnerMaxConcuncurrentReconciles is the maximum number of concurrent Reconciles which can be run
+	// by the EphemeralRunnerController.
+	RunnerMaxConcuncurrentReconciles int
+}
+
+// OptionsWithDefault returns the default options.
+// This is here to maintain the options and their default values in one place,
+// rather than having to correlate those in multiple places.
+func OptionsWithDefault() Options {
+	return Options{
+		RunnerMaxConcuncurrentReconciles: 1,
+	}
+}
+
+// LoadEnv loads the options from the environment variables.
+// This updates the option value only if the environment variable is set.
+// If the option is already set (via a command-line flag), the value from the environment variable takes precedence.
+func (o *Options) LoadEnv() error {
+	v, err := o.getEnvInt("RUNNER_MAX_CONCURRENT_RECONCILES")
+	if err != nil {
+		return err
+	}
+
+	if v != nil {
+		o.RunnerMaxConcuncurrentReconciles = *v
+	}
+
+	return nil
+}
+
+func (o *Options) getEnvInt(name string) (*int, error) {
+	s := os.Getenv(name)
+	if s == "" {
+		return nil, nil
+	}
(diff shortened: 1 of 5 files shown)


controllers/actions.github.com/options.go
```diff
@@ -0,0 +1,90 @@
...
+	return Options{
+		RunnerMaxConcuncurrentReconciles: 1,
+	}
+}
+
+// LoadEnv loads the options from the environment variables.
+// This updates the option value only if the environment variable is set.
+// If the option is already set (via a command-line flag), the value from the environment variable takes precedence.
+func (o *Options) LoadEnv() error {
+	v, err := o.getEnvInt("RUNNER_MAX_CONCURRENT_RECONCILES")
```
mumoshu: I added this env just to make it easy to give it a shot without modifying the chart.

Instead, we can modify the manager args template there https://github.com/actions/actions-runner-controller/blob/8b36ea90ebe81710fcdcb4f96424b43203d24f1e/charts/gha-runner-scale-set-controller/templates/deployment.yaml#L50

controllers/actions.github.com/options.go
```diff
@@ -0,0 +1,90 @@
...
+	// by the EphemeralRunnerController.
+	RunnerMaxConcuncurrentReconciles int
+}
+
+// OptionsWithDefault returns the default options.
+// This is here to maintain the options and their default values in one place,
+// rather than having to correlate those in multiple places.
+func OptionsWithDefault() Options {
+	return Options{
+		RunnerMaxConcuncurrentReconciles: 1,
```
mumoshu: Making 2 the new default so that users can benefit from the update without any config changes.
Formerly it was treated 1 as not specified, and having 2 shouldn't be harmful as you should be able to limit CPU req/lim at the K8s level anyway.

controllers/actions.github.com/options.go
```diff
@@ -0,0 +1,90 @@
...
+	// by the EphemeralRunnerController.
+	RunnerMaxConcuncurrentReconciles int
+}
+
+// OptionsWithDefault returns the default options.
+// This is here to maintain the options and their default values in one place,
+// rather than having to correlate those in multiple places.
+func OptionsWithDefault() Options {
+	return Options{
+		RunnerMaxConcuncurrentReconciles: 2,
```
mumoshu: Making 2 the new default so that users can benefit from the update without any config changes.
Formerly it was treated 1 as not specified, and having 2 shouldn't be harmful as you should be able to limit CPU req/lim at the K8s level anyway.

controllers/actions.github.com/options.go
Link-: @mumoshu - I recommend we go with modifying the chart and adding this as an extra element under `flags`: https://github.com/actions/actions-runner-controller/blob/8b36ea90ebe81710fcdcb4f96424b43203d24f1e/charts/gha-runner-scale-set-controller/values.yaml#L97

mumoshu: @Link- Thanks! Addressed in [0bebb66](https://github.com/actions/actions-runner-controller/pull/3832/commits/0bebb66bc277a3bf9a65381ea8a4580bad42531f)

Link- (approved):
Thank you @mumoshu 🙏

PR #3833, opened 2024-12-02
Make k8s client rate limiter parameters configurable

tfujiwar:
## WHAT

Make k8s client rate limiter parameters configurable.

## WHY

To reduce latency of runner pod creation (https://github.com/actions/actions-runner-controller/issues/3276).

EphemeralRunnerReconciler requires several k8s API calls until pod creation and those API calls are rate limited by a client library. It would be helpful if users can optimize the rate limiter parameter on their responsibility.

## TESTING

We have tested the change with QPS=80 and Burst=120 (4 times higher than the default) while running 50-100 runners and confirmed that the reconcile time of EphemeralRunnerReconciler has decreased from 60 ms to 20 ms.

Screenshot 2024-12-02 at 21 43 19

### DIFF
--- a/main.go
+++ b/main.go
@@ -103,6 +103,9 @@ func main() {
 		autoScalerImagePullSecrets stringSlice
 
 		commonRunnerLabels commaSeparatedStringSlice
+
+		k8sClientRateLimiterQPS   int
+		k8sClientRateLimiterBurst int
 	)
 	var c github.Config
 	err = envconfig.Process("github", &c)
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -145,6 +148,8 @@ func main() {
 	flag.BoolVar(&autoScalingRunnerSetOnly, "auto-scaling-runner-set-only", false, "Make controller only reconcile AutoRunnerScaleSet object.")
 	flag.StringVar(&updateStrategy, "update-strategy", "immediate", `Resources reconciliation strategy on upgrade with running/pending jobs. Valid values are: "immediate", "eventual". Defaults to "immediate".`)
 	flag.Var(&autoScalerImagePullSecrets, "auto-scaler-image-pull-secrets", "The default image-pull secret name for auto-scaler listener container.")
+	flag.IntVar(&k8sClientRateLimiterQPS, "k8s-client-rate-limiter-qps", 20, "The QPS value of the K8s client rate limiter.")
+	flag.IntVar(&k8sClientRateLimiterBurst, "k8s-client-rate-limiter-burst", 30, "The burst value of the K8s client rate limiter.")
```
tfujiwar, line 152: These default values are the current value defined in controller-runtime.
https://github.com/kubernetes-sigs/controller-runtime/blob/f529320b4c0fa5205067dec4eaeabc3c6cf62d25/pkg/client/config/config.go#L101-L106

Link-:
Thank you @tfujiwar - I had to move this change to another branch so that I can run all the workflows and validate the change.

You will still be the author of the change and the commit will remain attributed to you.

https://github.com/actions/actions-runner-controller/pull/3848

pulkitanz:
This has sped up my reconciles a fair bit, thank you for adding this!! :)

Issue #3709, opened 2024-08-15
Ability to customize the dind image in gha-runner-scale-set


victor-cetinic:
### What would you like added?

I have been trying to customise the gha-runner-scale-set so that i can set the docker registry mirror globally for all the runners using the Docker DIND approach.  

Most instructions point to modifying the registry settings of the docker service by adding a command line like :
``` --registry-mirror=https://mirror.gcr.io```
(further options - https://cloud.google.com/artifact-registry/docs/pull-cached-dockerhub-images )

The issue im facing is that these need to be configured on the **dind** container, and not the runner container., and when i created a template like this

```

template:
  spec: 
    nodeSelector:
      cloud.google.com/gke-preemptible: "true"
    initContainers:
      - name: dockerconfigwriter
        image: alpine
        command:
          - sh
          - -c
          - |
            echo "{\"registry-mirrors\": [ \"https://mirror.gcr.io\"] }" > /home/runner/.docker/daemon.json
        volumeMounts:
          - name: docker-config-volume
            mountPath: /home/runner/.docker
    containers:
      - name: runner
        command:
          - /home/runner/run.sh
        image:
          ghcr.io/actions/actions-runner:latest
        env:
          - name: DOCKER_HOST
            value: unix:///var/run/docker.sock
          - name: RUNNER_WAIT_FOR_DOCKER_IN_SECONDS
            value: \"120\"
        resources:
          limits:
            cpu: 6
            memory: 5Gi
        volumeMounts:
          - name: work
            mountPath: /home/runner/_work
          - name: dind-sock
            mountPath: /var/run
          - name: docker-config-volume
            mountPath: /etc/docker
      - name: dind
        image: docker:dind
        args:
          - dockerd
          - --host=unix:///var/run/docker.sock
          - --group=$(DOCKER_GROUP_GID)
          - --registry-mirror=https://mirror.gcr.io
        env:
          - name: DOCKER_GROUP_GID
            value: \"123\"
        securityContext:
          privileged: true
        volumeMounts:
          - name: work
            mountPath: /home/runner/_work
          - name: dind-sock
            mountPath: /var/run
          - name: dind-externals
            mountPath: /home/runner/externals
          - name: docker-config-volume
            mountPath: /etc/docker
    volumes:
      - name: docker-config-volume
        emptyDir: { }
```

[...]

PR #3645, opened 2024-07-05
feat(chart): refactor helm chart

norman-zon:
The reactions and follow up discussion in #3216 seemed to signal users are interested in a design for the helm chart, where pod configuration is done using `extra...` values, instead of the `template.spec` approach.
I understand the design decision for the current implementation, but wanted to provide an alternative.

PR #3851, opened 2024-12-16
Make DIND image configurable and remove non-runner-containers helper to avoid unintended DIND inclusion

matthew-collins-dailypay:
This PR updates the `gha-runner-scale-set` Helm chart templates to allow the DIND image to be configurable based on `.Values.template.spec.containers`. With this change, the chart will continue to use `docker:dind` by default, but if a container named `dind` is defined, that image is used instead. This provides greater flexibility for users who need to customize their DIND image.

As a result of making the DIND image configurable, the `non-runner-containers` helper now inadvertently includes the DIND container in scenarios where it’s not intended, specifically when using Kubernetes mode. To address this, we remove the `non-runner-containers` helper and rely exclusively on `non-runner-non-dind-containers`. This ensures that DIND remains properly scoped and does not appear where it should not.

### DIFF
--- a/charts/gha-runner-scale-set/templates/_helpers.tpl
+++ b/charts/gha-runner-scale-set/templates/_helpers.tpl
@@ -96,7 +96,14 @@ volumeMounts:
 {{- end }}
 
 {{- define "gha-runner-scale-set.dind-container" -}}
-image: docker:dind
+{{- $dindImage := "docker:dind" -}}
+{{- range $i, $val := .Values.template.spec.containers }}
+  {{- if eq $val.name "dind" }}
+    {{- $dindImage = $val.image }}
+  {{- end }}
+{{- end }}
+
+image: {{ $dindImage }}
 args:
   - dockerd
   - --host=unix:///var/run/docker.sock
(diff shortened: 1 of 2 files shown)


niallhdailypay (approved):

norman-zon:
also covered here: https://github.com/actions/actions-runner-controller/pull/3645. no reaction from the maintainers for months...

FearlessHyena:
@Link- @nikola-jokic @rentziass 
Would really like this and #3645 to be merged so #3709 can be supported
This seems to be a very straightforward change so could someone from GH please review and merge this?

MKarytka-gastro:
Any update here, please? Important one.

djetelina:
Hi, this would be really useful :)

kohbis:
What's blocking the merge?
This change is convenient for users.

cyrilico:
+1 for this, it's annoying to have to override a full spec when we just need to point to an internal registry

nikola-jokic:
Please refer to this comment: https://github.com/actions/actions-runner-controller/pull/4219#issuecomment-3236359120

I will close this PR since it would require a redesign of the chart instead of a temporary fix. I understand that customizing just one field would result in larger values.yaml file, but I would disagree that it is highly inconvenient.

PR #3869, opened 2025-01-07
fix: hardcoded namespace and add support for multiple listener pods

fed-sv:
this was added in v0.10.0

https://github.com/actions/actions-runner-controller/compare/gha-runner-scale-set-0.9.3...gha-runner-scale-set-0.10.0#diff-fa8c7b8dda021d469814bb2923cb93b10a55a2852b7cd3496a3176e64a7dabe3R195

### DIFF
--- a/.github/actions/execute-assert-arc-e2e/action.yaml
+++ b/.github/actions/execute-assert-arc-e2e/action.yaml
@@ -192,8 +192,11 @@ runs:
       shell: bash
       if: always()
       run: |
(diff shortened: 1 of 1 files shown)


nikola-jokic:
Hey @fed-sv 

Can you please help me understand why is this change needed? There is only one listener per autoscaling runner set. We deployed it on the desired namespace, and install one scale set to that namespace.

fed-sv:
hi @nikola-jokic , there are two changes, the first one it's replacing the hardcoded namespace `arc-systems` with the user entered namespace `${{inputs.arc-controller-namespace}}`.
The other change, it's so that this task works when you have multiple listeners. For example in our workplace we have 5 different listeners. Not sure why this task was assuming only one listener?

PR #3870, opened 2025-01-07
cmd/ghalistener/config: export Validate

hanwen-flow:
Since the `Config` type is fully exported, it can be instantiated directly, in which case the caller will also need the `Validate` functionality.


nikola-jokic (approved):
LGTM, thanks!

PR #3891, opened 2025-01-23
Use Ready from the pod conditions when setting it to the EphemeralRunner

### LINKED ISSUE #3823: EphemeralRunner `status.ready` should reflect Pod readiness
### Checks

### Controller Version

0.9.3

### Deployment Method

Other

### Checks

### To Reproduce

```markdown
N/A
```

### Describe the bug

EphemeralRunner's readiness (`status.ready` field) is calculated as follows.
It becomes `true` when the associated Pod starts running (`status.phase: Running`).

https://github.com/actions/actions-runner-controller/blob/8b36ea90ebe81710fcdcb4f96424b43203d24f1e/controllers/actions.github.com/ephemeralrunner_controller.go#L701

However, EphemeralRunner's readiness should reflect the Pod's readiness (`status.conditions[type="Ready"].status`).
Otherwise, an EphemeralRunner is recognized as ready while its Pod is actually not ready yet.

EphemeralRunners become ready once their associated Pods become ready, not just running.

### Additional Context

```yaml
N/A
```

### Controller Logs

```shell
N/A
```

### Runner Pod Logs

```shell
N/A
```

nikola-jokic:
Fixes #3823


controllers/actions.github.com/ephemeralrunner_controller.go
```diff
@@ -712,22 +712,41 @@ func (r *EphemeralRunnerReconciler) updateRunStatusFromPod(ctx context.Context,
...
 	err := patchSubResource(ctx, r.Status(), ephemeralRunner, func(obj *v1alpha1.EphemeralRunner) {
 		obj.Status.Phase = pod.Status.Phase
-		obj.Status.Ready = obj.Status.Ready || (pod.Status.Phase == corev1.PodRunning)
+		obj.Status.Ready = ready
 		obj.Status.Reason = pod.Status.Reason
 		obj.Status.Message = pod.Status.Message
 	})
 	if err != nil {
-		return fmt.Errorf("failed to update runner status for Phase/Reason/Message: %v", err)
+		return fmt.Errorf("failed to update runner status for Phase/Reason/Message/Ready: %v", err)
```
rentziass, line 746: nit but shouldn't this be `%w`?

rentziass (approved):
LGTM, just left a nit comment 😄

controllers/actions.github.com/ephemeralrunner_controller.go
nikola-jokic: It doesn't make a difference, but this suggestion is great! I think I should create a PR to modify every error in every controller returned to be wrapped instead of `%v`.

Then we can always know we can inspect the underlying error, and the purpose of these private methods are so that the controller loop is cleaner. Again, I really like the suggestion :smile: !

PR #3902, opened 2025-01-29
chore: add OwnerReference to created resources

foyerunix:
Hello,

This is basically an alternative implementation of https://github.com/actions/actions-runner-controller/pull/3575 taking into account the points raised by @nikola-jokic.

It improves upon it by:
- Setting the OwnerReference on all resources created in https://github.com/actions/actions-runner-controller/blob/master/controllers/actions.github.com/resourcebuilder.go
- Avoiding to hardcode `APIVersion` and `Kind`

Please note that I had to move the namespace where we create the `autoscalinglisteners` from the namespace where we run the controller to the namespace where we run the runners. Without this change, the `autoscalinglisteners` cannot be owned by their respective `AutoscalingRunnerSet`. If this is incorrect, please advise on the alternative approach so I can rework my PR accordingly.

It is running in my environment with no issues so far.

Best Regards.


kahirokunn:
duplicates: https://github.com/actions/actions-runner-controller/pull/3575

PR #3917, opened 2025-02-07
Fix typos

omahs:
Fix typos


nikola-jokic:
Hey @omahs,

You probably need to re-generate CRDs since doc-comments are included in them.

PR #3927, opened 2025-02-13
docs: Add verification step after pod deployment

younsl:
docs PR:

- Add verification step after webhook pod deployment


younsl:
Closing due to long inactivity

PR #3729, opened 2024-09-04
Use gha-runner-scale-set-controller.chart instead of .Chart.Version

Mikey032:
Fixes #3728

PR #3961, opened 2025-03-09
fix: oci chart rendering version

buroa:
This fixes a simple issue where upon rendering the chart with OCI manifests adds a +sha to it. This results in a failure to install because it's not a valid kubernetes label with the + in it.

[Other charts](https://github.com/stakater/Reloader/blob/master/deployments/kubernetes/chart/reloader/templates/_helpers.tpl#L25C27-L25C65) replace the + with a _ and that seems to be the standard fix.

```
actions-runner-controller    0.10.1+9f0f5cfb9a0e    False        False    Helm install failed for release actions-runner-system/actions-runner-controller with chart gha-runner-scale-set-controller@0.10.1+9f0f5cfb9a0e: 1 error occurred:
                                                                              * Deployment.apps "actions-runner-controller" is invalid: spec.template.labels: Invalid value: "0.10.1+9f0f5cfb9a0e": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```

### DIFF
--- a/charts/gha-runner-scale-set-controller/templates/deployment.yaml
+++ b/charts/gha-runner-scale-set-controller/templates/deployment.yaml
@@ -25,7 +25,7 @@ spec:
       labels:
         app.kubernetes.io/part-of: gha-rs-controller
         app.kubernetes.io/component: controller-manager
-        app.kubernetes.io/version: {{ .Chart.Version }}
+        app.kubernetes.io/version: {{ .Chart.Version | replace "+" "_" }}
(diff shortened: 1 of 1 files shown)


buroa:
Closing in favor of #3729

PR #3975, opened 2025-03-18
Create configurable metrics

### LINKED ISSUE #3153: High cardinality of metrics
### Checks

### Controller Version

0.7.0

### Deployment Method

Helm

### Checks

### To Reproduce

1. enable metrics to controller by using:

metrics:
  listenerAddr: ":8080"
  listenerEndpoint: "/metrics"

2. you need to have enough high traffic build environment that it will build a lot from different repos and prs

### Describe the bug

Listeners contains metrics which is great. However, the cardinality of these metrics are just something that is not going to work. There needs to be way to disable high cardinality labels from the metrics.

Screenshot 2023-12-14 at 18 19 39
Screenshot 2023-12-14 at 18 19 52

so just during one hour we got over 15k new timeseries to our prometheus which is going to explode if we keep these enabled even 12 hours.

The expected behaviour is that high cardinality labels is REMOVED from the metrics. Also metrics buckets should be configurable in case of bucket metrics.

Lets take the worst metric `gha_job_execution_duration_seconds_bucket`

it do have `job_workflow_ref` which is basically almost always unique, which means that it will always create new timeseries to prometheus which is really expensive. Worth of thinking is even `job_name` needed. I would like to disable both of these labels. Also number of the default buckets are something which is going to explode our prometheus.

### Additional Context

```yaml
replicaCount: 2

flags:
  logLevel: "debug"

metrics:
  listenerAddr: ":8080"
  listenerEndpoint: "/metrics"
```

### Controller Logs

```shell
not relevant
```

### Runner Pod Logs

```shell
not relevant
```

nikola-jokic:
This PR adds configurable metrics to the listener, allowing users to trim down labels, remove metrics and modify histogram buckets.

:warning: The default values.yaml file includes all the metrics. It serves both as a documentation, and as a starting point. However, if no metrics are specified, no metrics will be registered.

One important thing to note is for histograms, if buckets are not specified, the default buckets will be used.

Fixes #3153


cmd/ghalistener/metrics/metrics.go
```diff
@@ -311,20 +168,112 @@ func NewExporter(config ExporterConfig) ServerPublisher {
...
 			Addr:    config.ServerAddr,
 			Handler: mux,
 		},
 	}
 }
 
+var errUnknownMetricName = errors.New("unknown metric name")
+
+func installMetrics(config v1alpha1.MetricsConfig, reg *prometheus.Registry, logger logr.Logger) *metrics {
+	logger.Info(
```
Link-, line 189: 👍😁👍

Link- (approved):
🚀

PR #4005, opened 2025-03-28
Add response body to error when fetching access token

nashluffy:
We spent some time trying to debug why we were getting a 403 from GitHub when the credentials worked fine elsewhere, would be super useful if the body was present in the error.


github/actions/client.go
```diff
@@ -1056,10 +1056,14 @@ func (c *Client) fetchAccessToken(ctx context.Context, gitHubConfigURL string, c
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusCreated {
+		body, err := io.ReadAll(resp.Body)
```
nikola-jokic: If the response status is not created, and if reading the body fails, we would lose the error type `GitHubAPIError` which contains the `RequestID`. This request ID is important because we can troubleshoot the issues related to the request on the back end. 

Maybe something along the lines of:
```go
errMsg := fmt.Sprintf("failed to get access token for GitHub App auth (%v)", resp.Status)
if body, err := io.ReadAll(resp.Body); err == nil {
  errMsg = fmt.Sprintf("%s: %s", errMsg, string(body))
}

return nil, &GitHubAPIError{
  StatusCode: resp.StatusCode,
  RequestID:  resp.Header.Get(HeaderGitHubRequestID),
  Err:        errors.New(errMsg),
}
```

rentziass (approved):

PR #4018, opened 2025-04-04
bugfix: allow DinD image to be configured via values

yashrajdighe:
This PR allows the `DinD` image to be configured via Helm values by checking the container spec for a container named `dind` and using the defined image if present. If no custom image is specified, it defaults to `docker:dind`.

This change is necessary because we're hitting Docker Hub rate limits when pulling the default DinD image. By making the image configurable, we can use a custom image from our own registry, which helps avoid these limitations and improves reliability in CI workflows.

The change is already tested in the cloned chart and is working as expected.

### DIFF
--- a/charts/gha-runner-scale-set/templates/_helpers.tpl
+++ b/charts/gha-runner-scale-set/templates/_helpers.tpl
@@ -89,14 +89,22 @@ image: {{ $val.image }}
 command: ["cp"]
 args: ["-r", "/home/runner/externals/.", "/home/runner/tmpDir/"]
 volumeMounts:
-  - name: dind-externals
+  - name: -externals
     mountPath: /home/runner/tmpDir
   {{- end }}
 {{- end }}
 {{- end }}
 
 {{- define "gha-runner-scale-set.dind-container" -}}
-image: docker:dind
+
+{{- $dindImage := "docker:dind" -}}
+{{- range $i, $val := .Values.template.spec.containers }}
+  {{- if eq $val.name "dind" }}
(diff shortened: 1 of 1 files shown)


yashrajdighe:
@nikola-jokic @mumoshu @toast-gear @rentziass Apologies for tagging everyone, but if possible, could someone review this PR? It's required to resolve the Docker Hub rate limiting issue we're (and many more) currently facing.

anthony-yeong-partior:
Isnt this already configurable through the template spec in the value file? We already using custom registry and image for dind with that.

yashrajdighe:
> Isnt this already configurable through the template spec in the value file? We already using custom registry and image for dind with that.

@anthony-yeong-partior Nope, its not. Its hardcoded to use `docker:dind` image from dockerhub which results in rate limiting.

Reference: https://github.com/actions/actions-runner-controller/blob/f832b0b2544630651f2c7631d045bc36c56cd269/charts/gha-runner-scale-set/templates/_helpers.tpl#L99

nikola-jokic:
Closing this one since we documented that customisations to the dind mode should be applied by leaving the container mode commented out, and applying the spec we provided inside the commented section in values.yaml file.
We are aware that container mode customisations are a pain point, but there is a workaround. 
Unless we support customising all fields, supporting just some fields is error-prone and hard to maintain.
See https://github.com/actions/actions-runner-controller/blob/f832b0b2544630651f2c7631d045bc36c56cd269/charts/gha-runner-scale-set/values.yaml#L91

PR #4033, opened 2025-04-11
Delete config secret when listener pod gets deleted

### LINKED ISSUE #4029: scaleset listener's config is not deleted on cleanup
### Checks

### Controller Version

0.11.0

### Deployment Method

Helm

### Checks

### To Reproduce

```markdown
- have a scaleset with a GHA token
- expire the token
- scaleset listener pod crashes (correct)
- controller deletes listener pod
- listener config is kept in place, with old copy of the secret
- new listener pod is created with as context the new secret, but as the config already exists, it's not replaced
- listener pod crashes again, because token is still invalid
```

### Describe the bug

https://github.com/actions/actions-runner-controller/blob/462db4dfc8f203dab714195aa5d3ae0e13654782/controllers/actions.github.com/autoscalinglistener_controller.go#L296 is never reached (I don't see the logged line in my logs).
So the lines after also never run, which seem to be in charge of deleting the config secret.

Which results into the following, when the token gets refreshed / replaced in the original secret (configured with the `autoscalingrunnerset.spec.githubConfigSecret`):
```bash
$ kubectl -n github-runners get secret test-github-runners-55655b45-listener-config  -o jsonpath="{.data['config\.json']}"|base64 -d|jq .token | md5sum 
c4e76cd6a3e556f2348fa05590baf4a7  -
$ kubectl -n github-runners get secret test-github-runners-55655b45-listener  -o jsonpath={.data.github_token}|base64 -d|md5sum 
91ff6fda9734754b5fed52613bb15d13  -
```

I would expect that if I replace the token in the secret configured in `autoscalingrunnerset.spec.githubConfigSecret`, the listener pod will eventually run with the new token.

### Additional Context

```yaml
.
```

### Controller Logs

```shell
https://gist.github.com/hsmade/9f21b38f9680b7c3a42f690ecb6ddf65
```

### Runner Pod Logs

```shell
https://gist.github.com/hsmade/82d14b70299d878322e420c605733760
```

hsmade:
When the GitHub token gets updated, and the old one expires, the listener pod dies. Only to be recreated with the existing listener config that holds the now-expired token. This PR fixes that by deleting the config secret as well, so the reconciler recreates it with the updated token.

Fixes #4029


controllers/actions.github.com/autoscalinglistener_controller.go
```diff
@@ -260,6 +260,19 @@ func (r *AutoscalingListenerReconciler) Reconcile(ctx context.Context, req ctrl.
...
+			err := r.Get(ctx, types.NamespacedName{Namespace: autoscalingListener.Namespace, Name: scaleSetListenerConfigName(autoscalingListener)}, &configSecret)
+			switch {
+			case err == nil:
+				if configSecret.ObjectMeta.DeletionTimestamp.IsZero() {
+					log.Info("Deleting the listener config secret")
+					if err := r.Delete(ctx, &configSecret); err != nil {
+						return ctrl.Result{}, fmt.Errorf("failed to delete listener config secret: %w", err)
+					}
+				}
+			}
```
nikola-jokic:
```suggestion
			err := r.Get(ctx, types.NamespacedName{Namespace: autoscalingListener.Namespace, Name: scaleSetListenerConfigName(autoscalingListener)}, &configSecret)
			switch {
			case err == nil && configSecret.ObjectMeta.DeletionTimestamp.IsZero():
				log.Info("Deleting the listener config secret")
				if err := r.Delete(ctx, &configSecret); err != nil {
					return ctrl.Result{}, fmt.Errorf("failed to delete listener config secret: %w", err)
				}
			case !kerrors.IsNotFound(err):
			     return ctrl.Result{}, fmt.Errorf("failed to get the listener config secret: %w", err)
			}
```

nikola-jokic (dismissed):
Can you please add this case in the switch and add the controller test :)

nikola-jokic (approved):

rentziass (approved):

PR #4055, opened 2025-04-22
feat: add ability to customize DinD volume mounts

kirillmakhonin-brt:
This PR adds ability to customize DinD volume mounts and CLI arguments. This is needed to be able to resolve issues with MTU (as discussed in [discussions/2993](https://github.com/actions/actions-runner-controller/discussions/2993))

### DIFF
--- a/charts/gha-runner-scale-set/templates/_helpers.tpl
+++ b/charts/gha-runner-scale-set/templates/_helpers.tpl
@@ -113,6 +113,11 @@ volumeMounts:
     mountPath: /var/run
   - name: dind-externals
     mountPath: /home/runner/externals
+  {{- with $.Values.dind.volumeMounts }}
+      {{- range $i, $volMount := . }}
+  - {{ $volMount | toYaml | nindent 4 | trim }}
+      {{- end }}
+  {{- end }}
 {{- end }}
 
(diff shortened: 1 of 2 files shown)


kirillmakhonin-brt:
Please review @mumoshu  @toast-gear  @nikola-jokic  @rentziass

nikola-jokic:
Closing this one since we documented that customisations to the dind mode should be applied by leaving the container mode commented out, and applying the spec we provided inside the commented section in values.yaml file.
We are aware that container mode customisations are a pain point, but there is a workaround. 
Unless we support customising all fields, supporting just some fields is error-prone and hard to maintain.
See https://github.com/actions/actions-runner-controller/blob/f832b0b2544630651f2c7631d045bc36c56cd269/charts/gha-runner-scale-set/values.yaml#L91

kirillmakhonin-brt:
When you mention `Unless we support customising all fields` what do you mean under `all fields` ?

nikola-jokic:
What I mean by that is the entire `dind` container spec. But we already provided the expanded spec that our chart expands when you specify `containerMode: dind`. So you can just uncomment out that part and apply customisations that you want.

PR #4057, opened 2025-04-24
Allow use of client id as an app id

### LINKED ISSUE #3667: Runner Scale Set Auth Secret Does Not Support GH App Client ID
### Checks

### Controller Version

0.9.3

### Deployment Method

Helm

### Checks

### To Reproduce

```markdown
1. Create an auth secret to authenticate a runner scale set, use the GH App's Client ID (not App ID) in the "github_app_id" secret parameter.
2. Attempt to use this secret when deploying the helm chart for a new runner scale set.  Observe the reconciler error stating `"strconv.ParseInt: parsing \"[GH APP CLIENT ID HERE]\": invalid syntax"`.
```

### Describe the bug

While deploying a runner scale set, I attempted to use the GH App client ID rather than the GH App ID.  This produces the following error in the reconciler:

> ERROR Reconciler error {"controller": "autoscalingrunnerset", "controllerGroup": "actions.github.com", "controllerKind": "AutoscalingRunnerSet", "AutoscalingRunnerSet": {"name":"gh-arc-runners","namespace":"arc-runners"}, "namespace": "arc-runners", "name": "gh-arc-runners", "reconcileID": "bad05748-8745-447f-bac3-62d771451fe0", "error": "strconv.ParseInt: parsing \"I[REDACTED FOR SECURITY REASONS}\": invalid syntax"}

However, per [the GH App Docs](https://docs.github.com/en/enterprise-cloud@latest/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app) an App ID and Client ID should both be allowed in the `iss` parameter of a JWT token.  Additionally, GitHub seems to be guiding people to use the Client ID for authenticating GH Apps instead of the App ID (messaging such as:

> Using your App ID to get installation tokens? You can now [use your Client ID instead](https://docs.github.com/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app). 

appears on the app configuration page).

When I run this with an App ID instead of a Client ID the runner scale set deploys as expected.

I'd expect the Client ID to be able to be used when deploying a new Runner Scale Set.

### Additional Context

```yaml
None.
```

### Controller Logs

```shell
See note above about the relevant logs output by the controller.
```

### Runner Pod Logs

```shell
None, since the runner pod never was deployed because of this error.
```

nikola-jokic:
This change allows client ID not to be an integer only, but also to be a client ID (see https://github.blog/changelog/2024-05-01-github-apps-can-now-use-the-client-id-to-fetch-installation-tokens/).

This change is backwards-compatible.

Fixes #3667


cmd/ghalistener/config/config.go
```diff
@@ -18,7 +18,7 @@ import (
 
 type Config struct {
 	ConfigureUrl                string                  `json:"configure_url"`
-	AppID                       int64                   `json:"app_id"`
```
Link-, line 21: We should probably clarify in the values.yaml file that this field can be either the app id or client id

github/actions/multi_client.go
```diff
@@ -23,7 +23,7 @@ type multiClient struct {
 }
 
 type GitHubAppAuth struct {
-	AppID             int64
+	AppID             string
```
Link-, line 27: nit: Do we want to clarify that this could be both AppID or ClientID?

cmd/ghalistener/config/config.go
```diff
@@ -62,26 +62,26 @@ func (c *Config) Validate() error {
...
+		return fmt.Errorf(`MinRunners "%d" cannot be greater than MaxRunners "%d"`, c.MinRunners, c.MaxRunners)
 	}
 
 	hasToken := len(c.Token) > 0
-	hasPrivateKeyConfig := c.AppID > 0 && c.AppPrivateKey != ""
+	hasPrivateKeyConfig := len(c.AppID) > 0 && c.AppPrivateKey != ""
 
 	if !hasToken && !hasPrivateKeyConfig {
-		return fmt.Errorf("GitHub auth credential is missing, token length: '%d', appId: '%d', installationId: '%d', private key length: '%d", len(c.Token), c.AppID, c.AppInstallationID, len(c.AppPrivateKey))
+		return fmt.Errorf(`GitHub auth credential is missing, token length: "%d", appId: %q, installationId: "%d", private key length: "%d"`, len(c.Token), c.AppID, c.AppInstallationID, len(c.AppPrivateKey))
```
Link-, line 81: Do we want to clarify here that we expect an AppID or a ClientID?

Link- (changes requested):
Minor comments

Link- (changes requested):
Minor comments

nikola-jokic:
I pushed a few comments explaining that the app_id can be a client id
I deliberately didn't change the name, so previous charts would still work while extending the functionality.
If the app ID gets deprecated, we can change the field. Otherwise, I wanted to keep minimal changes and at the same time, extend the behavior (since the client id should be used in every place where app id was used previously)

Link- (approved):

PR #4061, opened 2025-04-28
Update ubuntu 22.04 images to have the same software as 20.04

sidneyprins:
This will make sure that `ubuntu22.04` contains the same software as `ubuntu20.04` to make upgrading a breeze.

## Existing difference
- The user / group ID's are different, this may still cause issues.

### DIFF
--- a/runner/actions-runner-dind-rootless.ubuntu-22.04.dockerfile
+++ b/runner/actions-runner-dind-rootless.ubuntu-22.04.dockerfile
@@ -20,17 +20,39 @@ RUN apt-get update -y \
     && add-apt-repository -y ppa:git-core/ppa \
     && apt-get update -y \
     && apt-get install -y --no-install-recommends \
+    build-essential \
     curl \
     ca-certificates \
+    dnsutils \
+    ftp \
     git \
     iproute2 \
+    iputils-ping \
     iptables \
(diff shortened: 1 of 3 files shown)


lucianog-telnyx:
I don't need `build-essential` and many others packages you propose... why mandatory?

sidneyprins:
> I don't need `build-essential` and many others packages you propose... why mandatory?

The proposal was to make `22.04` behave the same as `20.04`, so that you can upgrade without having to worry about breaking your pipeline. For us, the pipeline broke due to these changes, so we had to fix on multiple places due to these mandatory changes

valeriy-soloviov-wd:
Hi @lucianog-telnyx,

We ran into a similar issue recently. Our workaround was to pin our runners to the `summerwind/actions-runner:v2.329.0-ubuntu-20.04` tag, which resolved the problem for now.

It appears the newer `latest` image tag is missing packages that were included in `...-ubuntu-20.04`.

If these packages were removed intentionally (perhaps to create a slimmer base image), would the team consider publishing an additional tag (e.g., `...-full` or `latest-full` or similar) that restores these packages? This would be a great help for teams who have a dependency on them.

jonathanf-eleos (approved):

PR #4080, opened 2025-05-08
Relax version requirements to allow patch version mismatch

nikola-jokic:
This change reduces the version requirements to allow for a patch version mismatch. Previously, the resource version of each `AutoscalingRunnerSet` had to match the build version.

This change allows the patch version mismatch.

### DIFF
--- a/apis/actions.github.com/v1alpha1/version.go
+++ b/apis/actions.github.com/v1alpha1/version.go
@@ -1,7 +1,9 @@
 package v1alpha1
 
+import "strings"
+
 func IsVersionAllowed(resourceVersion, buildVersion string) bool {
-	if buildVersion == "dev" || resourceVersion == buildVersion {
+	if buildVersion == "dev" || resourceVersion == buildVersion || strings.HasPrefix(buildVersion, "canary-") {
 		return true
 	}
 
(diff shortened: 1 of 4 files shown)


apis/actions.github.com/v1alpha1/version.go
```diff
@@ -0,0 +1,70 @@
...
+type semver struct {
+	major string
+	minor string
+}
+
+func parseSemver(v string) (p semver, ok bool) {
+	if v == "" {
+		return
+	}
+	p.major, v, ok = parseInt(v[:])
```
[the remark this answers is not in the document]
Link-: 😄 It is not wrong here

apis/actions.github.com/v1alpha1/version_test.go
```diff
@@ -0,0 +1,60 @@
...
+func TestIsVersionAllowed(t *testing.T) {
+	t.Parallel()
+	tt := map[string]struct {
+		resourceVersion string
+		buildVersion    string
+		want            bool
+	}{
+		"dev should always be allowed": {
+			resourceVersion: "0.11.0",
+			buildVersion:    "dev",
```
Link-, line 19: We probably need to accept `canary-sha` here as well: https://github.com/actions/actions-runner-controller/blob/a1a8dc5606d7d4e5532994f7c4a245354384d8b3/.github/workflows/global-publish-canary.yaml#L132

nikola-jokic: Yeah, I thought this one was covered by the exact match, but I guess I can add `HasPrefix` on `canary-` to allow mismatches in canary deployments

Link- (approved):

PR #4089, opened 2025-05-14
feat(charts): Allow customizing the dind image of the runner-scale-set

pablotp:
# what

* Allow customizing the docker image used for the dind container.

# why

* To download it from different repositories or use a custom image.

# related issues

* https://github.com/actions/actions-runner-controller/issues/3709

### DIFF
--- a/charts/gha-runner-scale-set/templates/_helpers.tpl
+++ b/charts/gha-runner-scale-set/templates/_helpers.tpl
@@ -96,7 +96,7 @@ volumeMounts:
 {{- end }}
 
 {{- define "gha-runner-scale-set.dind-container" -}}
-image: docker:dind
+image: {{ .Values.containerMode.dindImage | default "docker:dind" }}
 args:
   - dockerd
   - --host=unix:///var/run/docker.sock
--- a/charts/gha-runner-scale-set/values.yaml
+++ b/charts/gha-runner-scale-set/values.yaml
@@ -116,6 +116,8 @@ githubConfigSecret:
 ## empty, and configuration should be applied to the template.
 # containerMode:
 #   type: "dind"  ## type can be set to dind or kubernetes
+#   ## Optional. To use a custom image for dind, set the image name here.
+#   dindImage: "docker:dind"
 #   ## the following is required when containerMode.type=kubernetes
 #   kubernetesModeWorkVolumeClaim:
 #     accessModes: ["ReadWriteOnce"]


evilhamsterman:
I think you're the 5th person to submit a PR for this #3645 #3851 #3540 #3394 #3216

pablotp:
> I think you're the 5th person to submit a PR for this #3645 #3851 #3540 #3394 #3216

This seems correct. Nevertheless, those PRs are adding more changes than what this PR does.

gustavoquintao-hotmart:
This will really help us, since docker hub has a very low rate limit for our scenario. Using a dind coming from other source is a necessity.

arohter (approved):
👍   We need this to get around dockerhub rate limiting.

krisdock (approved):

krisdock:
we really need this. docker's rate limiting on this one image is going to mean we can't use this chart until a custom image is supported.

nikola-jokic:
Hey everyone, you don't have to use `containerMode`, you can simply [uncomment](https://github.com/actions/actions-runner-controller/blob/4326693888fa5e1e8591f1d7f9bc1ee5e81d9873/charts/gha-runner-scale-set/values.yaml#L302) the dind spec and modify it

pablotp:
> Hey everyone, you don't have to use `containerMode`, you can simply [uncomment](https://github.com/actions/actions-runner-controller/blob/4326693888fa5e1e8591f1d7f9bc1ee5e81d9873/charts/gha-runner-scale-set/values.yaml#L302) the dind spec and modify it

Hi, thanks for pointing this out! That's what I'm currently working on. The downside of that approach is that future updates to this chart will require reviewing changes in [50+ lines](https://github.com/actions/actions-runner-controller/blob/4326693888fa5e1e8591f1d7f9bc1ee5e81d9873/charts/gha-runner-scale-set/values.yaml#L306-L359) to ensure the configuration still works.

This change aims to specify a different repository for the image while still benefiting from the great defaults.

nikola-jokic:
Hey @pablotp,

I completely understand, and this is a completely valid request. Please don't get me wrong, it would be great to have these things customizable. Ideally, we want to land on a solution that makes it easy for users to enter values.YAML file, as well as for us to maintain these charts and debug issues.

If we start accepting partial customisations, it would be much harder for us to maintain and support. Therefore, it seems like a better way forward to not allow customisations this way, but to provide the spec which can be modified.

pablotp:
> Hey @pablotp,
> 
> I completely understand, and this is a completely valid request. Please don't get me wrong, it would be great to have these things customizable. Ideally, we want to land on a solution that makes it easy for users to enter values.YAML file, as well as for us to maintain these charts and debug issues.
> 
> If we start accepting partial customisations, it would be much harder for us to maintain and support. Therefore, it seems like a better way forward to not allow customisations this way, but to provide the spec which can be modified.

Thanks for the prompt response! Understood 👍🏻 In that case, I will proceed and close this PR.

PR #4101, opened 2025-05-23
Avoid nil point when config.Metrics is nil and expose all metrics if none are configured

### LINKED ISSUE #3993: SIGSEGV after update to 0.11.0 on listener
### Checks

### Controller Version

0.11.0

### Deployment Method

Helm

### Checks

### To Reproduce

```markdown
1. Stay on 0.10.1 on controller and scale set
2. Upgrade to both 0.11.0
3. Check listener logs
4. Revert back to 0.10.1
```

### Describe the bug

During upgrade listener was recreated with the image version 0.11.0 and restarting due to error:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1844383]

goroutine 1 [running]:
github.com/actions/actions-runner-controller/cmd/ghalistener/app.New({{0xc0000465e0, 0x1a}, 0x0, 0x0, {0x0, 0x0}, {0xc000430120, 0x28}, {0xc0003053b0, 0x10}, ...})
    github.com/actions/actions-runner-controller/cmd/ghalistener/app/app.go:73 +0x3e3
main.main()
    github.com/actions/actions-runner-controller/cmd/ghalistener/main.go:27 +0x1f8
Stream closed EOF for gha-runners-controller/k8s-dind-6746d74f-listener (listener)
```
Reverting back to 0.10.1 fixes the issue.
Config wasn't changed during the upgrade.

PS. Controller has no backward compatibility for 0.11.0 version with scale set on 0.10.1. Version should be aligned.

Smooth upgrade from 0.10.1 to 0.11.0

### Additional Context

```yaml
Ubuntu 24.04.1 LTS (GNU/Linux 6.8.0-45-generic x86_64)
k8s version: v1.31.1+rke2r1
```

### Controller Logs

```shell
Not provided
```

### Runner Pod Logs

```shell
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1844383]

goroutine 1 [running]:
github.com/actions/actions-runner-controller/cmd/ghalistener/app.New({{0xc0000465e0, 0x1a}, 0x0, 0x0, {0x0, 0x0}, {0xc000430120, 0x28}, {0xc0003053b0, 0x10}, ...})
    github.com/actions/actions-runner-controller/cmd/ghalistener/app/app.go:73 +0x3e3
main.main()
    github.com/actions/actions-runner-controller/cmd/ghalistener/main.go:27 +0x1f8
Stream closed EOF for gha-runners-controller/k8s-dind-6746d74f-listener (listener)
```

TingluoHuang:
When metrics are configured on the controller, but no metrics are applied to the scale set, the listener would panic with nil dereference error.

This change applies all metrics by default if metrics are configured, but no customisation is being done.

To avoid server errors in the future, defaults are applied to the server address and the default endpoint. This code path should not be hit by default, since the `app.go` decides if it should apply metrics based on the server address. But it made sense to use the default `:8080` port and `/metrics` endpoint in the default function to future-proof the code.

Fixes #3993

### DIFF
--- a/cmd/ghalistener/metrics/metrics.go
+++ b/cmd/ghalistener/metrics/metrics.go
@@ -155,134 +155,137 @@ type ExporterConfig struct {
 	Metrics           *v1alpha1.MetricsConfig
 }
 
+var defaultMetrics = v1alpha1.MetricsConfig{
+	Counters: map[string]*v1alpha1.CounterMetric{
+		MetricStartedJobsTotal: {
+			Labels: []string{
+				labelKeyEnterprise,
+				labelKeyOrganization,
+				labelKeyRepository,
+				labelKeyJobName,
+				labelKeyEventName,
+			},
+		},
+		MetricCompletedJobsTotal: {
+			Labels: []string{
+				labelKeyEnterprise,
+				labelKeyOrganization,
+				labelKeyRepository,
+				labelKeyJobName,
+				labelKeyEventName,
+				labelKeyJobResult,
+			},
+		},
+	},
+	Gauges: map[string]*v1alpha1.GaugeMetric{
+		MetricAssignedJobs: {
+			Labels: []string{
+				labelKeyEnterprise,
+				labelKeyOrganization,
+				labelKeyRepository,
+				labelKeyRunnerScaleSetName,
+				labelKeyRunnerScaleSetNamespace,
+			},
+		},
+		MetricRunningJobs: {
+			Labels: []string{
+				labelKeyEnterprise,
+				labelKeyOrganization,
+				labelKeyRepository,
+				labelKeyRunnerScaleSetName,
+				labelKeyRunnerScaleSetNamespace,
+			},
+		},
+		MetricRegisteredRunners: {
+			Labels: []string{
+				labelKeyEnterprise,
+				labelKeyOrganization,
+				labelKeyRepository,
+				labelKeyRunnerScaleSetName,
+				labelKeyRunnerScaleSetNamespace,
+			},
+		},
+		MetricBusyRunners: {
+			Labels: []string{
+				labelKeyEnterprise,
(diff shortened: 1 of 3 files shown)


cmd/ghalistener/metrics/metrics.go
```diff
@@ -152,13 +152,145 @@ type ExporterConfig struct {
 	ServerAddr        string
 	ServerEndpoint    string
 	Logger            logr.Logger
-	Metrics           v1alpha1.MetricsConfig
+	Metrics           *v1alpha1.MetricsConfig
+}
+
+func (e *ExporterConfig) defaults() {
```
Link-: Can we update the PR description to document the default behaviour please?

cmd/ghalistener/metrics/metrics.go
```diff
@@ -152,13 +152,145 @@ type ExporterConfig struct {
...
+					labelKeyJobResult,
+				},
+				Buckets: defaultRuntimeBuckets,
+			},
+		},
+	}
 }
 
 func NewExporter(config ExporterConfig) ServerExporter {
+	config.defaults()
```
Link-, line 293: Do we want to add some more tests in `metrics_test.go` to cover all the possible configuration options?

cmd/ghalistener/metrics/metrics.go
nikola-jokic: done :relaxed:

cmd/ghalistener/metrics/metrics.go
nikola-jokic: Added few tests :+1:

rentziass (approved):

PR #4142, opened 2025-06-24
Remove check if runner exists after exit code 0

### LINKED ISSUE #4136: EphemeralRunners are stuck in failed state after the job succeeds
### Checks

### Controller Version

0.12.0

### Deployment Method

ArgoCD

### Checks

### To Reproduce

```markdown
Not consistently reproducible, there's a small chance for it to happen whenever a job succeeds.
Just run some workload, then check that no EphemeralRunners are stuck in a failed state.
```

### Describe the bug

After the workload succeeds the controller marks the EphemeralRunner as failed and it doesn't create a new pod, just hangs in a failed state until manually removed. There is always exactly 1 failure with a timestamp: `"status": { "failures": {"<uuid>": "<timestamp>"}}`.

It probably lingers in the Github API a bit longer after the pod dies and the controller treats it as a failure. It calls [deletePodAsFailed](https://github.com/actions/actions-runner-controller/blob/0b2534ebc906c8e14d0affbdccafb922e2a5c2db/controllers/actions.github.com/ephemeralrunner_controller.go#L316) which is what's visible in the log excerpt.

After that it goes into backoff but it is never processed again. After the backoff period elapses there are no further logs available referencing the EphemeralRunner and it remains stuck and unmanaged.

For now we are removing these orphans periodically but the orphans seem to negatively impact CI job startup times.

The runners are eventually removed from Github API because I manually checked them and they were no longer present in Github. Yet the EphemeralRunners remain stuck.

The EphemeralRunner is cleanly removed once Github releases it. It should keep reconciling after the backoff period elapses instead of giving up on it silently.

### Additional Context

```yaml
Using a simple runnerset with DinD mode, cloud Github organization installation (via Github app).
```

### Controller Logs

```shell
https://gist.github.com/Dawnflash/0a3fc1da0f99dfe67fc17b6987821a53
```

### Runner Pod Logs

```shell
Don't have those but the jobs normally succeed. All green in Github.
```

nikola-jokic:
When the runner exits, we check if it still exists within the actions service. If it doesn't, we mark it as finished.

There is a delay between runners being deleted, so the check can rarely return an incorrect result.

This change aims to remove this check completely. The ephemeral runner controller will only react to the exit code of the runner. If the runner exits with the exit code 0, that means that the runner has successfully executed the job.

Fixes #4136

### DIFF
--- a/controllers/actions.github.com/ephemeralrunner_controller.go
+++ b/controllers/actions.github.com/ephemeralrunner_controller.go
@@ -293,28 +293,10 @@ func (r *EphemeralRunnerReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		}
 		return ctrl.Result{}, nil
 
-	default:
-		// pod succeeded. We double-check with the service if the runner exists.
-		// The reason is that image can potentially finish with status 0, but not pick up the job.
-		existsInService, err := r.runnerRegisteredWithService(ctx, ephemeralRunner.DeepCopy(), log)
-		if err != nil {
-			log.Error(err, "Failed to check if runner is registered with the service")
-			return ctrl.Result{}, err
-		}
-		if !existsInService {
-			// the runner does not exist in the service, so it must be done
-			log.Info("Ephemeral runner has finished since it does not exist in the service anymore")
-			if err := r.markAsFinished(ctx, ephemeralRunner, log); err != nil {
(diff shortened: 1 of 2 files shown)


JohnYoungers:
would it be possible to publish a patch release based on this fix? I think everyone currently playing the recurring `delete ephemeralrunner {old-runner}` game would appreciate it

rentziass (approved):

PR #4181, opened 2025-07-15
Update example GitHub URLs in values.yaml to include an example for enterprise account-level runners

corycalahan:
The ARC Docs (such as [Choosing runner destinations](https://docs.github.com/en/enterprise-cloud@latest/actions/tutorials/actions-runner-controller/deploying-runner-scale-sets-with-actions-runner-controller#choosing-runner-destinations)) outlines it is possible to register runners at the enterprise account level and points back to `values.yaml`, but does not currently include an example for this.


adjn:
LGTM!

nikola-jokic (approved):

PR #4186, opened 2025-07-23
Restore runner name label in listener metrics

air-hand:
This PR restores the `runner_name` label to job metrics as an opt-in feature for inspect pod performance issue purpose.
The label can be enabled by configuring it in the metrics configuration.

Also adds unit tests that maintain existing structure and avoid complex mocking.

Origin: https://github.com/orgs/community/discussions/167243


air-hand:
@nikola-jokic 
Hi nikola.
No rush, but would appreciate feedback when convenient plz.

PR #4201, opened 2025-08-08
Requeue if create pod returns already exists error

nikola-jokic:
There are situations when the newly created pod is not yet delivered through the watcher, causing subsequent reconciliations to issue a second create pod event.

This change tests for the condition, logs at the info level, and re-queues the event in 5s.
It is not necessary to do it for secrets since the secrets are not watched; therefore, the `GET` request was being issued directly.

It will reduce the noise in the log while backing off the reconciliation for a bit.


erinhannum:
@nikola-jokic Do we have an updated ETA on when we could expect this to be deployed to prod? Is there anything I can help unblock to expedite? Thanks!
cc: @kevinegordon03

Steve-Glass (approved):

Link- (approved):

otfung:
@nikola-jokic , which version will include this PR#4201? Any ETA? Thanks!
cc: @kevinegordon03, @erinhannum

nikola-jokic:
Hey @otfung,

This will be included in `0.13.0`. The deadline for the next release is 13th of October, but it might be sooner, depending on the progress of the rest of the work planned for the 0.13.0 release

Issue #4189, opened 2025-07-27
[Security]  Improve secureness of dind root containers with "User Namespaces" feature.


YvesEarnix:
### What would you like added?

Because the DinD (Docker-in-Docker) pod runs as root, it should leverage Kubernetes “User Namespaces” for better isolation and security (see https://kubernetes.io/docs/tasks/configure-pod-container/user-namespaces/).

Enabling this feature only requires adding the field _**hostUsers: false**_ to the pod template, for example here:
https://github.com/actions/actions-runner-controller/blob/master/charts/gha-runner-scale-set/templates/autoscalingrunnerset.yaml#L157

### Why is this needed?

User Namespaces are a new Kubernetes security feature that significantly reduces the risk of privilege escalation. Security is a must-have. 😊

### Additional context

• The spec.hostUsers field is available starting with Kubernetes 1.30 (to be confirm)
• Setting _**hostUsers: false**_ activates User Namespaces only if clusters support them; otherwise clusters the field is simply ignored, so compatibility is maintained.

yashrajdighe:
Good suggestion to fix the VULN, I tried enabling this while doing a POC but I am getting below error post configuring `hostUsers: false` and the runner pod was unable to start.

`Warning  FailedCreatePodSandBox  5s (x7 over 90s)  kubelet            Failed to create pod sandbox: the handler "" is not known`

mkushakov:
I will not recommend to enable it yet as default template as not all hosts supports user namespace isolation. Was trying it in EKS v1.33 with `osImage: Amazon Linux 2023.8.20250721` and the error was 
```
Warning  FailedCreatePodSandBox  3s    kubelet            Failed to create pod sandbox: can't set `spec.hostUsers: false`, runtime does not support user namespaces
```
Same for `osImage: Bottlerocket OS 1.44.0 (aws-k8s-1.33)` with user namespace activated by
```
[settings.kernel.sysctl]
"user.max_user_namespaces" = "16384"
```
and error is `[rootlesskit:parent] error: failed to setup UID/GID map: newuidmap 33 [0 1001 1 1 100000 65536] failed: newuidmap: write to uid_map failed: Operation not permitted`

Hi-Fi:
True, it's easier to keep the default as-is for now, but still it would be nice to be able to enable user namespace usage directly from chart.

As mentioned, this would be crucial especially for DinD usage, but also improve security in general.

[...]

PR #4212, opened 2025-08-16
feat: runners can run on user namespace

Hi-Fi:
Partial solution to #4189

### DIFF
--- a/charts/gha-runner-scale-set/templates/autoscalingrunnerset.yaml
+++ b/charts/gha-runner-scale-set/templates/autoscalingrunnerset.yaml
@@ -153,6 +153,9 @@ spec:
       {{ $key }}: {{ $val | toYaml | nindent 8 }}
         {{- end }}
       {{- end }}
+      {{- if kindIs "bool" .Values.template.spec.hostUsers }}
+      hostUsers: {{ .Values.template.spec.hostUsers }}
+      {{- end }}
       {{- if not .Values.template.spec.restartPolicy }}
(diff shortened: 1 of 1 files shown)


charts/gha-runner-scale-set/templates/autoscalingrunnerset.yaml
```diff
@@ -153,6 +153,9 @@ spec:
       {{ $key }}: {{ $val | toYaml | nindent 8 }}
         {{- end }}
       {{- end }}
+      {{- if kindIs "bool" .Values.template.spec.hostUsers }}
+      hostUsers: {{ .Values.template.spec.hostUsers }}
```
nikola-jokic, line 157: Hey, but shouldn't the line above cover this case?
Since the key is not containers, and is not volumes, and is not initContainers, and is not serviceAccountName, it should be applied, right?

If not, I'm wondering why

Hi-Fi: I feel ashamed, but yes, you're right. Additional spec entries are already flowing to template without changes.

Hi-Fi:
This is already possible through template.spec, as only specific keys are handled separately.

charts/gha-runner-scale-set/templates/autoscalingrunnerset.yaml
nikola-jokic: No worries!

PR #4219, opened 2025-08-19
[Helm] Expose dind image

SheldonTsen:
This image is not configurable. This seemed the simplest to expose this to be configurable. 

All other images are configurable but this one. I need to use an internal container registry, and hit this issue. 

Tried my best to come up with a sensible place to put this key and this is what I came up with. 

See issue here: https://github.com/actions/actions-runner-controller/issues/4232

### DIFF
--- a/charts/gha-runner-scale-set/templates/_helpers.tpl
+++ b/charts/gha-runner-scale-set/templates/_helpers.tpl
@@ -96,7 +96,8 @@ volumeMounts:
 {{- end }}
 
 {{- define "gha-runner-scale-set.dind-container" -}}
-image: docker:dind
+image: {{ .Values.dind.image }}
+imagePullPolicy: {{ .Values.dind.imagePullPolicy }}
 args:
   - dockerd
   - --host=unix:///var/run/docker.sock
--- a/charts/gha-runner-scale-set/values.yaml
+++ b/charts/gha-runner-scale-set/values.yaml
@@ -295,6 +295,11 @@ githubConfigSecret:
 #           3600.0,
 #         ]
 
+# if using dind - you can overwrite the image used
+dind:
+  image: docker:dind
+  imagePullPolicy: IfNotPresent
+
 ## template is the PodSpec for each runner Pod
 ## For reference: https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#PodSpec
 template:


SheldonTsen:
Hello @mumoshu @toast-gear @rentziass @nikola-jokic, small bump!

SheldonTsen:
Hello @mumoshu @toast-gear @rentziass @nikola-jokic, small bump again! Thanks!

nikola-jokic:
Hey @SheldonTsen,

As mentioned a few times now, we don't want to create a patch that will simply modify the dind image. The more reasonable path forward would be to rework our helm chart so it is better suited to handle these cases.

[This PR](https://github.com/actions/actions-runner-controller/pull/3645) looks much closer to what we need to do, instead of allowing top-level objects to modify inner objects.

Since modifying dind image is possible, you simply don't specify the containerMode and uncomment out the spec that would be applied if the containerMode is set to dind, I'll close this PR for now.

I admit that the charts could be better designed, but this is not the way to do it. And since it doesn't block users from modifying the spec, I would hesitate before applying the temporary patch.

SheldonTsen:
@nikola-jokic - no problem. Thanks for the response. I should have checked for other PRs first before creating this. 

> Since modifying dind image is possible, you simply don't specify the containerMode and uncomment out the spec that would be applied if the containerMode is set to dind, I'll close this PR for now.

Let me look at this more closely. I thought it was not currently possible. 

Thanks!

nikola-jokic:
No problem! You can see in the template section that we have commented out the spec that would be expanded if you pick the container mode dind for example. You just uncomment out that section and modify the fields you need in order to make it work :relaxed:

SheldonTsen:
Actually, @nikola-jokic, I took a deeper look, it might not be possible.

So in `_helpers.tpl` we have:
```
image: docker:dind
args:
  - dockerd
  - --host=unix:///var/run/docker.sock
  - --group=$(DOCKER_GROUP_GID)
 ...
```

In `autoscalingrunnerset.yaml`:
```
...
      initContainers:
        {{- if eq $containerMode.type "dind" }}
      - name: init-dind-externals
          {{- include "gha-runner-scale-set.dind-init-container" . | nindent 8 }}
          {{- if (ge (.Capabilities.KubeVersion.Minor | int) 29) }}
      - name: dind
          {{- include "gha-runner-scale-set.dind-container" . | nindent 8 }}
          {{- end }}
        {{- end }}
        {{- with .Values.template.spec.initContainers }}
      {{- toYaml . | nindent 6 }}
        {{- end }}
      {{- end }}
...
```

And in `values.yaml` if we do this:
```
   template:
     spec:
       initContainers:
       - name: init-dind-externals
         command: ["cp", "-r", "/home/runner/externals/.", "/home/runner/tmpDir/"]
         volumeMounts:
           - name: dind-externals
       - name: dind
         image: docker:dind
         args:
           - dockerd
           - --host=unix:///var/run/docker.sock
           - --group=$(DOCKER_GROUP_GID)
         ...
```

If I am not mistaken, we will have 4 containers under `initContainers`, 2x `dind` 2x `init-dind-externals`, which I think will throw an error. Because from here:
```
      - name: init-dind-externals
          {{- include "gha-runner-scale-set.dind-init-container" . | nindent 8 }}   #   ---> will give: name: init-dind-externals
          {{- if (ge (.Capabilities.KubeVersion.Minor | int) 29) }}
      - name: dind
          {{- include "gha-runner-scale-set.dind-container" . | nindent 8 }}          # ---> will give: name: dind
          {{- end }}
        {{- end }}
        {{- with .Values.template.spec.initContainers }}      # ---> will give: name: init-dind-externals and name: dind
```

A possible easy fix would be to add a flag to know when to take from `_helpers` vs `values.yaml`:
```
initContainers:
  {{- if .Values.template.spec.manualOverwrite }}
  {{- toYaml .Values.template.spec.initContainers | nindent 2 }}
  {{- else }}
  - name: init-dind-externals
    {{- include "gha-runner-scale-set.dind-init-container" . | nindent 4 }}
  - name: dind
    {{- include "gha-runner-scale-set.dind-container" . | nindent 4 }}
  {{- end }}
```

Or a more implicit version (without a flag):
```
{{- $containerMode := .Values.containerMode }}
{{- $userInit := default (list) .Values.template.spec.initContainers -}}

{{- if or (gt (len $userInit) 0) (eq $containerMode.type "dind") }}
initContainers:
  {{- if gt (len $userInit) 0 }}
  {{- toYaml $userInit | nindent 2 }}
  {{- else if eq $containerMode.type "dind" }}
  - name: init-dind-externals
    {{- include "gha-runner-scale-set.dind-init-container" . | nindent 4 }}
  {{- if (ge (.Capabilities.KubeVersion.Minor | int) 29) }}
  - name: dind
    {{- include "gha-runner-scale-set.dind-container" . | nindent 4 }}
  {{- end }}
  {{- end }}
{{- end }}

```

nikola-jokic:
Please read this comment: https://github.com/actions/actions-runner-controller/blob/0a0be027fd3eb5c5b029f8f3a74b11ab12c48e1e/charts/gha-runner-scale-set/values.yaml#L115

SheldonTsen:
Ah right, I need to disable even more things. Okay, yeah, then if I re-defined that again, then it's possible. Okay - my bad! Apologies! Thanks for the clarification.

Though in my case, it's basically introducing lots of lines, simply to change one line. Which is not the most ideal way to leverage values file where each line is intended to change something underlying. But I understand also the broader context from the other PR mentioning a nicer way to make things more configurable.

Issue #4031, opened 2025-04-10
AutoscalingListener doesn't have an owner reference refering to its autoscalingrunnerset


paul-civitas:
### Checks

- [x] I've already read https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/troubleshooting-actions-runner-controller-errors and I'm sure my issue is not covered in the troubleshooting guide.
- [x] I am using charts that are officially provided

### Controller Version

0.11.0

### Deployment Method

Helm

### Checks

- [x] This isn't a question or user support case (For Q&A and community support, go to [Discussions](https://github.com/actions/actions-runner-controller/discussions)).
- [x] I've read the [Changelog](https://github.com/actions/actions-runner-controller/blob/master/docs/gha-runner-scale-set-controller/README.md#changelog) before submitting this issue and I'm sure it's not due to any recently-introduced backward-incompatible changes

### To Reproduce

```markdown
Deploy ARC using argocd. Then it's visually easy to see that owner references haven't been fully implemented yet.
```

### Describe the bug

The autoscalingrunnerset has an associated autoscalinglistener that's its child.

However the controller isn't creating the autoscalinglistener with owner references to the associated autoscalingrunnerset.

If this were to be implemented, it would make it much more clear which autoscalinginglistener relates to which autoscalingrunnerset and also it would make things look much more clear in argocd.

### Describe the expected behavior

Owner references exist for autoscalinglisteners.

### Additional Context

```yaml
Can be provided upon request.
```

### Controller Logs

```shell
Can be provided upon request.
```

### Runner Pod Logs

```shell
Can be provided upon request.
```

nikola-jokic:
Hey @paul-civitas,

This can't be done properly (see https://github.com/actions/actions-runner-controller/pull/3575#issuecomment-2703191490). One of the requirements we have is to separate the controller's namespace from the installation namespace. The listener exists in the controllers namespace while the autoscaling runner set exists in the installation namespace. Therefore, the owner reference cannot be set that way.

paul-civitas:
Understood. Thanks!

Nastaliss:
@nikola-jokic is merging the two namespaces planned ? 
In my case I'm using the same namespace for the controller and the scale set, is there a workaround for me to make the ownerreferences work for listeners ?

nikola-jokic:
Hey @Nastaliss,

We might change that in future releases, but I cannot say for sure at this time.

[...]

PR #4224, opened 2025-08-21
chore: conditionally add `OwnerReferences` during resource creation for `AutoScalingListener`

jhychan:
Addressing issue #4031 and building upon PR #3575

As per this [comment](https://github.com/actions/actions-runner-controller/pull/3575#issuecomment-2703191490), owner references cannot be cross-namespace. Since the default architecture for ARC is to install the runner scale sets in separate namespaces to the controllers and listeners, the owner reference for AutoScalingListeners cannot be set.

However, ARC can run in single-namespace configuration, which can be the same namespace as the runner scale sets. In this scenario, the ownerReference can be set on the AutoScalingListener.

### DIFF
--- a/controllers/actions.github.com/resourcebuilder.go
+++ b/controllers/actions.github.com/resourcebuilder.go
@@ -140,6 +140,19 @@ func (b *ResourceBuilder) newAutoScalingListener(autoscalingRunnerSet *v1alpha1.
 		},
 	}
 
+	if namespace == autoscalingRunnerSet.Namespace {
+		autoscalingListener.ObjectMeta.OwnerReferences = []metav1.OwnerReference{
+			{
+				APIVersion:         autoscalingRunnerSet.GetObjectKind().GroupVersionKind().GroupVersion().String(),
+				Kind:               autoscalingRunnerSet.GetObjectKind().GroupVersionKind().Kind,
+				UID:                autoscalingRunnerSet.GetUID(),
+				Name:               autoscalingRunnerSet.GetName(),
+				Controller:         boolPtr(true),
+				BlockOwnerDeletion: boolPtr(true),
+			},
+		}
+	}
+
 	return autoscalingListener, nil
 }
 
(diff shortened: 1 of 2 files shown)


enricojonas:
Is it possible to get this approved and merged? Thank you!

MichaelMuz:
@nikola-jokic Can we have this merged? Seems to be finished with multiple issues pointing to it

https://github.com/actions/actions-runner-controller/issues/4265
https://github.com/actions/actions-runner-controller/issues/4031

I am sure there are more too, but the change seems small no? (ran into this issue today)

mfuhrmeisterDM:
This would help me too

vparmeland:
Hi Any news ? @nikola-jokic @Okabe-Junya

controllers/actions.github.com/resourcebuilder.go
Okabe-Junya, line 1: Your base is a bit too old, so could you rebase with fresh master?

controllers/actions.github.com/resourcebuilder.go
```diff
@@ -140,6 +140,19 @@ func (b *ResourceBuilder) newAutoScalingListener(autoscalingRunnerSet *v1alpha1.
 		},
 	}
 
+	if namespace == autoscalingRunnerSet.Namespace {
+		autoscalingListener.ObjectMeta.OwnerReferences = []metav1.OwnerReference{
+			{
+				APIVersion:         autoscalingRunnerSet.GetObjectKind().GroupVersionKind().GroupVersion().String(),
+				Kind:               autoscalingRunnerSet.GetObjectKind().GroupVersionKind().Kind,
```
Okabe-Junya, line 147: FYI: I dropped this in #4475, so would be better to avoid using this

controllers/actions.github.com/resourcebuilder.go
Okabe-Junya: Probably cherry-picking and recreate another one would be easier

Okabe-Junya (commented):
I'm not an official reviewer / GitHub employee, but kindly reviewed.

@nikola-jokic 

This PR would be helpful, so I'd appreciated if you checked this PR

controllers/actions.github.com/resourcebuilder.go
nikola-jokic: I agree, we should use the same approach @Okabe-Junya used

nikola-jokic (changes requested):
Hey @jhychan 

Could you please rebase on the master and apply the suggestion from @Okabe-Junya ?

PR #4252, opened 2025-09-17
Allow setting the MTU of nested Docker daemon in DinD mode

tsuna:
When running DinD it's common to need a lower MTU on the nested Docker bridge.

Use a default of 1450 as it's generally safer than not doing anything, and pretty harmless if done when not needed.

### DIFF
--- a/charts/gha-runner-scale-set/templates/_helpers.tpl
+++ b/charts/gha-runner-scale-set/templates/_helpers.tpl
@@ -101,6 +101,7 @@ args:
   - dockerd
   - --host=unix:///var/run/docker.sock
   - --group=$(DOCKER_GROUP_GID)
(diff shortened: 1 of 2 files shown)


tsuna:
Ping. Please :)

Draco7587 (approved):

Draco7587 (approved):

Draco7587 (approved):

benoit-nexthop:
@Draco7587 you approved this change 3 weeks ago but some workflows remained waiting for approval. Would it be possible to let those run so this PR can merge? Thanks in advance.

nikola-jokic:
Closing since experimental chart now contains the ability to modify each parameter of the dind

PR #4059, opened 2025-04-25
Create backoff mechanism for failed runners and allow re-creation of failed ephemeral runners

nikola-jokic:
Fixes #2721
Fixes #3685
Fixes #3700

PR #4272, opened 2025-10-11
retry ephemeral runner even if pod creation fails

badstreff:
Currently if the pod fails to create, for example if a resource quote is blocking it or some other admission controller then the controller will not retry. 

There is code to handle event when the pod fails here PR #4059 but this doesn't work when the pod fails to actually create

This PR adds a failure count to the EphemeralRunner resource even if the pod fails to create so we back off and try again during reconciliation

I'm also not sure if it would be better to just emit an event that the pod creation failed, keep the ephemeralrunner in pending, and let the reconciler try again - this seems like a better pattern to me but the existing code is here to implement this retry logic so I think keeping it consistent is best


badstreff:
@nikola-jokic I see there is a 0.13.0 PR - are these changes still relevant? It looks like some of the changes in that PR may fix this (but not 100% sure)

nikola-jokic:
Hey @badstreff, I think so too, at least it would fix some of the issues I was able to reproduce. I'm pretty sure that this case is already covered.

badstreff:
So I just tested because I wasn't sure and it looks like the issue is still present, I'll try to get a test written today that covers the behavior. 

So far I have been testing manually by deploying arc and creating a namespace with a highly restrictive resource quota and trying to run jobs. You will see the ephemeral pod enters a failed state and gets stuck there even after the resource quota is removed, you need to manually delete the failed ephemeral runner to get the controller to make new pods

Attached is a SS of the error I see - 

image

badstreff:
@nikola-jokic Added a test case that covers the scenario I'm attempting to resolve, this test fails on the master branch without these changes

badstreff:
@nikola-jokic If you have some time can you review? We ran into this again today

nikola-jokic:
Hey @badstreff,

I'm not sure that this is the right way to fix the issue. If the error is invalid, I'd prefer not retrying. The reason is we don't need to keep hitting the API for something that is not going to succeed.
I created a PR fixing the issue when resource quota is being hit: https://github.com/actions/actions-runner-controller/pull/4305
This can potentially be improved to handle other errors, but I don't think retrying always is the way

nikola-jokic:
Hey @badstreff,

I took another look, if the ephemeral runner phase is set to failed to not be retried. We should introduce different phase names, but for now, phase failed always require manual intervention. This basically means we don't know what the failure is, but it is unlikely that it will self-recover. So you might have to do something manually on the cluster and start over.

badstreff:
Hey @nikola-jokic, thanks for looking into this. Yeah I agree I think a different phase would make sense - maybe we set them as Pending if the reason for a failure is a quota or some other temporary error? Right now when they are marked as failed they are essentially just sitting there until someone manually cleans them up and new jobs are not picked up until this is done.

nikola-jokic:
Exactly
The quota error has been fixed with #4305 and will be part of the next release. I hope it solves some of your issues.

nikola-jokic:
Closing this one since we shouldn't always retry, but thank you @badstreff for being so proactive! If you find any issues down the line that we should retry on, please submit an issue and we should handle retries here probably on case-by-case basis.

PR #4282, opened 2025-10-23
Add support for giving kubernetes mode scaleset service account additional permissions

marcusramberg:
This is useful for instance to allow buildx kubernetes driver to work, of if you use a custom container hook that needs additional permissions like pod watch.

Tested locally with ct against a kind cluster.


marcusramberg:
Any chance of some feedback on this PR? For more background on the use case, we've implemented an alternative container hook in go that uses watch - https://github.com/reMarkable/actions-k8shook

nikola-jokic (approved):
Thank you, this looks great!

PR #4320, opened 2025-11-20
Add ephemeral runner finalizer during creation and check finalizer without requeue

nikola-jokic:
Ephemeral runner first:
1. Adds the cleanup finalizer and returns
2. Adds the actions finalizer and returns
3. Proceeds with the secret

This is unnecessary. Especially during spikes, we would have many events in the queue to get to the running state. This change:
1. Adds finalizers during ephemeral runner creation
2. Checks if finalizers exist (They will exist, but since this is a state machine, I'd prefer not to assume anything)

This change intends to make it faster to get to the token creation as soon as ephemeral runner is created.

### DIFF
--- a/controllers/actions.github.com/ephemeralrunner_controller.go
+++ b/controllers/actions.github.com/ephemeralrunner_controller.go
@@ -154,31 +154,17 @@ func (r *EphemeralRunnerReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, nil
 	}
 
-	if !controllerutil.ContainsFinalizer(ephemeralRunner, ephemeralRunnerFinalizerName) {
-		log.Info("Adding finalizer")
+	addFinalizers := !controllerutil.ContainsFinalizer(ephemeralRunner, ephemeralRunnerFinalizerName) || !controllerutil.ContainsFinalizer(ephemeralRunner, ephemeralRunnerActionsFinalizerName)
+	if addFinalizers {
+		log.Info("Adding finalizers")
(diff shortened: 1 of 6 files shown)


controllers/actions.github.com/ephemeralrunner_controller.go
```diff
@@ -154,31 +154,17 @@ func (r *EphemeralRunnerReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, nil
 	}
 
-	if !controllerutil.ContainsFinalizer(ephemeralRunner, ephemeralRunnerFinalizerName) {
-		log.Info("Adding finalizer")
+	addFinalizers := !controllerutil.ContainsFinalizer(ephemeralRunner, ephemeralRunnerFinalizerName) || !controllerutil.ContainsFinalizer(ephemeralRunner, ephemeralRunnerActionsFinalizerName)
+	if addFinalizers {
```
densto88, line 158: if controllerUtil contains one finalizer but not the other, what happens when it meets this condition and tries to re-add the one it already has again below?

nikola-jokic: If the finalizer is already added, it would return false and not add another one. That why I feel this is okay. Here is the library code:
```go
func AddFinalizer(o client.Object, finalizer string) (finalizersUpdated bool) {
	f := o.GetFinalizers()
	for _, e := range f {
		if e == finalizer {
			return false
		}
	}
	o.SetFinalizers(append(f, finalizer))
	return true
}
```
So if finalizer is found, it won't be set. This code path should not be hit anyway, but I wanted to make sure that `EphemeralRunnerReconciler` knows how to fix itself even if something else updates unwanted fields.

densto88 (approved):

PR #4345, opened 2025-12-23
First changes for igeolise's fork

gergelytraveltime:
Upgrading docker.
Changing default Ubuntu version.
Removing git-lfs (we'll add it on top).
We don't need docker-compose.


gergelytraveltime:
This was a mistake.

PR #4359, opened 2026-01-20
Fix `AcivityId` typo in error strings

caius:
## What?

- [x] Fix typo in error strings
- [x] Update tests

## Why?

Noticed this whilst looking at the logs for a different issue in our cluster and ended up going down a rabbit hole investigating if something was comparing "AcivityId" to "ActivityId" and perhaps that was the error.

Turned out to just be an error string with a typo in, figured I'd update them to stop someone else going down the same rabbit hole in future.

### DIFF
--- a/github/actions/errors.go
+++ b/github/actions/errors.go
@@ -36,7 +36,7 @@ type ActionsError struct {
 }
 
 func (e *ActionsError) Error() string {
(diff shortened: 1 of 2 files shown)


nikola-jokic (approved):
Thanks!

PR #4373, opened 2026-02-07
Introduce experimental chart release

### LINKED ISSUE #3471: `gha-runner-scale-set` chart comment about `containerMode` is wrong and misleading
### Checks

### Controller Version

0.9.1

### Deployment Method

Helm

### Checks

### To Reproduce

```markdown
- Read these lines: https://github.com/actions/actions-runner-controller/blob/49490c4421aa8d58a8eb375fe7a539bdfe28b7a6/charts/gha-runner-scale-set/values.yaml#L76-L77
- Specify `template.spec` while keeping `containerMode.type=kubernetes`
- Everything works
```

### Describe the bug

> If any customization is required for dind or kubernetes mode, containerMode should remain
> empty, and configuration should be applied to the template.

This is plain not true.

If you want kubernetes mode you have to specify `containerMode.type=kubernetes` otherwise you won't get the RBAC setup for you.
You are also pretty much forced to customize the `template` on EKS with because otherwise things do not work by default. (See: https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/troubleshooting-actions-runner-controller-errors#error-access-to-the-path-homerunner_work_tool-is-denied)
Or if you want to add other environment variables to the runner..

Going into the chart you also see that the `template` is honoured even if `containerMode.type` is set.

Proper comment that is not misleading.
Link to relevant documentation if necessary.

### Additional Context

`values.yml` that works in EKS for `containerMode.type=kubernetes`
```yaml
    controllerServiceAccount:
      namespace: ...
      name: ...
    runnerScaleSetName: ...
    githubConfigUrl: ...
    maxRunners: 1
    containerMode:
      type: kubernetes
        storageClassName: "ephemeral-storage"
        resources:
          requests:
            storage: 5Gi
    template:
      spec:
        # See https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/troubleshooting-actions-runner-controller-errors#error-access-to-the-path-homerunner_work_tool-is-denied
        securityContext:
          fsGroup: 123
        containers:
          - name: runner
```

### Controller Logs

```shell
N/A
```

### Runner Pod Logs

```shell
N/A
```

### LINKED ISSUE #3819: Newer versions of Kustomize fail to override githubConfigSecret to String
### Checks

### Controller Version

Chart version 0.9.3

### Deployment Method

Kustomize

### Checks

### To Reproduce

```markdown
Kustomization.yaml

apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

commonLabels:
  app.kubernetes.io/name: arc

helmCharts:
  - name: gha-runner-scale-set-controller
    repo: oci://ghcr.io/actions/actions-runner-controller-charts
    version: 0.9.3
    releaseName: gha-runner-scale-set-controller
    namespace: arc-systems
    includeCRDs: true
  - name: gha-runner-scale-set
    repo: oci://ghcr.io/actions/actions-runner-controller-charts
    version: 0.9.3
    namespace: arc-runners
    valuesInline:
      githubConfigUrl: https://github.com/
      githubConfigSecret: gha-runner-scale-set
      controllerServiceAccount:
        namespace: arc-systems
        name: gha-runner-scale-set-controller-gha-rs-controller
      runnerGroup: foo-bar-baz
      runnerScaleSetName: foo-bar-baz
```

Using Kustomize v.5.4.0 or greater:
`kustomize build --enable-helm`

Will lead to this error:
```
Error: could not merge values: wrong node kind: expected MappingNode but got ScalarNode: node contents:
gha-runner-scale-set
```
```


### Describe the bug

This begins breaking in _Kustomize_ `v.5.4.0` and later.

It seems this could be related to a change in behavior in _Kustomize_ that begins unmarshalling the yaml with Strict typing prior to applying the values and requires the `valuesInLine` types to match what is in values.yaml.
https://github.com/kubernetes-sigs/kustomize/pull/5550

In the resource templates, for the `gha-runner-scale-set` chart, there are expectations made around this value being overridden to a String in order to use your own provided secret.

I would expect to be able to set `githubConfigSecret` to the name of a secret in Kubernetes and have it patched appropriately.

### Additional Context

```yaml
Tested on Kustomize versions:
* v5.2.1 - works as expected
* v5.3.0 - works as expected
* v5.4.0 - patching githubConfigSecret with String value no longer works
* v5.5.0 - patching githubConfigSecret with String value no longer works
```

### Controller Logs

```shell
Not a runtime issue.
```

### Runner Pod Logs

```shell
Not a runtime issue.
```

### LINKED ISSUE #4350: Can't config listeners nodeSelector and tolerations whitout rewriting the whole spec block
### Checks

### Controller Version

0.13.0

### Deployment Method

Helm

### Checks

### To Reproduce

```markdown
Define listenerTemplate spec with only nodeSelector for ex and it don't be templated
```

### Describe the bug

When trying to set nodeSelector and/or tolerations, the helm templating breaks with error
```
listenerTemplate:
  spec:
    nodeSelector:
      dedicated: "compute"
    tolerations:
      - key: "kubernetes.azure.com/scalesetpriority"
        operator: "Equal"
        value: "spot"
        effect: "NoSchedule"
```

`one or more objects failed to apply, reason: AutoscalingRunnerSet.actions.github.com is invalid: spec.listenerTemplate.spec.containers: Required value`

[In this issue,](https://github.com/actions/actions-runner-controller/pull/3947) @nikola-jokic mentioned that we could configure it like this but it's not working because of the templating

Defining nodeSelector, toleration or any other standard k8s fields should be allowed and (a bit better) documented ?

In the controller chart, it's very easy to do so

### Additional Context

```yaml
Example with a subchart of a failing configuration
gha-runner-scale-set:
      listenerTemplate:
        spec:
          nodeSelector:
            dedicated: "compute"
          tolerations:
            - key: "kubernetes.azure.com/scalesetpriority"
              operator: "Equal"
              value: "spot"
              effect: "NoSchedule"
```

### Controller Logs

```shell
N/A since because it's on templating level
```

### Runner Pod Logs

```shell
N/A since because it's on templating level
```

nikola-jokic:
The chart redesign intention is to:
- Make it easier to customize charts
- Allow nicer integration with supported modes, mainly container hooks
- Cover extension of all fields. Controller should be able to run side-car containers, modify pod fields etc.
- Refactor helper functions making them easier to maintain

The new chart comes with one limitation: There is no longer kubernetes resolution searching for the controller service account.

This chart should be released next to the current one. In case there is a bug in the chart, users should have a way to keep using the stable version of the chart while we are working on the fix.

Fixes #3819
Fixes #4350
Fixes #3471


.github/workflows/gha-validate-chart.yaml
```diff
@@ -18,7 +18,7 @@ on:
   workflow_dispatch:
 env:
   KUBE_SCORE_VERSION: 1.16.1
-  HELM_VERSION: v3.17.0
+  HELM_VERSION: v3.19.4
```
Steve-Glass, line 21: Out of curiosity, why not upgrade to the latest stable version?

charts/gha-runner-scale-set-controller-experimental/Chart.yaml
```diff
@@ -0,0 +1,33 @@
...
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.13.1
```
Steve-Glass, line 18: Would the chart version be 0.14.0?

nikola-jokic: I can bump it when during the release, but I can bump it now. I just used this for local testing so I don't need to re-build the image with 0.14.0 version to test charts.

.github/workflows/gha-validate-chart.yaml
nikola-jokic: We tend to pin the version for all releases but we can change this if you want :relaxed:

charts/gha-runner-scale-set-experimental/values.yaml
```diff
@@ -0,0 +1,409 @@
...
+      initContainers: []
+      volumes: []
+
+  # container field is applied to the container named "runner". You cannot override the name of the runner container
+  container:
+    image: "ghcr.io/actions/actions-runner:latest"
+    command: ["/home/runner/run.sh"]
+
+  dind:
+    copyExternals: true
```
kenmuse: Not sure customers will understand this or whether it's required. If we always need it, then could leave it off. Otherwise, we might want a name like `configureDindForActions` that makes it more clear what the purpose is.

charts/gha-runner-scale-set-experimental/values.yaml
```diff
@@ -0,0 +1,409 @@
...
+    dockerGroupId: "123"
+    dockerSock: "unix:///var/run/docker.sock"
+    waitForDockerInSeconds: 120
+    container:
+      image: "docker:dind"
+
+  kubernetesMode:
+    serviceAccountName: ""
+    hookPath: "/home/runner/k8s/index.js"
+    requireJobContainer: true
```
kenmuse, line 207: This one should have a big warning if it's exposed. Changing this substantially reduces the security of the environment by giving the runner access to the K8S APIs. Most customers don't understand the consequence of changing this today. They just understand that doing that means they don't have to use a job container.

charts/gha-runner-scale-set-experimental/templates/_mode_dind.tpl
kenmuse, line 1: Really like the organization of the charts! Much more readable and provides opportunities for further "modes" in the future.

charts/gha-runner-scale-set-experimental/values.yaml
```diff
@@ -0,0 +1,409 @@
...
+    image: "ghcr.io/actions/actions-runner:latest"
+    command: ["/home/runner/run.sh"]
+
+  dind:
+    copyExternals: true
+    dockerGroupId: "123"
+    dockerSock: "unix:///var/run/docker.sock"
+    waitForDockerInSeconds: 120
+    container:
+      image: "docker:dind"
```
kenmuse, line 197: It looks like you have support for volumes and environment variables for the runner. Are those supported for configuring the DinD container as well?

kenmuse (commented):
Really like the direction on this one!

charts/gha-runner-scale-set-experimental/values.yaml
nikola-jokic: We should, good call!

Steve-Glass (approved):

rentziass (approved):

PR #4377, opened 2026-02-11
feat: add default linux nodeSelector to listener pod

gateixeira:
The listener is a Linux-only binary. In mixed-OS clusters (e.g. with Windows node pools), the listener pod can be scheduled on a Windows node where it fails with a platform mismatch.

Issue #2652 was closed by #2758 which gave users a workaround via listenerTemplate, but the default behavior still allows the listener to be scheduled on Windows nodes. This PR adds a safe default (kubernetes.io/os: linux) as a defensive approach so users don't need to explicitly configure listenerTemplate just to prevent the listener from landing on a non-Linux node.

Changes:
- Add default nodeSelector kubernetes.io/os: linux to listener pod spec
- Define LabelKeyKubernetesOS constant alongside existing k8s labels
- Make mergeListenerPodWithTemplate preserve the default when the listenerTemplate does not explicitly set a nodeSelector (nil check)
- Add unit tests for all nodeSelector merge scenarios

Refs: #2652


nikola-jokic (approved):
LGTM

Issue #4397, opened 2026-03-06
listener: stale TotalAssignedJobs from GitHub Actions service causes permanent over-provisioning after platform incidents


jameshounshell:
**Describe the bug**

After a GitHub Actions platform incident, the `TotalAssignedJobs` value in `RunnerScaleSetStatistic` gets stuck at an inflated count that never recovers. The listener passes this value directly to the scaling calculation, causing the controller to permanently over-provision runners for phantom demand. The only fix is deleting the AutoscalingRunnerSet CR — listener pod restarts don't clear the state because it comes from the GitHub Actions service, not the listener.

We're running ARC 0.13.1 on GKE with an org-level runner group serving ~100 repos.

**How we tracked it down**

Started with Prometheus. The listener metrics were showing a persistent gap between `gha_assigned_jobs` and `gha_running_jobs`:

| Metric | Value | Source |
|--------|-------|--------|
| `gha_assigned_jobs` | 34 | Listener (`TotalAssignedJobs`) |
| `gha_running_jobs` | 19 | Listener (`TotalRunningJobs`) |
| `gha_desired_runners` | 39 | Listener (calculated) |
| `gha_busy_runners` | 19 | Listener (`TotalBusyRunners`) |
| `gha_idle_runners` | 0 | Listener (`TotalIdleRunners`) |
| `gha_min_runners` | 5 | Config |
| `gha_max_runners` | 80 | Config |

The gap — `assigned(34) - running(19) = 15` stale jobs — was the first clue. The listener was inflating `desired_runners` to 39 instead of the expected ~24 (19 busy + 5 minRunners).

This wasn't a transient burst. Looking at the `desired_runners - busy_runners` gap over time:

- **Pre-incident (72-48h prior):** gap was 5-8 during work hours, dropping to 5 (minRunners) overnight. Normal.
- **Post-incident (last 24h):** gap jumped to 16-25 and stayed there permanently, never dropping below ~16 even overnight when real job volume was near zero.

Grafana dashboard showing the 48h timeline across runners (busy/idle/desired/max), jobs (assigned/running), and job results (cancelled/failed/succeeded). The persistent divergence between `assigned` and `running` in the Jobs panel is the stale state, and the elevated `desired` in the Runners panel is the resulting over-provisioning:

<img width="1597" height="1032" alt="Image" src="https://github.com/user-attachments/assets/d468cd42-dbe0-4418-b9bb-ef9d2b9ae870" />

[...]

Issue #4396, opened 2026-03-06
ARC runners did not recover automatically after Github Outage


rob-howie-depop:
### Checks

- [x] I've already read https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/troubleshooting-actions-runner-controller-errors and I'm sure my issue is not covered in the troubleshooting guide.
- [x] I am using charts that are officially provided

### Controller Version

0.13.1

### Deployment Method

Helm

### Checks

- [x] This isn't a question or user support case (For Q&A and community support, go to [Discussions](https://github.com/actions/actions-runner-controller/discussions)).
- [x] I've read the [Changelog](https://github.com/actions/actions-runner-controller/blob/master/docs/gha-runner-scale-set-controller/README.md#changelog) before submitting this issue and I'm sure it's not due to any recently-introduced backward-incompatible changes

### To Reproduce

```markdown
This is not reproducible directly as it is the byproduct of an incident.
see https://www.githubstatus.com/incidents/g9j4tmfqdd09
```

### Describe the bug

During the GitHub Actions degraded availability incident on 2026-03-05 [githubstatus.com/incidents/g9j4tmfqdd09](https://www.githubstatus.com/incidents/g9j4tmfqdd09)), our ARC deployment experienced runners that became stuck in a bad state and were **not automatically recovered** after GitHub's services came back online.

These runners were registered before the incident, lost their registration during GitHub's degradation, and ARC never reconciled them back to a healthy state.

This was resolved by manually deleting all of the stuck ARC runner pods.

### Describe the expected behavior

ARC has **no garbage collection loop** that reconciles GitHub-side runner registrations against actual Kubernetes pod state. The `EphemeralRunnerReconciler` only handles the forward path (create secret -> create pod -> monitor pod). It does not:

1. Periodically verify that a runner's GitHub registration is still valid
2. Detect runners whose registration was invalidated by a GitHub-side incident
3. Clean up and re-provision runners that are running but no longer recognized by GitHub

I suggest that we should add periodic health check to the `EphemeralRunnerReconciler` that verifies the GitHub-side registration is still valid for running EphemeralRunners.

### Additional Context

```yaml
1. GitHub's `api.github.com/actions/runner-registration` endpoint became unavailable.
2. The ARC `EphemeralRunnerReconciler` failed to generate JIT configs for new runners, retrying 5 times with backoff before giving up.
[...]

Issue #4395, opened 2026-03-06
GHA Self-hosted runner pods are running but the runner status is showing offline.


bhat39a:
### Checks

- [x] I've already read https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/troubleshooting-actions-runner-controller-errors and I'm sure my issue is not covered in the troubleshooting guide.
- [x] I am using charts that are officially provided

### Controller Version

0.13.1

### Deployment Method

Helm

### Checks

- [x] This isn't a question or user support case (For Q&A and community support, go to [Discussions](https://github.com/actions/actions-runner-controller/discussions)).
- [x] I've read the [Changelog](https://github.com/actions/actions-runner-controller/blob/master/docs/gha-runner-scale-set-controller/README.md#changelog) before submitting this issue and I'm sure it's not due to any recently-introduced backward-incompatible changes

### To Reproduce

```markdown
1. Login to the k8s cluster and can see long running runner pods
2. Checked the pods, runner container logs and can see no movement.
3. Go to the Runners section of the repository and there we can see offline runners
```

### Describe the bug

We are running GitHub actions runners in a self-hosted mode and using AWS EKS to host the runners. We have also Istio configured in our k8s cluster to take care of the egress traffic routing. 

I can see some of the runner pods running for more than 3-4 hours but are showing as "offline" in our repository. When I can check the runner container logs its seems to be not doing anything at all. 

```
<runner-label>-d2v4v-runner-zrp5g   3/3     Running   0               7h40m
<runner-label>-d2v4v-runner-zkpgm   3/3     Running   0               5h4m
<runner-label>-d2v4v-runner-jd78x   3/3     Running   0               5h4m
<runner-label>-d2v4v-runner-qq274   3/3     Running   0               5h3m
<runner-label>-d2v4v-runner-2tvmq   3/3     Running   0               5h
<runner-label>-d2v4v-runner-p7fhw   3/3     Running   0               4h59m
<runner-label>-d2v4v-runner-9wbfw   3/3     Running   0               4h59m
<runner-label>-d2v4v-runner-h5s42   3/3     Running   0               4h58m
<runner-label>-d2v4v-runner-pdkv5   3/3     Running   0               4h45m
<runner-label>-d2v4v-runner-jpncv   3/3     Running   0               4h43m
```

### Describe the expected behavior

Runner pods should be cleared if they are offline.

### Additional Context

```yaml
N/A
```

[...]

Issue #4155, opened 2025-06-27
EphemeralRunner and its pods left stuck Running after runner OOMKILL


kennedy-whytech:
### Checks

- [x] I've already read https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/troubleshooting-actions-runner-controller-errors and I'm sure my issue is not covered in the troubleshooting guide.
- [x] I am using charts that are officially provided

### Controller Version

v0.12.1

### Deployment Method

ArgoCD

### Checks

- [x] This isn't a question or user support case (For Q&A and community support, go to [Discussions](https://github.com/actions/actions-runner-controller/discussions)).
- [x] I've read the [Changelog](https://github.com/actions/actions-runner-controller/blob/master/docs/gha-runner-scale-set-controller/README.md#changelog) before submitting this issue and I'm sure it's not due to any recently-introduced backward-incompatible changes

### To Reproduce

```markdown
1. Create an AutoScalingRunnerSet targeting an arm64 runner with limited memory.
2. Trigger a GitHub Actions job from a repository that consumes high memory during startup.
3. The pod will get OOMKilled, but:
- It remains in a Running state.
- The controller does not detect the failure.
- The EphemeralRunner CRD is not deleted.
4. Observe that new jobs remain stuck in the queued state due to the zombie runner.
```

### Describe the bug

Ephemeral runner pods that are OOMKilled do not get properly cleaned up by the controller. Although the pod is no longer functioning (due to OOMKilled), it stays in a Running state and the associated EphemeralRunner CRD is not removed. This leads to zombie runners that block new job assignments, since the controller believes an active runner is still available.

(In v0.12.0, at least it's easier to detect the killed pod because the EphemeralRunner will be left without any pods )

### Describe the expected behavior

When an ephemeral runner pod is OOMKilled, the controller should detect the failure, mark the associated EphemeralRunner CRD as failed, clean up the pod, and (optional)recreate a new runner if needed. This ensures no stale CRDs or zombie runners block new job assignments.

### Additional Context

```yaml
githubConfigUrl: https://github.com/<REDACTED>

controllerServiceAccount:
  namespace: arc-system
  name: arc-controller-gha-rs-controller

githubConfigSecret:
  github_app_id: <REDACTED>
  github_app_installation_id: <REDACTED>
  github_app_private_key: <REDACTED>

containerMode:
  type: "dind"

minRunners: 4

runnerGroup: "k8s"

[...]

PR #4399, opened 2026-03-09
fix: add health check to detect and recover stale EphemeralRunner registrations

rob-howie-depop:
## Summary

Adds a GitHub-side registration health check to the `EphemeralRunnerReconciler`. During reconciliation of a running EphemeralRunner that has a valid `RunnerId`, the controller now calls `GetRunner()` against the GitHub Actions API to verify the registration still exists. If the API returns 404 (registration gone), the runner is marked as failed so the `EphemeralRunnerSet` can provision a replacement.

This also fixes `deleteRunnerFromService` to tolerate 404 from `RemoveRunner`, since a runner whose registration was already invalidated will also 404 on removal. Previously this caused `markAsFailed` to error and trigger a requeue loop, preventing cleanup of stale runners.

A redundant pod phase check was removed from the health check path — `checkRunnerRegistration` is only called from within the `cs.State.Terminated == nil` branch (container confirmed running), so the phase check was unnecessary and could cause false negatives since the local `EphemeralRunner` object may not have the updated phase yet.

## Changes

- **`ephemeralrunner_controller.go`**: New `checkRunnerRegistration()` method that calls `GetRunner()` and returns unhealthy only on confirmed 404. Transient API errors (500, timeouts, etc.) are logged and ignored to avoid false positives. Called during reconciliation when a running pod has a non-zero `RunnerId`.
- **`ephemeralrunner_controller.go`**: `deleteRunnerFromService()` now treats 404 from `RemoveRunner` as success, since the runner is already gone.
- **`ephemeralrunner_controller_test.go`**: Two new test cases — one verifying a 404 from `GetRunner` marks the runner as failed, another verifying a 500 does not. Both simulate a running pod container status before the health check triggers.
- **`fake/client.go`**: New `WithRemoveRunnerError` option for the fake client to support testing the 404 removal path.

### Directly fixes

- **#4396** — *ARC runners did not recover automatically after GitHub Outage*: Runners that lost their GitHub-side registration during the [2026-03-05 GitHub Actions incident](https://www.githubstatus.com/incidents/g9j4tmfqdd09) remained stuck indefinitely because the `EphemeralRunnerReconciler` never verified registration validity. This change adds exactly that verification — on each reconciliation of a running runner, `GetRunner()` confirms the registration exists. If it returns 404, the runner is marked failed and replaced.

- **#4395** — *GHA Self-hosted runner pods are running but the runner status is showing offline*: Runner pods were running for 5-7+ hours showing "Listening for Jobs" but marked offline in GitHub. The pods' registrations were invalidated server-side (same incident), but the controller had no mechanism to detect this. The new health check detects the 404 and triggers cleanup, preventing these zombie runners from accumulating.

### Partially addresses

- **#4397** — *Stale TotalAssignedJobs causes permanent over-provisioning after platform incidents*: This issue has two components: (1) zombie runner pods occupying slots, and (2) the listener's `TotalAssignedJobs` remaining inflated. This PR fixes component (1) — by detecting and cleaning up runners with invalidated registrations, the controller stops accumulating zombie pods. However, the listener-side `TotalAssignedJobs` inflation (a separate code path in `listener.go` → `worker.go`) is not addressed by this change and still requires either a session reconnect mechanism or CR deletion to clear.

- **#4307** — *Ephemeral Runners seem to get stuck when job is canceled or interrupted*: Reports runners getting stuck with `Registration <uuid> was not found` errors in BrokerServer backoff loops after job cancellation. The health check would detect that the registration is gone (404 from `GetRunner`) and mark these runners as failed rather than leaving them in an infinite backoff loop.

- **#4155** — *EphemeralRunner and its pods left stuck Running after runner OOMKill*: Runners that are OOMKilled can end up in a state where the pod is technically running but the runner process is non-functional, and the registration may become stale. The health check provides a secondary detection mechanism — if the GitHub-side registration is invalidated for a non-functional runner, it will be caught and cleaned up.

- **#3821** — *AutoscalingRunnerSet gets stuck thinking runners exist when they do not*: Reports the AutoscalingRunnerSet believing runners exist when no pods are present, requiring CR deletion to recover. While the root cause may vary, stale registrations that the controller cannot clean up (due to `RemoveRunner` 404 errors causing requeue loops) are one contributing factor. The 404 tolerance in `deleteRunnerFromService` directly addresses this cleanup path.

### DIFF
--- a/controllers/actions.github.com/constants.go
+++ b/controllers/actions.github.com/constants.go
@@ -75,8 +75,9 @@ const DefaultScaleSetListenerLogFormat = string(logging.LogFormatText)
 // ownerKey is field selector matching the owner name of a particular resource
 const resourceOwnerKey = ".metadata.controller"
 
-// EphemeralRunner pod creation failure reasons
+// EphemeralRunner failure reasons
 const (
-	ReasonTooManyPodFailures = "TooManyPodFailures"
-	ReasonInvalidPodFailure  = "InvalidPod"
+	ReasonTooManyPodFailures       = "TooManyPodFailures"
+	ReasonInvalidPodFailure        = "InvalidPod"
+	ReasonRegistrationInvalidated  = "RegistrationInvalidated"
 )
--- a/controllers/actions.github.com/ephemeralrunner_controller.go
+++ b/controllers/actions.github.com/ephemeralrunner_controller.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/actions/actions-runner-controller/apis/actions.github.com/v1alpha1"
@@ -44,12 +45,22 @@ const (
 	ephemeralRunnerActionsFinalizerName = "ephemeralrunner.actions.github.com/runner-registration-finalizer"
 )
 
+const defaultHealthCheckInterval = 5 * time.Minute
+
 // EphemeralRunnerReconciler reconciles a EphemeralRunner object
 type EphemeralRunnerReconciler struct {
 	client.Client
 	Log    logr.Logger
 	Scheme *runtime.Scheme
 	ResourceBuilder
+
+	// HealthCheckInterval controls how often the GitHub registration
+	// health check runs per runner. Defaults to defaultHealthCheckInterval.
+	HealthCheckInterval time.Duration
+
+	// healthCheckLastRun tracks the last time each runner's registration
+	// was checked, keyed by "namespace/name".
+	healthCheckLastRun sync.Map
 }
 
 // precompute backoff durations for failed ephemeral runners
@@ -82,6 +93,9 @@ func (r *EphemeralRunnerReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 
 	ephemeralRunner := new(v1alpha1.EphemeralRunner)
 	if err := r.Get(ctx, req.NamespacedName, ephemeralRunner); err != nil {
+		if client.IgnoreNotFound(err) == nil {
+			r.healthCheckLastRun.Delete(req.Namespace + "/" + req.Name)
+		}
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
@@ -357,16 +371,19 @@ func (r *EphemeralRunnerReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 			return ctrl.Result{}, err
 		}
 
-		// Check if runner's GitHub registration is still valid
-		healthy, err := r.checkRunnerRegistration(ctx, ephemeralRunner, log)
-		if err != nil {
-			return ctrl.Result{}, err
-		}
-		if !healthy {
-			if err := r.markAsFailed(ctx, ephemeralRunner, "Runner registration no longer exists on GitHub", "RegistrationInvalidated", log); err != nil {
+		// Only check registration when the container is actually running
+		// (not in Waiting states like ImagePullBackOff or CrashLoopBackOff)
+		if cs.State.Running != nil {
+			healthy, err := r.checkRunnerRegistration(ctx, ephemeralRunner, log)
+			if err != nil {
 				return ctrl.Result{}, err
 			}
-			return ctrl.Result{}, nil
+			if !healthy {
+				if err := r.markAsFailed(ctx, ephemeralRunner, "Runner registration no longer exists on GitHub", ReasonRegistrationInvalidated, log); err != nil {
+					return ctrl.Result{}, err
+				}
+				return ctrl.Result{}, nil
+			}
 		}
 
 		return ctrl.Result{}, nil
@@ -817,13 +834,29 @@ func (r *EphemeralRunnerReconciler) updateRunStatusFromPod(ctx context.Context, 
 	return nil
 }
 
+func (r *EphemeralRunnerReconciler) healthCheckIntervalOrDefault() time.Duration {
+	if r.HealthCheckInterval > 0 {
+		return r.HealthCheckInterval
+	}
+	return defaultHealthCheckInterval
+}
+
 // checkRunnerRegistration verifies the runner's GitHub-side registration is still valid.
 // Returns true if the runner is healthy or we can't determine status (API errors).
 // Returns false if the registration is confirmed gone (404).
+// Calls are throttled to at most once per HealthCheckInterval per runner.
 func (r *EphemeralRunnerReconciler) checkRunnerRegistration(ctx context.Context, ephemeralRunner *v1alpha1.EphemeralRunner, log logr.Logger) (bool, error) {
 	// Only check runners that have been registered (have a RunnerId)
 	if ephemeralRunner.Status.RunnerId == 0 {
 		return true, nil
+	}
+
+	// Throttle: skip if we checked this runner recently
+	key := ephemeralRunner.Namespace + "/" + ephemeralRunner.Name
+	if lastRun, ok := r.healthCheckLastRun.Load(key); ok {
+		if time.Since(lastRun.(time.Time)) < r.healthCheckIntervalOrDefault() {
+			return true, nil
+		}
 	}
 
 	actionsClient, err := r.GetActionsService(ctx, ephemeralRunner)
@@ -833,6 +866,7 @@ func (r *EphemeralRunnerReconciler) checkRunnerRegistration(ctx context.Context,
 	}
 
 	_, err = actionsClient.GetRunner(ctx, int64(ephemeralRunner.Status.RunnerId))
+	r.healthCheckLastRun.Store(key, time.Now())
 	if err == nil {
 		return true, nil
 	}
@@ -847,7 +881,7 @@ func (r *EphemeralRunnerReconciler) checkRunnerRegistration(ctx context.Context,
 	}
 
 	// For non-404 errors (transient API issues), don't take action
-	log.Info("Health check API call failed, will retry on next reconciliation",
+	log.Info("Health check API call failed, will retry after health check interval",
 		"runnerId", ephemeralRunner.Status.RunnerId,
 		"error", err.Error(),
 	)
--- a/controllers/actions.github.com/ephemeralrunner_controller.go
+++ b/controllers/actions.github.com/ephemeralrunner_controller.go
@@ -356,6 +356,19 @@ func (r *EphemeralRunnerReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 			log.Info("Failed to update ephemeral runner status. Requeue to not miss this event")
 			return ctrl.Result{}, err
 		}
+
+		// Check if runner's GitHub registration is still valid
+		healthy, err := r.checkRunnerRegistration(ctx, ephemeralRunner, log)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		if !healthy {
+			if err := r.markAsFailed(ctx, ephemeralRunner, "Runner registration no longer exists on GitHub", "RegistrationInvalidated", log); err != nil {
+				return ctrl.Result{}, err
+			}
+			return ctrl.Result{}, nil
+		}
+
 		return ctrl.Result{}, nil
 
 	case cs.State.Terminated.ExitCode != 0: // failed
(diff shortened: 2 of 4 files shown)


nikola-jokic:
Hey @rob-howie-depop,

Thank you for creating this PR but this is not the way we should approach the issue. The original issue is that the runner doesn't exit. If we were to check the runner on each reconcile loop, you would quickly get rate limited on a large enough scale. The source of the issue is that the runner doesn't exit. Therefore, the runner should dictate how we should react, not the controller.

rob-howie-depop:
Hey @nikola-jokic - 
> The original issue is that the runner doesn't exit

Which specific issue are you talking about? There are several linked and mine had runners that were running fine, with no github registration - hence could never have jobs scheduled on them. This, coupled with some poorly implemented scaling logic within ARC meant that I had stale runners which could never receive jobs, simultaneously to not being able to scale. This deadlocked my CI, and based on the issues raised post the Github incident, other peoples' CIs

> If we were to check the runner on each reconcile loop, you would quickly get rate limited on a large enough scale.

This is a valid point and something I was considering. The default sync period is ~1m, so this is an extra 60 API calls per runner. With a limit of 5,000 requests/hour we do quickly run into that issue. Sadly github provide no batch endpoint to get all runner registrations, so we will need to implement a time gate of some kind to avoid blowing up the API rate limit. Once every 10 mins could work? This would balance the rate of API calls with the ability of a runners to recover after the next github outage 

I am open to other suggestions, but we need to fix this issue. ARC needs to be resilient to github outages given how common these have been in the past couple of years.

I am happy to do the work to implement & test the PR if you can give me some direction on how best to implement this?

nikola-jokic:
> There are several linked and mine had runners that were running fine, with no github registration 

This should not occur. This is the point I'm trying to bring. The runners should not run without registration. They should start, register, and talk to the API. If they can't, they should try to recover. If they can't, they should exit. It is a simple contract between the controller and the runner. That is why I've said the runner should exit here, which would allow controller to properly react. Checking the state with the API is simply not a good path forward.

May I ask, what do you mean by "some poorly implemented scaling logic within ARC"?

rob-howie-depop:
> some poorly implemented scaling logic within ARC

I guess this is another way to view the incident - we had one runner out of 50 which had an active registration (so could accept jobs), the rest were functionally dead (were not accepting jobs as they were not registered with Github) but existing from the viewpoint of the ARC controller. In my mind ARC should scale 50 -> upwards since there were jobs pending (even if a number of of existing jobs are idle, from its view), but as i understand it, it was comparing the queue size of pending jobs (on the Github side) to the number of available runners (from its view) so didn't scale as there were less jobs than available runners.

There is another perspective here on what this looked like (much better write up)
https://github.com/actions/actions-runner-controller/issues/4397

> This should not occur. This is the point I'm trying to bring. The runners should not run without registration. They should start, register, and talk to the API. If they can't, they should try to recover. If they can't, they should exit. It is a simple contract between the controller and the runner. 

Yep agreed on this. The runners should not run without registration.

> That is why I've said the runner should exit here, which would allow controller to properly react. Checking the state with the API is simply not a good path forward.

 This is where I am a bit confused - we can move the logic to verify that the registration is still valid into the container (either directly or by proxy) but its just that - shifting the polling of github into the container. Does that make sense?

Remember we are talking about when a container starts, registers, and talks to the API, then the registration goes missing AFTER its started, in our case this happened due to a github incident

sethjackson:
Wanted to chime in on this thread as well.

We too have experienced similar issues with running ARC. Regardless of whether caused by a GitHub incident or other reasons sometimes the runners do lose their registration and the pods do not terminate which basically means the cluster fills up with unusable runner pods and will eventually cause jobs to not schedule at all (what I like to call a catastrophic queuing issues) as cluster resource constraints (CPU, memory) are reached. Currently, we run out of band automation that greps the pod logs and kills the underlying `ephemeralrunner` to prevent catastrophic queueing issues. This logic is based on many other issues that have been reported against the [actions/runner](https://github.com/actions/runner) and [actions/actions-runner-controller](https://github.com/actions/actions-runner-controller) repos.

If I understand what you are saying @nikola-jokic the issue is somewhere in the runner logic (eg. in [actions/runner](https://github.com/actions/runner)) where the runner doesn't exit with a non-zero exit code when the registration is lost. Is that a correct understanding? But to @rob-howie-depop's point I'm a bit confused as to how that would work over in the `runner` repo as the registration goes missing **after** the runner pod starts either due to a GitHub incident or other reasons.

nikola-jokic:
Hey, so the idea is that the runner should exit when it cannot recover. The controller shouldn't keep checking the runner all the time to make sure the runner is online, it is the job of the runner to run with its configuration, and if it can't, it should exit. The controller should be responsible to properly react on runner exits. While the runner is active, it should assume that it is working on something, where something might be registering itself, listening on jobs, running the job...

What I am trying to say here is not that we shouldn't fix this, we definitely should. I'm just saying that the controller is not the right place to do so.
The controller has a simple contract with the runner. The container named "runner" is assumed to run the runner binary. Then, we react to exits by the pod inspecting the exit code of the runner. We apply specific environment variables known to the runner to configure how it should behave. For example, in the next release, we set the environment variable that configures runner to exit with exit code 7 if it is outdated. Then, exit code 7 of the runner container would stop the autoscaling runner set since we know that if the runner is deprecated, it cannot self-recover.

Similar logic applies here. If the runner cannot recover, it should exit. It should not be a job of the controller to monitor the runner states. It should be a responsibility of the runner to inform the outer environment (in this case, the ARC managing runner pods) that it cannot receive jobs.

sethjackson:
Got it. Thanks for the clarification @nikola-jokic! And apologies if my post came across like I was implying that I thought there was pushback on fixing the issue. That was not my intent whatsoever. I was just trying to understand where the issue should get fixed. I did see the exit code 7 changes in the runner and that's why I was wondering if the changes to fix this would actually live in [actions/runner](https://github.com/actions/runner). So I'm assuming the fix is in that repo somewhere?

I wasn't sure if the [actions/runner-container-hooks](https://github.com/actions/runner-container-hooks) was involved in the runner state in any way. From a quick glance it doesn't seem to be but was just curious.

Thanks!

nikola-jokic:
No problem, I'm trying to explain my reasoning as much as I can, and also to check if I understand the problem correctly. So pushback is always appreciated because when done nicely (like you did), I can clarify my reasoning and explain it properly. It also allows me to better understand what the issue is, so please don't shy away from asking questions :relaxed: . Sometimes, I misunderstand the issue, and asking questions helps clarify it!

Cjewett:
We're also running into this issue. Some of the workarounds mentioned aren't working for us though.

What we're currently seeing:
- 3 pods spun up that are running in k8s but show up on the GitHub runners page as being offline.
- Deleting the pods results in them immediately spinning back up. That doesn't fix anything and the runners still show as offline.
- Canceling the workflows related to the pods and then deleting the pods doesn't work. It's still auto spinning them back up.
- Canceling the workflows related to the pods and restarting it doesn't work.

Is there a guaranteed workaround for this issue?

I'm also a little confused on what the next steps are here. Does the focus need to be on https://github.com/actions/runner to fix this issue?

nikola-jokic:
Hey @Cjewett,

You should inspect the reason of failures. If it gets deleted right away, it might be something related to the entrypoint or runner config.

rob-howie-depop:
Closing this pull request since the actions runner repository does not take contributions. 
https://github.com/actions/runner

Cjewett:
@nikola-jokic The errors we received are pretty much the exact same as this report: https://github.com/actions/actions-runner-controller/issues/4395

I can't post any information related to our logs on public forums but I can provide GitHub Support ticket numbers and coordinate with our GitHub Enterprise support reps to facilitate any discussions if that would help debug this issue and lead to resolution.

MarkPorohnja:
@rob-howie-depop Could you please elaborate more, why PR was closed. These changes are really needed for us. Every morning I wake up and remove idle runners (github shows them as offline, but they are up in AKS cluster and they block new jobs scheduling).

MarkPorohnja:
Guys who are also disappointed by this PR closure, feel free to +1 comment I left in actions/runner repository. Comment:
https://github.com/actions/runner/issues/4312#issuecomment-4279912771

**Also upvote the issue itself here**: https://github.com/actions/runner/issues/4312

```text
Updating certificates in /etc/ssl/certs...
rehash: warning: skipping ca-certificates.crt,it does not contain exactly one certificate or CRL
rehash: warning: skipping ca.pem,it does not contain exactly one certificate or CRL
1 added, 0 removed; done.
Running hooks in /etc/ca-certificates/update.d...
done.
[RUNNER 2026-03-11 14:38:25Z INFO HostContext] No proxy settings were found based on environmental variables (http_proxy/https_proxy/HTTP_PROXY/HTTPS_PROXY)
[RUNNER 2026-03-11 14:38:25Z INFO HostContext] Well known directory 'Bin': '/home/runner/bin'
[RUNNER 2026-03-11 14:38:25Z INFO HostContext] Well known directory 'Root': '/home/runner'
[RUNNER 2026-03-11 14:38:25Z INFO HostContext] Well known config file 'Credentials': '/home/runner/[redacted]'
[RUNNER 2026-03-11 14:38:25Z INFO HostContext] Well known config file 'Runner': '/home/runner/[redacted]'
[RUNNER 2026-03-11 14:38:25Z INFO HostContext] Adding extra user agent 'actions-runner-controller/0.12.1' to all HTTP requests.
[RUNNER 2026-03-11 14:38:25Z INFO Listener] Runner is built for Linux (X64) - linux-x64.
[RUNNER 2026-03-11 14:38:25Z INFO Listener] RuntimeInformation: Ubuntu 24.04.3 LTS.
[RUNNER 2026-03-11 14:38:25Z INFO Listener] Version: 2.331.0
[RUNNER 2026-03-11 14:38:25Z INFO CommandSettings] Env arg 'jitconfig': '***'
[RUNNER 2026-03-11 14:38:25Z INFO Runner] RunAsync
[RUNNER 2026-03-11 14:38:25Z INFO Runner] Falling back to original .runner settings
[RUNNER 2026-03-11 14:38:25Z INFO Runner] Using BrokerMessageListener
[RUNNER 2026-03-11 14:38:25Z INFO BrokerMessageListener] Using provided settings
[RUNNER 2026-03-11 14:38:25Z INFO BrokerMessageListener] {
[RUNNER 2026-03-11 14:38:25Z INFO BrokerMessageListener]   "AgentId": [redacted],
[RUNNER 2026-03-11 14:38:25Z INFO BrokerMessageListener]   "AgentName": "[redacted]",
[RUNNER 2026-03-11 14:38:25Z INFO BrokerMessageListener]   "PoolId": [redacted],
[RUNNER 2026-03-11 14:38:25Z INFO BrokerMessageListener]   "DisableUpdate": true,
[RUNNER 2026-03-11 14:38:25Z INFO BrokerMessageListener]   "Ephemeral": true,
[RUNNER 2026-03-11 14:38:25Z INFO BrokerMessageListener]   "ServerUrl": "[redacted]",
[RUNNER 2026-03-11 14:38:25Z INFO BrokerMessageListener]   "WorkFolder": "_work",
[RUNNER 2026-03-11 14:38:25Z INFO BrokerMessageListener]   "UseV2Flow": true,
[RUNNER 2026-03-11 14:38:25Z INFO BrokerMessageListener]   "ServerUrlV2": "https://broker.actions.githubusercontent.com/"
[RUNNER 2026-03-11 14:38:25Z INFO BrokerMessageListener] }
[RUNNER 2026-03-11 14:38:26Z INFO BrokerMessageListener] Attempt to create session.
[RUNNER 2026-03-11 14:38:26Z INFO BrokerMessageListener] Connecting to the Broker Server...
[RUNNER 2026-03-11 14:38:26Z INFO BrokerMessageListener] VssConnection created
[RUNNER 2026-03-11 14:38:26Z INFO BrokerMessageListener] Connecting to the Runner server...
[RUNNER 2026-03-11 14:38:27Z INFO Terminal] WRITE LINE:

√ Connected to GitHub
[RUNNER 2026-03-11 14:38:28Z ERR  GitHubActionsService] POST request to https://broker.actions.githubusercontent.com/session failed. HTTP Status: Conflict
[RUNNER 2026-03-11 14:38:28Z ERR  BrokerMessageListener] Catch exception during create session.
[RUNNER 2026-03-11 14:38:28Z ERR  BrokerMessageListener] GitHub.DistributedTask.WebApi.TaskAgentSessionConflictException: Error: Conflict
[RUNNER 2026-03-11 14:38:28Z INFO BrokerMessageListener] The session for this runner already exists.
[RUNNER 2026-03-11 14:38:28Z ERR  Terminal] WRITE ERROR: A session for this runner already exists.
A session for this runner already exists.
[RUNNER 2026-03-11 14:38:28Z ERR  Terminal] WRITE ERROR: 2026-03-11 14:38:28Z: Runner connect error: Error: Conflict. Retrying until reconnected.
2026-03-11 14:38:28Z: Runner connect error: Error: Conflict. Retrying until reconnected.
[RUNNER 2026-03-11 14:38:28Z INFO BrokerMessageListener] Sleeping for 30 seconds before retrying.
...
[RUNNER 2026-03-11 14:41:08Z INFO Terminal] WRITE LINE:

√ Connected to GitHub
[RUNNER 2026-03-11 14:41:09Z INFO BrokerMessageListener] Session created.
[RUNNER 2026-03-11 14:41:09Z INFO Terminal] WRITE LINE: 2026-03-11 14:41:09Z: Runner reconnected.
2026-03-11 14:41:09Z: Runner reconnected.
[RUNNER 2026-03-11 14:41:09Z INFO Terminal] WRITE LINE: Current runner version: '2.331.0'
Current runner version: '2.331.0'
[RUNNER 2026-03-11 14:41:09Z INFO Terminal] WRITE LINE: 2026-03-11 14:41:09Z: Listening for Jobs
2026-03-11 14:41:09Z: Listening for Jobs
[RUNNER 2026-03-11 14:41:09Z INFO JobDispatcher] Set runner/worker IPC timeout to 30 seconds.
```

rob-howie-depop:
@MarkPorohnja - I see you took this forward, and they merged your pull request. Thanks for doing that! I closed it because of the following line in the README, which, as it happens, was the first thing I read when I opened the repo with the intention to contribute: 

>  Thank you for your interest in this GitHub repo, however, right now we are not taking contributions.

At which point I simply did not continue, since I expected any pull requests that I raised would be ignored. 


As an aside, in the incident yesterday we saw more of the same problem. We only see this issue during instance, and we're not sure why you're seeing it in your AKS cluster more frequently than that. 

https://www.githubstatus.com/incidents/qcvjkzcs7j74

What's interesting, though, is that this issue that we have been discussing here for months has finally been flagged as something that they intend to fix:

> During the incident, some Actions Runner Controller (ARC) runner pods became stuck in an idle state. Affected users can delete those pods using kubectl or redeploy their Actions Runner Controller application. ARC will automatically create replacement runners.
> 
> The next releases of Actions Runner and Actions Runner Controller will include an automatic recovery mechanism, preventing the need for these manual steps in the future.

PR #4408, opened 2026-03-16
Add multi-label support to scalesets

nikola-jokic:
Add support for scaleset to set multiple labels similar to the runner: https://docs.github.com/en/actions/how-tos/manage-runners/self-hosted-runners/apply-labels


rentziass (approved):

francesco-beccaria:
Hi, probably this is not the best place to ask this question. I'm happy to move it elsewhere if you'd prefer.

Was there a reason for choosing `scaleSetLabels` instead of `runnerScaleSetLabels` for the key in the Helm Chart values? I found it a bit confusing next to `runnerScaleSetName`.

Issue #4425, opened 2026-03-26
EphemeralRunners created through an AutoscalingRunnerSet do not have labels set in the GitHub.com API


ivucica:
### Checks

- [x] I've already read https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/troubleshooting-actions-runner-controller-errors and I'm sure my issue is not covered in the troubleshooting guide.
- [x] I am using charts that are officially provided

### Controller Version

ghcr.io/actions/gha-runner-scale-set-controller 0.14.0

### Deployment Method

Helm

### Checks

- [x] This isn't a question or user support case (For Q&A and community support, go to [Discussions](https://github.com/actions/actions-runner-controller/discussions)).
- [x] I've read the [Changelog](https://github.com/actions/actions-runner-controller/blob/master/docs/gha-runner-scale-set-controller/README.md#changelog) before submitting this issue and I'm sure it's not due to any recently-introduced backward-incompatible changes

### To Reproduce

```markdown
See #4424.

Ensure that AutoscalingRunnerSet has the labels set:

$ kubectl get autoscalingrunnerset  -n arc-runners -o json | jq .items[0].spec.runnerScaleSetLabels
[
  "self-hosted",
  "linux",
  "arc-runner-set"
]

(I wish I could easily prove that running scale set has this set server side, but the API is not documented and the [scaleset client](https://github.com/actions/scaleset) does not have an obvious endpoint documented beyond [`_apis/runtime/runnerscalesets`](https://github.com/actions/scaleset/blob/ebedf0bbbf6245cf5b1ff3c8a1297a08e133caf8/client.go#L24) (however I have not managed to guess what the host is). It's there on the web.

Let an ephemeral runner be created -- it won't have labels. Apply to runners (for example like this, starting with 0th one, repeat as needed)


curl -X POST -d '{"labels":["arc-runner-set"]}' \
    -H "Authorization: Bearer $(grep github_token: 03-pat.secrets.yaml | cut -d':' -f 2)" \
    https://api.github.com/repos/PERSONUSER/REPONAME/actions/runners/$(curl \
        -H "Authorization: Bearer $(grep github_token: 03-pat.secrets.yaml | cut -d':' -f 2)" \
    https://api.github.com/repos/PERSONUSER/REPONAME/actions/runners |  jq -r .runners[0].id)/labels


Run an action.

See the runner get nuked / recycled.

See a new runner get created again without labels.
```

[...]

Issue #4424, opened 2026-03-26
Editing (instead of destroying+recreating) an AutoscalingRunnerSet object's runnerScaleSetLabels does not update the registered runner set's labels


ivucica:
### Checks

- [x] I've already read https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/troubleshooting-actions-runner-controller-errors and I'm sure my issue is not covered in the troubleshooting guide.
- [x] I am using charts that are officially provided

### Controller Version

0.14.0

### Deployment Method

Helm

### Checks

- [x] This isn't a question or user support case (For Q&A and community support, go to [Discussions](https://github.com/actions/actions-runner-controller/discussions)).
- [x] I've read the [Changelog](https://github.com/actions/actions-runner-controller/blob/master/docs/gha-runner-scale-set-controller/README.md#changelog) before submitting this issue and I'm sure it's not due to any recently-introduced backward-incompatible changes

### To Reproduce

```markdown
1. In chart `oci://ghcr.io/actions/actions-runner-controller-charts/gha-runner-scale-set`, forget[^1] to set `scaleSetLabels` [sic] which propagates into `runnerScaleSetLabels`.
2. Edit the chart values and correctly set `scaleSetLabels`.
3. Confirm this is updated in local `AutoscalingRunnerSet`.
4. On Github.com web UI, observe how the neither the runner set nor the runner have any labels on it.
5. Destroy the `AutoscalingRunnerSet` (and separately remove the Helm chart installation matching it).
6. Create it again by applying the chart again.
7. Observe how the runner set is recreated with correct charts visible on Github.com web UI.

[^1]: I have actually set `runnerScaleSetLabels` per documentation, but that's not what the chart uses.
```

### Describe the bug

The change in labels does not apply the new labels to the runner set via the api, nor (if that is required) recreate the ephemeral runner set, nor does it recreate the runner itself.

### Describe the expected behavior

Updating `spec.runnerScaleSetLabels` on the custom resource for `AutoscalingRunnerSet` should update labels for runners, as correct labels are critical for running GHA actually finding the correct runner to run the workload on. Whether this is done by destroying and recreating the `EphemeralRunnerSet` and `EphemeralRunner` objects, and destroying and re-creating the runner scale set entry in the github.com API (visible on the web), is less important -- but it is essential that the runners have labels actually applied to them when the API calls are made.

### Additional Context

```yaml
[...]

PR #4426, opened 2026-03-27
Update labels on ScaleSet and Runners

pauloconnor:
This should address #4424 and #4425


ybrovka-cw (approved):

nikola-jokic:
I'm not sure if `strings.Join` on labels is the right thing to do to set the labels, because the combination of labels joined with "," must have less than 64 characters.
> Valid label value:
  must be 63 characters or less (can be empty),
  unless empty, must begin and end with an alphanumeric character ([a-z0-9A-Z]),
  could contain dashes (-), underscores (_), dots (.), and alphanumerics between.

We can set the annotation on these resources, because it doesn't have such restrictions as far as I can tell.

PR #4451, opened 2026-04-15
Add option to disable workqueue bucket rate limiter

Okabe-Junya:
cf. https://github.com/mercari/actions-runner-controller/pull/6

## What

Add a `--disable-workqueue-bucket-rate-limiter` flag to the controller manager. When set, the controller workqueue uses only per-item exponential backoff (`DefaultTypedItemBasedRateLimiter`) instead of the default `DefaultTypedControllerRateLimiter`, which includes an overall token bucket limiter (10 QPS, 100 bucket size).

## Why

In large-scale environments with many runner scale sets, the default token bucket rate limiter (10 QPS overall) can become a bottleneck for reconciliation throughput. When many `EphemeralRunnerSet` or `AutoscalingRunnerSet` resources need to be reconciled simultaneously, the 10 QPS cap delays processing and causes runner provisioning latency.

The per-item exponential backoff is still preserved to prevent tight retry loops on individual failing resources. Removing only the overall bucket constraint allows the controller to process independent reconciliation requests without artificial global throttling.

Note: This flag defaults to `false`, so the existing behavior is unchanged unless explicitly opted in.

### DIFF
--- a/charts/gha-runner-scale-set-controller/templates/deployment.yaml
+++ b/charts/gha-runner-scale-set-controller/templates/deployment.yaml
@@ -93,6 +93,9 @@ spec:
         {{- with .Values.flags.k8sClientRateLimiterBurst }}
         - "--k8s-client-rate-limiter-burst={{ . }}"
         {{- end }}
+        {{- if .Values.flags.disableWorkqueueBucketRateLimiter }}
+        - "--disable-workqueue-bucket-rate-limiter"
+        {{- end }}
         command:
         - "/manager"
         {{- with .Values.metrics }}
--- a/charts/gha-runner-scale-set-controller/templates/deployment.yaml
+++ b/charts/gha-runner-scale-set-controller/templates/deployment.yaml
@@ -93,8 +93,10 @@ spec:
         {{- with .Values.flags.k8sClientRateLimiterBurst }}
         - "--k8s-client-rate-limiter-burst={{ . }}"
         {{- end }}
-        {{- if .Values.flags.disableWorkqueueBucketRateLimiter }}
-        - "--disable-workqueue-bucket-rate-limiter"
+        {{- with .Values.flags.rateLimiter }}
+        {{- with .name }}
+        - "--workqueue-rate-limiter={{ . }}"
+        {{- end }}
         {{- end }}
         command:
         - "/manager"
--- a/charts/gha-runner-scale-set-controller/values.yaml
+++ b/charts/gha-runner-scale-set-controller/values.yaml
@@ -142,3 +142,10 @@ namespaceOverride: ""
 ## Defines the K8s client rate limiter parameters.
   # k8sClientRateLimiterQPS: 20
   # k8sClientRateLimiterBurst: 30
+
+  ## Disable the overall token bucket rate limiter for the controller workqueue.
+  ## By default, controller-runtime uses a combined rate limiter with both a per-item
+  ## exponential backoff and an overall token bucket (10 QPS, 100 bucket size).
+  ## In large-scale environments, the token bucket can become a bottleneck for
+  ## reconciliation throughput. Set to true to use only per-item exponential backoff.
+  # disableWorkqueueBucketRateLimiter: false
--- a/charts/gha-runner-scale-set-controller/values.yaml
+++ b/charts/gha-runner-scale-set-controller/values.yaml
@@ -136,12 +136,12 @@ flags:
   # excludeLabelPropagationPrefixes:
   #   - "argocd.argoproj.io/instance"
 
-  ## Disable the overall token bucket rate limiter for the controller workqueue.
+  ## Workqueue rate limiter configuration.
   ## By default, controller-runtime uses a combined rate limiter with both a per-item
   ## exponential backoff and an overall token bucket (10 QPS, 100 bucket size).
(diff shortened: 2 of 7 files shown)


charts/gha-runner-scale-set-controller/values.yaml
```diff
@@ -142,3 +142,10 @@ namespaceOverride: ""
 ## Defines the K8s client rate limiter parameters.
   # k8sClientRateLimiterQPS: 20
   # k8sClientRateLimiterBurst: 30
+
+  ## Disable the overall token bucket rate limiter for the controller workqueue.
+  ## By default, controller-runtime uses a combined rate limiter with both a per-item
+  ## exponential backoff and an overall token bucket (10 QPS, 100 bucket size).
+  ## In large-scale environments, the token bucket can become a bottleneck for
+  ## reconciliation throughput. Set to true to use only per-item exponential backoff.
+  # disableWorkqueueBucketRateLimiter: false
```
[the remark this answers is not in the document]
Okabe-Junya: ah good catch, updated in [767e8e5](https://github.com/actions/actions-runner-controller/pull/4451/commits/767e8e59db0f3fda37d41ed289c92f890f56f1f0)

controllers/actions.github.com/options.go
```diff
@@ -1,10 +1,16 @@
...
 import (
+	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+// Request is a type alias for reconcile.Request, used to parameterize
+// the generic workqueue rate limiter types from client-go.
+type Request = reconcile.Request
```
nikola-jokic: Why do we alias reconcile.Request?

charts/gha-runner-scale-set-controller/templates/deployment.yaml
```diff
@@ -93,6 +93,9 @@ spec:
         {{- with .Values.flags.k8sClientRateLimiterBurst }}
         - "--k8s-client-rate-limiter-burst={{ . }}"
         {{- end }}
+        {{- if .Values.flags.disableWorkqueueBucketRateLimiter }}
+        - "--disable-workqueue-bucket-rate-limiter"
```
nikola-jokic: Since there are multiple rate limiters, can we use rate limiter name?

By default, it would be a  `bucket_rate_limiter` or `typed_rate_limiter`? The config in values.yaml could be an object such as:
```
rateLimiter:
  name: "typed_rate_limiter"
  # options: {}
```

controllers/actions.github.com/options.go
```diff
@@ -37,6 +43,23 @@ func WithMaxConcurrentReconciles(n int) Option {
...
+// By default, the controller-runtime uses DefaultControllerRateLimiter, which
+// combines an exponential backoff per-item limiter with a token bucket overall
+// limiter (10 QPS, 100 bucket size). In large-scale environments with many
+// runner scale sets, the token bucket limiter can become a bottleneck for
+// reconciliation throughput.
+//
+// Use this option to override the default rate limiter, for example, to use
+// DefaultItemBasedRateLimiter which removes the overall token bucket constraint
+// while keeping the per-item exponential backoff.
+func WithWorkqueueRateLimiter(rateLimiter workqueue.TypedRateLimiter[Request]) Option {
```
nikola-jokic: I'd name this `WithTypedRateLimiter`

nikola-jokic (changes requested):
Hey, I left some comments. I like the idea to allow different rate limiters, this is great. But since we are doing it, I'd love to design it so we can implement other rate limiters when necessary ☺️ Thank you for the work you put in!

controllers/actions.github.com/options.go
Okabe-Junya: Thanks, good catch,
we don't need this alias, fixed in [d878d10](https://github.com/actions/actions-runner-controller/pull/4451/commits/d878d108331245746fa9db50b32ceb3975fdf281)

controllers/actions.github.com/options.go
Okabe-Junya: Got it, fixed in [f99e4de](https://github.com/actions/actions-runner-controller/pull/4451/commits/f99e4de2aac5adba6861647ac96af6d564fe222f)

charts/gha-runner-scale-set-controller/templates/deployment.yaml
Okabe-Junya: Thanks, good suggestion,

updated implementation in [0c6868c](https://github.com/actions/actions-runner-controller/pull/4451/commits/0c6868c16e8caa746c7a716e0274bbe4f8e25ae5), PTAL again when you have time

Okabe-Junya:
@nikola-jokic 

Thank you for your review, updated some points from my side.
Please taek a look again when you have time, thanks!

nikola-jokic:
Can you please go fmt?

Okabe-Junya:
@nikola-jokic 

Oh sorry... updated

nikola-jokic (approved):
LGTM, thanks!

PR #4477, opened 2026-04-30
feat(helm/gha-runner-scale-set-experimental): add $extra field passthrough for runner container in dind and kubernetes modes

stokkie90:
## Summary

- In `dind` and `kubernetes` modes, the runner container template now reads `runner.container` as a map and passes any fields not explicitly rendered (`securityContext`, `resources`, `imagePullPolicy`, `args`, etc.) through to the container spec via the `$extra = omit ... | toYaml` pattern.
- This mirrors the behaviour already present in `mode: ""` (empty mode) via `_mode_empty.tpl`, and in the dind sidecar container in `_mode_dind.tpl`.
- Input validation (wrong type guards) is included: `runner.container` must be a map, `env`/`volumeMounts`/`args` must be lists, `securityContext` must be a map, and `volumes` is rejected with a clear error pointing to `runner.pod.spec.volumes`.

## Risks & rollout

- Purely additive: existing values that don't set `runner.container` are unaffected (the map defaults to `{}`).
- No breaking changes to existing behaviour; the `name` key in `runner.container` is silently ignored to preserve the hardcoded `runner` container name.

## Validation

- 14 new `helm-unittest` test cases added (7 per mode) covering: extra field passthrough, name-override suppression, and all validation failure paths.
- Full test suite passes: **171 tests, 0 failures**.

```
Charts:      1 passed, 1 total
Test Suites: 35 passed, 35 total
Tests:       171 passed, 171 total
Time:        427ms
```

### DIFF
--- a/charts/gha-runner-scale-set-experimental/templates/_defaults.tpl
+++ b/charts/gha-runner-scale-set-experimental/templates/_defaults.tpl
@@ -122,6 +122,40 @@ Hook extension ConfigMap name for kubernetes runner mode.
 If runner.kubernetesMode.extension.metadata.name is set, use it.
 Otherwise, default to a name derived from the scale set name.
 */}}
+{{/*
+Validate runner.container fields.
+
+Fails with a descriptive error if:
+- runner.container is set but is not a map/object
+- runner.container.env is set but is not a list
+- runner.container.volumeMounts is set but is not a list
+- runner.container.args is set but is not a list
+- runner.container.securityContext is set but is not a map/object
+- runner.container.volumes is set (unsupported; use runner.pod.spec.volumes)
+*/}}
+{{- define "runner.container.validate" -}}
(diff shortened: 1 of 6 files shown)


gazab:
We would love to be able to extends `env` and `volumeMounts` in the `runner` container when using mode `dind` as well!

stokkie90:
Addressed both Copilot review comments:

1. **`_mode_dind.tpl` — use `$runner` intermediate variable**: fixed by introducing `$runner := (.Values.runner | default dict)` before accessing `runner.container`, consistent with the kubernetes mode pattern.

2. **Duplicate validation block**: extracted the six type-guard checks into a shared `runner.container.validate` helper in `_defaults.tpl`. Both `_mode_dind.tpl` and `_mode_kubernetes.tpl` now call `{{- include "runner.container.validate" . -}}` instead of duplicating the block.

All 171 tests continue to pass.

PR #4535, opened 2026-06-19
Add status conditions and observedGeneration

Okabe-Junya:
## What

- Add `status.observedGeneration` and a standard "Ready" condition to `AutoscalingRunnerSet`, `EphemeralRunnerSet`, `EphemeralRunner`, and `AutoscalingListener`.
- Every status write stamps `observedGeneration`
  - "Ready" condition is updated on phase/readiness transitions. phase is unchanged for backward compatibility.
-  (CRD regenerated with `make manifests`)

## Why

The resources expose only a free-form "phase" string, so clients cannot tell whether the status reflects the latest spec or wait for readiness in a standard way. With this change, `k wait --for=condition=Ready` and `generation == status.observedGeneration` checks work, per the k8s API conventions for status.


Okabe-Junya:
Just noticed that some files are not marked as “generated” in gitattribute. Let me create another PR to fix this

Okabe-Junya:
> Just noticed that some files are not marked as “generated” in gitattribute. Let me create another PR to fix this

https://github.com/actions/actions-runner-controller/pull/4536

PR #4536, opened 2026-06-19
Mark generated files as linguist-generated

Okabe-Junya:
## What

- Add `.gitattributes` rules that mark machine-generated files as
`linguist-generated=true`


nikola-jokic (approved):
This makes sense, thanks!

PR #4541, opened 2026-06-24
(feat) Rootless dind mode

alexrecuenco:
Currently we can't overwrite the dind deployment to use rootless mode, this updates the dind deployment to use rootless mode instead of privileged mode when required


alexrecuenco:
The rootless mode configuration was extracted from the rootless mode that is created when you run `docker buildx create --bootstrap --name=kube --driver=kubernetes --driver-opt=namespace=buildkit,rootless=true`. Just replicating the same confinement configurations that are there

PR #4556, opened 2026-07-05
Make controller terminationGracePeriodSeconds configurable and align it with graceful shutdown timeout

Okabe-Junya:
## What

- Replace the hardcoded `terminationGracePeriodSeconds: 10` in scale set controller chart Deployment with a value configurable via `.Values.terminationGracePeriodSeconds`.
- Raise the default to `35` so it exceeds the manager's graceful shutdown timeout.

## Why

The controller manager relies on controller-runtime's default `GracefulShutdownTimeout` of 30s: after SIGTERM it waits up to 30s for in-flight reconciles to finish. However, the chart pinned `terminationGracePeriodSeconds` to 10s, so the kubelet sends SIGKILL 20s before that window ends. 

Every rolling update or node drain could kill the controller mid-reconcile (e.g. during EphemeralRunner finalizer handling or GitHub API calls), causing avoidable re-reconciles, API retries, and delayed status updates on restart.

### DIFF
--- a/charts/gha-runner-scale-set-controller-experimental/templates/deployment.yaml
+++ b/charts/gha-runner-scale-set-controller-experimental/templates/deployment.yaml
@@ -40,7 +40,7 @@ spec:
         -
       {{- toYaml . | nindent 10 }}
         {{- end }}
-      terminationGracePeriodSeconds: {{ default 10 (index $podSpec "terminationGracePeriodSeconds") }}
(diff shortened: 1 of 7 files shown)


nikola-jokic:
Hey @Okabe-Junya,

Thanks for your contribution! Could you please add this option to the experimental chart as well?

Okabe-Junya:
Oh, sorry...
I usually accidentally forget about the experimental chart... 😓 🙇

nikola-jokic (approved):
Thank you for your contributions!

Issue #4561, opened 2026-07-10
0.14.2: AutoscalingListener can retain deleted EphemeralRunnerSet after runner template update


DocX:
### Checks

- [x] I've already read https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/troubleshooting-actions-runner-controller-errors and I'm sure my issue is not covered in the troubleshooting guide.
- [x] I am using charts that are officially provided (rendered from the official 0.14.2 charts and applied through Argo CD).

### Controller Version

0.14.2

### Deployment Method

ArgoCD

### Checks

- [x] This isn't a question or user support case.
- [x] I've read the changelog and this is not an expected backward-incompatible change.

### To Reproduce

1. Deploy `gha-runner-scale-set-controller` and `gha-runner-scale-set` 0.14.2 with `--update-strategy=eventual`, `minRunners: 1`, and `maxRunners: 10`.
2. Wait until the `AutoscalingRunnerSet`, `AutoscalingListener`, and `EphemeralRunnerSet` are healthy.
3. Change `AutoscalingRunnerSet.spec.template.spec` so ARC rolls the EphemeralRunnerSet. In our case, a GitOps deployment removed an obsolete `runtimeClassName` and changed the runner container security context.
4. Observe that the old EphemeralRunnerSet is deleted and a different EphemeralRunnerSet survives, while `AutoscalingListener.spec.ephemeralRunnerSetName` still names the deleted set.
5. The controller repeatedly recreates the listener pod from that unchanged `AutoscalingListener`. Every pod attempts its initial patch against the deleted EphemeralRunnerSet, exits, and enters a permanent recreate/crash loop.

This happened during an update to an existing scale set, rather than a fresh install.

### Describe the bug

On ARC 0.14.2, an AutoscalingListener remained bound to a deleted EphemeralRunnerSet after a runner-template update. The surviving EphemeralRunnerSet had a different generated name, but the listener CR, config, and RBAC still referenced the deleted name.

Every listener restart failed identically:

```text
Application returned an error: handling initial message failed: could not patch ephemeral runner set,
patch JSON: {"spec":{"patchID":0,"replicas":1}},
error: ephemeralrunnersets.actions.github.com "<scale-set>-qb49n" not found
```

The controller then deleted and recreated the listener pod, but did not repair or replace the stale AutoscalingListener CR, so the loop continued indefinitely and jobs could not be assigned.

[...]

Issue #4370, opened 2026-02-06
gha-runner-scale-set-0.13.1 installation fails with RunnerScaleSetNotFoundException


erichorwath:
### Checks

- [x] I've already read https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/troubleshooting-actions-runner-controller-errors and I'm sure my issue is not covered in the troubleshooting guide.
- [x] I am using charts that are officially provided

### Controller Version

0.13.1

### Deployment Method

Helm

### Checks

- [x] This isn't a question or user support case (For Q&A and community support, go to [Discussions](https://github.com/actions/actions-runner-controller/discussions)).
- [x] I've read the [Changelog](https://github.com/actions/actions-runner-controller/blob/master/docs/gha-runner-scale-set-controller/README.md#changelog) before submitting this issue and I'm sure it's not due to any recently-introduced backward-incompatible changes

### To Reproduce

```markdown
Install 0.12.1, then upgrade to 0.13.1
```

### Describe the bug

After upgrade, listeners are not getting ready:

```
Application returned an error: createSession failed: failed to create session: failed to do the session request: actions error: StatusCode 404, ActivityID "xxx" Github.Actions.Runtime.WebApi.RunnerScaleSetNotFoundException, Github.Actions.Runtime.WebApi: No runner scale set found with identifier 22.
```

<img width="1087" height="80" alt="Image" src="https://github.com/user-attachments/assets/b0d43361-1399-405a-85f6-8bfc6d7bb6b5" />

### Describe the expected behavior

listener get ready and helm upgrade succeeds.

### Additional Context

```yaml
K8s 1.32
GHES 3.16
```

### Controller Logs

```shell
.
```

### Runner Pod Logs

```shell
.
```

nikola-jokic:
Hey @erichorwath,

What are the exact steps you took to upgrade?

bavugisainath:
We are experiencing the same issue on our end. We recently created a new EKS cluster v1.35 and installed the ARC controller version 0.13.1, then deployed the scaleset v0.13.1.

After deploying the listeners, we noticed that the pods are behaving differently compared to version 0.12.1 (which we previously used). One of the listener pods continuously retries and appears to be stuck in a loop, while other listeners on the same node are functioning correctly.

The behavior seems random, as only specific listener pods are affected.

Could you please help us investigate this issue?

K8s 1.35
GHES 3.17.3
[...]

Issue #3942, opened 2025-02-21
Listener Pod Session Creation Fails with RunnerScaleSetNotFoundException During GitHub API Outage


ali-kafel:
### Checks

- [x] I've already read https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/troubleshooting-actions-runner-controller-errors and I'm sure my issue is not covered in the troubleshooting guide.
- [x] I am using charts that are officially provided

### Controller Version

0.10.1

### Deployment Method

Helm

### Checks

- [x] This isn't a question or user support case (For Q&A and community support, go to [Discussions](https://github.com/actions/actions-runner-controller/discussions)).
- [x] I've read the [Changelog](https://github.com/actions/actions-runner-controller/blob/master/docs/gha-runner-scale-set-controller/README.md#changelog) before submitting this issue and I'm sure it's not due to any recently-introduced backward-incompatible changes

### To Reproduce

```markdown
- This error was observed during an unplanned GitHub API outage and has not been reproducible via deliberate steps. 
- We suspect that GitHub API availability plays a critical role in triggering the bug; therefore, simulating an outage (e.g., by blocking network access to GitHub API endpoints) might help in reproducing it. 
- As it stands, the bug appears to be a timing/availability issue rather than a deterministic code defect.
```

### Describe the bug

We encountered an issue in the listener pod (pod name: xlarge-dind-spot-6566dbd7-listener) of the GitHub Actions controller during a GitHub API/Actions outage. The listener pod’s logs indicate that while the pod initializes correctly, it fails when attempting to create a session with GitHub Actions.

Error Log Snippet from the Listener Pod:
```
│ stream logs failed container "listener" in pod "xlarge-dind-spot-6566dbd7-listener" is waiting to start: ContainerCreating for gha-runner-controller/xlarge-dind-spot-6566dbd7-listener (listener)  │
│ 2025-02-17T08:37:14Z    INFO    listener-app    app initialized                                                                                                                                     │
│ 2025-02-17T08:37:14Z    INFO    listener-app    Starting metrics server                                                                                                                             │
│ 2025-02-17T08:37:14Z    INFO    listener-app    Starting listener                                                                                                                                   │
│ 2025-02-17T08:37:14Z    INFO    listener-app    refreshing token    {"githubConfigUrl": "https://github.com/****"}                                                                                │
[...]

PR #4571, opened 2026-07-14
Re-register the runner scale set when it is gone from the Actions service

diogotorres97:
Partially addresses #4370.

New scale sets — and any whose listener has not been created yet — recover automatically. An `AutoscalingListener` that is *already* crash-looping on a dead ID is not healed on its own; deleting it with `kubectl delete autoscalinglistener <name> -n <controller-namespace>` makes the controller rebuild it against a revalidated ID. Before this change the only recovery was deleting the whole `AutoscalingRunnerSet`. See "Follow-up worth considering" below for why, and what would close the gap.

## The problem

A runner scale set can disappear from the Actions service out of band: deleted from the organization settings, or dropped and registered again with a fresh ID. The ID recorded in the `runner-scale-set-id` annotation then points at nothing.

Nothing notices:

- `shouldCreateScaleSet` registers a scale set again only when the annotation is **missing, unparseable, or non-positive** (`autoscalingrunnerset_controller.go:798-805`). A valid-but-dead ID survives every reconcile.
- The ID is baked into the `AutoscalingListener` spec once, when it is built (`resourcebuilder.go:137`), and each listener pod reads it on startup and exits on any error. Every pod fails with `RunnerScaleSetNotFoundException` and the listener crash-loops.

Restarting pods does not help, because the ID is baked into the `AutoscalingListener` spec. Correcting the annotation by hand does not help either. The only way out today is to delete the `AutoscalingRunnerSet` and let it be created again, which several people in #4370 report is not viable in production.

Deleting it gets stuck too: `DeleteRunnerScaleSet` returns not-found for a scale set that is already gone, `deleteRunnerScaleSet` propagates that error, the finalizer never completes, and the resource sits in `Terminating` until the finalizer is removed by hand.

## The fix

Confirm the recorded scale set still exists before creating a listener with its ID, and register it again when it does not. `createRunnerScaleSet` already reuses an existing scale set with the same name, so a scale set that came back with a new ID is adopted rather than duplicated.

**Only a definitive not-found triggers this.** Any other failure is returned to the caller, so a transient Actions service outage cannot cause a second scale set to be registered — the failure mode reported in #3942. There is a test for exactly this.

The lookup runs only when a listener is about to be created, not on every reconcile, so steady state costs no extra API calls.

The delete path now treats an already-deleted scale set as the desired state, so an `AutoscalingRunnerSet` whose scale set is gone no longer needs its finalizer stripped by hand.

## A note on the dependency bump

Both paths need a not-found sentinel from `github.com/actions/scaleset`. `wrapResponseErrorType` on `main` maps HTTP 404 to `scaleset.NotFoundError`, which is exactly what is needed, but it is **not in v0.4.0** — the latest tag — so this PR moves the dependency to a pseudo-version of `main`.

Happy to re-point it at a tag if you cut a `scaleset` release, or to rework this if you would rather not carry a pseudo-version. Without the sentinel the alternative is matching on the error string, which is out of step with how the rest of the controllers detect these conditions (`errors.Is(err, scaleset.JobStillRunningError)` and friends).

## Tests

Three specs in `autoscalingrunnerset_controller_test.go`:

- a stale scale set is registered again, and the listener is created with the new ID — fails on `main`, where the listener is created with the dead ID
- a transient Actions service failure does **not** re-register the scale set, and creates nothing
- deleting an `AutoscalingRunnerSet` whose scale set is already gone removes the finalizer instead of hanging in `Terminating` — fails on `main`, where deletion never completes

Also adds `WithGetRunnerScaleSetByIDFunc` to the fake client, mirroring the existing `Func` options, so a test can answer per-ID.

The full `TestAPIs` suite passes locally, including the three new specs.

## Follow-up worth considering

This heals a listener at creation time. A listener that is *already* crash-looping is not healed automatically, because its spec never changes and so it is never rebuilt — deleting the `AutoscalingListener` is now enough to recover, where before the whole `AutoscalingRunnerSet` had to go. Making the controller notice a persistently failing listener and revalidate would close that gap, but it is a larger change and I left it out of this PR.

#4561 looks like the same shape with a different field, and may already be resolved on `main` now that the `EphemeralRunnerSet` is patched in place under a deterministic name.


controllers/actions.github.com/autoscalingrunnerset_controller_test.go
```diff
@@ -14,6 +14,9 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"strconv"
+	"sync/atomic"
```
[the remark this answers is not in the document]
diogotorres97: Thanks, but this one doesn't hold up. `strconv`, `sync`, and `sync/atomic` are in the stdlib import group (lines 4–14), separated from the k8s / controller-runtime group by the blank line at line 15 — they aren't mixed into the non-stdlib imports. Both `gofmt -l` and `goimports -l` report the file as clean (no diff), so the formatting checks pass. Closing this as not applicable, but happy to revisit if I'm missing something.

diogotorres97:
@nikola-jokic @Okabe-Junya can you take a look?

go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.22.0
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.14.0
 	github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azsecrets v1.5.0
-	github.com/actions/scaleset v0.4.0
+	github.com/actions/scaleset v0.4.1-0.20260706130337-9b2803251ede
```
Okabe-Junya, line 9: I think would be better to use SemVer officeial release - listed in https://github.com/actions/scaleset/releases

but let's wait the officeial response from team (since I'm not working in GitHub)

Okabe-Junya (commented):
I'm not a official reviewer, maintainer in this repo, but left a minor comment
Let's wait the response from Nicola!

diogotorres97:
ping

buildpulser:
We ran into what I think is the persistent version of this: one org where the broker evicts every message session, so the listener 404s on every token refresh and dies — every ~50 minutes for two weeks, and readiness checks never noticed because the pod comes right back. Opened actions/scaleset#123 to make the session client re-create the session on a refresh 404 instead of bailing; we've been running it in production and it turned the crash loop into a log line. Might be complementary to the re-registration logic here.

nikola-jokic (approved):
LGTM, thanks!

go.mod
nikola-jokic: Before the release, we will publish the new scaleset version so we will use the semver but this is a good call

rentziass (approved):

PR #4580, opened 2026-07-23
Use lazy copy to patch resources to ensure multiple modifications are applied to the base resource

nikola-jokic:
Reconcilers read an object far more often than they change it, but every reconcile paid for a full `DeepCopy` up front just to have a patch base available for the rare write. `lazyCopy` defers that copy until the moment something is actually about to be mutated, so only the reconciles that patch pay for it.

## Behaviour change worth reviewing

`EphemeralRunner` previously added only one of its two finalizers on the reconcile where both were missing. The old code was:

```go
var addedFinalizers bool
addedFinalizers = addedFinalizers || controllerutil.AddFinalizer(&ephemeralRunner, ephemeralRunnerFinalizerName)
addedFinalizers = addedFinalizers || controllerutil.AddFinalizer(&ephemeralRunner, ephemeralRunnerActionsFinalizerName)
```

`||` short-circuits. With both finalizers missing, the first `AddFinalizer` adds one and returns `true`, so the second call is never evaluated and its finalizer is not added.

It is not lost permanently: the patch triggers another reconcile, where the first `AddFinalizer` returns `false` because its finalizer is now present, so the second call runs and adds the remaining one. The effect is a gap of one reconcile — which is the window in which it matters, because a runner deleted while `ephemeralRunnerActionsFinalizerName` is absent skips the cleanup that unregisters it from the Actions service.

After this PR both are added on the first pass. This is the "ensure multiple modifications are applied" in the title, and no line of the diff announces it — the change reads as a stylistic rewrite — so it is called out here rather than left to be found later.

## How it is used

```go
runner := newLazyCopy(&ephemeralRunner)
if !controllerutil.ContainsFinalizer(&ephemeralRunner, name) {
    controllerutil.AddFinalizer(runner.Mutate(), name)
}
if runner.Modified() {
    err := r.Patch(ctx, &ephemeralRunner, runner.MergeFrom())
}
```

`Mutate()` returns the live object and snapshots it on first call. Reads go through the original as before; every mutation the patch should carry goes through `Mutate`. `MergeFrom` forwards controller-runtime's merge options, so a call site needing `client.MergeFromWithOptimisticLock{}` can ask for it.

The ordering is a caller invariant, not something the type enforces — the caller keeps the pointer it passed to `newLazyCopy`. A write landing before the first `Mutate` is already in the snapshot, so the patch comes out empty and the write is silently dropped. That is documented on the type and pinned by a test rather than only described.

## Fixes found in review

- **Classifier read a superseded revision.** `patchAppliedActionableRevisionStatus` built its desired status as a value copy. The monotonicity guard wrote the advance into the copy while the classifier below still read the applied revision off the fetched object — a line that is textually unchanged and therefore absent from the diff. A runner left over from the superseded revision, missed by the cache-read cleanup, was then counted `Outdated` against the pre-advance revision and that phase saved next to the advanced marker. Nothing recomputes it: `Reconcile` returns on the Outdated path before `updateStatus`, and this function only runs while spec is ahead of applied, so the set stayed switched off until the next spec change. Caught by @salmanmkc.

- **The split itself is gone.** Keeping the revision in two places is what made the above expressible, so the function now patches through the lazy copy and the revision lives in one place. The value copy was also shallow; all three status fields are scalars today, but a pointer field added later would have been shared between the copy and the object it was meant to be compared against.

Every existing test of that function drove the guard-not-fired branch, where the two revisions agree, so all of them passed with the bug live. The added tests cover the guard firing and the no-patch-when-nothing-changed path, and both were checked against a deliberately broken build to confirm they can fail.

---

Based on #4575


rentziass (dismissed):

rentziass (dismissed):

controllers/actions.github.com/lazycopy.go
```diff
@@ -0,0 +1,73 @@
...
+//
+// Reconcilers observe an object far more often than they change it, so taking
+// the snapshot up front means paying for a full deep copy on every reconcile
+// just to serve the rare patch. lazyCopy pays for it only on the reconciles
+// that patch.
+//
+// The snapshot must be taken before the first mutation, otherwise the merge
+// patch is computed against the already mutated object and comes out empty.
+// Mutate is the only way to reach the object, which makes that ordering
+// impossible to get wrong:
```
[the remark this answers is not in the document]
nikola-jokic: Valid, and it contradicted a claim two lines above it. Fixed in `aba0de22`.

The doc said *"Mutate is the only way to reach the object, which makes that ordering impossible to get wrong."* That's false — the caller keeps the pointer it passed to `newLazyCopy`, and the worked example in that very comment goes on using it, both to read (`ContainsFinalizer(&ephemeralRunner, …)`) and to address the patch (`r.Patch(ctx, &ephemeralRunner, …)`). All four production call sites do the same. So the object is demonstrably reachable without `Mutate`; nothing structural prevents a write through it.

The code is correct today — every call site only *reads* through the raw pointer — but you're right that the guarantee isn't one. A write landing before the first `Mutate` is captured by the snapshot, so the merge patch against that snapshot is empty and the write is silently dropped rather than sent to the API server.

Two changes:

- The doc now states the ordering as a caller invariant, spells out that the caller retains the pointer, and names the failure mode (empty patch, silent no-op) instead of promising it can't happen.
- Added `TestLazyCopyDropsWritesMadeBeforeTheFirstMutate`, which writes through the original before the first `Mutate` and asserts that key is **absent** from the resulting patch. That turns the prose warning into an executable claim, so the hazard is pinned rather than just described.

No behaviour change. Worth fixing precisely because the comment is what a future caller reads instead of the implementation.

controllers/actions.github.com/ephemeralrunnerset_controller.go
```diff
@@ -353,14 +354,16 @@ func (r *EphemeralRunnerSetReconciler) patchAppliedActionableRevisionStatus(ctx
...
+		// Build the desired status as a value so no copy of the object is taken
+		// on the common path where the status already matches.
+		desiredStatus := latest.Status
 
 		// Only an advance means the idle and pending runners were just deleted and
 		// the listener restarted. Guarding both writes on it keeps this callable
 		// as a plain "make sure status reflects revision N" without disturbing a
 		// marker that still describes the live patch sequence.
 		if latest.Status.AppliedActionableRevision < targetAppliedRevision {
-			latest.Status.AppliedActionableRevision = targetAppliedRevision
```
salmanmkc, line 363: This updates `desiredStatus`, but the `newEphemeralRunnersByStates` call below still uses `latest.Status.AppliedActionableRevision`.


If an old runner was missed by the cached cleanup list, we can count its `Outdated` status against the old revision, then save that phase alongside the new revision. The set gets stuck because the `Outdated` reconcile path doesn't recalculate the phase.


We should pass `desiredStatus.AppliedActionableRevision` to the classifier instead.

salmanmkc (changes requested):

controllers/actions.github.com/ephemeralrunnerset_controller.go
salmanmkc: To clarify, I mean the `newEphemeralRunnersByStates` call on line 423 (420 on the left). Its second argument should be `desiredStatus.AppliedActionableRevision` rather than `latest.Status.AppliedActionableRevision`.

nikola-jokic: Correct, and it's a regression this PR introduced. Fixed in `c03a60e2` — thank you, this was a genuine land blocker.

Confirmed against `master`: the guard used to write in place, so the classifier reading `latest.Status.AppliedActionableRevision` observed the advanced value. Building the desired status as a value copy redirected the write without moving the read. The classifier line is **textually identical** and therefore absent from the diff, which is why it survived review and a green suite — the line changed meaning without changing text.

Your reachability argument holds exactly as you describe. The revisions only diverge when the guard fires, and a runner missed by the cache-read cleanup carries a revision equal to the pre-advance marker, so `< appliedActionableRevision` is false and it lands in `outdated` instead of `staleOutdated`. The resulting `Outdated` phase is then persisted next to the advanced revision, and nothing recomputes it: `Reconcile` returns on the Outdated path before `updateStatus`, and this function only runs while spec is ahead of applied. Stuck until the next spec change, as you said.

Changes:

- Pass `desiredStatus.AppliedActionableRevision`, as you suggested.
- Added `TestPatchAppliedActionableRevisionStatusClassifiesAgainstTheAdvancedRevision`. I verified it actually pins the bug rather than merely passing: reverted to `latest.Status...` and it fails with `expected: "Running", actual: "Outdated"` at revision 2 — your stuck state, reproduced.
- Extended the comment to say the value must come from `desiredStatus`, since the existing comment asserted `max(live, target)` and had been quietly falsified.

Worth noting why the existing coverage missed it: `TestPatchAppliedActionableRevisionStatusIgnoresATargetBehindTheLiveRevision` exercises the target-behind-live case, where the guard does **not** fire and both revisions agree — structurally incapable of observing this. It still passed with the bug present. The new test is its complement.

I also swept the rest of the diff for the same shape (a write redirected to a copy while a later unchanged line still reads the original). This is the only occurrence. `updateStatus`'s `desiredStatus` predates this PR and derives its classifier input from a parameter, and the `newLazyCopy` conversions are immune because `Mutate()` returns the live object, so reads through the original cannot drift from writes through `Mutate`.

Full `./controllers/... ./cmd/...` green.

rentziass (dismissed):

salmanmkc (approved):

PR #4583, opened 2026-07-26
Remove AutoscalingRunnerSet printer columns referencing removed status fields

Okabe-Junya:
## What

Remove the 5 `AutoscalingRunnerSet` printer columns that reference status fields missing from the schema, and regenerate the CRDs with `make manifests`. 
The hand-written change is 5 deleted marker lines; the rest is generated.

Before:

```console
$ k get autoscalingrunnersets
NAME              MINIMUM RUNNERS   MAXIMUM RUNNERS   CURRENT RUNNERS   PHASE   PENDING RUNNERS   RUNNING RUNNERS   FINISHED RUNNERS   DELETING RUNNERS
dummy-scale-set   1                 5
```

After:

```console
$ k get autoscalingrunnersets
NAME              MINIMUM RUNNERS   MAXIMUM RUNNERS   PHASE
dummy-scale-set   1                 5
```

## Why

#4557 removed the runner-count fields from `AutoscalingRunnerSetStatus` (they are now exposed as controller metrics) and dropped the matching printer columns on `EphemeralRunnerSet`, but the `AutoscalingRunnerSet` markers were left behind so the 5 columns above always render empty.

### DIFF
--- a/apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
+++ b/apis/actions.github.com/v1alpha1/autoscalingrunnerset_types.go
@@ -36,12 +36,7 @@ import (
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:JSONPath=".spec.minRunners",name=Minimum Runners,type=integer
 // +kubebuilder:printcolumn:JSONPath=".spec.maxRunners",name=Maximum Runners,type=integer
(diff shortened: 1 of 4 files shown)


Draco7587 (approved):

nikola-jokic (approved):

nikola-jokic:
Thank you!

PR #4516, opened 2026-06-02
Upgrade resources in-place, causing 1-1 mapping between autoscaling runner set and ephemeral runner set

nikola-jokic:
This PR changes upgrades significantly. Now, the upgrades unrelated to the listener would keep the listener running.
The upgrades to the ephemeral runner set will keep the ephemeral runner set. The ephemeral runner set on the other hand will delete idle runners (so it can quickly create new ones with the new spec), without touching the scale target. This is also important to allow ephemeral runner set to scale up to min runners (target count + min runners) so new runners are warm to pick up new jobs. If the change to the ephemeral runner set spec is related to patch metadata, the ephemeral runners **will not** get recreated and the new runners will eventually respect the updated metadata. The metadata patch is assumed to not be anything crucial, therefore, we should avoid removing idle runners.

However, when the listener-related spec is modified, the listener is re-created. This is not a change in behavior. However, it should make it quicker because the update patches the secret and re-creates the new listener.

If the change is unrelated to the ephemeral runner spec, the ephemeral runner keeps running normally.

The rest of the changes are related to checking the spec of resources. If they get modified, we try to revert them to the desired state, while keeping already applied annotations and labels from the outdated resource.

This is a breaking change. The upgrade strategy is no longer necessary.

Fixes #4432
Fixes #4346
Fixes #4424
Fixes #4489

PR #4584, opened 2026-07-26
Remove spec hash helpers orphaned by 4516

Okabe-Junya:
## What

Removes 

- `AutoscalingRunnerSet.ListenerSpecHash()`
- `AutoscalingRunnerSet.RunnerSetSpecHash()`

## Why

#4516 moved resource upgrades to the in-place model and replaced the`runner-spec-hash` annotation comparison with the integrity-hash annotation

That PR removed all four call sites of these two methods (`autoscalingrunnerset_controller.go`, `resourcebuilder.go`) but left the definitions behind; nothing references them anymore.


Okabe-Junya:
#4636 already covered this PR, so let me close then

PR #4585, opened 2026-07-27
Force-remove EphemeralRunner registration finalizer after deletion timeout

vnazzaro-cpu:
## What

- When an `EphemeralRunner` is being deleted and the Actions service keeps returning `JobStillRunningError` from `RemoveRunner`, stop requeueing forever: once deletion has been pending longer than a timeout **and** the runner pod is gone or in a terminal phase, force-remove `runner-registration-finalizer` so cleanup can finish.
- The timeout defaults to 10 minutes and is injectable via a new `RegistrationFinalizerForceTimeout` field on `EphemeralRunnerReconciler` (used by tests; a controller flag could follow if there is interest).
- Add envtest coverage for the three paths (force after timeout with pod gone, force after timeout with pod `Failed`, no force before timeout).

| Condition | Behavior |
|-----------|----------|
| `JobStillRunning`, pod exists and is not terminal | requeue every 30s (unchanged) |
| `JobStillRunning`, pod gone or terminal, deletion age < timeout | requeue every 30s (unchanged) |
| `JobStillRunning`, pod gone or terminal (`Succeeded`/`Failed`), deletion age ≥ timeout | force-remove registration finalizer, continue cleanup |

## Why

The deletion path currently has no deadline:

```go
if !ok {
	log.Info("Runner is not finished yet, retrying in 30s")
	return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
}
```

If the runner pod died mid-job (kubelet ephemeral-storage eviction, OOMKill, node loss) while the service still considers the job active, the `EphemeralRunner` stays `Terminating` with finalizers indefinitely, and operators end up running external CronJobs that patch `finalizers: null` (#4307). The controller owns this lifecycle; it should not need an external sweeper for a known stuck state.

A pod that is gone or in a terminal phase can never finish the job and unregister on its own, and the deletion path never deletes the pod before the registration finalizer clears (pod cleanup runs after finalizer removal, and ownerRef GC is blocked by the CR's own finalizers), so there is nothing left to wait for. We never force-remove while the pod is running — the runner may still finish and unregister normally.

Observed in production (2026-07, 0.14.0): a runner container hit its ephemeral-storage limit mid-job and was evicted; the `EphemeralRunner` remained `Terminating` until finalizers were manually cleared.

One trade-off worth stating: force-removing the finalizer skips a successful unregister call, so the runner record may linger on the service side until it reaps it. That seems preferable to a CR that can never finish deleting.

Related: #4155, #4307. #4340 improved failed-pod restart handling but does not bound the delete-path wait on `JobStillRunning`.

## Verification

- New envtest specs (`Registration finalizer force removal`) cover the three paths with `RemoveRunner` stubbed to return `JobStillRunningError`:
  - deleting + pod deleted + past timeout → registration finalizer force-removed, CR fully deletes
  - deleting + pod in `Failed` phase + past timeout → same
  - deleting + pod deleted + before timeout → registration finalizer kept
- The first spec also asserts the CR is never force-finalized while a non-terminal pod exists, even past the timeout. Specs are event-driven and use the injectable timeout, so they add ~15s to the suite.
- Full `controllers/actions.github.com` suite run repeatedly against envtest (Kubernetes 1.36): all specs pass; the new specs pass 5/5 consecutive focused runs.
- `go build`, `go vet`, `gofmt` clean.

### DIFF
--- a/controllers/actions.github.com/ephemeralrunner_controller.go
+++ b/controllers/actions.github.com/ephemeralrunner_controller.go
@@ -46,6 +46,10 @@ const (
 	ephemeralRunnerActionsFinalizerName = "ephemeralrunner.actions.github.com/runner-registration-finalizer"
 )
 
+// how long a deleting EphemeralRunner may stay blocked on JobStillRunning
+// before the registration finalizer is force-removed
+const defaultRegistrationFinalizerForceTimeout = 10 * time.Minute
+
 // EphemeralRunnerReconciler reconciles a EphemeralRunner object
 type EphemeralRunnerReconciler struct {
 	client.Client
@@ -53,6 +57,16 @@ type EphemeralRunnerReconciler struct {
 	Scheme         *runtime.Scheme
 	PublishMetrics bool
 	ResourceBuilder
+
+	// RegistrationFinalizerForceTimeout overrides defaultRegistrationFinalizerForceTimeout when set
+	RegistrationFinalizerForceTimeout time.Duration
+}
+
+func (r *EphemeralRunnerReconciler) registrationFinalizerForceTimeout() time.Duration {
+	if r.RegistrationFinalizerForceTimeout > 0 {
+		return r.RegistrationFinalizerForceTimeout
+	}
+	return defaultRegistrationFinalizerForceTimeout
 }
 
 var ephemeralRunnerPhaseMetrics = struct {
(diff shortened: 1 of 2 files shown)


bquenin:
We hit this exact scenario in production yesterday and can confirm the failure mode this PR addresses, including the missing deadline.

Setup: gha-runner-scale-set mode, ARC 0.13.1, runner 2.330.0, GitHub Enterprise Server 3.18, EKS (Kubernetes 1.34).

Sequence (all times UTC, single incident):

- 21:31:37 — EphemeralRunner created on scale-up; pod Ready, runner registered and listening.
- 21:33:25 — runner receives a job message and spawns `Runner.Worker`; job request lease renewed ("job is valid till 21:43:26").
- 21:33:56 — the worker never reads the job from the IPC channel within the 30s timeout; the listener kills the process tree ("both cancel and terminate signal has been ignored").
- 21:34:26 — the listener deletes its session and exits 0; ARC records "finished successfully (exitCode 0)". Pod containers reaped by 21:34:59.
- 21:35:02 → 21:46 — EphemeralRunner deletion blocked: `RemoveRunner` fails with `JobStillRunningError` on every attempt; the controller logs `Runner is not finished yet, retrying in 30s` indefinitely. The pod was gone for this entire window, so the force-finalizer conditions in this PR's table (`JobStillRunning`, pod gone/terminal, deletion age ≥ timeout) would have applied and released it at ~21:45.
- 21:46:31 — a human cancelled the stuck workflow job; the service immediately accepted the runner removal and cleanup completed within seconds.

One observation for reviewers: the workflow job itself stayed in `queued`, invisibly bound to the dead runner the whole time — the service never re-queued it, and only the manual cancel released it. That part is service-side and out of scope here, but it means this force-finalizer path will likely fire in exactly the situations where a job is also wedged. The 10-minute default matches the job lease duration we observed ("valid till" = start + 10m), which seems like a sensible floor.

The `Runner.Worker` IPC hang that triggered this is filed as actions/runner#4598.

PR #4620, opened 2026-09-06
Remove orphaned testserver package

Okabe-Junya:
## What

- Deletes `github/actions/testserver/`

## Why

The package was added in #2281 as a test double for the old
github/actions client. #4390 moved the controllers to the scaleset client and #4405 removed the old client together with its tests.


nikola-jokic (approved):
LGTM Thanks!

PR #4621, opened 2026-09-07
Add guards on timings so we don't publish incorrect metrics

### LINKED ISSUE #2506: Workflow job run duration metric shows wrong value if the job was canceled before it was started
### Checks

### Controller Version

0.27.2

### Helm Chart Version

0.23.1

### CertManager Version

1.9.1

### Deployment Method

Helm

### cert-manager installation

Cert manager was installed using the official helm chart from the jetstack repository.

### Checks

### Resource Definitions

```yaml
kind: RunnerSet
metadata:
  name: ga-runner-deployment-default-test-poc
  namespace: system-actions-runner
spec:
  labels:
  - test-poc
  selector:
    matchLabels:
      type: test-poc
  organization: #######
  serviceName: test-poc
  template:
    metadata:
      labels:
        type: test-poc
    spec:
      containers:
      - name: runner
        image: summerwind/actions-runner-dind:v2.303.0-ubuntu-20.04-3417c5a
        resources:
          limits:
            cpu: 2
            memory: 7Gi
          requests:
            cpu: 2
            memory: 7Gi
      tolerations:
      - effect: NoSchedule
        key: dedicated
        operator: Equal
        value: github_actions
      nodeSelector:
        apps: github-actions-runners
---
metadata:
  namespace: system-actions-runner
  name: ga-runner-deploy-autoscaler-default-test-poc
spec:
  scaleTargetRef:
    kind: RunnerSet
    name: ga-runner-deployment-default-test-poc
  scaleUpTriggers:
  - githubEvent:
      workflowJob: {}
    amount: 1
    duration: "5m"
  minReplicas: 1
  maxReplicas: 1
```

### To Reproduce

```markdown
1. Scale down your runners to 1;
2. Run two simple workflow runs (workflow jobs) that runs sleep 60 in a step;
3. Wait until one of workflows will be started on the runner;
4. Cancel second job which was not manage to start on the runner (waited in a queue);
5. Check the github_workflow_job_run_duration_seconds_sum metric value of your job - you will find it by job_conclusion="cancelled".
6. Check the actions-metrics-server /metrics endpoint. You will see giant value of your job run duration. Something like 9.223372036854776e+09 after the first try.
```

### Describe the bug

It seems we have a bug in a `actions-metrics-server` component.

It generates wrong value of the `github_workflow_job_run_duration_seconds_sum` metric if the job was queued but canceled before it started on some of runners.

It could happen if you have all of your runners are busy by other workflow jobs and one of queued jobs cancelled by user. So the job never started, but the metrics server increases the value of `github_workflow_job_run_duration_seconds_sum` belong for the cancelled workflow_job by some giant value.
In my case this is 9.223372036854776e+09. The job could be found by `job_conclusion="cancelled"` label.

image

Events of queued but not started (cancelled) job that goes from the GitHub to the metrics-server:

- workflow_run `requested`;
- workflow_job `queued`;
- workflow_run `in_progress`. It has `workflow_run.status: in_progress` and `workflow_run.conclusion: cancelled`;
- workflow_job `completed`.  It has `workflow_job.status: in_progress` and `workflow_job.conclusion: cancelled`.

It is easy to reproduce.
Scale down runners to 1 and start two separate workflow runs (workflow job). These workflows should have a `sleep 60` or something like that to let second job wait for the runner. Once you start these two jobs, just cancel one of them from the UI.

So your first workflow_job goes through the event `workflow_job.in_progress` with "status": "in_progress"`, and `"conclusion": null,`. But second workflow_job event will not have `in_progress` in action, but `completed` - never started on a runner. 

I am not sure where the problem is, it might be [here](https://github.com/actions/actions-runner-controller/blob/v0.27.2/pkg/actionsmetrics/event_reader.go#L156-L157) or [there](https://github.com/actions/actions-runner-controller/blob/v0.27.2/pkg/actionsmetrics/event_reader.go#L270) and [there](https://github.com/actions/actions-runner-controller/blob/v0.27.2/pkg/actionsmetrics/event_reader.go#L283) or somewhere else. I expect there is no start time in logs so may be some kind of conditions needs to be added.

I would expect that `github_workflow_job_run_duration_seconds_sum` metric will not be increased if the workflow job did not manage to run - when the metrics server never receive `workflow_job` event with `in_progress` status. Or if the job logs are blank or have no start time.

### Whole Controller Logs

```shell
https://gist.github.com/serhij-matvejev/1c56de86ddf9d00ccc0a4474ac6f9f41
```

### Whole Runner Pod Logs

```shell
Metrics server topic. The job never started on a runner. So there are no runner logs available.
```

### Additional Context

_No response_

### LINKED ISSUE #4611: gha_job_execution_duration_seconds records a large negative duration for jobs canceled after runner assignment
### What happened

`gha_job_execution_duration_seconds_sum` accumulates large negative values (roughly -6.4e10 per occurrence) for some canceled jobs, which makes `rate()` and `increase()` on the counter return negative results.

### Version

`gha-runner-scale-set` 0.14.2 (controller and listener).

### Cause

In `cmd/ghalistener/metrics/metrics.go`:

```go
func (e *exporter) RecordJobCompleted(msg *scaleset.JobCompleted) {
	if msg.RunnerAssignTime.IsZero() {
		return
	}

	l := e.completedJobLabels(msg)
	e.incCounter(MetricCompletedJobsTotal, l)
	e.observeHistogram(MetricJobExecutionDurationSeconds, l, float64(msg.FinishTime.Unix()-msg.RunnerAssignTime.Unix()))
}
```

`RunnerAssignTime` is checked but `FinishTime` is not. When a job is canceled after a runner has been assigned but before it finishes, `FinishTime` is the zero value and `FinishTime.Unix()` is `-62135596800`, so the observed duration is about `-62135596800 - <now>`.

Observed increments match that exactly: `-62135596800 - 1787769600 = -63923366400`.

This depends on when the cancellation happens:

| canceled | RunnerAssignTime | FinishTime | result |
| --- | --- | --- | --- |
| before runner assigned | zero | zero | returns early, correct |
| after assigned, before finish | set | zero | large negative observation |
| after finishing | set | set | correct |

#4472 addressed the first row. The second row still reproduces on 0.14.2.

`gha_job_startup_duration_seconds` is not affected, since both of its inputs are populated when it is recorded.

### Reproduction

Cancel a workflow run after a runner has picked up the job but before the job completes, then read `gha_job_execution_duration_seconds_sum` from the listener.

### Possible fix

Extending the existing guard would cover it:

```go
if msg.RunnerAssignTime.IsZero() || msg.FinishTime.IsZero() {
	return
}
```

That skips the observation entirely, consistent with how the `RunnerAssignTime` case is handled today. Recording a zero duration, or deriving the end time from another field, may be preferable if the completion should still be counted. Happy to defer to what fits the intended semantics.

Related: #3731, #2506, #4444

nikola-jokic:
Validate message data before publishing metrics to avoid publishing incorrect data.

Fixes #4611
Fixes #2506


rentziass (dismissed):

cmd/ghalistener/metrics/metrics.go
```diff
@@ -499,7 +502,9 @@ func (e *exporter) RecordJobCompleted(msg *scaleset.JobCompleted) {
 
 	l := e.completedJobLabels(msg)
 	e.incCounter(MetricCompletedJobsTotal, l)
-	e.observeHistogram(MetricJobExecutionDurationSeconds, l, float64(msg.FinishTime.Unix()-msg.RunnerAssignTime.Unix()))
+
+	finishDuration := max(msg.FinishTime.Unix()-msg.ScaleSetAssignTime.Unix(), 0)
```
salmanmkc: Doesn't the change from finish time minus assign time to finish time minus scale set assign time mean that a job waiting 60 seconds for a runner and executing for 10 seconds now reports 70 seconds of execution?

salmanmkc: Is `ScaleSetAssignTime` guaranteed to be populated whenever  RunnerAssignTime  is set? If not, max(..., 0)  would still allow a huge positive duration when the start timestamp is zero. In that case, could we skip the histogram observation while still incrementing the job counter? If it's always populated though  no additional guard is needed I think

salmanmkc (approved):
looks good to me, thanks for addressing the comments.

rentziass (approved):

Issue #3021, opened 2023-10-24
Allow option to configure the kubernetes controllers MaxConcurrentReconciles


StevenPJ:
### What would you like added?

*A clear and concise description of what you want to happen.*

We would like to be able to configure the [MaxConcurrentReconciles](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/config/controller.go#L36) of the runnerpod and runner controllers.

### Why is this needed?

The `runnerpod_controller` is currently making a lot of individual patches per runner pod (each one adds a different annotation to the pod). The controller is[ not configured to pass options to the k8s controller](https://github.com/actions/actions-runner-controller/blob/510b1d82e511d54283d1f35f5a182dd3debf5cc0/controllers/actions.summerwind.net/runner_pod_controller.go#L277), so defaults to a MaxConcurrentReconciles of `1`. There may be a way to configure this using system configuration, but could not find any docs in ARC or the controller-runtime project, and from looking at the code it looks like it should be passed in on creation. There may be an alternative means of configuring/injecting the setting as are not Golang or kubenetes-controller experts, in which case we would appreciate a pointer to the correct docs.

The high number of requests the controller makes to the k8s api, and the single threaded nature of the controller worker means it is constantly behind and puts a lot of pressure during scaling. We have run multiple configurations with the HRA, from:
* `min: 0. max:60`
* `min: 60. max:60`
* `min: 100. max:200`
* `min: 100. max:100`

and we have found it performs best when no scaling is applied i.e. `min == max`. While we see the queue depth stay constant (at around 1000 requests in the queue), we observed more consistent performance this way.

*A clear and concise description of any alternative solutions or features you've considered.*

- We have tried using ARC V2 but ran into issues using a single org runner group due to limited availability of the listener pod (a seperate issue but its why we havent upgraded just yet)
- We have configured different resources for the controller
- We have configured different replica/min/max settings and observe the same behaviour for all.

### Additional context

We are using the latest version of the summerwinds controller, using ephmeral runners with a `RunnerDeployment` and webhook driven scaling. We have a few different runnerDeployments, including our main one in the default organisation runner group. We forked the code and manually configured the  [MacConcurrentReconciles](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/config/controller.go#L36) to 8 and noticed a very clear correlation.

[...]

PR #4626, opened 2026-09-08
Add default and per-controller max-concurrent-reconciles flags

### LINKED ISSUE #4624: gha-runner-scale-set-controller: `--runner-max-concurrent-reconciles` only applies to EphemeralRunner; AutoscalingListener/AutoscalingRunnerSet/EphemeralRunnerSet reconcilers are hard-wired to 1 worker
### Checks

### Controller Version

gha-runner-scale-set-controller 0.14.2 (latest release), and `master` at 54147cf. Also present in `gha-runner-scale-set-controller-experimental`.

### Deployment Method

Helm

### Checks

### To Reproduce

```markdown
1. Install gha-runner-scale-set-controller with `flags.runnerMaxConcurrentReconciles: 20` (and, to rule out the token-bucket limiter, `flags.workqueueRateLimiter: typed_rate_limiter`) and `metrics.controllerManagerAddr: ":8080"`.
2. Install ~50 gha-runner-scale-set releases in one go (one `AutoscalingRunnerSet` each).
3. Scrape the controller-manager metrics endpoint.
4. `controller_runtime_max_concurrent_reconciles{controller="ephemeralrunner"}` is 20, but
   `{controller="autoscalinglistener"}`, `{controller="autoscalingrunnerset"}` and `{controller="ephemeralrunnerset"}` are all 1, regardless of the flag.
5. `workqueue_depth{name="autoscalinglistener"}` grows to the number of scale sets and listener Pods appear one at a time, while controller CPU stays low.
```

### Describe the bug

The controller exposes a single concurrency knob, `--runner-max-concurrent-reconciles` (chart value `flags.runnerMaxConcurrentReconciles`, default 2), but in `main.go` it is appended only to the options passed to `EphemeralRunnerReconciler`:

https://github.com/actions/actions-runner-controller/blob/54147cfa5eda5d9bbd798b99d40bf3f4b4402e82/main.go#L340-L347

```go
runnerOpts := append(controllerOpts, actionsgithubcom.WithMaxConcurrentReconciles(opts.RunnerMaxConcurrentReconciles))
if err = (&actionsgithubcom.EphemeralRunnerReconciler{...}).SetupWithManager(mgr, runnerOpts...); err != nil {
```

`AutoscalingRunnerSetReconciler` (L327-L335), `EphemeralRunnerSetReconciler` (L352-L357) and `AutoscalingListenerReconciler` (L362-L369) receive only `controllerOpts`, which carries just the rate-limiter choice (L312-L325). They therefore run with controller-runtime's default `MaxConcurrentReconciles = 1`.

Nothing in the reconcilers prevents higher concurrency: all three `SetupWithManager` methods already take `opts ...Option` and go through `builderWithOptions` (`controllers/actions.github.com/options.go#L66-L77`; `autoscalinglistener_controller.go#L841`, `autoscalingrunnerset_controller.go#L854`, `ephemeralrunnerset_controller.go#L636`). The option is simply never wired for them.

This matters in clusters that run many scale sets (for example one `AutoscalingRunnerSet` per team/repo, or short-lived scale sets created from CI). Each `AutoscalingListener` reconcile performs several API round trips (ServiceAccount, Role, RoleBinding, mirrored secret, listener Pod) and a single worker serialises all of them, so with tens of scale sets churning, listener Pod creation for a new scale set can take many minutes even though the controller is otherwise idle. `--workqueue-rate-limiter=typed_rate_limiter` (#4451) removes the global 10 QPS bucket but does not help here, because the limit is the single worker, not the queue rate. The in-place upgrade work in #4516 does not change this either; the wiring above is identical before and after it.

Related history: #3021 (original request, legacy controller), #3832 (added the flag for `EphemeralRunner` only). The flag's help text and `options.go` say "EphemeralRunner controller", so the current behaviour is as documented, but there is no way at all to raise concurrency for the other three reconcilers, and the chart value name (`runnerMaxConcurrentReconciles`) makes it easy to assume it covers the whole controller.

Either:

1. `--runner-max-concurrent-reconciles` (or a new general flag) also applies to the `AutoscalingRunnerSet`, `EphemeralRunnerSet` and `AutoscalingListener` reconcilers, or
2. separate flags/values exist for them, e.g. `--listener-max-concurrent-reconciles` and `--scale-set-max-concurrent-reconciles` (chart: `flags.listenerMaxConcurrentReconciles`, `flags.scaleSetMaxConcurrentReconciles`), defaulting to 1 so existing behaviour is unchanged.

Option 2 seems cleaner because the three reconcilers have different write patterns (`EphemeralRunnerSet` in particular does status/scale updates that are more sensitive to conflicts), so operators may want to tune them independently.

#### Proposed fix

- `controllers/actions.github.com/options.go`: add `ListenerMaxConcurrentReconciles` / `ScaleSetMaxConcurrentReconciles` to `Options` (default 1) and to `OptionsWithDefault()`.
- `main.go`: register the two flags next to `runner-max-concurrent-reconciles` (L154), and build per-reconciler option slices the same way `runnerOpts` is built at L340, passing them to the three `SetupWithManager` calls at L335, L357 and L369.
- `charts/gha-runner-scale-set-controller/values.yaml` (next to L117-L120) and `templates/deployment.yaml` (next to L70-L71): new `flags.*` values and `{{- with }}` blocks; same for `charts/gha-runner-scale-set-controller-experimental/values.yaml` (L31-L32) and `templates/_controller_template.tpl` (L50-L51).
- A case in `charts/gha-runner-scale-set-controller/tests/template_test.go` asserting the new args are rendered.

Roughly 40-60 lines plus tests; no CRD or RBAC changes.

### Additional Context

```yaml
# gha-runner-scale-set-controller values used
flags:
  runnerMaxConcurrentReconciles: 20
  workqueueRateLimiter: typed_rate_limiter
metrics:
  listenerAddr: ":8080"
  listenerEndpoint: "/metrics"
```

### Controller Logs

```shell
# Startup shows the option is read, but it is only applied to the EphemeralRunner controller (see main.go references above):
INFO Using options {"runner-max-concurrent-reconciles": 20}
INFO Using typed rate limiter (per-item only, no global token bucket)
# Metrics endpoint after startup:
controller_runtime_max_concurrent_reconciles{controller="ephemeralrunner"} 20
controller_runtime_max_concurrent_reconciles{controller="ephemeralrunnerset"} 1
controller_runtime_max_concurrent_reconciles{controller="autoscalingrunnerset"} 1
controller_runtime_max_concurrent_reconciles{controller="autoscalinglistener"} 1
```

### Runner Pod Logs

```shell
N/A - the behaviour is in the controller manager's controller registration, not in runner pods.
```

Found while operating this in production; happy to send a PR if maintainers agree with the analysis.

KR-Ravindra:
## Problem

The controller manager exposes one concurrency knob, `--runner-max-concurrent-reconciles` (chart value `flags.runnerMaxConcurrentReconciles`), but it only applies to the `EphemeralRunner` controller. The `AutoscalingListener`, `AutoscalingRunnerSet` and `EphemeralRunnerSet` controllers always run with controller-runtime's default of a single worker, and there is no flag or chart value that can change that.

In clusters with many runner scale sets (one `AutoscalingRunnerSet` per team/repo, or short-lived scale sets created from CI), the single `AutoscalingListener` worker serialises every listener reconcile (ServiceAccount, Role, RoleBinding, mirrored secret, listener Pod), so `workqueue_depth{name="autoscalinglistener"}` grows and listener Pods for new scale sets appear one at a time while the controller is otherwise idle. `--workqueue-rate-limiter=typed_rate_limiter` (#4451) does not help because the bottleneck is the worker count, not the queue rate.

`controller_runtime_max_concurrent_reconciles` on `master` (54147cf) with `runnerMaxConcurrentReconciles: 20`:

```
controller_runtime_max_concurrent_reconciles{controller="ephemeralrunner"} 20
controller_runtime_max_concurrent_reconciles{controller="ephemeralrunnerset"} 1
controller_runtime_max_concurrent_reconciles{controller="autoscalingrunnerset"} 1
controller_runtime_max_concurrent_reconciles{controller="autoscalinglistener"} 1
```

## Root cause

`main.go:340` builds `runnerOpts := append(controllerOpts, actionsgithubcom.WithMaxConcurrentReconciles(opts.RunnerMaxConcurrentReconciles))` and passes it only to `EphemeralRunnerReconciler.SetupWithManager` (`main.go:347`). The other three reconcilers (`main.go:335`, `main.go:357`, `main.go:369`) receive plain `controllerOpts`, which carries only the workqueue rate-limiter choice, so `controller.Options.MaxConcurrentReconciles` stays at 0 and controller-runtime defaults it to 1.

All three `SetupWithManager` methods already accept `opts ...Option` and go through `builderWithOptions` (`controllers/actions.github.com/options.go:66`), so nothing in the reconcilers prevents higher concurrency; the option was simply never wired for them (#3832 added it for `EphemeralRunner` only).

## Fix

Following review feedback, the concurrency knobs are now one default plus one flag per controller. Any per-controller flag left unset falls back to the default, and the default is 1. Following the second review round, the charts no longer carry the legacy `runnerMaxConcurrentReconciles` value: the `EphemeralRunner` override is `ephemeralRunnerMaxConcurrentReconciles` and, like the other overrides, is unset by default. A default chart install therefore renders no concurrency flags and every controller runs with 1 worker (`EphemeralRunner` previously rendered 2). The bare binary behaves the same. `Options.Resolve()` also replaces a default of 0 or less with 1 explicitly rather than relying on controller-runtime.

- `--default-max-concurrent-reconciles` (default 1), applied to every controller without its own value
- `--autoscaling-runner-set-max-concurrent-reconciles`
- `--autoscaling-listener-max-concurrent-reconciles`
- `--ephemeral-runner-set-max-concurrent-reconciles`
- `--ephemeral-runner-max-concurrent-reconciles` (renamed from `--runner-max-concurrent-reconciles`; the old flag is removed, 0.15.0 does not promise flag compatibility)

Changes:

- `controllers/actions.github.com/options.go`: one `Options` field per flag, `OptionsWithDefault()` sets only the default, and `Options.Resolve()` clamps a default of 0 or less to 1 and replaces every zero per-controller value with the default. Unit test in `options_test.go`.
- `main.go`: register the five flags, call `opts.Resolve()` after `flag.Parse()`, log all five resolved values, and pass a `WithMaxConcurrentReconciles(...)` option to each of the four `SetupWithManager` calls.
- `charts/gha-runner-scale-set-controller`: new `flags.defaultMaxConcurrentReconciles`, `flags.autoscalingRunnerSetMaxConcurrentReconciles`, `flags.autoscalingListenerMaxConcurrentReconciles`, `flags.ephemeralRunnerSetMaxConcurrentReconciles`, `flags.ephemeralRunnerMaxConcurrentReconciles` (documented, commented out). `flags.runnerMaxConcurrentReconciles` is removed.
- `charts/gha-runner-scale-set-controller-experimental`: the same five values under `controller.manager.config` (`null` by default), rendered by `templates/_controller_template.tpl`.

No CRD or RBAC changes.

## How tested

- `controllers/actions.github.com/options_test.go`: `TestOptionsResolve` covers the all-default case, a default of 0 or -1, and a changed default with one per-controller override.
- `charts/gha-runner-scale-set-controller/tests/template_test.go`: `TestTemplate_ControllerDeployment_MaxConcurrentReconciles` sets all five values and asserts the rendered args; the default-args assertions expect no concurrency flag.
- `charts/gha-runner-scale-set-controller-experimental/tests/controller_deployment_args_test.yaml` (helm-unittest): defaults render no concurrency flag; setting all five values renders all five flags.

```
go build ./... && go vet . ./controllers/actions.github.com/   # clean
go test ./controllers/actions.github.com/ -run TestOptionsResolve   # ok
go test ./charts/gha-runner-scale-set-controller/...              # ok
helm unittest ./charts/gha-runner-scale-set-controller-experimental/
Test Suites: 18 passed, 18 total
Tests:       42 passed, 42 total
helm lint charts/gha-runner-scale-set-controller/ charts/gha-runner-scale-set-controller-experimental/   # 0 failed

go run . --help | grep max-concurrent-reconciles
  -autoscaling-listener-max-concurrent-reconciles int
  -autoscaling-runner-set-max-concurrent-reconciles int
  -default-max-concurrent-reconciles int   (default 1)
  -ephemeral-runner-max-concurrent-reconciles int
  -ephemeral-runner-set-max-concurrent-reconciles int
```

Note for reviewers: `go test ./controllers/actions.github.com/...` currently fails on `master` (54147cf) in the pre-existing spec "AutoscalingRunnerSet with a stale runner scale set ... registers the runner scale set again" (same failure in the upstream `test` job for master, run 34118603425); this PR does not touch that path.

## Links

- Fixes https://github.com/actions/actions-runner-controller/issues/4624
- Related: #3021 (original request), #3832 (flag added for `EphemeralRunner` only), #4451 (workqueue rate limiter)

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

### DIFF
--- a/charts/gha-runner-scale-set-controller/values.yaml
+++ b/charts/gha-runner-scale-set-controller/values.yaml
@@ -114,20 +114,20 @@ flags:
   ## Defaults to watch all namespaces when unset.
   # watchSingleNamespace: ""
 
+  ## The maximum number of concurrent reconciles applied to every controller that does not set its own value below.
+  # Increase this value to improve the throughput of the controller when running many runner scale sets.
+  # It may also increase the load on the API server and the external service (e.g. GitHub API).
+  # defaultMaxConcurrentReconciles: 1
+
+  ## Per-controller overrides. Each defaults to defaultMaxConcurrentReconciles when unset.
+  # autoscalingRunnerSetMaxConcurrentReconciles: 1
+  # autoscalingListenerMaxConcurrentReconciles: 1
+  # ephemeralRunnerSetMaxConcurrentReconciles: 1
+
   ## The maximum number of concurrent reconciles which can be run by the EphemeralRunner controller.
   # Increase this value to improve the throughput of the controller.
   # It may also increase the load on the API server and the external service (e.g. GitHub API).
   runnerMaxConcurrentReconciles: 2
-
-  ## The maximum number of concurrent reconciles which can be run by the AutoscalingListener controller.
-  # Increase this value to improve the throughput of the controller when running many runner scale sets.
-  # It may also increase the load on the API server.
-  # listenerMaxConcurrentReconciles: 1
-
-  ## The maximum number of concurrent reconciles which can be run by the AutoscalingRunnerSet and EphemeralRunnerSet controllers.
-  # Increase this value to improve the throughput of the controller when running many runner scale sets.
-  # It may also increase the load on the API server.
-  # scaleSetMaxConcurrentReconciles: 1
 
   ## Defines a list of prefixes that should not be propagated to internal resources.
   ## This is useful when you have labels that are used for internal purposes and should not be propagated to internal resources.
--- a/charts/gha-runner-scale-set-controller/values.yaml
+++ b/charts/gha-runner-scale-set-controller/values.yaml
@@ -119,6 +119,16 @@ flags:
   # It may also increase the load on the API server and the external service (e.g. GitHub API).
   runnerMaxConcurrentReconciles: 2
 
+  ## The maximum number of concurrent reconciles which can be run by the AutoscalingListener controller.
+  # Increase this value to improve the throughput of the controller when running many runner scale sets.
+  # It may also increase the load on the API server.
+  # listenerMaxConcurrentReconciles: 1
+
+  ## The maximum number of concurrent reconciles which can be run by the AutoscalingRunnerSet and EphemeralRunnerSet controllers.
+  # Increase this value to improve the throughput of the controller when running many runner scale sets.
+  # It may also increase the load on the API server.
+  # scaleSetMaxConcurrentReconciles: 1
+
   ## Defines a list of prefixes that should not be propagated to internal resources.
   ## This is useful when you have labels that are used for internal purposes and should not be propagated to internal resources.
   ## See https://github.com/actions/actions-runner-controller/issues/3533 for more information.
--- a/charts/gha-runner-scale-set-controller/values.yaml
+++ b/charts/gha-runner-scale-set-controller/values.yaml
@@ -123,11 +123,7 @@ flags:
   # autoscalingRunnerSetMaxConcurrentReconciles: 1
   # autoscalingListenerMaxConcurrentReconciles: 1
   # ephemeralRunnerSetMaxConcurrentReconciles: 1
-
-  ## The maximum number of concurrent reconciles which can be run by the EphemeralRunner controller.
-  # Increase this value to improve the throughput of the controller.
-  # It may also increase the load on the API server and the external service (e.g. GitHub API).
-  runnerMaxConcurrentReconciles: 2
+  # ephemeralRunnerMaxConcurrentReconciles: 1
 
   ## Defines a list of prefixes that should not be propagated to internal resources.
   ## This is useful when you have labels that are used for internal purposes and should not be propagated to internal resources.
--- a/controllers/actions.github.com/options.go
+++ b/controllers/actions.github.com/options.go
@@ -40,10 +40,14 @@ func OptionsWithDefault() Options {
 	}
 }
 
-// Resolve returns a copy of the options where every per-controller
+// Resolve returns a copy of the options where a DefaultMaxConcurrentReconciles
+// of zero or less is replaced by 1, and every per-controller
 // MaxConcurrentReconciles that is left at zero is replaced by
 // DefaultMaxConcurrentReconciles.
 func (o Options) Resolve() Options {
+	if o.DefaultMaxConcurrentReconciles <= 0 {
+		o.DefaultMaxConcurrentReconciles = 1
+	}
 	orDefault := func(n int) int {
 		if n > 0 {
 			return n
--- a/controllers/actions.github.com/options.go
+++ b/controllers/actions.github.com/options.go
@@ -10,17 +10,25 @@ import (
 // Options is the optional configuration for the controllers, which can be
 // set via command-line flags or environment variables.
 type Options struct {
-	// RunnerMaxConcurrentReconciles is the maximum number of concurrent Reconciles which can be run
-	// by the EphemeralRunnerController.
-	RunnerMaxConcurrentReconciles int
+	// DefaultMaxConcurrentReconciles is the maximum number of concurrent Reconciles
+	// applied to every controller that does not have its own value set below.
+	DefaultMaxConcurrentReconciles int
 
-	// ListenerMaxConcurrentReconciles is the maximum number of concurrent Reconciles which can be run
-	// by the AutoscalingListenerController.
-	ListenerMaxConcurrentReconciles int
+	// AutoscalingRunnerSetMaxConcurrentReconciles is the maximum number of concurrent Reconciles
+	// which can be run by the AutoscalingRunnerSetController. Zero means DefaultMaxConcurrentReconciles.
+	AutoscalingRunnerSetMaxConcurrentReconciles int
 
-	// ScaleSetMaxConcurrentReconciles is the maximum number of concurrent Reconciles which can be run
-	// by the AutoscalingRunnerSetController and the EphemeralRunnerSetController.
-	ScaleSetMaxConcurrentReconciles int
+	// AutoscalingListenerMaxConcurrentReconciles is the maximum number of concurrent Reconciles
+	// which can be run by the AutoscalingListenerController. Zero means DefaultMaxConcurrentReconciles.
+	AutoscalingListenerMaxConcurrentReconciles int
+
+	// EphemeralRunnerSetMaxConcurrentReconciles is the maximum number of concurrent Reconciles
+	// which can be run by the EphemeralRunnerSetController. Zero means DefaultMaxConcurrentReconciles.
+	EphemeralRunnerSetMaxConcurrentReconciles int
+
+	// EphemeralRunnerMaxConcurrentReconciles is the maximum number of concurrent Reconciles
+	// which can be run by the EphemeralRunnerController. Zero means DefaultMaxConcurrentReconciles.
+	EphemeralRunnerMaxConcurrentReconciles int
 }
 
 // OptionsWithDefault returns the default options.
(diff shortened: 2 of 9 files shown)


KR-Ravindra:
Round 1 self-review.

Checked against `master` (54147cf): `main.go` passes `WithMaxConcurrentReconciles` only to `EphemeralRunnerReconciler`, the other three reconcilers get plain `controllerOpts`, and all three `SetupWithManager` methods already route options through `builderWithOptions`, so no reconciler changes are needed. `go build ./...`, `go vet . ./controllers/actions.github.com/` and `go test ./charts/gha-runner-scale-set-controller/...` pass here. Defaults of 1 keep current behaviour, and the charts render nothing when the new values are unset. The only mention of `runnerMaxConcurrentReconciles` outside the charts is a changelog line in `docs/gha-runner-scale-set-controller/README.md`, so no docs update is needed. Marking ready for review.

nikola-jokic:
Hey,

I think this is very useful! Just one change I'd make. Can we expose the following:
```
--default-max-concurrent-reconciles  Default concurrent reconciles applied, defaults to 1
--autoscaling-runner-set-max-concurrent-reconciles Applied to autoscaling runner set defaults to --default-max-concurrent-reconciles
--autoscaling-listener-max-concurrent-reconciles Applied to the listener, defaults to --default-max-concurrent-reconciles
--ephemeral-runner-set-max-concurrent-reconciles Applied to ephemeral runner set, defaults to --default-max-concurrent-reconciles
--ephemeral-runner-max-concurrent-reconciles Applied to ephemeral runner, just rename the runner-max-concurrent-reconciles, also defaults to --default-max-concurrent-reconciles
```

The reason I'd go with renaming is because the next release is 0.15.0 which doesn't promise backward compatibility, so let's be consistent with the naming.

KR-Ravindra:
Thanks,  @nikola-jokic  done in 7748547. The flags are now exactly the five you listed: `--default-max-concurrent-reconciles` (default 1) plus `--autoscaling-runner-set-`, `--autoscaling-listener-`, `--ephemeral-runner-set-` and `--ephemeral-runner-max-concurrent-reconciles`, each falling back to the default when unset. `--runner-max-concurrent-reconciles` is removed. `Options.Resolve()` does the fallback after `flag.Parse()` and has a unit test.

Both charts render the new flags. I kept the existing `runnerMaxConcurrentReconciles` chart value name (it now renders `--ephemeral-runner-max-concurrent-reconciles`) so current values files keep working; happy to rename it to `ephemeralRunnerMaxConcurrentReconciles` as well if you prefer consistency there too.

KR-Ravindra:
On the two automated review points, for whoever picks this up:

1. Bare-binary default for `EphemeralRunner`: correct, it goes from the previous hard-coded 2 in `OptionsWithDefault()` to 1, because every controller now falls back to `--default-max-concurrent-reconciles` as requested above. Both charts still render `--ephemeral-runner-max-concurrent-reconciles=2` from `runnerMaxConcurrentReconciles: 2`, so chart deployments are unchanged. I have updated the description to say this explicitly rather than "exactly as before". Happy to seed 2 for `EphemeralRunner` instead if you would rather keep the binary's old default.
2. Values below 1: controller-runtime v0.24 sets `MaxConcurrentReconciles` to 1 when the option is `<= 0` (`pkg/controller/controller.go`, `NewTypedUnmanaged`), so `0` or a negative value cannot disable reconciliation; it behaves like the documented default. I left out an extra guard to keep the change small, but can add a fail-fast in `Resolve()` if you prefer.

KR-Ravindra:
@nikola-jokic CI note: the `test` job on 7748547 fails only in the pre-existing spec "AutoscalingRunnerSet with a stale runner scale set ... registers the runner scale set again" (75 of 76 specs pass; the failure is a nil-pointer panic in the autoscalingrunnerset reconciler during that spec). The same spec fails identically on master at 54147cf in run 34118603425, so it is unrelated to the flag changes here.

humh25:
536975948763

humbertohlgl

charts/gha-runner-scale-set-controller/values.yaml
nikola-jokic: We should remove this flag

controllers/actions.github.com/options.go
```diff
@@ -10,18 +10,51 @@ import (
...
-		RunnerMaxConcurrentReconciles: 2,
+		DefaultMaxConcurrentReconciles: 1,
+	}
+}
+
+// Resolve returns a copy of the options where every per-controller
+// MaxConcurrentReconciles that is left at zero is replaced by
+// DefaultMaxConcurrentReconciles.
+func (o Options) Resolve() Options {
+	orDefault := func(n int) int {
```
nikola-jokic, line 51: I know that controller runtime has a check for values <= 0, but I'd add the following line:
if o.DefaultMaxConcurrentReconciles <= 0 {
  o.DefaultMaxConcurrentReconciles = 1
}
Let's not rely on the default controller-runtime behavior when we can be explicit here

nikola-jokic:
Not right now, I'd love to take time to set the default value to something that would make sense. 2 is a bit low, so it should either by default be 1 (like controller runtime has it) or should be something tested that should be okay on most clusters

nikola-jokic (approved):
LGTM! Thanks!

PR #4054, opened 2025-04-22
Add job_workflow_ref label to listener metrics

rskmm0chang:
fix: https://github.com/actions/actions-runner-controller/issues/4053

PR #4633, opened 2026-09-08
Document which listenerMetrics example labels are not exposed by default

KR-Ravindra:
## Problem

`charts/gha-runner-scale-set/values.yaml` (and the `listener.metrics` block in `charts/gha-runner-scale-set-experimental/values.yaml`) ship a commented-out `listenerMetrics` example introduced as the defaults ("Default buckets are provided here for documentation purposes") and tell users to uncomment it when they want custom histogram buckets.

The example is not the default configuration. It adds `job_workflow_ref`, `job_workflow_name` and `job_workflow_target` to both counters and both histograms, while the listener's built-in `defaultMetrics` has none of them. Anyone who uncomments the block to tune buckets silently starts exporting per-ref/per-PR labels on histograms with 45 buckets, which multiplies the series count on busy repositories by orders of magnitude. Nothing in the file mentions the cardinality cost of those labels.

## Root cause

- `cmd/ghalistener/metrics/metrics.go:162-281` (`defaultMetrics`): counters and histograms carry only `enterprise`, `organization`, `repository`, `job_name`, `event_name` (+ `job_result`).
- The values example lists the three `job_workflow_*` labels on top of the defaults while the surrounding comment calls it the defaults.

`job_workflow_ref` was dropped from the defaults in #3671 because of cardinality (#3670, #3153) and re-added in #4054 as an opt-in label plus this example; #4240 then added `job_workflow_name`/`job_workflow_target` to the example only.

## Fix

Values comments plus a test; no template or Go behaviour changes. Per the review, the example keeps every metric and every label the listener can export, and documents which labels are not part of the defaults instead of removing them.

- In both charts, mark `job_workflow_ref`, `job_workflow_name` and `job_workflow_target` with a `# not exposed by default:` comment inside each label list, and rewrite the header: the example lists every metric and label, it is not identical to the built-in defaults, remove the marked labels to keep the default label set, and the two ref-based labels create a new series per branch/pull request per job (times the bucket count on histograms).
- Add `cmd/ghalistener/metrics/chart_values_test.go`: uncomments the example block from both values files and asserts that (a) the metric names equal `defaultMetrics`, (b) each example label list equals the full label set the exporter populates for that metric (`startedJobLabels`, `completedJobLabels`, `scaleSetLabels`), (c) the labels beyond `defaultMetrics` are exactly the three documented opt-in labels (none for gauges), and (d) the buckets equal the defaults. A new exporter label that is not added to the example, or a change to the defaults that is not reflected in the comment, fails the test.

## How tested

- `go test ./cmd/ghalistener/metrics/`, `go vet`, `gofmt -l`: pass/clean.
- Negative check: removing `job_workflow_name` from one example fails the test on both the "labels" and "labels not exposed by default" assertions.
- `helm lint` on both charts: clean.
- `helm template` with default values for both charts: rendered output is byte-identical to `master` (the block is commented out; the chart diff is comment-only).
- `helm template -f <uncommented example>` on the main chart: renders both counters with 8/9 labels including the three `job_workflow_*` labels, 8 gauges, and both histograms with 45 buckets, so the in-list `# not exposed by default:` comment parses as a YAML comment.

## Links

- Issue: https://github.com/actions/actions-runner-controller/issues/4625
- Related history: #3671, #4054, #4240
- Related open PRs touching the same comment block: #4525 (label sources), #4606 (new labels)

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

### DIFF
--- a/charts/gha-runner-scale-set-experimental/values.yaml
+++ b/charts/gha-runner-scale-set-experimental/values.yaml
@@ -323,15 +323,22 @@ listener:
   ## metrics configuration for the listener.
   ## In order to avoid helm merging these fields, we left the metrics commented out.
   ## When configuring metrics, please uncomment the metrics object below.
-  ## You can modify the configuration to remove the label or specify custom buckets for histogram.
+  ## You can modify the configuration to add or remove labels or specify custom buckets for histograms.
   ##
-  ## If the buckets field is not specified, the default buckets will be applied. Default buckets are
-  ## provided here for documentation purposes
+  ## The configuration below matches the built-in defaults used when listener.metrics is not set, so
+  ## uncommenting it as-is does not change the exported metrics. If the buckets field is not specified,
+  ## the default buckets will be applied.
+  ##
+  ## Counters and histograms also accept the optional labels "job_workflow_ref", "job_workflow_name"
+  ## and "job_workflow_target". "job_workflow_ref" and "job_workflow_target" contain the git ref the
+  ## workflow ran from (for example refs/heads/<branch> or refs/pull/<number>/merge), so every branch and
+  ## pull request creates a new time series per job. On histograms this is multiplied by the number of
+  ## buckets. Only add these labels when your metrics backend can handle the resulting cardinality.
   # metrics:
   #   counters:
   #     gha_started_jobs_total:
   #       labels:
-  #         ["repository", "organization", "enterprise", "job_name", "event_name", "job_workflow_ref", "job_workflow_name", "job_workflow_target"]
+  #         ["repository", "organization", "enterprise", "job_name", "event_name"]
   #     gha_completed_jobs_total:
   #       labels:
   #         [
@@ -341,9 +348,6 @@ listener:
   #           "job_name",
   #           "event_name",
   #           "job_result",
-  #           "job_workflow_ref",
-  #           "job_workflow_name",
-  #           "job_workflow_target",
   #         ]
   #   gauges:
   #     gha_assigned_jobs:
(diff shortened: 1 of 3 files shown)


KR-Ravindra (commented):
Round 1 self-review.

1. Checked `cmd/ghalistener/metrics/metrics.go` `defaultMetrics` on master: counters and histograms carry `enterprise`, `organization`, `repository`, `job_name`, `event_name` (+ `job_result`) only, no `job_workflow_*`, so the example in both values files was a superset of the defaults as described.
2. Ran `go test ./cmd/ghalistener/metrics/ -run TestChartValuesMetricsExampleMatchesDefaults` on this branch (pass) and again with the two values files replaced by their master versions (fails with `extra elements in list B: job_workflow_ref, job_workflow_name, job_workflow_target` for all four metrics), so the test guards the drift it is meant to.
3. The uncommenting helper keys on the exact `# listenerMetrics:` / `# metrics:` line and stops at the first line without the comment prefix, which matches the block layout in both files.

Default render is unchanged since the block stays commented out. Workflows need maintainer approval to run here; marking ready.

charts/gha-runner-scale-set-experimental/values.yaml
```diff
@@ -365,7 +369,7 @@ listener:
   #   histograms:
   #     gha_job_startup_duration_seconds:
   #       labels:
-  #         ["repository", "organization", "enterprise", "job_name", "event_name","job_workflow_ref", "job_workflow_name", "job_workflow_target"]
```
nikola-jokic, line 368: Hey, this is a good suggestion, but I'd prefer if we include all and document which ones are not exposed by default. Can you do that please?

KR-Ravindra: @nikola-jokic done in a9e2b43: both charts keep all labels in the example, with the three job_workflow_* labels marked `# not exposed by default:` inside each label list and the header rewritten to say the example is the full set, not the defaults. The test now checks the example lists every label the exporter populates and that the labels beyond `defaultMetrics` are exactly those three. Could you take another look?

nikola-jokic (approved):
Thanks you!

PR #4637, opened 2026-09-09
Render label and annotation values as strings

nikola-jokic:
## The bug

Kubernetes only accepts **string** values for labels and annotations. Both `gha-runner-scale-set` and `gha-runner-scale-set-experimental` rendered metadata values with a raw `toYaml` or a bare `| quote`, so a values file like:

```yaml
annotations:
  argocd.argoproj.io/sync-wave: 1
labels:
  enabled: true
```

rendered a YAML number and a YAML boolean. The resulting manifests are rejected by the API server.

## The fix

Every metadata value now goes through a `metadataValue` helper (`gha-runner-scale-set.metadataValue` in the stable chart, `metadata-value` in the experimental one), and every metadata map through a matching `stringMap` / `string-map` helper. The two charts are deliberate mirrors with different naming conventions.

## The float64 trap

The non-obvious part, and the reason the helper is not just `printf "%v"`:

Values loaded from a `-f values.yaml` file arrive in Helm as **float64**, while `--set` yields int64/string. So `printf "%v"` turns `12345678901234` into `"1.2345678901234e+13"` — a silently corrupted value the user never asked for. Integral floats are therefore formatted with `printf "%.0f"`.

The same asymmetry means Go tests using `SetValues` **cannot** exercise this behavior at all, so the new `TestTemplateRenderedAutoScalingRunnerSet_ScalarMetadataValuesAreRenderedAsStrings` drives a real values file (`tests/values_scalar_metadata.yaml`) and asserts `chart-big-int-annotation` renders as `"12345678901234"`.

## Verification

- `go test ./charts/... -count=1` — pass
- `helm unittest charts/gha-runner-scale-set-experimental` — 176 tests pass
- `helm lint` on both charts — pass
- Mutation-tested: dropping the `%.0f` branch fails the test with `expected: "12345678901234", actual: "1.2345678901234e+13"`; reverting `stringMap` back to `toYaml` fails the decode outright. Restored, both pass.

## Stack

This is **layer 1 of a 3-PR stack** splitting #4630 into reviewable pieces. It contains string coercion only. Render-time validation of label/annotation keys and values (layer 2) and listener metadata coverage (layer 3) follow on top of this branch.

### DIFF
--- a/charts/gha-runner-scale-set-experimental/templates/_autoscalingrunnerset.tpl
+++ b/charts/gha-runner-scale-set-experimental/templates/_autoscalingrunnerset.tpl
@@ -32,11 +32,11 @@ Render a ResourceMeta block for AutoscalingRunnerSet spec fields.
 {{- define "autoscaling-runner-set.spec-resource-metadata" -}}
 {{- with .labels }}
 labels:
-  {{- toYaml . | nindent 2 }}
+  {{- include "string-map" . | nindent 2 }}
 {{- end }}
 {{- with .annotations }}
 annotations:
-  {{- toYaml . | nindent 2 }}
+  {{- include "string-map" . | nindent 2 }}
 {{- end }}
 {{- end }}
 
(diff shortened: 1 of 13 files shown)


humh25:
536975948763

humbertohlgl

humh25:
536975948763

humbertohlgl

rentziass (approved):

PR #4673, opened 2026-09-21
Hand a finished runner's pod back as soon as its job is over

nikola-jokic:
## What this changes

When a runner pod's containers have all exited there is nothing left to shut down, but a normal delete still leaves the object in `Terminating` for the duration of its `terminationGracePeriodSeconds`. During that wait the pod keeps its name, its scheduling slot and its share of any `ResourceQuota`, so the runners queued behind it cannot start. The grace period buys nothing here: the kubelet unmounts volumes and tears down the sandbox whether or not the API object is still there.

Finished runner pods are now deleted with an explicit grace period of zero, so the capacity comes back as the delete is issued rather than a termination grace period later.

## How it works

`podTerminated` decides whether a pod has anything left running in it, and `deletePodOptions` turns that into the delete options. The grace period is only ever skipped for a pod with nothing running; a pod that is still alive is always deleted normally, since dropping the object early would leave the kubelet killing containers with nothing in the API accounting for what they hold.

The pod is released during finalization, which the runner's own deletion triggers on the very next reconcile. Deleting it earlier, on the reconcile that observes the clean exit, would emit a pod deletion on the pod informer just after the status patch goes to the runner informer, and nothing orders those two streams — the reconcile that deletion wakes could read a runner that is not yet `Succeeded` and build a replacement pod from a JIT config that has already been used.

The delete also carries a `Preconditions{UID}`. Runner pods are named after their `EphemeralRunner`, so the name is reused across pod generations, and the pod being judged is read from the informer cache. Under a single leader that cache cannot show a pod that no longer exists while its successor does, but the single-writer assumption breaks during a leader-election handover: an outgoing leader's in-flight reconcile is not cancelled the moment it loses the lease. The precondition makes such a delete fail rather than land on the wrong pod, which matters more now that the delete is unsurvivable instead of a `SIGTERM` the runner could deregister on.

## Configuration

`--terminated-runner-pod-grace-period-seconds` (chart value `terminatedRunnerPodGracePeriodSeconds`) controls the grace period, default `0`. Raise it to keep finished pods around longer, for example for a log collector. A negative value sends no delete option at all, restoring the previous behaviour.

## Also in here

**Batched creates and deletes in the EphemeralRunnerSet controller.** Scaling used to create and delete runners one API call at a time, so a large scale-up was a long serial walk. They now go out in batches of 8. This is concurrency in a single reconcile's API calls, not reconcile concurrency, so it is a different axis from the benchmark in #4670, and `EphemeralRunnerMaxConcurrentReconciles` is untouched at its benchmarked default of 4.

**A deep copy in `newEphemeralRunner`.** The builder handed every runner a spec aliasing the one in the `EphemeralRunnerSet`, which is a shared cache object. That was survivable while runners were built serially; with batching it is not.

**A scheme fix in `setControllerReference`.** A builder constructed without a scheme used to build one and assign it back to itself. Under batching that assignment is a write shared with every goroutine reading the same field, and a goroutine can read another's scheme after assignment but before its types are registered, failing the ownership call with an unknown kind. It now builds into a local and leaves the builder alone.

## Testing

Unit tests cover `podTerminated` and `deletePodOptions`, including that a still-running pod yields no options at all and that the delete carries the UID of the pod the reconcile actually looked at. The UID precondition has to be pinned at the options level: the controller-runtime fake client only honours the `ResourceVersion` precondition, so a fake delete would pass either way. The envtest suite runs against a real apiserver, which does enforce it, and confirms a finished runner's pod disappears promptly through finalization.


controllers/actions.github.com/ephemeralrunner_controller.go
```diff
@@ -484,14 +497,108 @@ func (r *EphemeralRunnerReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+// Anything that does not go to plan falls back to that route by leaving the
+// finalizers alone: the deletion still happens, and the deletion reconcile
+// finishes the work exactly as it does today. The pod and the secret are owned
+// by the runner, so even a crash between the patch and the deletion leaves them
+// to the garbage collector rather than behind.
+func (r *EphemeralRunnerReconciler) deleteSucceededRunner(ctx context.Context, ephemeralRunner *v1alpha1.EphemeralRunner, pod *corev1.Pod, log logr.Logger) error {
+	if r.releaseFinishedRunnerPod(ctx, pod, log) && r.releaseSucceededRunnerFinalizers(ctx, ephemeralRunner, log) {
+		// Bookkeeping the deletion reconcile would have done. Both are local to
+		// this process, and both are restored by the next reconcile if the
+		// deletion below fails: the phase metric is republished from the recorded
```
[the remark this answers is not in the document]
nikola-jokic: Correct, and already gone in f92cd8e9 — this thread is against the outdated diff.

You and a second reviewer both landed on this one. The finalizer fast path was removed entirely: `releaseFinishedRunner` now releases the pod and then deletes the `EphemeralRunner` normally, finalizers intact, so the deletion reconcile runs `cleanupContainerHooksResources` exactly as it does on master. Container-hook pods and Secrets are selected by the `runner-pod` label rather than an owner reference, so nothing else would have collected them.

The other reason it had to go: removing a succeeded runner outright also took it out of both buckets `ephemeralrunnerset_controller.go` counts against a scale-up — `FinishedRunnerCleanupPatchID` is only written inside `if len(finished) > 0`, and `scaleUpTotal` is `total + len(deleting)` — so the next patch would have created replacements for jobs that had already finished.

controllers/actions.github.com/ephemeralrunner_controller.go
```diff
@@ -1135,6 +1242,70 @@ func (r *EphemeralRunnerReconciler) SetupWithManager(mgr ctrl.Manager, opts ...O
...
+//
+// Native sidecars run as init containers that outlive the regular ones, so they
+// are checked too. A pod still running one of those, or a legacy sidecar
+// alongside the runner, is not terminated no matter what the runner container
+// did.
+func podTerminated(pod *corev1.Pod) bool {
+	switch pod.Status.Phase {
+	case corev1.PodSucceeded, corev1.PodFailed:
+		return true
+	}
```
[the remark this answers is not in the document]
nikola-jokic: Fixed in 6fa5189e.

The cited test isn't quite the case — `It should not set the phase to succeeded without pod termination status` puts the pod in `Running`, not `Succeeded`, so it never reached the phase shortcut. But the underlying point holds for `Failed`: the control plane moves a pod to `Failed` when its node is lost or shut down, while the last status the kubelet managed to send still shows a container running on the other side of the partition. Taking the phase's word there drops the pod out of the API while something may still be alive under it.

I didn't drop the phase check entirely, because it is doing a second job: deciding whether the containers that *haven't* reported a status are still to come. A pod whose init container failed never starts its regular containers, so `len(ContainerStatuses) != len(Spec.Containers)` forever, and requiring completeness alone would make every one of those wait out a full grace period before the retry can have its slot back.

So the two checks are now split by what they each answer. Every container the kubelet has reported on must be terminated — always, whatever the phase says. The phase only waives the requirement that every container has reported at all. Cases added: `Failed` with the runner still reported as `Running` → graceful, and `Failed` with no container statuses → immediate.

charts/gha-runner-scale-set-controller-experimental/tests/controller_deployment_args_test.yaml
```diff
@@ -89,7 +89,7 @@ tests:
             - "--autoscaling-runner-set-max-concurrent-reconciles=2"
             - "--autoscaling-listener-max-concurrent-reconciles=2"
             - "--ephemeral-runner-set-max-concurrent-reconciles=2"
-            - "--ephemeral-runner-max-concurrent-reconciles=4"
+            - "--ephemeral-runner-max-concurrent-reconciles=16"
```
Okabe-Junya: hmm, I haven't tested recent changes, but I think this conflicts with another your PR. Is this okey?

cf. https://github.com/actions/actions-runner-controller/pull/4670

> Why EphemeralRunner gets 4
Varying only EphemeralRunner concurrency at a non-binding rate limit, 1 → 4: EphemeralRunner-to-pod latency -92%, cleanup -62%, job-to-pod -59%, Kubernetes API requests -8%, and 409 conflicts halved from 14 to 7.
> 
> Going 4 → 16 is past the knee: no throughput gain, mean reconcile time +67%, and p99 variance explodes. 4 sits at the knee; 8 is about the ceiling.

main.go
```diff
@@ -163,9 +165,10 @@ func main() {
...
 	flag.BoolVar(&autoScalingRunnerSetOnly, "auto-scaling-runner-set-only", false, "Make controller only reconcile AutoRunnerScaleSet object.")
 	flag.Var(&autoScalerImagePullSecrets, "auto-scaler-image-pull-secrets", "The default image-pull secret name for auto-scaler listener container.")
-	flag.IntVar(&k8sClientRateLimiterQPS, "k8s-client-rate-limiter-qps", 20, "The QPS value of the K8s client rate limiter.")
-	flag.IntVar(&k8sClientRateLimiterBurst, "k8s-client-rate-limiter-burst", 30, "The burst value of the K8s client rate limiter.")
+	flag.IntVar(&k8sClientRateLimiterQPS, "k8s-client-rate-limiter-qps", 100, "The QPS value of the K8s client rate limiter. A burst of finishing jobs costs roughly ten API calls per runner, so the client-go default of 20 turns a few hundred runners into a minute of client-side queueing that delays both their cleanup and the runners replacing them.")
+	flag.IntVar(&k8sClientRateLimiterBurst, "k8s-client-rate-limiter-burst", 200, "The burst value of the K8s client rate limiter. Sized to absorb the API calls of a scale set turning over its runners at once, rather than spreading them over the QPS above.")
```
Okabe-Junya: I could understand the motivation to increase number (As this original one is from our org -  https://github.com/actions/actions-runner-controller/pull/3833), but do we actually need value?

I think would be better to put some results/scores to show improvements 👍

Okabe-Junya (commented):
Thank you for your work, I really like this improvement and understand the motivation 👍 
I haven't checked deeply, but let me left some comments

controllers/actions.github.com/ephemeralrunner_controller.go
```diff
@@ -1135,6 +1193,77 @@ func (r *EphemeralRunnerReconciler) SetupWithManager(mgr ctrl.Manager, opts ...O
...
+//
+// A pod that is still running is deleted normally. Skipping the grace period
+// there would drop the object while its containers were still alive, leaving
+// the kubelet to kill them with nothing in the API to account for the resources
+// they hold in the meantime.
+func (r *EphemeralRunnerReconciler) deletePodOptions(pod *corev1.Pod) []client.DeleteOption {
+	if !podTerminated(pod) || r.TerminatedPodGracePeriodSeconds < 0 {
+		return nil
+	}
+	return []client.DeleteOption{client.GracePeriodSeconds(r.TerminatedPodGracePeriodSeconds)}
```
Okabe-Junya: Minor suggestion, how about passing  `client.Preconditions{UID: &pod.UID}` here?

The runner pod is always named after the EphemeralRunner, and a failed pod is recreated under the same name (from my understanding). 

If the informer cache still holds the old pod when this force delete is issued, it can land on the new pod with that name

nikola-jokic: Good catch, adopted in a368e2dc.

You're right about the premise: the pod is named after the EphemeralRunner (`resourcebuilder.go` sets `Name: runner.Name`), a replacement reuses that name, and pods do come out of the informer cache — `main.go` only disables caching for Secrets and ConfigMaps. So a delete by name is a delete of whatever holds the name when the API server reads the request.

Where I'd push back slightly is the specific mechanism. Under a single leader the cache can't hand us pod A while pod B exists: the controller is the only creator of that name, controller-runtime serializes reconciles per key, it only creates after a Get returns NotFound, and informer events are ordered, so the cache can't regress from "A absent" back to "A present". B can't exist while we still see A.

But that argument is only worth the single-writer assumption behind it, and that assumption isn't load-bearing in practice: leader election is enabled whenever `replicaCount > 1`, and an outgoing leader's in-flight reconcile isn't killed the moment the lease moves. A stale old leader deleting the new leader's replacement pod is reachable.

What tipped it is that this PR changed the blast radius. That mis-targeted delete used to be graceful — SIGTERM plus the terminationGracePeriod, enough for the runner to deregister. At grace 0 it's an instant kill of a pod running a job, with nothing downstream to catch it. Cheap guard, so it earns its place.

Two implementation notes:

- I guarded on `pod.UID != ""`. All four call sites pass API-read pods (three `Get`s, one `List`), so it never trips today, but an empty-string UID precondition is one the API server can never satisfy, so a future synthesized pod would turn into a guaranteed 409 rather than a normal delete.
- A UID mismatch is 409 Conflict, not NotFound, so it surfaces as an error and a requeue at every call site. I left that as is deliberately. A conflict means the live pod isn't the one we read — and that pod still needs deleting. Swallowing it would let `cleanupResources` finish finalization with a pod left behind. The requeue re-reads and converges. It's also consistent with `deletePodAsFailed`, which keys `Status.Failures` by `pod.UID`: recording a failure against a UID we no longer own would be wrong anyway.

On testing: the fake client only honours the ResourceVersion precondition, not UID, so a fake-client delete would pass either way. The new `TestDeletePodOptionsDeletesOnlyThePodItLookedAt` pins it at the options level instead. The envtest suite runs against a real API server that does enforce UID preconditions, and it's green (117 specs), which is the actual evidence the live delete paths still work.

controllers/actions.github.com/resourcebuilder.go
```diff
@@ -849,7 +849,13 @@ func (b *ResourceBuilder) newEphemeralRunner(ephemeralRunnerSet *v1alpha1.Epheme
...
 			},
 		},
-		Spec: ephemeralRunnerSet.Spec.EphemeralRunnerSpec,
+		// Copied rather than shared. A plain assignment is a shallow copy, which
+		// leaves every runner built from this set pointing at the same container,
+		// volume and map values. Creating a runner writes the API server's
+		// response back into the object it was given, and the decoder reuses the
+		// maps and slices it finds there, so runners built in parallel would be
+		// writing into each other. Concurrently written maps end the process.
+		Spec: *ephemeralRunnerSet.Spec.EphemeralRunnerSpec.DeepCopy(),
```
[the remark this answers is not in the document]
nikola-jokic: Added in 4e4cb571 — `TestNewEphemeralRunnerDoesNotShareItsSpec` in `resourcebuilder_test.go`.

You were right that the existing tests couldn't catch this. The `EphemeralRunnerSet` they build has one container and no maps, so there is no shared memory for a second runner to collide with, and envtest passes either way.

The new test builds a set carrying the things the decoder reuses rather than allocates — pod labels and annotations, `NodeSelector`, a volume, a container with env, plus `Proxy.NoProxy` and `EphemeralRunnerConfigSecretMetadata.Labels` — then writes over one runner the way the API server's reply to its creation would, and asserts the runner built beside it is untouched.

I verified it fails on a shallow copy rather than assuming it would: reverting the `DeepCopy()` fails 10 of its assertions. The last of those is the one I hadn't expected — `assert.Equal(t, unchanged.Spec, set.Spec)` fails too, because a shallow copy means the decode writes straight back into the `EphemeralRunnerSet` the runner was built from, so every later runner in the same batch would inherit the first one's decoded values. That part isn't even concurrency-dependent; it corrupts serially within one reconcile.

charts/gha-runner-scale-set-controller-experimental/tests/controller_deployment_args_test.yaml
nikola-jokic: You're right, and this was our inconsistency. Reverted to 4 in 0ec1d5cd.

#4670 settled this with a benchmark, and the finding was specifically that 16 is past the knee: no throughput gain, mean reconcile time up by about two thirds, and p99 variance coming apart. This PR then raised the default to exactly that value with nothing measured to put against it. That's not a defensible way to overturn a benchmarked decision, especially one this recent.

What makes it worse is the direction of the regression. This PR exists to hand a finished runner's pod back promptly, and the pod is released by the reconcile that observes the exit — so a slower, less predictable reconcile is a slower release of the pod the next job is waiting on. We'd have been paying for the thing we were trying to fix.

The k8s client rate limiter bump went with it, for the reason you raised separately.

The batching in `EphemeralRunnerSet` stays, since it's a different axis from what #4670 measured: it removes serial round trips within a single reconcile of a single object, rather than changing how many objects reconcile at once. If the per-object concurrency is worth revisiting, that should be its own PR with numbers.

main.go
nikola-jokic: Fair, and we don't have the data. Reverted to 20/30 in 0ec1d5cd.

The justification was an estimate of API calls per runner lifecycle multiplied out — not a measurement. We never confirmed client-side throttling was actually occurring, so it was a tuning change aimed at a bottleneck nobody had observed. Raising QPS fivefold and burst nearly sevenfold for every ARC user on that basis isn't something the PR earned, and if the real constraint is elsewhere it buys nothing while putting more load on the API server.

If it's worth revisiting it should be its own PR with numbers behind it, the way #4670 did it — ideally with client-side wait latency shown before and after, so there's evidence the limiter was the constraint in the first place.

controllers/actions.github.com/ephemeralrunner_controller.go
```diff
@@ -484,14 +497,59 @@ func (r *EphemeralRunnerReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+//
+// Deleting the runner is what the controller has always done here, and the
+// deletion reconcile is what releases its registration, its pod and its
+// jitconfig secret. The pod is deleted first so that the cluster gets it back
+// now rather than a reconcile later. That reconcile is not worth skipping: the
+// EphemeralRunnerSet reads the runners that are finishing to tell a gap it
+// opened itself from fresh demand, and a runner that disappears before it is
+// read is a runner the set replaces for a job that is already done.
+func (r *EphemeralRunnerReconciler) releaseFinishedRunner(ctx context.Context, ephemeralRunner *v1alpha1.EphemeralRunner, pod *corev1.Pod, log logr.Logger) error {
+	r.releaseFinishedRunnerPod(ctx, pod, log)
```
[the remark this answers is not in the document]
nikola-jokic: Good catch — this is real, and it is ours. Removed in 2f84748.

Walking the chain, every link holds:

- `ephemeralRunnerOwnedPodPredicate` in `predicates.go` defines only `UpdateFunc`. `predicate.Funcs` defaults the rest to true, so pod DELETE events do enqueue the owning runner.
- `Reconcile` reads the runner with a plain cached `r.Get`. Pods and EphemeralRunners are separate informers on separate watch streams, and same-key workqueue serialization gives no ordering across resources, so that read can miss the `markAsSucceeded` patch.
- If it does, nothing downstream stops the recreate. `IsDone()` is false, the secret `Get` finds and reuses the JIT secret that has already been consumed, `Status.RunnerID` is non-zero so that block is skipped, there are no recorded failures, and the pod `Get` returns NotFound — straight into `createPod`.

The part worth being explicit about: master does not delete the pod on the success path at all. Its pod delete happens later, at finalization, long after Succeeded is in cache. This branch replaced both success sites with a helper that deleted the pod first, which is what put a pod deletion microseconds after the status patch on a different stream. So this was introduced here, not uncovered here.

Both sites are back to master's `r.Delete(ctx, &ephemeralRunner)` and the helper is gone. The grace period survives untouched, because `cleanupResources` already applies `deletePodOptions` and the runner delete triggers finalization on the very next reconcile — the capacity still comes back without a termination grace period, just without the new event ordering. The reconcile test now pins exactly that: the reconcile observing the clean exit must not delete the pod, and the finalizing one must delete it with grace 0. Envtest confirms the pod still disappears promptly.

On your alternative — an uncached read before recreating a missing pod — I agree it is the more complete fix, and it closes the latent version that exists on master independently of this PR. I would rather not fold it in here: it adds an API call to every runner creation, on the hot path this PR is trying to relieve, so it deserves its own PR with its own tests and a clear account of that cost.

rentziass (approved):

PR #4676, opened 2026-09-22
Remove dead GitHubConfig surface left by scaleset migration

Okabe-Junya:
## What

- Removes `GitHubConfig.GitHubAPIURL()`, the unread `Scope`/`IsHosted` fields with the `GitHubScope` type, and `isHostedGitHubURL()` from `github/actions/actions.go`

## Why

#4405 deleted `github/actions/client.go`, the only reader of `GitHubAPIURL()` and `Scope`; `IsHosted` and `GitHubScopeUnknown` were never read anywhere. 
The remaining callers of `ParseGitHubConfigFromURL` only use `ConfigURL`, `Enterprise`, `Organization`, and `Repository`. This also removes the only read of `GITHUB_ACTIONS_FORCE_GHES`, which no longer had any effect.


nikola-jokic (approved):
Thanks!

